PR #92, opened 2018-07-10
Document installation and usage of logging, monitoring and tracing components

mdemirhan:
Document installation and usage of logging, monitoring and tracing components


nrshrivatsan:
Excellent work @mdemirhan. Perfect example of how the Monitoring of Kubernetes itself has to be!

PR #234, opened 2018-07-23
Update Knative Serving GCS bucket

adrcunha:
Knative Serving release was moved to its own directory in knative/serving#1654.


mattmoor:
/assign @rgregg 

I think we should use the yamls attached to the `v0.1.0` release instead of HEAD so that folks aren't as disrupted by nightly issues.

mattmoor:
/approve

We can update these again later

PR #265, opened 2018-07-25
Add build component instructions throughout

### LINKED ISSUE #246: Re-add install and getting-started walkthrough for build
## Expected Behavior

Able to find a simple command to install Knative Build onto my cluster

## Actual Behavior

No simple setup instructions

## Additional Info

https://github.com/knative/build/pull/283 updated the build repo's README, and removed the install instructions and "Getting started" walkthrough.

That information should be added to the docs repo instead, with links from the build repo's README.

RichieEscarez:
Fixes #246 

Added stand alone "add build component to existing installation"
Added mention of install in README.md (which is linked to from knative/build code repo)
Added build component option in knative/install GKE instructions

### DIFF
--- a/install/Knative-with-GKE.md
+++ b/install/Knative-with-GKE.md
@@ -60,16 +60,16 @@ You need a GCP project to create a Google Kubernetes Engine cluster.
       gcloud projects create my-knative-project --set-as-default
       ```
       Replace `my-knative-project` with the name you'd like to use for your GCP project.
-      
+
       You also need to [enable billing](https://cloud.google.com/billing/docs/how-to/manage-billing-account)
       for your new project.
-      
+
     * If you already have a GCP project, make sure your project is set as your
       `gcloud` default:
       ```bash
       gcloud config set project my-knative-project
       ```
-      
+
       > Tip: Enter `gcloud config get-value project` to view the ID of your default GCP project.
 1. Enable the necessary APIs:
    ```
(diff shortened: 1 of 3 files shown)


install/Knative-with-GKE.md
```diff
@@ -135,32 +135,56 @@ rerun the command to see the current status.
...
     ```
-1. Monitor the Knative components, until all of the components show a `STATUS` of
-`Running`:
+1. Monitor the Knative serving components until all of the components show a
+   `STATUS` of `Running`:
     ```bash
     kubectl get pods -n knative-serving
     ```
 
+### Installing Knative build
```
ImJasonH: I'm not sure we need to update this readme just to mention that you can install build by itself. I think just having `installing-build-component.md` is sufficient for users that only want build; most users coming to Knative at this point probably want all of serving.

RichieEscarez: Information architecture wise, the Knative/install repo stands alone and therefore, users are going to expect to find install info about build here. I also want to avoid copying all of this file (all the prereq's like k8s, istio, and roles). Otherwise `installing-build-component.md` lacks that info (we dont have an end-to-end set of instructions for installing build by itself).

Ive not ramped up on eventing yet but I imagine we might need another section here too, for installing that component?

ImJasonH: That makes sense. In that case, can we start standardizing on documenting how to set up Knative components against _any_ k8s cluster, e.g., https://github.com/knative/docs/blob/master/install/Knative-with-any-k8s.md

Then at the top of that page we can have links to setup for GKE, AKS, PKS, etc., which can just be links to each provider's own evergreen documentation?

I don't want to have to copy "installing build" into each provider-specific tutorial, and I don't want to have it only documented in the GKE setup doc.

RichieEscarez: Ive created https://github.com/knative/docs/issues/270 to circle back around to creating shareable Knative specific install steps.

RichieEscarez:
@samodell will immediately begin working on issue #270 once this content is merged.

samodell:
/lgtm
/approve

I'll take on re-factoring the install guides so that we're not repeating ourselves.

install/Knative-with-GKE.md
```diff
@@ -135,32 +135,56 @@ rerun the command to see the current status.
...
+## Installing Knative components
 
-1. Next, we will install [Knative Serving](https://github.com/knative/serving)
-and its dependencies:
+You have the option to install and use only the Knative components that you
+want. You can install only the component of Knative if you need that
+functionality, for example Knative serving is not required to create and run
+builds.
+
+### Installing Knative Serving
```
ImJasonH: Installing serving also installs build, so it might be confusing to users who think they have to do both.

RichieEscarez: Sam and I discussed changing the language to "serving and build" or "build only" to clarify.

PR #271, opened 2018-07-25
Fix typo in comment

SaaldjorMike:
Fixes #(issue-number)

## Proposed Changes

* 
* 
*

### DIFF
--- a/serving/samples/blue-green-deployment.md
+++ b/serving/samples/blue-green-deployment.md
@@ -134,7 +134,7 @@ spec:
   - revisionName: blue-green-demo-00001
     percent: 100 # All traffic still going to the first revision
   - revisionName: blue-green-demo-00002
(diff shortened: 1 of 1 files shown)


samodell:
Thanks for your PR!

This is actually not a typo; the demo shows adding a new revision and at first not routing any traffic to it. In the next step, the traffic gets a 50-50 split.

SaaldjorMike:
@samodell I understand that the next step changes it to 50/50 split, but to me it looks like that exact point in time (where the comment is) defines traffic is 100% for blue-green-demo-00001 and 0% for blue-green-demo-00002

samodell:
@SaaldjorMike  Ah, yes! Apologies; I was reading this as adding the 5 instead of deleting it. Thanks for catching that.

SaaldjorMike:
I signed it!

PR #276, opened 2018-07-26
Update role reference in sample

ericbottard:
This may require double check / confirmation, but my cluster did not have a `viewer` role. It did have a `view` role though, which once used allowed me to run a Flow


evankanderson:
/unassign @vaikas-google 
/assign @evankanderson

Ville is out on a well-deserved break.

You're correct that this is `view`

/approve
/lgtm

Issue #170, opened 2018-07-18
Create a glossary of terms with definitions


labadav:

aslom:
This is excellent idea especially for more technical parts such as defining components in Knative projects.

For example defining queue-proxy:
https://github.com/knative/serving/blob/08b386d8fbcef2751cd2cc221098526354da4d37/docs/scaling/SYSTEM.md
https://github.com/knative/serving/pull/7126

omerbensaadon:
Started here: https://docs.google.com/spreadsheets/d/1p1_kBUd6ZvonxHkMcEJPayf6QIpExuFf5cFq0ptar7I/edit#gid=0

abrennan89:
> Started here: https://docs.google.com/spreadsheets/d/1p1_kBUd6ZvonxHkMcEJPayf6QIpExuFf5cFq0ptar7I/edit#gid=0

I don't really think this is a glossary; it's just for the style guide and doesn't actually contain descriptions of what the words are / mean

omerbensaadon:
Ah, I stand corrected, could be a good place to source words from though!

abrennan89:
Added a doc to collect terms that need definitions and possibly snippets: https://docs.google.com/spreadsheets/d/1ayoHb0Y833smQzIoe_Cj9NqwxaHpaYvaJ2tYtk3xw7A/edit?usp=sharing

soham283:
@abrennan89 I would like to work on the issue, can you please assign this issue to me?
(the changes must be done into word list page or on the spreadsheet which you have mentioned above?)

abrennan89:
@soham283 thanks! I don't think we ever decided where it should live, I think I may have deleted the old file since I never really got anywhere with it, so feel free to create any kind of document on the Google Drive for Knative that suits you (spreadsheet might be easiest or doc) to collect information and then we can decide where this will live later.

@snneji @omerbensaadon was a glossary included anywhere in the card sort, or should this live in github etc maybe?

soham283:
@abrennan89  I would like to do it on Spreadsheet.
 I have one query, I should start writing glossary by referencing only word list  or I have to go through from some different list??

snneji:
> @snneji @omerbensaadon was a glossary included anywhere in the card sort, or should this live in github etc maybe?

No, the glossary is not on the card sort. It can be put wherever you think best.

abrennan89:
> No, the glossary is not on the card sort. It can be put wherever you think best.

[...]

PR #285, opened 2018-07-27
Add glossary.md

ImJasonH:
This briefly describes some terms commonly thrown around when discussing Knative development and usage, and can serve as a handy reference guide when talking to new users and developers.

The list of terms is faaaar from exhaustive, these are just the few that came up in the last Build WG meeting, which we thought might be useful to document.


samodell:
Solves Issue #170

samodell:
After taking a look, I think this is a bit different than what we were imagining with issue #170 . It might make more sense to call this a "tools list" or something else other than "glossary," since we're not really defining terms here so much as explaining what tools are helpful when working with Knative.

What do you think about renaming this from "glossary" to something else, @ImJasonH ?

ImJasonH:
I don't have a strong affinity to the name, I just can't think of any better alternatives. :smile: 

For now these are all tools (except `go-containerregistry`, which is a _library_ used by some tools), but I suspected this would expand to include other things over time.

#170 seems to describe a glossary of Knative's own terms (Configuration, Route, Build, etc.), which definitely seems useful, but I agree probably doesn't belong in the same list alongside `ko` and `skaffold`.

If you have a better name in mind, I'm happy to change it, I just don't know to what. :thinking:

samodell:
Hmm... something like "helpful technologies" or "related tools and technologies"? We definitely will have a glossary of terms sometime in the near future, so it'd be great to reserve glossary.md for that!

resources.md
```diff
@@ -1,11 +1,77 @@
-# Glossary
+# Resources
 
-There is a dizzying array of tools used by Knative, and during Knative
-development, and it can be confusing to keep them all straight. This is an
-attempt to briefly document some terms that you might hear while working with
-or using Knative.
+This page contains informations about various tools and technologies
```
ImJasonH: Typo: information

resources.md
```diff
@@ -0,0 +1,132 @@
...
+that are useful to anyone developing on Knative.
+
+## Community Resources
+
+This section contains tools and technologies developed by members of the
+Knative community specifically for use with Knative.
+
+### [`knctl`](https://github.com/cppforlife/knctl)
+
+`knctl` is an under-development command line tool for working with
```
ImJasonH: Link to more information? Otherwise if there isn't any more than this I think we should remove it until there is more info

samodell: There is a link to the knctl repo, and there's some documentation there about how to work with it. What other information would you want included here? I could add that it's an alternative to working directly with kubectl.

ImJasonH: Each of the other sections has at least a couple explanatory/introductory paragraphs, even though you could link to each repo and just tell people to read there.

It's not a big deal, it just feels light compared to the others.

PR #287, opened 2018-07-27
Update install instructions to use release 0.1.0 files

### LINKED ISSUE #286: Use 0.1.0 in setup instructions
Our setup instructions point at a nightly release, these should reference our 0.1.0 release.

e.g. https://github.com/knative/docs/blob/master/install/Knative-with-AKS.md#installing-knative-serving

trisberg:
Fixes #286

## Proposed Changes

* Update install instructions to use files from the Serving 0.1.0 release instead of latest

### DIFF
--- a/install/Knative-with-AKS.md
+++ b/install/Knative-with-AKS.md
@@ -123,7 +123,7 @@ Knative depends on Istio.
 
 1. Install Istio:
     ```bash
-    kubectl apply -f https://storage.googleapis.com/knative-releases/serving/latest/istio.yaml
+    kubectl apply -f https://raw.githubusercontent.com/knative/serving/v0.1.0/third_party/istio-0.8.0/istio.yaml
     ```
 1. Label the default namespace with `istio-injection=enabled`:
     ```bash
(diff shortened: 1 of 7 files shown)


samodell:
Curious about the rationale behind this change -- why is /latest/ no longer a good strategy? @trisberg 

If we swap out latest for a specific version number, we'll need to update these commands regularly moving forward.

mattmoor:
@samodell it's a function of stability and putting our best foot forward for newcomers.

We know these samples work with 0.1.0, but now we might make breaking changes (intentionally or not).  I don't want prospective users to have a bad experience because of a bad nightly release the day before.

Does that make sense?  WDYT?

samodell:
Makes sense! Definitely agree we want to always direct users to stable versions. Would it be possible to have a "latest-stable" pointer similar to "latest" instead of hard-coding the versions? I don't know the complexity that might be involved, but if it's not a huge amount of work it might be worth it. @mattmoor

mattmoor:
cc @adrcunha 

I think the question is how we define "stable" :)

/hold cancel

krancour:
And if it's perceived as a maintenance burden to update docs with each release... that can always be automated.

rgregg:
Glad we're making this change. 

/lgtm

PR #301, opened 2018-08-01
Making a change to default azure location to reflect the correct Azure

ealeyner:
Changing what seems like a typo in the documentation, which would likely trip others.

Fixes #(issue-number)

## Proposed Changes

* 
* 
*


ealeyner:
I believe I have already signed the CLA for this project.

PR #312, opened 2018-08-03
Add performance to scaling working group

josephburnett:
## Proposed Changes

* Add performance to the responsibilities of the scaling working group.

### DIFF
--- a/community/WORKING-GROUPS.md
+++ b/community/WORKING-GROUPS.md
@@ -80,7 +80,7 @@ Slack Channel              | [#eventing](https://knative.slack.com/messages/C9JP
 
 ## Scaling
 
-Autoscaling
+Autoscaling, Performance
 
 Artifact                   | Link
 -------------------------- | ----


evankanderson:
/assign @mattmoor 

Matt can give the LGTM as a 2-of-3 ToC approval.

/approve

mattmoor:
tl;dr I feel like "Performance" is a cross-cutting thing, and autoscaling already encompasses a lot work.

I think the right separation of concerns is that this WG should be focused on the metrics pipeline and decision making to quickly right-size `Scale.Spec.Replicas` (incl. both KPA and Activator); the performance of which is critical, but just one link in the chain.  There is also the vertical scaling analog to this.

I don't think `Spec.Replicas => Status.Replicas` should be this working group's focus, e.g. the latency of reconciliations, image pulls, scheduling, network programming, container startup, hibernation, ... all feel like they belong in other working groups.

An analog to this kind of cross-cutting WG is `sig-scale` in Kubernetes, which looks across areas for bottlenecks scaling to bigger clusters.

My $0.02.

josephburnett:
What about "Scaling Performance and Efficiency"?

PR #321, opened 2018-08-07
Update Scaling WG VC link

greghaynes:
## Proposed Changes
* The scaling WG VC link does not match what is currently in the calendar invite,
updating to match with the assumption the calendar invite is correct.


greghaynes:
Looks like someone beat me to the update :)

PR #325, opened 2018-08-10
Doc to introduce how to install ExternalDNS within Knative

ZhiminXiang:
Fixes #(issue-number)
https://github.com/knative/serving/issues/1617

## Proposed Changes

* Add a doc to introduce how to install ExternalDNS within Knative.

[ExternalDNS](https://github.com/kubernetes-incubator/external-dns) is a tool that synchronizes exposed Kubernetes Services and Ingresses with DNS providers.


serving/using-external-dns.md
```diff
@@ -0,0 +1,171 @@
...
+Choose a DNS provider from the list.
+
+### Create a DNS zone for managing DNS records
+
+Skip this step if you already have a zone for managing DNS records of your 
+custom domain.
+
+A DNS zone which will contain the managed DNS records needs to be created.
+Assume your custom domain is `external-dns-test.my-org.do`. You can use the 
+following command to create a DNS zone if you are using Google Cloud DNS.
```
tcnghia: Can we split a sub section here for Google Cloud DNS specific stuff?  That way other people could contribute other provider instructions easily.

serving/using-external-dns.md
```diff
@@ -0,0 +1,171 @@
...
+gcloud dns record-sets transaction start --zone "my-org-do"
+gcloud dns record-sets transaction add ns-cloud-e{1..4}.googledomains.com. \
+    --name "external-dns-test.my-org.do." --ttl 300 --type NS --zone "my-org-do"
+gcloud dns record-sets transaction execute --zone "my-org-do"
+```
+
+### Deploy ExternalDNS
+Apply the following manifest to install ExternalDNS. Note that you need to set 
+the argument `domain-filter` to your custom domain. 
+```
```
tcnghia: can we check this in as a yaml file and link to it, so people can download?  alternatively if we have a link straight to external-dns's version that would be great.

tcnghia: if we check this in, we could provide a one-liner to replace the `domain-filter` property and install

tcnghia: similarly the one-liner could replace `provider=google` to something else, may be an env var.

ZhiminXiang: I chose to link straight to external-dns's version since we have to maintain our version to keep consistent with external-dns's version if we create our own one, which seems not very necessary.

serving/using-external-dns.md
ZhiminXiang: Done

ZhiminXiang:
@samodell would you mind taking a look at this PR? Thanks!

samodell:
@ZhiminXiang  Will do! Planning to take a look today. Thanks for your patience, and sorry for the wait.

serving/using-external-dns.md
```diff
@@ -0,0 +1,119 @@
...
+that synchronizes exposed Kubernetes Services and Ingresses with DNS providers.
+
+This doc explains how to set up ExternalDNS within a Knative cluster to 
+automate the process of publishing the domain used in Knative.
+
+## Prerequisite
+
+1. A Kubernetes cluster with [Knative Serving](https://github.com/knative/docs/blob/master/install/README.md) installed.
+1. A public domain that will be used in Knative.
+1. Follow the steps at [Setting up a custom domain](https://github.com/knative/docs/blob/master/serving/using-a-custom-domain.md#edit-using-kubectl) in the "Edit with kubectl" section to make Knative use your domain.
```
samodell: It might be easier to just pull in the necessary steps to this document instead of linking out to another page, especially since it's only a few steps.

serving/using-external-dns.md
```diff
@@ -0,0 +1,119 @@
...
+gcloud dns record-sets transaction start --zone "my-org-do"
+gcloud dns record-sets transaction add ns-cloud-e{1..4}.googledomains.com. \
+    --name "external-dns-test.my-org.do." --ttl 300 --type NS --zone "my-org-do"
+gcloud dns record-sets transaction execute --zone "my-org-do"
+```
+
+### Deploy ExternalDNS
+
+#### Google Cloud DNS
+Apply the [manifest](https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/gke.md#manifest-for-clusters-without-rbac-enabled) to install ExternalDNS. Note that you need
```
samodell: I would actually include the command here.

serving/using-external-dns.md
```diff
@@ -0,0 +1,119 @@
...
+You should see ExternalDNS is installed by running:
+```shell
+kubectl get deployment external-dns
+```
+
+### Configuring Knative Gateway service
+
+In order to publish the Knative Gateway service, the annotation
+`external-dns.alpha.kubernetes.io/hostname: '*.external-dns-test.my-org.do'`
+needs to be added into Knative gateway service:
```
samodell: I would show where this is added; you can include an example final version of the file after this addition has been made.

serving/using-external-dns.md
```diff
@@ -0,0 +1,119 @@
+# Using ExternalDNS to automate DNS setup
```
samodell: Just to make sure I understand, why is it necessary to automate DNS setup? It seems like publishing DNS records is a one-time thing you have to do when you first setup a custom domain.

Also, would it also be helpful to set up a static IP address for the Knative cluster?

serving/using-external-dns.md
samodell: Pull them in from the "Setting up a custom domain" document, that is. The "Edit with kubectl" section is pretty short.

serving/using-external-dns.md
ZhiminXiang: This is a good question. The benefits we can get from ExternalDNS are:
1. If user cannot (or does not want to) set up a static IP for Knative in their working environment, then user does not need to manually update their DNS record every time when they reinstall/restart Knative.
2. It provides a possibility for user to dynamically and programmatically control the Knative DNS.

ExternalDNS is not a strong requirement for Knative. But I think it is good to have an introduction for it.

serving/using-external-dns.md
ZhiminXiang: Done

serving/using-external-dns.md
ZhiminXiang: Done

PR #329, opened 2018-08-13
Clarifies working directory to run command

doddatpivotal:
## Proposed Changes

* I found that I didn't know the starting point to run the command to add monitoring components, given it referenced relative file locations.  By specifying the starting point as the root of the serving repository, it is clear to readers.


doddatpivotal:
I signed it!

PR #347, opened 2018-08-22
Update ROLES.md

dprotaso:


evankanderson:
I think we've been a bit more open with "member" status -- I'd either combine "Collaborator" and "Member", or change this to "recommended by 2 members".

dprotaso:
Done

PR #353, opened 2018-08-23
Vendor the test-infra scripts

adrcunha:
* shared scripts from test-infra live in //vendor/github.com/knative/test-infra/scripts;
* created `//hack/update-deps.sh` to properly update deps, including test-infra;
* all bash scripts were updated to use the vendored scripts;

Part of knative/test-infra#30.


adrcunha:
Ah.... `istio-galley` flakiness... :(

/retest

PR #354, opened 2018-08-24
add productivity wg

jessiezcc:
Fixes #(343)

added productivity wg description


jessiezcc:
had to get rid of previous PR since git commit --amend won't remove commits with wrong author

PR #360, opened 2018-08-28
services.serving.knative.dev -> ksvc, adding check install guide

### LINKED ISSUE #324: Docs should use the `kservice` short name instead of `services.serving.knative.dev`
We now have short names for Knative resources (see https://github.com/knative/serving/pull/1723)
In particular, the service object can now be queried with `kservice`.

This simplifies our docs, for example: `kubectl get services.serving.knative.dev helloworld-go` becomes `kubectl get kservice helloworld-go`

CC @rgregg

samodell:
Fixes #324 

## Proposed Changes

* Changes services.serving.knative.dev -> ksvc, adding a note about what version is necessary
   to use the short name.
* Adds guide on how to check the currently installed version of Knative
* Links to new guide from /install/README.md


install/README.md
```diff
@@ -51,6 +51,10 @@ and set up an Istio IP range for outbound network access:
 * [Configure outbound network access](../serving/outbound-network-access.md)
 * [Configuring HTTPS with a custom certificate](../serving/using-an-ssl-cert.md)
 
+## Checking Your Installed Version of Knative
```
RichieEscarez: Maybe: 

"Viewing the Version Number of Your Installation"

install/README.md
```diff
@@ -51,6 +51,10 @@ and set up an Istio IP range for outbound network access:
...
+
+* [Checking Your Installed Version of Knative](check-install-version.md)
```
RichieEscarez: Since this topic only covers serving, maybe we can be clearer in this text ( -> "Knative Serving")

install/check-install-version.md
```diff
@@ -0,0 +1,33 @@
+# Checking Knative Installation Version
```
RichieEscarez: I tried to run this command with only build installed (and know it says for serving in the first paragraph) but think that we should be clearer with the title too? ("Knative Serving")

PR #370, opened 2018-09-04
Mention MutatingAdmissionWebhook is required for serving

### LINKED ISSUE #369: Need to mention knative-serving depend on MutatingAdmissionWebhook
When I just follow "install/Knative-with-any-k8s.md" document with my own k8s cluster, I couldn't let knative-serving work unless added 2 additional spec to revisionTemplate in service.yaml. These additional values are ServingState, ConcurrencyMode.  After some investigation, it turned out that's because I didn't enable  MutatingAdmissionWebhook in kube-apiserver. these value are supposed to be injected by Webhook.
 
I think some of them can face same situation, so it's better to mention that knative depend on MutatingAdmissionWebhook in "install/Knative-with-any-k8s.md".

ukinau:
Fixes #369 

## Proposed Changes
* Mention MutatingAdmissionWebhook is required for serving in "Knative-with-any-k8s.md" doc

### DIFF
--- a/install/Knative-with-any-k8s.md
+++ b/install/Knative-with-any-k8s.md
@@ -5,9 +5,10 @@ using pre-built images.
 
 ## Before you begin
 
-Knative requires a Kubernetes cluster v1.10 or newer which enabled MutatingAdmissionWebhook.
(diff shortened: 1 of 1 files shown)


ukinau:
I signed it!

Issue #290, opened 2018-07-27
[samples/helloworld-go] remove redundant "flag" import|parse


DazWilkin:
## Expected Behavior

Since `"flag"` is not used, remove the import and the `Parse`.

## Actual Behavior

`"flag"` is imported in #20 and `Parse`'d.
https://github.com/knative/docs/blob/master/serving/samples/helloworld-go/helloworld.go

vdemeester:
This one is closeable (PR is merged)
/close

PR #373, opened 2018-09-05
Remove flag package import

tylerauerbeck:
Fixes #(issue-number)
#290

## Proposed Changes

Removes the `flag` package import and flag parsing as it is noted that this is done elsewhere.


RichieEscarez:
Reassigning to [serving lead](https://github.com/knative/docs/blob/master/community/WORKING-GROUPS.md#observability)

Issue #379, opened 2018-09-06
ksvc alias doesn't work with knative v0.1.1 as described


ahmetb:
## Expected Behavior

https://github.com/knative/docs/blob/eab66f7c86bb1b6c0ee570094565c9b081f5a4d2/serving/samples/helloworld-go/README.md says:

> Note: ksvc is an alias for services.serving.knative.dev. If you have an older version (version 0.1.0) of Knative installed, you'll need to use the long name until you upgrade to version 0.1.1 or higher.

## Actual Behavior

```
kubectl get ksvc helloworld-go  -o=custom-columns=NAME:.metadata.name,DOMAIN:.status.domain

error: the server doesn't have a resource type "ksvc"
```

even though knative 0.1.1 was installed in the first place.

## Steps to Reproduce the Problem

1. Install knative v0.1.1 to GKE [from this doc](https://github.com/knative/docs/blob/eab66f7c86bb1b6c0ee570094565c9b081f5a4d2/install/Knative-with-GKE.md#whats-next)
2. `kubectl get ksvc helloworld-go  -o=custom-columns=NAME:.metadata.name,DOMAIN:.status.domain` 
3. Get error.

RichieEscarez:
in all docs we do mention "*ksvc is an alias for services.serving.knative.dev*" so not P0 but `ksvc` is not working right now

steren:
`ksvc` has been introduced in https://github.com/knative/serving/pull/1723
If it does not work, please open this issue on the https://github.com/knative/serving repo

ahmetb:
Opened https://github.com/knative/serving/issues/1995 to track API issue.

I recommend removing this from the documentation for now until this works.

The abbreviated form simply isn't there in `kubectl api-resources`. I suspect either the commit didn't make it to the release or some other issue.

ahmetb:
Just verified commit https://github.com/knative/serving/commit/401b1ba25e8187325a1ce7f6e431bc5fbf4ce8ec didn't make it to the v0.1.1 release.

steren:
I think the underlying question here is: how to handle docs versioning? When should a feature be documented?

ahmetb:
That's a bit off-topic but for this simple instance, release changelog should be reflecting whether the feature is in v0.1.1 or not.

Docs versioning is a non-trivial problem, kubernetes solved it in v1.8 and until then (and still) they use "this feature is in available in vX.Y or later" notices all over the place.

[...]

PR #383, opened 2018-09-08
ksvc -> service.serving.knative.dev in readme

balopat:
Fixes a simple typo.


balopat:
Ah. I just realized ksvc is an alias - it just doesn't work (https://github.com/knative/docs/issues/379). closing.

PR #384, opened 2018-09-10
use base64 -w0 instead of some random high number

ahmetb:
This will yield the same result (single-line output as -w 0 disables line
breaks). However note that, -w flag doesn't exist in BSD base64, but this PR
is not about that.


ahmetb:
ping @labadav

ahmetb:
I don't think this needs eng input. It's pretty straightforward.

PR #389, opened 2018-09-10
source-to-url-go: support multi-line base64 output

ahmetb:
if my password is long "echo password | base64" will yield in a multiline
string that can confuse users about how to put that thing in Kubernetes
Secret manifest.

Adding -w0 (and for macOS/BSD base64(1), -b0) solves that problem.


ahmetb:
Revision is lgtm. :shipit:

PR #391, opened 2018-09-10
fix case sensitive #section anchors

ahmetb:
Looks like anchors are case-sensitive. GitHub Markdown rendering creates
lower-case-section-anchors.

How many more of these there are, I'm not sure. But when I searched like

    ag --md '#[A-Z]'

This is the only such markdown page with such uppercase anchor references,
so I'm just fixing this page. Sorry for long commit message.


mdemirhan:
Thanks for the fix!

/lgtm
/approve

PR #396, opened 2018-09-12
Fix typo in accessing-metrics command

howardjohn:


howardjohn:
I signed it!

howardjohn:
I think you may have added the needs cla label at the same time that I signed it

Issue #407, opened 2018-09-21
Update references of the `monitoring` namespace to `knative-monitoring`


tanzeeb:
Once https://github.com/knative/serving/pull/1851 is merged, we will need to update the documentation to reflect the new name change. 

All references to the `monitoring` namespace should be changed to `knative-monitoring`.

### What needs to be changed (not exhaustive):
* URLs (eg.  `/api/v1/namespaces/monitoring/` -> `/api/v1/namespaces/knative-monitoring/`)
* Command line flags (eg. `kubectl get pods --namespace monitoring --watch` -> `kubectl get pods --namespace knative-monitoring --watch`)

### What does *not* need to be changed:
* File paths (eg. `kubectl apply -Rf config/monitoring/` does not need to change)

tanzeeb:
Making a list:
- [ ] serving/accessing-logs.md
- [ ] serving/debugging-application-issues.md
- [ ] install/Knative-with-OpenShift.md
- [ ] serving/installing-logging-metrics-traces.md
- [ ] serving/accessing-metrics.md
- [ ] serving/setting-up-a-logging-plugin.md
- [ ] install/scripts/knative-with-openshift.sh
- [ ] serving/samples/telemetry-go/README.md
- [ ] serving/samples/telemetry-go/telemetrysample.go
- [ ] serving/samples/autoscale-go/README.md
- [ ] eventing/samples/k8s-events/README.md

RichieEscarez:
FYI: I am currently editing a few the logging related files, including:

serving/installing-logging-metrics-traces.md 
serving/accessing-metrics.md
serving/setting-up-a-logging-plugin.md

You can consider these done/in-progress. Please let me handle these changes since the files have changed significantly and a separate PR will cause merge conflicts.

tanzeeb:
Hey @RichieEscarez, no problem! Should I submit a PR for the other changes, or leave them?

RichieEscarez:
I was actually wonder when the code related changes will actually work in someones install?
For example, when will the commands/URLs in the docs actually work (I'd like to make sure the docs include "working commands" until the changes actual appear in the product)?

That said, im happy to take this on and handle the other files in a separate PR.

yanweiguo:
If someone installs knative from the release version, which currently is https://github.com/knative/serving/releases/download/v0.1.1/release.yaml, then the code relative changes will actually work after next release which includes the changes.

If someone installs knative serving from master branc(as a developer), then the code relative changes should actually work.

PR #416, opened 2018-09-25
Update references to monitoring namespace (#407)

tanzeeb:
Renames the `monitoring` namespace to `knative-monitoring` to reflect the changes in https://github.com/knative/serving/pull/1851. 

Some files have not been updated, see comments in #407.

cc @RichieEscarez

### DIFF
--- a/eventing/samples/k8s-events/README.md
+++ b/eventing/samples/k8s-events/README.md
@@ -96,7 +96,7 @@ kubectl apply --filename serviceaccount.yaml
     ```
 
     Then visit the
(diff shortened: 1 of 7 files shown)


RichieEscarez:
Marking as **Hold** for now since our docs are intended to serve the current release and none of these updates will work until the next release is cut. 

FYI: We are working towards defining a docs versioning model (so that there is an in-development version as well as "current release" version), and will merge this in to the correct place when ready.

matzew:
@RichieEscarez I get that docs should be versioned, but the `yaml` file for install, does actually use the renamed namespace - So... not sure why this is on hold ... should the install file be changed before ?

RichieEscarez:
Sorry, did I misunderstand? I thought that our install docs, for example on [GKE](https://github.com/knative/docs/blob/master/install/Knative-with-GKE.md#installing-knative-serving-and-build-components), where we specify a specific release version (like:
`https://raw.githubusercontent.com/knative/serving/v0.1.1/third_party/config/build/release.yaml`), won't actually pick up that change.

Regarding docs versioning, this might be happening soon (which will unblock these types of changes).

PR #427, opened 2018-09-27
update calendar link for productivity wg

jessiezcc:
The current link is invalid due to calendar event cancellation happened before. 
Updated the link pointing to valid knative team meeting calendar.


bbrowning:
Thanks for updating this! As a non-Google employee, I can confirm the previous link did not work for me and the new link to the shared calendar does work.

/lgtm

PR #431, opened 2018-09-28
use 0.1.1 release for Openshift instructions

### LINKED ISSUE #430: Openshift: use 0.1.1 release instead of latest
## Expected Behavior

Openshift doc/script use `latest`, but should use a versioned install (current `0.1.1`)

## Actual Behavior

usage of latest, which is not the desired approach

matzew:
Fixes #430

## Proposed Changes

* usage of latest is replaced by _actual_ release
* added a script to waiting for pods on the `knative-build` namespace to become ready
*

### DIFF
--- a/install/Knative-with-OpenShift.md
+++ b/install/Knative-with-OpenShift.md
@@ -133,9 +133,9 @@ rerun the command to see the current status.
 > Note: Instead of rerunning the command, you can add `--watch` to the above
   command to view the component's status updates in real time. Use CTRL+C to exit watch mode.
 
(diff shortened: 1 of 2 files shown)


pmorie:
@matzew thanks for this PR - but @kameshsampath has been working on the same area. Will you two get in touch and coordinate on these changes?

matzew:
@pmorie those are mainly _minishift_ updates.

I wonder if we should either keep `oc cluster up` and `minishift start` different, or make one doc, to basically just point out the diffs - but not sure which is really better

PR #432, opened 2018-09-28
Close response body in autoscale-go sample test program

sgotti:
## Proposed Changes

The test program doesn't close the http response Body causing the connection to
not be released and the creation of new connections until the exhaustion of the
process file descriptors.

This patch closes the response body (actually a defer also if not really needed but useful if in future the body will be read)


sgotti:
gentle ping :smile:

sgotti:
@josephburnett Thanks!

PR #435, opened 2018-10-01
Update accessing-metrics.md with knative-metrics namespace.

josephburnett:
## Proposed Changes

* Update accessing-metrics.md with knative-monitoring namespace.


RichieEscarez:
Marking a Hold until a separate docs branch is created so that the in-progress (new development) content is separate out from the docs the represent the behavior of the latest release.

PR #436, opened 2018-10-02
build/auth.md: add link for GCR/ECR helpers

AkihiroSuda:
Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>

### DIFF
--- a/build/auth.md
+++ b/build/auth.md
@@ -185,6 +185,9 @@ When this build executes, before steps execute, a `~/.docker/config.json` will
 be generated containing the credentials configured in the `Secret`, and these
 credentials are then used to authenticate with the Docker registry.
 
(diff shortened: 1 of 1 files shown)


samodell:
/hold

@AkihiroSuda , thanks for this! This seems a little confusing at the moment, though, and I'm not sure how it relates to the other steps in this guide.

Couple of questions:

1. Did you mean to add this under the "Basic authentication (Docker)" section, or should this be in a new section on its own?

2. Can you work this into a set of steps and add more explanation for how the script you've added works? It would be helpful to call out areas in the script where you're expecting users to replace dummy values with their own correct values, for example.

AkihiroSuda:
opened https://github.com/knative/build-templates/pull/72 and replaced this PR with just a link to the build-templates repo

PR #443, opened 2018-10-07
Make samples less scary when no env var is provided

steren:
When no env var is provided (for example because the developer only has deployed the code without using the `service.yaml`), instead of displaying "Hello World: NOT SPECIFIED", display the more friendly message "Hello Word".
This means that when the `TARGET` env var is set, instead of displaying "Hello World: <TARGET>", the samples will display "Hello <TARGET>", which seems totally OK.

All sample changes have been tested.


bbrowning:
The integration test failure was unrelated to this change and due to an issue impacting all PRs in the docs repo that is now fixed.

/test pull-knative-docs-integration-tests

PR #434, opened 2018-10-01
Update knative/test-infra

adrcunha:
Major changes:
* fixes and updates to E2E test framework
* better parsing of E2E test results

PR #446, opened 2018-10-08
Update vendored test-infra version

bbrowning:
This repo has fallen behind others in the vendored version of
test-infra and as a result integration tests are erroring out.

I'm using `dep` version `v0.5.0` locally, which matches what other
Knative repositories are using for generation of their Gopkg.lock
files. That does cause a bit more noise to this PR as the `digest` and
`pruneopts` fields in Gopkg.lock get filled in.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,88 +3,113 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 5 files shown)


bbrowning:
I'm closing this one in favor of #434 - I didn't see that was already out there.

PR #451, opened 2018-10-09
Node.js serving sample README.md has inconsistent response example

jonasfj:
**WARNING:** I didn't actually test this. **Consider this a bug report** with a suggestion, more than a fully tested patch.

Please, check that I didn't missing something, it's plausible that missed something obvious.

-----

We specify `TARGET='Node.js Sample v1'` in `service.yaml`, so it would be surprising if this isn't printed in the response.

### DIFF
--- a/serving/samples/helloworld-nodejs/README.md
+++ b/serving/samples/helloworld-nodejs/README.md
@@ -186,7 +186,7 @@ folder) you're ready to build and deploy the sample app.
 
     ```shell
     curl -H "Host: helloworld-nodejs.default.example.com" http://{IP_ADDRESS}
-    Hello World: NOT SPECIFIED
+    Hello Node.js Sample v1!
     ```
 
 ## Removing the sample app deployment


RichieEscarez:
Needs to be tested before we can merge. Assigning to original author (someone more familiar with curl and nodejs).

serving/samples/helloworld-nodejs/README.md
```diff
@@ -179,7 +179,7 @@ folder) you're ready to build and deploy the sample app.
 
     ```shell
     curl -H "Host: helloworld-nodejs.default.example.com" http://{IP_ADDRESS}
-    Hello World: NOT SPECIFIED
+    Hello World: Node.js Sample v1
```
averikitsch: Should be "Hello Node.js Sample v1!"

RichieEscarez: also please update the corresponding line in the [service.yaml](https://github.com/knative/docs/blob/c1be002db74a3874ba66cd1c2db9e39f00f85fe0/serving/samples/helloworld-nodejs/service.yaml#L15)

jonasfj: The `service.yaml` file is correct... the `"Hello "` and `"!"` originates from the `app.js` file...

PR #473, opened 2018-10-22
Removed anyuid addon

sjwoodman:
Removed `anyuid` addon for Minishift as it is not required. Istio manually adds some `anyuid` as required but setting it globally is best avoided if possible.

Tested by @matzew


matzew:
@pmorie with your Openshift experience, would you like to see this change, to remove the `anyuid` addon (aka `oc adm policy add-scc-to-group anyuid system:authenticated`) ?

sjwoodman:
Removing this as `anyuid` may still be required by Istio Galley pod

PR #489, opened 2018-10-31
Copy VENDOR-LICENSE over from eventing repo

### LINKED ISSUE #488: Broken VENDOR-LICENSE symlinks cause build to fail in sample
When following instructions in k8s-events sample the docker image build fails due to VENDOR-LICENSE symlinks in /vendor/github.com/knative/eventing/cmd packages pointing into a non-existent third_party directory.

gsf:
This file is referenced by symlinks in eventing/cmd. Fix #488

### DIFF
--- a/vendor/github.com/knative/eventing/third_party/VENDOR-LICENSE
+++ b/vendor/github.com/knative/eventing/third_party/VENDOR-LICENSE
@@ -1,0 +1,7742 @@
+
+===========================================================
+Import: github.com/knative/eventing/vendor/cloud.google.com/go
+
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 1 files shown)


gsf:
Alternatively, eventing/cmd could be removed, as all it contains are the kodata files. Do they need to be vendored?

mattmoor:
This is generated and generally in flux, so we shouldn't update it manually.

Does running `./hack/update-deps.sh` fix this?

mattmoor:
@n3wscott can you help out here?

gsf:
Running `./hack/update-deps.sh` does not fix this.

mattmoor:
I see, it looks like `dep` is too clever and prunes `third_party/`, but not the things that link there.

Does adding a `.dockerignore` file for this sample with `*/kodata/*` help?

gsf:
Looks like `hack/update-deps.sh` in the eventing repo relies on `update_licenses` to generate the VENDOR-LICENSE:

https://github.com/knative/eventing/blob/0191960e6bf7cc6d4b99d3e486b109d558f47d64/hack/update-deps.sh#L34-L35

n3wscott:
I think the correct thing to do is just remove the cmd dirs from eventing: https://github.com/knative/docs/pull/490

Issue #497, opened 2018-11-06
serving/samples/helloworld-*: Simplify, standardize, and explain Dockerfiles


grayside:
## Expected Behavior

Helloworld Dockerfiles should be clear, consistent, and self-documenting to help developers understand what each Dockerfile does and further learn by comparing different Dockerfiles for languages they know.

Here is an example of the kind of results I'm hoping for across languages:

```Dockerfile
# Use the latest release of the official Clojure image.
# https://hub.docker.com/_/clojure
FROM clojure

# Create the project and download dependencies.
WORKDIR /usr/src/app
COPY project.clj .
RUN lein deps

# Copy local code to the container image.
COPY . .

# Build the app as an uberjar.
RUN mv "$(lein uberjar | sed -n 's/^Created \(.*standalone\.jar\)/\1/p')" app-standalone.jar

# Configure and document the service HTTP port.
ENV PORT 8080
EXPOSE $PORT

# Run the web service on container startup.
CMD ["java", "-jar", "app-standalone.jar"]
```

1. The base image is briefly explained and a link to Docker Hub presented so the developer can dig deeper into how the image is built and used.
2. Each "section" of the Dockerfile is commented with a capitalized letter and a period using a somewhat standardized description.
3. Use COPY, not ADD
4. When copying into the current working directory, target `.`, not `./` nor an absolute path. This helps the Dockerfile feel more concise.
5. Ensure multiple stages are used and broken down in a way that feels like a familiar build pipeline.
6. Consistently use port handling so the use of the PORT environment variable is clear and actually overridable by, e.g., `docker run -e PORT=9090`.
7. Use CMD not ENTRYPOINT wherever possible so docker run can be used with different commands without an --entrypoint flag.
8. End on CMD so the last thing developers read is what happens when they start the container.
9. If a multi-stage build is used, clearly comment both stages, the COPY statement, and name the first stage. Link to multi-stage build reference docs since that is a more intermediate-level feature, even though we are encouraging it in a basic example.
10. Minimize the number of layers, except where it promotes easier change-and-rebuild on top of Docker layer caching. **Should we should keep rebuild/layer caching concerns in helloworld? Punting to the future.**

## Actual Behavior

Helloworld Dockerfiles are inconsistent, making a variety of decisions around comments & syntax and not providing enough comments to orient container learners.

## Todo List

* [x] Clojure
* [x] C#
* [x] Dart
* [ ] Elixir
* [x] Go
* [x] Haskell
* [x] Java
[...]

PR #501, opened 2018-11-06
serving/samples/helloworld-dart: simplify and standardize Dockerfile

grayside:
Addresses #497 for Dart.

Tested with $PORT and $TARGET.


grayside:
This appears to be an infrastructure failure.

/retest

PR #512, opened 2018-11-06
serving/samples/helloworld-java: simplify and standardize Dockerfile

grayside:
Addresses #497 for Java.

Tested with $TARGET and $PORT, found $PORT support does not work. (That's a separate issue, as nothing in this change should affect PORT behavior.)


grayside:
/test pull-knative-docs-integration-tests

Looks like the failure is an infra fault.

PR #509, opened 2018-11-06
Remove outdated samples

vaikas-google:
## Proposed Changes

* Remove old samples and once new Github Source is working, add a sample here for it.
* https://github.com/knative/eventing-sources/issues/78

PR #519, opened 2018-11-07
Eventing - IoT Core Sample

Harwayne:
Fixes knative/eventing#551

## Proposed Changes

* Add the IoT Core sample for Knative Eventing.
    - Based on the sample removed in #509 and [mchmarny/iot-code-client](https://github.com/mchmarny/iot-code-client)


justingrayston:
I'd be interested in knowing if someone else gets this working? I followed the steps but the subscription never happened, previously when I tried the GcpPubSubSource sample I got a deploy and a new subscription in the name of that deploy in pubsub. 

I am going to keep battling, but it would be nice to know if it is just me or something fundamental :)

Also, I am not really a gopher (more a python and NodeJS person historically) so it could be a version thing, but the generator main.go refused to run as is. This was due to `uuid.NewV4()` returning multiple values and `.String()` seemingly wanting a single context.

I changed to the following and it worked.

```go
func makeEvent() string {

	s1 := rand.NewSource(time.Now().UnixNano())
	r1 := rand.New(s1)
	u, err := uuid.NewV4()
	id := u.String()

	if err != nil {
		log.Fatal(err)
	}

	event := struct {
		SourceID string  `json:"source_id"`
		EventID  string  `json:"event_id"`
		EventTs  int64   `json:"event_ts"`
		Metric   float32 `json:"metric"`
	}{
		SourceID: *eventSrc,
		EventID:  fmt.Sprintf("%s-%s", idPrefix, id),
		EventTs:  time.Now().UTC().Unix(),
		Metric:   r1.Float32(),
	}

	data, _ := json.Marshal(event)

	return string(data)

}
```

Harwayne:
> Also, I am not really a gopher (more a python and NodeJS person historically) so it could be a version thing, but the generator main.go refused to run as is. This was due to `uuid.NewV4()` returning multiple values and `.String()` seemingly wanting a single context.

That's odd. It looks like the vendored library, which is used by `main.go`, [doesn't return an error](https://github.com/knative/docs/pull/519/files#diff-d1e9b3a1fc1cd6c913900ac604ddf4d1R64). Is there a chance that your imports changed somehow? Or the vendored files changed?

vaikas-google:
/approve
/lgtm

I just ran through this example from a fresh GKE cluster after installing serving 0.2.1 (and istio based off the 0.2.1 docs) and the :latest of eventing, there are couple little cleanups we'll follow up with, but it did work for me.

Let's create a tracking bug if this doesn't work on the latest?

PR #538, opened 2018-11-12
Kubernetes event source sample: get message dumper logs by label selector

grantr:
Now user doesn't need to fill in the actual pod name.

/cc @vaikas-google


vaikas-google:
Nice! Didn't know about this!
/lgtm
/approve

PR #545, opened 2018-11-15
Require rack 2.0.6

grantr:
Replaces #544 

/cc @samodell


grantr:
I haven't run through the sample, but I did verify that `bundle install` worked locally.

PR #558, opened 2018-11-19
Update install instructions for eventing and eventing-sources.

evankanderson:
## Proposed Changes

* Ran prettier.io for formatting / fixed markdown lint
* `eventing-sources` controller requires Istio to be installed, and also knative-serving. Mentioned these dependencies.
* Updated links to point to released yaml rather than head.
* The PubSub source requires that a secret be created. Added a dummy secret. See my other PR for documentation on how to create the correct secret.

### DIFF
--- a/eventing/README.md
+++ b/eventing/README.md
@@ -3,15 +3,15 @@
 Knative Eventing is a system which is designed to address a common need for
 cloud native development:
 
-1.  Services are loosely coupled during development and deployed independently
-    on a variety of platforms (Kubernetes, VMs, SaaS or FaaS).
-1.  A producer can generate events before a consumer is listening, and a
-    consumer can express an interest in an event or class of events that is not
-    yet being produced.
-1.  Services can be connected to create new applications
-    - without modifying producer or consumer.
-    - with the ability to select a specific subset of events from a particular
-      producer
+1. Services are loosely coupled during development and deployed independently on
+   a variety of platforms (Kubernetes, VMs, SaaS or FaaS).
+1. A producer can generate events before a consumer is listening, and a consumer
+   can express an interest in an event or class of events that is not yet being
+   produced.
+1. Services can be connected to create new applications
+   - without modifying producer or consumer.
+   - with the ability to select a specific subset of events from a particular
+     producer
 
 The above concerns are consistent with the
 [design goals of CloudEvents](https://github.com/cloudevents/spec/blob/master/spec.md#design-goals),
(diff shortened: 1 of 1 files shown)


eventing/README.md
```diff
@@ -50,22 +50,33 @@ systems using Kubernetes
...
-the following yaml:
+In addition to the core sources, you can also use GCP PubSub as a source by
+creating a secret with the name `gcppubsub-source-key` with a `key.json` value
+and loading the released source yaml (the `-with-gcppubsub` release includes all
+the above sources, and adds GCP PubSub, which requires the listed secret):
 
 ```bash
-kubectl apply --filename https://storage.googleapis.com/knative-releases/eventing-sources/latest/release-with-gcppubsub.yaml
+kubectl create namespace knative-sources
+kubectl --namespace knative-sources create secret generic gcppubsub-source-key --from-literal=key.json=''
```
vaikas-google: seems like we should specify the key file here?

evankanderson: Creating the key file requires a GCP account and a half-dozen `gcloud` commands. This command is sufficient to cause the controller to schedule, leaving the actual secret generation to the gcppubsub sample (see #559).

IMO, we should fix the gcppubsub source to load the secret dynamically per-Source rather than putting it in a Downward API volume.

eventing/README.md
```diff
@@ -43,10 +48,13 @@ This allows message delivery in a cluster to vary based on requirements, so that
 some events might be handled by an in-memory implementation while others would
 be persisted using Kafka or NATS Streaming.
 
+### Future design goals
+
 The focus for the next Eventing release will be to enable easy implementation of
 event sources. Sources manage registration and delivery of events from external
 systems using Kubernetes
 [Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/).
+Learn more about Eventing development in the [Events work group](https://github.com/knative/docs/blob/master/community/WORKING-GROUPS.md#events).
```
evankanderson: s/Events/Eventing/

"Events" can be confused with attending something like Kubecon.

eventing/README.md
```diff
@@ -1,25 +1,28 @@
...
+   and across a variety of platforms (for example Kubernetes, VMs, SaaS or FaaS).
+1. Event producers and event sources are independent. Any producer (or source), can generate events 
+   before there are active event consumers that are listening. Any event consumer can express interest in an 
+   event or class of events, before there are producers that are creating those events.
+1. Other services can be connected to the Eventing system. These services can perform the following functions:
+   - Create new applications without modifying the event producer or event consumer. 
+   - Select and target specific subsets of the events from thier producers.
+1. Ensure cross-service interoperability. Knative Eventing is consistent with the 
+   [CloudEvents](https://github.com/cloudevents/spec/blob/master/spec.md#design-goals)
+   specification that is developed by the [CNCF Serverless WG](https://lists.cncf.io/g/cncf-wg-serverless).
```
evankanderson: This is fine, but I'll later (subsequent PR) run this through prettier.io with `prosewrap=always`.

eventing/README.md
RichieEscarez: I initially wrote "Eventing work group" but on that community page, its labelled "Events" work group (sounds like the WORKING-GROUPS.md page needs a update).

evankanderson:
(still needs an lgtm from not the author)

PR #563, opened 2018-11-21
Discover the K8S version for IKS deployment

### LINKED ISSUE #562: K8S version in IKS install doc is stale
## Expected Behavior
IKS install instruction work 

## Actual Behavior
Cluster creation fails since version 1.10.3 is not available on IKS.

1. ibmcloud cs cluster-create --name=$CLUSTER_NAME \
  --zone=$CLUSTER_ZONE \
  --kube-version=1.10.3 \
  --machine-type=b2c.4x16 \
  --workers=3 \
  --private-vlan $PRIVATE_VLAN_ID \
  --public-vlan $PUBLIC_VLAN_ID

afrittoli:
Version 1.10.3 is not available anymore in IKS. Having a fixed
version will become stale eventually. The default version is
now >1.10 but that may change, so ask installers to select an
appropriate version based on the minimum requirement.

Fixes #562

## Proposed Changes

* Add instructions on how to discover available k8s version
* Replace fixed k8s version with variable
*

### DIFF
--- a/install/Knative-with-IKS.md
+++ b/install/Knative-with-IKS.md
@@ -61,7 +61,7 @@ environment variables.
 To make sure the cluster is large enough to host all the Knative and Istio
 components, the recommended configuration for a cluster is:
 
-- Kubernetes version 1.10 or later
+- Kubernetes version  or later
 - 4 vCPU nodes with 16GB memory (`b2c.4x16`)
 
 1.  Set `ibmcloud` to the appropriate region:
@@ -71,6 +71,7 @@ components, the recommended configuration for a cluster is:
 1.  Select a Kubernetes version:
     ```bash
     ibmcloud cs kube-versions
+    export CLUSTER_K8S_VERSION=[a version from the list, must be >1.10]
     ```
 1.  Create a Kubernetes cluster on IKS with the required specifications:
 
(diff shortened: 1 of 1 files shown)


install/Knative-with-IKS.md
```diff
@@ -68,12 +68,16 @@ components, the recommended configuration for a cluster is:
...
+    ```bash
+    ibmcloud cs kube-versions
+    ```
 1.  Create a Kubernetes cluster on IKS with the required specifications:
 
     ```bash
     ibmcloud cs cluster-create --name=$CLUSTER_NAME \
       --zone=$CLUSTER_ZONE \
-      --kube-version=1.10.3 \
+      --kube-version=$CLUSTER_K8S_VERSION \
```
averikitsch: Do you need an additional step to set this env var?

afrittoli: Thanks for the review! 
I though it would be clear from the description in the previous step (similar as it's done later for PRIVATE_VLAN_ID and PUBLIC_VLAN_ID, but I would be happy to add a not in the previous step to make this more clear.

install/Knative-with-IKS.md
```diff
@@ -61,19 +61,24 @@ environment variables.
 To make sure the cluster is large enough to host all the Knative and Istio
 components, the recommended configuration for a cluster is:
 
-- Kubernetes version 1.10 or later
+- Kubernetes version  or later
```
averikitsch: Keep "1.10"

afrittoli: meh, sorry, nice catch

Issue #569, opened 2018-11-22
[sample/helloworld-go] helloworld-go pod cannot started


gyliu513:
## Expected Behavior
I was following steps at https://github.com/knative/docs/blob/master/serving/samples/helloworld-go/README.md and the pod of helloworld-go should be able to start.

## Actual Behavior
But the problem is that the pod of helloworld-go cannot start due to following errors.

```
root@gyliu-c11:~/go/src/github.com/gyliu513/knative-docs/serving/samples/helloworld-go# kubectl logs -f helloworld-go-00001-deployment-d4cb46b5-fwmkj -c user-container
standard_init_linux.go:190: exec user process caused "no such file or directory"
```

## Steps to Reproduce the Problem

Just follow steps at https://github.com/knative/docs/blob/master/serving/samples/helloworld-go/README.md

After `kubectl apply --filename service.yaml`, the pod of helloworld-go cannot start.

grayside:
As I mentioned in #571, I have encountered this issue and agree it needs fixing. The Go sample Dockerfiles without CGO_ENABLED=0 are currently broken.

PR #571, opened 2018-11-23
Use go build to build helloworld-go image.

gyliu513:
Fixes https://github.com/knative/docs/issues/569

## Proposed Changes

* Use CGO_ENABLED=0 to build docker image

/cc @grayside


serving/samples/helloworld-go/Dockerfile
```diff
@@ -3,19 +3,20 @@
 FROM golang as builder
```
vdemeester: nit: Or we could use `golang:alpine` here :angel:

gyliu513: Make sense, done.

grayside:
I have also observed this problem, though curiously Docker For Mac does not hit this issue.

/lgtm

grayside:
Ah, I just noticed that this changes the builder stage image to golang:alpine. I am not as sure about that, especially for helloworld. While valuable, it is an optimization that can require developers to need to install many basic components into their image in order to build more complex applications.

grayside:
Un-LGTM-ing for further discussion.

/lgtm cancel

gyliu513:
@grayside make sense, done. Thanks.

RichieEscarez:
@grayside did you mean to add the LGTM? I can approve and merge if thats the case?

gyliu513:
Ping @RichieEscarez for approve.

PR #572, opened 2018-11-23
Use golang:alpine instead of golang for example Dockerfile.

gyliu513:
This is a follow up action for https://github.com/knative/docs/pull/571

/cc @vdemeester

### DIFF
--- a/serving/samples/build-private-repo-go/README.md
+++ b/serving/samples/build-private-repo-go/README.md
@@ -187,7 +187,10 @@ The sample code is in a private Github repository consisting of two files.
 
 1. `Dockerfile`
    ```Dockerfile
-   FROM golang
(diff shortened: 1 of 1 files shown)


grayside:
As I commented on #571, using alpine as the base for any "building" image means developers that iterate on top of the Dockerfile may need to add their own basic development dependencies (ca-certificates, gcc, etc)

For these non-helloworld samples, I think it's more reasonable to do that, especially since these are not using multi-stage builds. However, it would be helpful for an inline comment to point out this concern and link to relevant documentation. Perhaps https://hub.docker.com/_/golang/ with a pointer to read the section about alpine images?

vdemeester:
@grayside sounds fair :wink:

gyliu513:
@grayside @vdemeester done.

PR #599, opened 2018-11-27
Fix --from-file reference

### LINKED ISSUE #594: Eventing IoT Core demo has incorrect --from-file value
## Expected Behavior
In GCP PubSub Source it should be:

kubectl -n knative-sources create secret generic gcppubsub-source-key --from-file=PATH_TO_KEY_FILE.json
kubectl -n default create secret generic google-cloud-key --from-file=PATH_TO_KEY_FILE.json

## Actual Behavior
Actual is:
kubectl -n knative-sources create secret generic gcppubsub-source-key --from-file=key.json=PATH_TO_KEY_FILE.json
kubectl -n default create secret generic google-cloud-key --from-file=key.json=PATH_TO_KEY_FILE.json

I will do a PR shortly.

justingrayston:
Change --from-file=key.json--from-file=PATH_TO_KEY_FILE.json
To --from-file=PATH_TO_KEY_FILE.json

Fixes #594

### DIFF
--- a/eventing/samples/iot-core/README.md
+++ b/eventing/samples/iot-core/README.md
@@ -89,8 +89,8 @@ export IOTCORE_TOPIC_DEVICE="iot-demo-device-pubsub-topic"
         the Receive Adapter):
 
         ```shell
-        kubectl -n knative-sources create secret generic gcppubsub-source-key --from-file=key.json=PATH_TO_KEY_FILE.json
-        kubectl -n default create secret generic google-cloud-key --from-file=key.json=PATH_TO_KEY_FILE.json
+        kubectl -n knative-sources create secret generic gcppubsub-source-key --from-file=PATH_TO_KEY_FILE.json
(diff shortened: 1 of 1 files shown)


ahmetb:
This change is not NOOP.
`--from-file=key.json=ACTUAL_FILE.json` ensures that the Secret object has a data field named `key.json` with the contents of ACTUAL_FILE.json.

If you drop the =key.json part, it will take the name of the file you're loading, which would break.

justingrayston:
Ah ha.

```
spec:
  gcpCredsSecret:
    name: google-cloud-key
    key: key.json
```

So the creds file was `myreallylongserviceaccountname.json` the above key would have to change to that name as key.json wouldn't exist. Leaving it key.json without the = would mean they would have to have key.json at the actual path. 

You know this could explain why my subscription failed to create!

I'll close this and the issue and get a 🚕

PR #601, opened 2018-11-27
Argument must be "--sink" exactly for the sink to be filled in for ContainerSource

### LINKED ISSUE #600: ContainerSource argument must be exactly "--sink" to fill in the sink
## Expected Behavior
The docs state "Any --sink= argument will be filled in with the DNS address of the sink object", implying I could provide "--sink=" in my list of arguments and it would be filled in with the URL to the Service/Channel defined in my "sink" section (ie. --sink=http://message-dumper.default.svc.cluster.local)

## Actual Behavior
If providing anything other than "--sink" exactly (ie. "--sink=", "--sink=placeholder", etc), the value of my sink is not filled in. I believe the docs could be a bit clearer at letting the reader know that the argument must be exactly "--sink", assuming this is the intended behavior.

1. Running eventing v0.2.0
2. Create a ContainerSource event source with a sink argument of anything other than "--sink"
3. Sink is not automatically populated

BrianMMcClain:
Fixes #600

## Proposed Changes

Assuming this is intended behavior,  I wanted to clarify the --sink argument for a ContainerSource a bit after getting tripped up a bit on it myself, where the argument must be exactly "--sink"

### DIFF
--- a/eventing/README.md
+++ b/eventing/README.md
@@ -197,8 +197,8 @@ FTP server for new files or generate events at a set time interval.
 **Spec fields**:
 
 - `image` (**required**): `string` A docker image of the container to be run.
-- `args`: `[]string` Command-line arguments. Any `--sink=` argument will be
-  filled in with the DNS address of the `sink` object.
+- `args`: `[]string` Command-line arguments. If no `--sink` flag is provided, 
+   one will be added and filled in with the DNS address of the `sink` object.
 - `env`: `map[string]string` Environment variables to be set in the container.
 - `serviceAccountName`: `string` The name of the ServiceAccount to run the
   container as.


RichieEscarez:
@evankanderson  help please?

I see "sink=" [here](https://github.com/knative/eventing-sources/blob/b2b4ee45c9227c2dc4e8ed7aee405764e0058595/pkg/controller/githubsource/resources/service.go#L48
) so im not sure if the statement needs to be clarified or if it indeed should change to just "--sink"

evankanderson:
I can take a look tomorrow or Thursday, or @Harwayne or @vaikas-google might be able to take a look sooner (the original author, @n3wscott is on leave until January).

BrianMMcClain:
So, super sorry if I'm completely misunderstanding everything, but digging into this a bit more, I believe the ContainerSource is written that so long as "--sink=" doesn't exist in any of the args, it adds a "--sink=<sink url>" entry to the list of args?

If I compare what I provide in my source YAML args array vs what the end result is when I run "kubectl describe pod", that seems to be what the pattern suggests:

**Provided**: ["--interval=1"]
Pod Description:
```
Args:
      --interval=1
      --sink=http://message-dumper-3.default.svc.cluster.local/
```


**Provided**:  ["--interval=1", "--sink"]
Pod Description:
```
    Args:
      --interval=1
      --sink
      --sink=http://message-dumper-3.default.svc.cluster.local/
```


**Provided**: ["--interval=1", "--sink="]
Pod Description: 
```
    Args:
      --interval=1
      --sink=
```

**Provided**: ["--interval=1", "--sink=placeholder"]
Pod Description: 
```
    Args:
      --interval=1
      --sink=placeholder
```


Also, if I'm understanding the code [here](https://github.com/knative/eventing-sources/blob/eb810f051a4562bf39973362dc24e20c36544822/pkg/controller/containersource/reconcile.go#L137) and [here](https://github.com/knative/eventing-sources/blob/master/pkg/controller/containersource/resources/deployment.go#L35) correctly, it seems to agree, though my Go is very rusty 😄. In which case, it seems like the docs should read something along the lines of "if NO --sink flag is provided, one will be appended with the sink URL". The reason for the confusion seems to have been that providing "--sink" in my source YAML didn't match the string compare to "--sink=", so Knative added a second --sink flag, and the source I wrote chose the second one when parsing the CLI flags. Also the omission of the --sink flag in the [heartbeat sample](https://github.com/knative/eventing-sources/blob/master/samples/sources_v1alpha1_containersources.yaml) seem to be in line with this.

If this all sounds right, I'll be happy to update the PR to accurately represent this. If I'm way off, I'm very sorry for the misunderstanding!

BrianMMcClain:
The more I looked into it, it does seem to be the case that the --sink flag is only provided and filled in with the URL to the Sink object only if "--sink=" is omitted, otherwise it is not overwritten. I've gone ahead and updated the PR to clarify this and cleaned up the commits a bit. Sorry if this got a bit sloppy!

PR #606, opened 2018-11-28
Add more content about granting CloudDNS permission to GKE cluster

ZhiminXiang:
Fixes #(issue-number)

## Proposed Changes
1. rename using-external-dns to using-external-dns-on-gcp since this doc is GCP specific.
2. Add the content of granting cluster the CloudDNS permission by usingservice account credential
3. Directly paste the manifest of ExternalDNS into the doc instead of asking users to find out from ExternalDNS site.


ZhiminXiang:
Looks like I should use "git mv" so that the diff btw old file and new file could be showed.
I will close this PR, and create a new one.

PR #608, opened 2018-11-29
Update some docs for installation of knative on GKE

vagababov:
- more bash annotations
- make project also a variable
- rearrange and re-word some docs about cluster name and zone

Fixes #(issue-number)

## Proposed Changes

* 
* 
*


install/Knative-with-GKE.md
```diff
@@ -59,7 +61,8 @@ You need a GCP project to create a Google Kubernetes Engine cluster.
    existing GCP project as your `gcloud` default:
     * If you don't already have a GCP project created, create a new project in `gcloud`:
       ```bash
-      gcloud projects create my-knative-project --set-as-default
+      export PROJECT=my-knative-project
```
averikitsch: This would be better set above this step, so it can apply to the "already have a project" step.

vagababov: hm, I had a few versions yesterday, but of course you're right.

install/Knative-with-GKE.md
```diff
@@ -40,16 +40,18 @@ commands will need to be adjusted for use in a Windows environment.
 To simplify the command lines for this walkthrough, we need to define a few
 environment variables.
 
-Set `CLUSTER_NAME` and `CLUSTER_ZONE` variables:
+Set `CLUSTER_NAME` and `CLUSTER_ZONE` variables, you can replace `knative` and
+`us-west1-c` with cluster name and zone of your choosing.
+
+The `CLUSTER_NAME` needs to be lowercase and unique among any other Kubernetes
+clusters in your GCP project. The zone can be
```
averikitsch: We shouldn't use acronyms, please use full name for GCP.

vagababov: it has been named GCP across the file before. Do you want me to go and update all of those?

vagababov:
Ptal.

install/Knative-with-GKE.md
averikitsch: Can you add "(GCP)" to the Google Cloud Platform in the before you begin section and use the full name at the beginning of Setting up a Google Cloud Platform project section. Thanks!

vagababov: Done in two places.

PR #611, opened 2018-11-30
Run goimports

mattmoor-sockpuppet:
Produced via: `goimports -w $(find -name '*.go' | grep -v vendor)`


mattmoor:
/hold

I'm pretty sure this is broken...

PR #634, opened 2018-12-05
samples/autoscale-go: simplify instructions

### LINKED ISSUE #633: autoscale-go sample should be published on gcr
When I read https://github.com/knative/docs/tree/098b76e0823a7eacbb54163031c7272617b258a3/serving/samples/autoscale-go there's very little value in forcing people to

- go get the repo
- build image from repo root (uploads the entire repo as docker build context to docker engine)
- enable gcr api
- run: gcloud auth configure-docker
- push to their repo
- edit service.yaml

I'll go ahead and remove those instructions and provide a prepublished image on `gcr.io/knative-samples/`

/cc @josephburnett

ahmetb:
This patch simplifies the autoscaling tutorial by eliminating steps to:
- go-get the repo
- set a $REPO
- enable GCR API
- show how to authenticate to GCR via docker client
- push image
- update service.yaml

Instead we just do:
- deploy service.yaml that has a pre-built public image

I just published this image to `gcr.io/knative-samples/autoscale-go:0.1` and I
have eliminated excessive Dockerfile instructions that cause the entire docs/
repo to be sent to docker-engine as part of the "docker build".

Verified manually that it works. 👍 

Fixes #633.

### DIFF
--- a/serving/samples/autoscale-go/Dockerfile
+++ b/serving/samples/autoscale-go/Dockerfile
@@ -14,14 +14,11 @@
 
 FROM golang AS builder
 
-WORKDIR /go/src/github.com/knative/docs/
-ADD . /go/src/github.com/knative/docs/
-
-RUN CGO_ENABLED=0 go build ./serving/samples/autoscale-go/
+WORKDIR /go/src/app
+COPY . /go/src/app
+RUN CGO_ENABLED=0 go build -o /bin/autoscale-go .
 
 FROM gcr.io/distroless/base
-
 EXPOSE 8080
-COPY --from=builder /go/src/github.com/knative/docs/autoscale-go /sample
-
+COPY --from=builder /bin/autoscale-go /sample
 ENTRYPOINT ["/sample"]
--- a/serving/samples/autoscale-go/Dockerfile
+++ b/serving/samples/autoscale-go/Dockerfile
@@ -17,8 +17,10 @@ FROM golang AS builder
 WORKDIR /go/src/app
 COPY . /go/src/app
 RUN CGO_ENABLED=0 go build -o /bin/autoscale-go .
+RUN CGO_ENABLED=0 go build -o /bin/load-generator ./test
 
 FROM gcr.io/distroless/base
 EXPOSE 8080
 COPY --from=builder /bin/autoscale-go /sample
+COPY --from=builder /bin/load-generator /load-generator
 ENTRYPOINT ["/sample"]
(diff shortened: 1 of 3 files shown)


ahmetb:
/hold
I'm realizing we actually use `test/test.go` as the load generator.

I'll probably refactor the image to include a /bin/load-generator that's compiled from `./test/test.go` and invoke it with `docker`.

ahmetb:
/hold cancel 

I figured I can just do

```
docker run --rm -i -t --entrypoint /load-generator -e IP_ADDRESS="${IP_ADDRESS}" \
     gcr.io/knative-samples/autoscale-go:0.1 \
     -args ...
```

serving/samples/autoscale-go/Dockerfile
```diff
@@ -14,14 +14,13 @@
 
 FROM golang AS builder
 
-WORKDIR /go/src/github.com/knative/docs/
-ADD . /go/src/github.com/knative/docs/
-
-RUN CGO_ENABLED=0 go build ./serving/samples/autoscale-go/
+WORKDIR /go/src/app
```
josephburnett: If you're not going to put the source into a golang workspace, why not just put it in root?  E.g. `WORKDIR /app`.

serving/samples/autoscale-go/README.md
```diff
@@ -89,7 +50,9 @@ Build the application container and publish it to a container registry:
 1. Ramp up traffic to maintain 10 in-flight requests.
 
    ```
-   go run serving/samples/autoscale-go/test/test.go -sleep 100 -prime 10000 -bloat 5 -qps 9999 -concurrency 300
+   docker run --rm -i -t --entrypoint /load-generator -e IP_ADDRESS="${IP_ADDRESS}" \
```
josephburnett: What's the advantage / simplification of using `docker run` over `go run`?  Now I cannot run this example on my Pixelbook running Crostini because I cannot run containers locally.

I rather like being able to build, run and test things entirely in user-space.  E.g. using `ko` go build instead of `docker`.

serving/samples/autoscale-go/Dockerfile
ahmetb: That's fair point. Just a convention to put into `/go/src` so folks can do `go install` within gopath. But I guess that has changed with go1.11. I think both are fien.

serving/samples/autoscale-go/README.md
ahmetb: It's likely more people will have docker than go in this audience.

I just wanted to shorten the tutorial. We can probably make the docker command an `alias` and maybe use that here.

josephburnett:
Thanks for answering the docker questions.  Looks good.

/lgtm
/approve

PR #635, opened 2018-12-06
Fix knative minikube installation doc

surajnarwade:
* Added `--disk-size=30g` since it was breaking
* Removed `--bootstrapper=kubeadm` because it is default

### DIFF
--- a/install/Knative-with-Minikube.md
+++ b/install/Knative-with-Minikube.md
@@ -36,7 +36,7 @@ For Linux use:
 minikube start --memory=8192 --cpus=4 \
   --kubernetes-version=v1.11.3 \
   --vm-driver=kvm2 \
-  --bootstrapper=kubeadm \
+  --disk-size=30g \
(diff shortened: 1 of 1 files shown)


install/Knative-with-Minikube.md
```diff
@@ -36,7 +36,7 @@ For Linux use:
 minikube start --memory=8192 --cpus=4 \
   --kubernetes-version=v1.11.3 \
   --vm-driver=kvm2 \
-  --bootstrapper=kubeadm \
+  --disk-size=30g \
```
gyliu513: How did you come to the number of 30g? What about setting an env variable for this?
```
export DISK_SIZE=30G
```
```
--disk-size=$DISK_SIZE
```

surajnarwade: I tried with default minikube values for disc size, it broke very badly, so I extended the value by 10 more and it worked, though minishift guide suggests 50g

PR #637, opened 2018-12-06
Update kubectl apply URLs to use release artifacts

grantr:
A few things I noticed while going through the install procedure.

## Proposed Changes

- Use the istio release artifact in serving (https://github.com/knative/serving/releases/download/v0.2.2/istio.yaml) to install Istio instead of a direct link to the file in the repo.
- Same as above for build-only.
- Install serving v0.2.2 instead of v0.2.1.
- Point users to the top-level install instructions from the eventing readme.


samodell:
Thanks, Grant!
/lgtm
/approve

PR #644, opened 2018-12-07
serving/samples: remove EXPOSE directive from helloworld

grayside:
Remove EXPOSE directive from all helloworld serving samples.

@steren indicated this was not required, so in the interest of simplicity lets remove it.


serving/samples/helloworld-dart/README.md
```diff
@@ -78,7 +78,6 @@ be created using the following instructions.
 
    # Configure and document the service HTTP port.
    ENV PORT 8080
```
steren: You should remove PORT, this is injected by Knative.

steren: (true for all samples)

grayside: I prefer to keep the explicit default as a way of documenting the contract and making it obvious how to override the port for experimentation.

steren: 8080 is not part of the runtime contract, only PORT is: https://github.com/knative/serving/blob/master/docs/runtime-contract.md#process

grayside: Explicitly showing this remains valuable, even if the value is a common convention rather than an official default.

grayside: Explicitly showing this remains valuable, even if the value is a common convention rather than an official default.

grayside: Just to make sure we have the same definition here, the ENV directive is a way to set a default environment variable value at build time which is overridden at run time by any mechanism which injects environment variables into the container.

I see including this line as a way of:

1. Having sensible defaults to ensure the container will work locally.
2. Documenting what the PORT env var is about for practitioners that read the Dockerfile but not the docs.

steren: My concerns is that users will assume these are the default values of `PORT`.

As far as I understand, the `ENV` directive is not needed for running locally when using `docker -p`.

Running locally is better done using `$ PORT=8080 docker run -p 8080:${PORT} -e PORT=${PORT} [IMAGE]`

grayside: This is setting a default value for the service, which is a good practice. It also allows the container to be spun up with a more typical and abbreviated command:

```
docker run -p 8080 [IMAGE]
```

A good amount of the value in this is if developers take the Dockerfile and pivot to using this as part of an ongoing service.

The Day 1 value is in surfacing a bit of the runtime contract for developers that look at code first. What if a comment were added indicating that 8080 is not a requirement but a sane default?

steren: OK to leave `ENV PORT 8080` in the Dockerfiles, but indeed please change the comment, "Configure the service HTTP port." is not correct. The service HTTP port is not configured via this env var. Use something like "Service must listen on $PORT, set value for local development"

serving/samples/helloworld-clojure/Dockerfile
```diff
@@ -13,9 +13,8 @@ COPY . .
 # Build an uberjar release artifact.
 RUN mv "$(lein uberjar | sed -n 's/^Created \(.*standalone\.jar\)/\1/p')" app-standalone.jar
 
-# Configure and document the service HTTP port.
+# Configure the service HTTP port.
```
averikitsch:
```suggestion
# Set the default value for PORT environment variable.
```

samodell: Should this change be consistent across all the samples?

averikitsch: Correct, just trying to figure out the phrasing.

serving/samples/helloworld-clojure/README.md
```diff
@@ -75,9 +75,8 @@ recreate the source files from this folder.
    # Build an uberjar release artifact.
    RUN mv "$(lein uberjar | sed -n 's/^Created \(.*standalone\.jar\)/\1/p')" app-standalone.jar
 
-   # Configure and document the service HTTP port.
+   # Configure the service HTTP port.
```
averikitsch: Can you update the ReadMes as well?

grayside: Working on it. This seems like a good spot for automation.

samodell:
It looks like our new build tests are a bit over-sensitive. @adrcunha 
1. Can we customize the checker so that it ignores links that point to http://localhost... ?
2. I'm less sure why www.dartlang.org isn't recognized as a valid link. Does the checker require "http://"?

adrcunha:
Probably the checker doesn't like an address without protocol, and I'm pretty sure localhost:4000 is not returning content at the time of checking. :)

adrcunha:
Actually the checker is correct: you must add the protocol to the darlang.org link, otherwise it will be treated as a "local" file. Just try clicking it in the README file and you'll see a nice 404 GitHub page.

samodell:
It looks like it's failing because of the merge conflicts. That test should pass once those are cleared up.

samodell:
@grayside  Ping to resolve merge conflicts. This should be mergeable after that.

PR #647, opened 2018-12-09
Added cronjob sample

whynowy:
## Proposed Changes

- Give a sample about how to configure CronJobSource event.


eventing/samples/cronjob-source/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Deployment Steps
+
+### Prerequisites
+
+1. Setup [Knative Serving](https://github.com/knative/docs/tree/master/serving).
+1. Setup
+   [Knative Eventing](https://github.com/knative/docs/tree/master/eventing).
+
+### Channel
```
evankanderson: Do we need to use a Channel here? Feedback on the GitHub source suggested that pointing directly at a Knative service was easier for simple cases: https://github.com/knative/docs/tree/master/eventing/samples/github-source#create-event-source-for-github-events

eventing/samples/cronjob-source/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. Create a `Channel`. You can use your own `Channel` or use the provided sample, which creates a channel called `cronjob-test`. If you use your own `Channel` with a different name, then you will need to alter other commands later.
+
+```shell
+kubectl apply -f channel.yaml
+```
+
+### Create Cron Job Event Source
+
+1. In order to receive events, you need to create a concrete Event Source for a specific namespace. If you want to consume events from a differenet namespace or using a different `Service Account`, you need to modify the yaml accordingly.
```
evankanderson:
```suggestion
1. For each set of cron events you want to request, you need to create a Event Source in a the same namespace as the destiantion. If you need a different ServiceAccount to create the Deployment, modify the entry in `cronjob-source.yaml`.
```

evankanderson: I think the tech writing team has also been encouraging actually including the yaml content inline for short and medium samples. The content reads better that way, instead of forcing users to bounce between different pages.

eventing/samples/cronjob-source/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+```shell
+kubectl apply -f cronjob-source.yaml
+```
+
+### Subscriber
+
+In order to check the `CronJobSource` is fully working, we will create a simple Knative Service that dumps incoming messages to its log and create a `Subscription` from the `Channel` to that Knative Service.
```
evankanderson: You should be able to directly create the Service and then point the Source at it (and, if you follow the GitHub case, you might actually deploy the Service first).

eventing/samples/cronjob-source/channel.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: eventing.knative.dev/v1alpha1
+kind: Channel
+metadata:
+  name: cronjob-test
+  namespace: default
```
evankanderson: I recommend omitting the namespace so that it can be picked up from the current `kubectl` context.

eventing/samples/cronjob-source/cronjob-source.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: sources.eventing.knative.dev/v1alpha1
+kind: CronJobSource
+metadata:
+  name: test-cronjob-source
```
evankanderson: (especially if you don't specify the namespace here on the source)

eventing/samples/cronjob-source/README.md
```diff
@@ -0,0 +1,58 @@
+# Cron Job Source example
+
+Cron Job Source example shows how to configure Cron Job as event source for functions.
```
evankanderson: I recommend using http://prettier.io/ to wrap this automatically with proseWrap=always (it makes it easier to make comments at the sub-paragraph level).

eventing/samples/cronjob-source/README.md
whynowy: Updated to point to service directly.

whynowy:
@evankanderson - can you please review again?

PR #670, opened 2018-12-19
Update instructions to access Prometheus UI

alculquicondor:
/cc @mdemirhan


mdemirhan:
This doesn't look right. The URL there should point to https://github.com/knative/docs/blob/a71f1b9ab42a889d6d48e265455060364ca1a885/serving/samples/telemetry-go/prometheus-test.yaml#L99

The current documentation is correct.

PR #685, opened 2018-12-27
Add annotations to autoscale-go sample.

josephburnett:
## Proposed Changes

- Add annotations to autoscale-go sample to demonstrate CPU-based autoscaling, setting custom targets and scale bounds.
- Use `hey` for load generation instead of custom tool.
- Link to Kubecon demo of autoscaler customization.

### DIFF
--- a/serving/samples/autoscale-go/README.md
+++ b/serving/samples/autoscale-go/README.md
@@ -31,7 +31,7 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
    export IP_ADDRESS=`kubectl get svc knative-ingressgateway --namespace istio-system --output jsonpath="{.status.loadBalancer.ingress[*].ip}"`
    ```
 
-## Send Load to the Service
+## Load the Service
 
 1. Make a request to the autoscale app to see it consume some resources.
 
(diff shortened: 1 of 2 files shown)


serving/samples/autoscale-go/README.md
```diff
@@ -46,57 +45,158 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
...
+     req write:    0.0001 secs, 0.0000 secs, 0.0045 secs
+     resp wait:    0.2766 secs, 0.1669 secs, 0.6633 secs
+     resp read:    0.0002 secs, 0.0000 secs, 0.0065 secs
+
+   Status code distribution:
+     [200] 5424 responses
+
+
+
+   NAME                                             READY   STATUS    RESTARTS   AGE
```
dgerd: Is this part really output by `hey`?

serving/samples/autoscale-go/README.md
```diff
@@ -46,57 +45,158 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
...
-   > Note: Use CTRL+C to exit watch mode.
-
 ## Analysis
 
 ### Algorithm
 
 Knative Serving autoscaling is based on the average number of in-flight requests
 per pod (concurrency). The system has a default
-[target concurrency of 100.0](https://github.com/knative/serving/blob/3f00c39e289ed4bfb84019131651c2e4ea660ab5/config/config-autoscaler.yaml#L35).
+[target concurrency of 100.0](https://github.com/knative/serving/blob/3f00c39e289ed4bfb84019131651c2e4ea660ab5/config/config-autoscaler.yaml#L35) but [we used 10](service.yaml#L26) for our service.  We loaded the service with 50 concurrent requests so the autoscaler created 5 pods (`50 concurrent requests / target of 10 = 5 pods`)
```
dgerd: nit: might be worth updating the first link to a relative link for consistency.

serving/samples/autoscale-go/README.md
```diff
@@ -112,43 +212,43 @@ kubectl port-forward --namespace knative-monitoring $(kubectl get pods --namespa
 
 ### Other Experiments
 
-1. Maintain 1000 concurrent requests.
+1. Maintain 100 concurrent requests.
```
dgerd: This says maintain still, but the command below runs for 60 second while the previous command does not look to have a timeout.

Same for the two below.

serving/samples/autoscale-go/README.md
josephburnett: This is the `&& kubectl get pods` part so you can see the effects of the load test.

serving/samples/autoscale-go/README.md
josephburnett: Is that the right thing to do between repos?  A relative link would be good within the same repo because it would stay within the same branch etc...

serving/samples/autoscale-go/README.md
dgerd: Ahh. Didn't notice that at the bottom.

serving/samples/autoscale-go/README.md
josephburnett: Good point.  I've updated to comments to reflect the limited test duration of 60 seconds.

serving/samples/autoscale-go/README.md
dgerd: Ah you are right. Thanks.

PR #692, opened 2019-01-02
Format markdown

mattmoor-sockpuppet:


serving/samples/autoscale-go/README.md
```diff
@@ -77,36 +77,25 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
      0.628 [14]    |
      0.686 [21]    |
      0.743 [13]    |
+   ```
 
+Latency distribution: 10% in 0.1805 secs 25% in 0.2197 secs 50% in 0.2801 secs
```
mattmoor: cc @evankanderson 

This is a strange bug in `prettier.io`.

cc @josephburnett Any interest in restructuring this so that the preformatting uses <pre>```</pre> instead of spacing?

PR #697, opened 2019-01-03
Improve Docker for Mac installation guide

### LINKED ISSUE #696: Docker for Mac installation guide has a few errors
+ It uses `knative v0.2.1` whereas minikube installation guide uses `v0.2.2`
+ The url to the istio manifests is not the same as in the minikube installation guide
    + For minikube, we are using `https://github.com/knative/serving/releases/download/v0.2.2/istio.yaml`
    + For Docker for Mac we are using `https://raw.githubusercontent.com/knative/serving/v0.2.1/third_party/istio-1.0.2/istio.yaml`
+ The instructions to install knative serving components are missing the `kubectl apply` part

dgageot:
## Proposed Changes

 + Use knative v0.2.2 like on minikube
 + Fix the instructions to install knative serving

Signed-off-by: David Gageot <david@gageot.net>

Fixes #696

### DIFF
--- a/install/Knative-with-Docker-for-Mac.md
+++ b/install/Knative-with-Docker-for-Mac.md
@@ -32,7 +32,7 @@ Knative depends on Istio. Run the following to install Istio. (This changes
 `LoadBalancer` to `NodePort` for the `istio-ingress` service).
 
 ```shell
-curl -L https://raw.githubusercontent.com/knative/serving/v0.2.1/third_party/istio-1.0.2/istio.yaml \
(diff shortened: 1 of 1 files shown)


Justin2997:
Maybe we should add a line with a command to validate that your Kubernetes cluster is at the good version? Docker for mac does not make it obvious with the version it is. Maybe just hire : https://github.com/dgageot/docs-1/blame/0ade4ea2384df5dbab566c19bc6cfb7029f16d46/install/Knative-with-Docker-for-Mac.md#L12

PR #698, opened 2019-01-03
Format markdown

mattmoor-sockpuppet:


samodell:
@mattmoor This seems a bit odd -- is prettier working OK?

mattmoor:
@samodell No, it is not...  I commented on a prior PR that it seems to be struggling with the indent-based preformatting.  We need a human to fix that, and it looks like @josephburnett is out.

PR #699, opened 2019-01-03
serving/samples/helloworld-php: use development php.ini

### LINKED ISSUE #695: serving/samples/helloworld-php: not using a recommended PHP configuration
## Expected Behavior

From the [PHP docker image documentation](https://hub.docker.com/_/php), any extending image is expected to use one of the following:

RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
RUN mv "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"

For flexibility, I would prefer this to be selected at runtime via an environment variable and a custom entrypoint.

## Actual Behavior

We are using the PHP default settings which are not well-suited to either development or production.

## Priority

The existing Dockerfile has the benefit of some simplicity. However, as a starting place for developers to build PHP services on knative it is inadequate. This requires a judgement call on the distinction between a quickstart and a minimal boilerplate.

grayside:
Fixes #695

## Proposed Changes

- Use the development php.ini shipped in the image but not "activated"
- Add comment pointing out the use of the production configuration
- Add comment linking to the documentation on the image README

### DIFF
--- a/serving/samples/helloworld-php/Dockerfile
+++ b/serving/samples/helloworld-php/Dockerfile
@@ -8,6 +8,12 @@ COPY index.php /var/www/html/
 # Use the PORT environment variable in Apache configuration files.
 RUN sed -i 's/80/${PORT}/g' /etc/apache2/sites-available/000-default.conf /etc/apache2/ports.conf
 
+# Configure PHP for development.
+# Switch to the production php.ini for production operations.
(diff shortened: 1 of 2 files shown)


averikitsch:
Can you add this to the ReadMe as well?

PR #701, opened 2019-01-03
Event Sources List

Harwayne:
## Proposed Changes

- Create a machine readable list of known event sources for Knative.
- Use the machine readable list to generate a human readable markdown document.


eventing/sources/Sources.md
```diff
@@ -0,0 +1,42 @@
+# Knative Event Sources
```
vaikas-google: Should this file also have the comment that this file should not be manually edited and one should edit the YAML instead? You know for people that don't always remember to read readmes (like me :) )

evankanderson: +1

Harwayne: Probably. I was trying to do it via a comment in the markdown, but can't find a good way to do so. Should I just write it so that it is visible in the rendered page as well?

eventing/sources/README.md
```diff
@@ -0,0 +1,17 @@
+# Event Sources
```
evankanderson: I would switch `Sources.md` and `README.md` such that the tool should generate `README.md` (possibly with the updating information in a comment) and then a `maintenance.md` file or the like describing how to update `README.md`.

eventing/sources/Sources.md
```diff
@@ -0,0 +1,42 @@
...
+
+This is a non-exhaustive list of Event sources for Knative. Inclusion in this list **does not**
```
evankanderson: I feel like we need at least one paragraph explaining why we have this.

> Event Sources are Kubernetes Custom Resources which provide a mechanism for registering interest in a class of events from a particular software system. Since different event sources may be described by different Custom Resources, this page provides an index of the available source resource types as well as links to installation instructions.
>
> This is a non-exhaustive...

eventing/sources/Sources.md
```diff
@@ -0,0 +1,42 @@
...
+imply an endorsement, nor does it imply any level of support.
+
+
+## Sources
+
+These are sources that are installed as `CRD`s.
+
+Name | Status | Support | Description
```
evankanderson: Should we include authorship here?

eventing/sources/Sources.md
```diff
@@ -0,0 +1,42 @@
...
+[GitHub](https://github.com/knative/eventing-sources/blob/master/pkg/apis/sources/v1alpha1/githubsource_types.go) | Proof of Concept | None | Registers for events of the specified types on the specified GitHub organization/repository. Brings those events into Knative.
+[GitLab](https://gitlab.com/triggermesh/gitlabsource) | Proof of Concept | None | Registers for events of the specified types on the specified GitLab repository. Brings those events into Knative.
+[Kubernetes](https://github.com/knative/eventing-sources/blob/master/pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go) | Active Development | Knative | Brings Kubernetes cluster events into Knative. Uses ContainerSource for underlying infrastructure.
+[Google Cloud Scheduler](https://github.com/vaikas-google/csr) | Active Development | None | Create, update, and delete [Google Cloud Scheduler](https://cloud.google.com/scheduler/) Jobs. When those jobs are triggered, receive the event inside Knative.
+
+
+
+## Meta Sources
+
+These are not directly usable, but make writing a Source much easier.
```
evankanderson: These feel like they should be on a separate page for "source authors".

eventing/sources/Sources.md
```diff
@@ -0,0 +1,42 @@
...
+
+Name | Status | Support | Description
+--- | --- | --- | ---
+[Container Source](https://github.com/knative/eventing-sources/blob/master/pkg/apis/sources/v1alpha1/containersource_types.go) | Active Development | Knative | Container Source is a generic controller. Given an Image URL, it will keep a single `Pod` running with the specified image, environment, and arguments. It is used by multiple other Sources as underlying infrastructure.
+[Auto Container Source](https://github.com/Harwayne/auto-container-source) | Proof of Concept | None | AutoContainerSource is a controller that allows the Source CRDs _without_ needing a controller. It notices CRDs with a specific label and starts controlling resources of that type. It utilizes Container Source as underlying infrastructure.
```
evankanderson: Do we have any sources (see above) written as *Auto Container Sources*?

eventing/sources/generator/SourcesTemplate.gomd
```diff
@@ -0,0 +1,39 @@
...
+These are sources that are installed as `CRD`s.
+
+Name | Status | Support | Description
+--- | --- | --- | ---
+{{ range .Sources -}}
+[{{ .Name }}]({{ .Url }}) | {{ .Status }} | {{ .Support }} | {{ .Description }}
+{{- end }}
+
+
+## Meta Sources
```
evankanderson: Not sure whether we need to generate these, since we shouldn't end up with a new meta source for each new source. 😬

eventing/sources/Sources.md
Harwayne: Done.

eventing/sources/Sources.md
```diff
@@ -0,0 +1,44 @@
+# Knative Event Sources
+
+This is a non-exhaustive list of Event sources for Knative. Inclusion in this list **does not**
+imply an endorsement, nor does it imply any level of support.
+
+This is a generated file and should not be changed manually. All changes should follow the procedure
```
evankanderson: You should be able to add a comment in `<!-- this markdown -->`

eventing/sources/Sources.md
vaikas-google: I think that would be nice, but in practice what would it mean for some of our existing ones? Knative authors? I think the URL should point to something that should have that context I think?

Harwayne: I agree with @vaikas-google, I don't know how to describe authorship in a better way than who owns the linked repo. Eventually we can use a distinct table for Knative supported sources.

eventing/sources/generator/SourcesTemplate.gomd
Harwayne: > since we shouldn't end up with a new meta source for each new source.

I think I agree with you, there should be far fewer meta sources than sources. MetaSources are more useful for source authors, not end users that just want to find a working source.

As for listing these, I think it is useful to have a place where we list all these meta sources. It doesn't have to be in the same place as all the end user usable sources, but for now I don't think it is problematic to have it here, especially as `Containers` refer to these by name.

eventing/sources/Sources.md
Harwayne: As for listing these, I think it is useful to have a place where we list all these meta sources. It doesn't have to be in the same place as all the end user usable sources, but for now I don't think it is problematic to have it here, especially as `Containers` refer to these by name.

eventing/sources/Sources.md
Harwayne: Added to the Containers section with a note in their description.

eventing/sources/Sources.md
Harwayne: Done.

eventing/sources/README.md
Harwayne: Done.

eventing/sources/Sources.md
Harwayne: Left it visible at the top of the document. Do you have a preference for it being a comment?

Harwayne: Made it a comment.

PR #702, opened 2019-01-03
Run `dep ensure`.

Harwayne:
## Proposed Changes

- Run `dep ensure`.


Harwayne:
Thank you for the explanation. This PR is totally unneeded as `hack/update-deps.sh` removes the files this PR was trying to add.

PR #703, opened 2019-01-04
fix(README): tiny cosmetic formatting fix

rhuss:
## Proposed Changes

- align CRD formatting with other links


samodell:
This also opens the larger question of consistency-- if you follow the links to these pages, they don't always put these words in ``, even across the same page. I created issue #707 so we can track this. For now, I do think merging this eliminates how bad this list looks with only one element not in code formatting, so I'll go ahead and merge this in.

/ok-to-test
/approve
/lgtm

PR #708, opened 2019-01-05
docs: add AWS SQS source to README

srvaroa:


eventing/README.md
```diff
@@ -191,6 +191,21 @@ The GcpPubSubSource fires a new event each time a message is published on a
...
+
+**Spec fields**:
+
+- `queueURL`: URL of the SQS queue to pull events from.
+- `awsCredsSecret`: credential to use to poll the AWS SQS queue.
+- `sink`:
+  [ObjectReference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#objectreference-v1-core)
+  A reference to the object that should receive events.
+- `serviceAccountName`: `string` The name of the ServiceAccount used to
+  access the `awsCredsSecret`.
```
vaikas-google: Might want to clarify that the Receive Adapter runs as this account. So, the effect is that we access the secret as this Service Account, but it's because the Receive Adapter runs as this Service Account.

Maybe add something like this (not verbatim, but shrug)

	// serviceAccountName is the name of the ServiceAccount that will be used to
	// run the Receive Adapter Deployment.

evankanderson: This seems to parallel line 184 for the GcpPubSubSource.

vaikas-google:
@evankanderson re: automating the parameter information, I've been playing with using the OpenAPI validation that's part of the CRDs with both Google Cloud Scheduler / Google Cloud Storage, and I think it's pretty good start.

PR #714, opened 2019-01-07
api-reference: Add API reference docs

### LINKED ISSUE #636: Proposal: start generating authoritative API reference docs
Currently, we don't have API reference documentation. 

For serving API we have [spec.md](https://github.com/knative/serving/blob/220eaf0ddabb2b4cc7a8ecc010ef71d94857e2ba/docs/spec/spec.md#resource-yaml-definitions) which I've been able to find multiple errors in it, and it's a manual maintenance.

For build API, we also have a hand-written informal syntax at [builds.md](https://github.com/knative/docs/blob/098b76e0823a7eacbb54163031c7272617b258a3/build/builds.md#syntax).

----

- Kubernetes [API reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/) docs generated with [gen-apidocs](https://github.com/kubernetes-incubator/reference-docs/tree/master/gen-apidocs) which uses `/swagger.json` or `/apis` paths exposed on Kubernetes API server

- Istio [API reference](https://istio.io/docs/reference/config/istio.networking.v1alpha3/) docs are generated with [protoc-gen-doc](https://github.com/pseudomuto/protoc-gen-doc) codegen plugin because Istio provides their api as protos

Knative doesn't fit either (we can use k8s /apis endpoint yet because it doesn't expose CRD OpenAPI specs).

**So I propose** we start generating these things from the source code (i.e pkg/apis at each repo) and import them to the knative/docs repo after a release happens.

I'll work on something right away if there are no objections.

/area API
/kind docs

ahmetb:
Generated with https://github.com/ahmetb/gen-crd-api-reference-docs and
this patch includes the script to self-serve this process. I hope to improve
this stuff in the future, for now it works fine.

Fixes #636.

### DIFF
--- a/hack/gen-api-reference-docs.sh
+++ b/hack/gen-api-reference-docs.sh
@@ -1,0 +1,169 @@
+#!/usr/bin/env bash
+#
+# This script is for generating API reference docs for Knative components.
+
+# Copyright 2018 Knative authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -euo pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+[[ -n "${DEBUG:-}" ]] && set -x
+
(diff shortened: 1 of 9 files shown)


RichieEscarez:
The script looks like it correctly targets the new "*reference*" directory but the PR currently uses **api-reference** as the file path for all files. Can you please rename (api-reference -> reference)?
I think its okay for the script/tool to reside elsewhere (if its coded to work that way)? I see that the other repos use a "*hack*" folder for their tools, maybe we follow that structure too? Otherwise if the script lives in "*reference*" that works for me too.

Thanks @ahmetb these files and the tool are great!

ahmetb:
I'll rename to reference.

I'll add the scripts to hack/.

ahmetb:
done. PTAL

RichieEscarez:
That was quick! This is such a great docs improvement, thanks again @ahmetb !

/lgtm

/approve

PR #723, opened 2019-01-08
Document default channels for users and operators

grantr:
[**Rendered**](https://github.com/grantr/docs/blob/default-channels/eventing/default-channels.md)

Features described were added in https://github.com/knative/eventing/pull/580.

Fixes https://github.com/knative/eventing/issues/632.

## Proposed Changes

- Add `default-channels.md` to eventing docs and link to it from `eventing/README.md`.

/cc @evankanderson 
for general documentation review

@Harwayne 
for accuracy review

### DIFF
--- a/eventing/README.md
+++ b/eventing/README.md
@@ -230,6 +230,9 @@ FTP server for new files or generate events at a set time interval.
 - [Install Eventing components](#installation)
 - [Run samples](samples/)
 
+## Configuration
+- [Default Channels](default-channels.md)
+
 ---
 
 Except as otherwise noted, the content of this page is licensed under the
--- a/eventing/README.md
+++ b/eventing/README.md
@@ -231,7 +231,7 @@ FTP server for new files or generate events at a set time interval.
 - [Run samples](samples/)
 
 ## Configuration
-- [Default Channels](default-channels.md)
+- [Default Channels](default-channels.md) provide a way to choose the persistence strategy for Channels across the cluster.
 
 ---
 
--- a/eventing/README.md
+++ b/eventing/README.md
@@ -93,6 +93,9 @@ This document will be updated as additional sources (which are custom resource
 definitions and an associated controller) and channels
 (ClusterChannelProvisioners and controllers) become available.
 
+Check out the [Configuration](#configuration) section to learn more about
+operating Knative Eventing.
+
 ## Architecture
 
 The eventing infrastructure supports two forms of event delivery at the moment:
@@ -231,7 +234,8 @@ FTP server for new files or generate events at a set time interval.
 - [Run samples](samples/)
 
 ## Configuration
-- [Default Channels](default-channels.md) provide a way to choose the persistence strategy for Channels across the cluster.
+- [Default Channels](default-channels.md) provide a way to choose the
+persistence strategy for Channels across the cluster.
 
 ---
 
--- a/eventing/default-channels.md
+++ b/eventing/default-channels.md
@@ -7,9 +7,8 @@ ConfigMap.
 
 ## Creating a default channel
 
-To create a default channel, use the same procedure as for creating a normal
-channel, but leave the `spec.provisioner` property blank. The `spec` property
-must be provided, but should be empty.
+To create a default channel, leave the `spec.provisioner` property blank. The
+`spec` property must be provided, but should be empty.
 
 _The content of `spec.arguments` will be cleared for default channels._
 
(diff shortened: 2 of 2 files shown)


eventing/README.md
```diff
@@ -230,6 +230,9 @@ FTP server for new files or generate events at a set time interval.
 - [Install Eventing components](#installation)
 - [Run samples](samples/)
 
+## Configuration
```
Harwayne: It might be worth linking to this section at the end of the install section on this page.

evankanderson: +1

eventing/README.md
```diff
@@ -230,6 +230,9 @@ FTP server for new files or generate events at a set time interval.
...
+- [Default Channels](default-channels.md)
```
evankanderson: Add a one-sentence explanation of what default channels are here, so people don't have to click the link to know if they want to read more? e.g.

```suggestion
- [Default Channels](default-channels.md) provide a way to choose the persistence strategy for Channels across the cluster.
```

eventing/default-channels.md
```diff
@@ -0,0 +1,129 @@
+# Default Channels
+
+The default channel configuration allows channels to be created without
+specifying a provisioner. This leaves the selection of channel provisioner and
+properties up to the operator. The operator controls the default settings via a
+ConfigMap.
+
+## Creating a default channel
+
+To create a default channel, use the same procedure as for creating a normal
```
evankanderson: This suggests that the normal mode of operation is to specify a `spec.provisioner`. Should we turn this around, and say that the normal mode is to not specify a provisioner, and to rely on this default?

eventing/default-channels.md
```diff
@@ -0,0 +1,129 @@
...
+channel, but leave the `spec.provisioner` property blank. The `spec` property
+must be provided, but should be empty.
```
evankanderson: Why does spec need to be provided, rather than left empty?

eventing/default-channels.md
```diff
@@ -0,0 +1,129 @@
...
+
+This is a valid default channel:
+
+```yaml
+apiVersion: eventing.knative.dev/v1alpha1
+kind: Channel
+metadata:
+  name: default-channel
+  namespace: default
+spec: {}
```
evankanderson: I think you can omit the curly brackets here and kubectl will do the right thing.

```suggestion
spec:
```

eventing/default-channels.md
```diff
@@ -0,0 +1,129 @@
...
+    apiversion: eventing.knative.dev/v1alpha1
+￼    kind: ClusterChannelProvisioner
+￼    name: default-provisioner
+```
+
+### Caveats
+
+#### Arguments cannot be specified by default channels
+
+If `spec.arguments` is set when creating a default channel, it will be cleared.
```
evankanderson: Do you want to explain why this is true, assuming that it is intentional rather than a bug?

eventing/default-channels.md
grantr: I think this is required due to the way go's json patch works. If the spec is omitted, it's set to nil instead of an empty map, but the patch assumes the map already exists. @Harwayne is this accurate?

eventing/README.md
grantr: Done.

eventing/default-channels.md
grantr: I just removed the clause about "use the same procedure...". Seems clearer. I agree that the normal mode should be to rely on the default.

eventing/default-channels.md
grantr: I get an error about `/spec/generation` doing that:
```
$ echo "apiVersion: eventing.knative.dev/v1alpha1
                                                      kind: Channel
                                                      metadata:
                                                        name: default-channel
                                                        namespace: default
                                                      spec:" | kubectl apply -f -
Error from server (InternalError): error when creating "STDIN": Internal error occurred: Internal error occurred: jsonpatch add operation does not apply: doc is missing path: /spec/generation
```

I think it's because kubectl is omitting the spec field when the brackets aren't included.

eventing/default-channels.md
grantr: Good point, this should be explained. IIUC it's a missing feature.

grantr:
@evankanderson Can you set the CLA label? Not sure if I'm allowed to assume you're ok with contributing. :laughing:

grantr:
(CLA status overridden because a review suggestion was accepted as a commit from the reviewer)

PR #736, opened 2019-01-10
Fix syntax of Knative install command

Freakin:
Fixes #N/A

## Proposed Changes
- Correct wrong slash used to separate commands
- Correct extraneous `--filename` flag


Freakin:
Signed CLA.

PR #738, opened 2019-01-10
Fix kubectl apply command for install on GKE

bradylove:
## Proposed Changes

This PR removes a duplicate `--filename` flag as well as replaces forward slashes with back slashes in the `kubectl apply` command for installing Knative.


samodell:
Thanks for your PR! PR #736  was opened a few hours earlier to fix that issue; so closing this one out.

Issue #740, opened 2019-01-10
Some install guides relied on -lite or -nomon releases, which no longer exist in 0.3.0


samodell:
The following guides have not been updated to install 0.3.0 versions of components, as they rely on -lite.yaml or -nomon.yaml releases that were not created for the 0.3.0 release.

- [ ] [Knative Install on Docker for Mac](Knative-with-Docker-for-Mac.md)
- [ ] [Knative Install on Minikube](Knative-with-Minikube.md)
- [ ] [Knative Install on Minishift](Knative-with-Minishift.md)
- [ ] [Knative Install on OpenShift](Knative-with-OpenShift.md)

PR #741, opened 2019-01-10
Update installs to 3.0 releases; fixing error in some install commands

samodell:
* Updates installs in most guides to 0.3.0 release. 

* Doesn't update guides that were using the release-lite.yamls to install Knative, since those release versions don't exist in 0.3.0. Created #740 to track this.

Some install commands had an extra --filename flag and were using incorrect slashes, same as #736 ; also fixes that issue

### DIFF
--- a/install/Knative-with-AKS.md
+++ b/install/Knative-with-AKS.md
@@ -132,8 +132,8 @@ Knative depends on Istio.
 
 1. Install Istio:
    ```bash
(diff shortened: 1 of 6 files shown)


bbrowning:
I manually verified the new install instructions for 0.3 on GKE and visually checked the other changes here. I did have to `kubectl apply` the serving and build yamls twice because of a known issue around defining and using CustomResourceDefinitions from a single yaml file. I'll separately ensure there's an issue filed to track that, as that change is unrelated to the version bumping here.

/lgtm

PR #734, opened 2019-01-09
Add real build/unit presubmit tests

adrcunha:
Addresses #66.
Addresses #67.

* Update test-infra to use the default build/unit test runners, which includes building go code and linting markdown files.
* Remove the integration tests, as they only start Knative Serving, which is irrelevant for this repo.

Bonuses:
* fix unbuildable `grpc-ping-go` sample
* remove test-infra import hack from `Gopack.lock` and `update-deps.sh`

PR #748, opened 2019-01-11
Format markdown

mattmoor-sockpuppet:


samodell:
@adrcunha  A lot of PRs are having their build tests fail. Is this a result of PR #734 ?

PR #751, opened 2019-01-12
Add hello world for Rust Nightly which uses Rocket

defcronyke:
[Rocket](https://rocket.rs) is nice because it makes the code smaller, and it 
makes URL route handling very elegant and easy. It currently
only works on [Rust Nightly](https://doc.rust-lang.org/1.5.0/book/nightly-rust.html).

This example also includes a best practice for speeding up
Rust Docker builds, and shows how to do a multi-stage build
so you can deploy the final application within a Docker
scratch container, leading to a final production container
that's only a couple of MB or less.

### DIFF
--- a/serving/samples/helloworld-rust-nightly/Dockerfile
+++ b/serving/samples/helloworld-rust-nightly/Dockerfile
@@ -1,0 +1,69 @@
+# ---------- Build Container ----------
+# This uses our prebuilt dependencies
+# container as a base to speed up builds.
+# It is used to build our app.
+FROM docker.io/{username}/helloworld-rust-nightly:deps AS build
+
+# Set working directory.
+WORKDIR /
+
+# Copy the source code.
+COPY ./src ./src
+
+# Create a user account to run the app with, 
+# remove files from the shell project,
+# build the app, and strip the binary to make
+# it smaller.
+RUN adduser --disabled-password --gecos "" appuser && \
+  rm target/x86_64-unknown-linux-musl/release/deps/helloworld_rust_nightly* && \
+  cargo build --release && \
+  strip target/x86_64-unknown-linux-musl/release/helloworld-rust-nightly
+# --------------------
+
+# ---------- Production Container ----------
+# Uses the smallest docker container as a base.
+# This container only holds what's strictly
+# needed to run our app.
+FROM scratch
+
+# Set working directory.
+WORKDIR /
+
+# Enable the user account we created in the other
+# container.
+COPY --from=build /etc/passwd /etc/
+
+# Use the ca-certs from the other container,
+# which we need if we want to be able to put
+# an https proxy in front of the service.
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+
+# Copy over the app we built in the other container.
+COPY --from=build target/x86_64-unknown-linux-musl/release/helloworld-rust-nightly ./
+
+# Copy the rocket config file.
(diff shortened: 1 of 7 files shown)


defcronyke:
I signed it!

serving/samples/helloworld-rust-nightly/Dockerfile
```diff
@@ -0,0 +1,69 @@
+# ---------- Build Container ----------
+# This uses our prebuilt dependencies
+# container as a base to speed up builds.
+# It is used to build our app.
+FROM docker.io/{username}/helloworld-rust-nightly:deps AS build
```
steren: Would you be able to use a publicly available image?

serving/samples/helloworld-rust-nightly/Dockerfile
```diff
@@ -0,0 +1,69 @@
+# ---------- Build Container ----------
```
steren: The goal of Hello World samples are to give simple examples. I am not sure we need a multi-stage Dockerfile for this use case.

serving/samples/helloworld-rust-nightly/Dockerfile
defcronyke: It is just the first image that the readme example has us build from the Dockerfile-deps file. So each user builds their own version before building the application container, and they can then choose when to update it.

serving/samples/helloworld-rust-nightly/Dockerfile
defcronyke: It is very simple, and also a best practice, and really more examples should be showing how to run code on scratch containers to keep resources and costs under control.

serving/samples/helloworld-rust-nightly/Dockerfile
steren: This adds a lot of complexity to the example IMO. Why cannot the sample use the official Rust base image?

defcronyke: Because to run on a scratch container, it must be built with rust nightly with the musl target, and must link against some C libraries which are themselves built with the musl target. These C libraries are not available on the official Rust base image and neither is rust nightly. Furthermore, compiling the dependencies takes a while so it is really nice to just build them once-in-a-while and have any changes to the actual application build really quickly.

steren: I fear that this might scare users. Hello world are supposed to convey simplicity, I am in favor of using the simplest Dockerfile and sample.

Maybe create a new sample instead of replacing the Hello World? Something like `rest-api-rust`?

defcronyke: Sure, I could move it to a different folder and change the name, and then I could even add an example to the source code showing how to handle a couple different URL routes, including with params, with Rocket. I think moving forward though, users need to learn about multi-stage builds right from the beginning, because otherwise everyone is pushing 300 MB+ images into production for no reason, which also leaves open a larger attack surface.

steren: I agree. let's create a new sample, but update the helloworld-rust sample to mention it as a beast practice example.

defcronyke: Okay, I will close this pull request and open a new one later today when I have the new example ready.

defcronyke:
Closing this because I will resubmit it with a bit more code, and renamed as rest-api-rust as discussed.

Issue #750, opened 2019-01-11
New markdown link checker and linter are overly sensitive


samodell:
Many PRs in the Docs repo right now have build test failures when it seems like they shouldn't.

Causes:
* Oversensitive link verification, such as #644 (more details in the comments of that PR; but to summarize, one link is valid and causing a failure, and one link is to a http://localhost URL, which we should probably exclude from link checks)

* Unknown: I can't see a root cause for some failures, such as #747 
Error logs are vague:
I0110 23:28:45.331] ============================
I0110 23:28:45.332] ==== BUILD TESTS FAILED ====
I0110 23:28:45.332] ============================
W0110 23:28:45.363] Run: ('/workspace/./test-infra/jenkins/../scenarios/../hack/coalesce.py',)
E0110 23:28:45.363] Command failed
I0110 23:28:45.363] process 643 exited with code 1 after 0.7m
E0110 23:28:45.363] FAIL: pull-knative-docs-build-tests

CC @srinivashegde86  @jessiezcc  @adrcunha

adrcunha:
About the links, yes, the checker might be over sensitive. Probably it doesn't like an address without protocol, and I'm pretty sure localhost:4000 is not returning content at the time of checking. :)

About the second, scroll up the logs and you'll see:

```
I0110 23:28:24.004] ------------------------------------
I0110 23:28:24.005] ---- Linting the markdown files ----
I0110 23:28:24.005] ------------------------------------
I0110 23:28:26.564] serving/samples/helloworld-go/README.md:141: MD007 Unordered list indentation
I0110 23:28:26.565] 
I0110 23:28:26.565] A detailed description of the rules is available at https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
```

samodell:
@adrcunha  with #747, I played around with the list formatting and wasn't able to satisfy the linter. The docs for the linter say that you must use 2 spaces for indented items in lists, but when I used that level of indentation, it broke the numbered list the unordered list is under by making the next numbered item in the list start over at 1. 

I think this linter might not be a great fit for our repo. Is there a way to disable the Markdown linter but keep the link checking capabilities?

adrcunha:
https://github.com/knative/test-infra/pull/391 allows disabling the linter for a repo.

PR #762, opened 2019-01-14
Use release.yaml instead of eventing.yaml

grantr:
`eventing.yaml` doesn't install the in-memory provisioner, `release.yaml` does. Since the provisioner is required to run the samples, we should be
telling users to install `release.yaml`.

This (hopefully) fixes the issues people are having running samples with 0.3.0.

### DIFF
--- a/install/Knative-with-AKS.md
+++ b/install/Knative-with-AKS.md
@@ -166,7 +166,7 @@ your Knative installation, see [Performing a Custom Knative Installation](Knativ
     ```bash
     kubectl apply --filename https://github.com/knative/serving/releases/download/v0.3.0/serving.yaml \
     --filename https://github.com/knative/build/releases/download/v0.3.0/release.yaml \
-    --filename https://github.com/knative/eventing/releases/download/v0.3.0/eventing.yaml \
+    --filename https://github.com/knative/eventing/releases/download/v0.3.0/release.yaml \
     --filename https://github.com/knative/eventing-sources/releases/download/v0.3.0/release.yaml
     ```
 1. Monitor the Knative components until all of the components show a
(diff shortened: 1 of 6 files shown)


samodell:
Fixes #762 

/lgtm
/approve

grantr:
Just noticed how easy it was to make this mistake: serving uses `serving.yaml` as its release yaml name, so it would make sense if `eventing.yaml` was also the preferred name. However build, eventing, and eventing-sources all use `release.yaml`. We should probably use the same naming scheme for all projects.

grantr:
Also updated `install/Knative-custom-install.md` to use 0.3.0 instead of 0.2.x, and replaced `eventing.yaml` with `release.yaml` in example urls.

samodell:
@grantr  The Knative custom install doc really needs to be rewritten for 0.3.0, which is why I hadn't updated it yet. It has a lot of now obsolete information referring to how the installs used to be bundled together, and makes references to monitoring being included in the Serving install, which is no longer the case. Just updating the links to the 0.3.0 components would cause a lot of confusion. Also, the -lite and -nomon releases for Serving don't exist in 0.3.0.

TL:DR; can you revert your changes to that file? It's on our to-do list to do a thorough review for 0.3.0.

install/Knative-custom-install.md
```diff
@@ -155,23 +155,23 @@ The following Knative installation packages are available:
 
 - **Multiple Component Bundle**:
 
-  - https://github.com/knative/serving/releases/download/v0.2.3/release.yaml
-  - https://github.com/knative/serving/releases/download/v0.2.3/release-no-mon.yaml
-  - https://github.com/knative/serving/releases/download/v0.2.3/release-lite.yaml
+  - https://github.com/knative/serving/releases/download/v0.3.0/release.yaml
```
samodell: Please revert this file. A lot of this info needs to be entirely rewritten for 0.3.0.

grantr: I rebased to remove the commit that changed this file. PTAL!

samodell: thanks! looks good.

samodell:
@evankanderson  The build test failures are a known issue, with a fix in progress. See issue #750 for more background info.

Issue #760, opened 2019-01-14
Configuring outbound network access topic is hard to discover


samodell:
## Expected Behavior
After installing Knative, users are able to easily figure out what they need to do to configure outbound network access.

## Actual Behavior
[Configuring Outbound Network Access](https://github.com/knative/docs/blob/master/serving/outbound-network-access.md) guide is not currently linked to from any existing install guides. However, this is a very common user task, and the errors a user will see if they attempt to access a resource outside of their cluster is non-intuitive.

One error that might pop up in this scenario:

```
Something went wrong: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```

PR #764, opened 2019-01-14
Note about outbound access

vaikas-google:
Addresses #760 
## Proposed Changes

- Add a note in eventing/README.md install about configuring outbound network access.
-
-

### DIFF
--- a/eventing/README.md
+++ b/eventing/README.md
@@ -70,8 +70,6 @@ Knative Eventing currently requires Knative Serving and Istio version 1.0 or
 later installed.
 [Follow the instructions to install on the platform of your choice](../install/README.md).
 
-### Configuring outbound network access
-
 Many of the sources require making outbound connections to create the event subscription,
 and if you have any functions that make use of any external (to cluster) services, you
 must enable it also for them to work.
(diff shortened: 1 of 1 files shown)


eventing/README.md
```diff
@@ -70,6 +70,13 @@ Knative Eventing currently requires Knative Serving and Istio version 1.0 or
 later installed.
 [Follow the instructions to install on the platform of your choice](../install/README.md).
 
+### Configuring outbound network access
+
+Many of the sources require making outbound connections to create the event subscription,
+and if you have any functions that make use of any external (to cluster) services, you
+must enable it also for them to work.
+[Follow the instructions to configure outbound network access](../serving/outbound-network-access.md).
+
```
evankanderson: You need a new sub-head here, or it looks like this is still about configuring outbound network access.

evankanderson: Alternately, you can move this to right before "This document will be updated as additional sources ..."

vaikas-google: I just got rid of the subheader. originally I just wanted to add bold, but our linter didn't like that. I just removed it and added it as a paragraph.

PR #771, opened 2019-01-15
Update custom install for v.0.3

samodell:
Work toward fixing #718

PR #779, opened 2019-01-16
Tweaks to custom install for v0.3

samodell:
Addresses @RichieEscarez 's comments from PR #771

### DIFF
--- a/install/Knative-custom-install.md
+++ b/install/Knative-custom-install.md
@@ -66,13 +66,13 @@ service mesh. If you install any of the following options, you must install
 
 #### Istio installation options
 
-| Istio Install Filename     | Description                                                            |
-| -------------------------- | ---------------------------------------------------------------------- |
-| **[`istio-crds.yaml`][a]** | Creates CRDs before installing Istio.                                  |
-| **[`istio.yaml`][b]**      | Install Istio with service mesh enabled (automatic sidecar injection). |
-| [`istio-lean.yaml`][c]     | Install Istio and disable the service mesh by default.                 |
+**Install files in _bold italics_ are the recommended standard install files suitable for most use cases.**
 
-**Install files in bold are the recommended standard install suitable for most use cases.**
+| Istio Install Filename       | Description                                                              |
+| ---------------------------- | ------------------------------------------------------------------------ |
+| **_[`istio-crds.yaml`][a]_** | Creates CRDs before installing Istio.                                  |
+| **_[`istio.yaml`][b]_**      | Install Istio with service mesh enabled (automatic sidecar injection). |
+| [`istio-lean.yaml`][c]       | Install Istio and disable the service mesh by default.                   |
 
 [a]: https://github.com/knative/serving/releases/download/v0.3.0/istio-crds.yaml
 [b]: https://github.com/knative/serving/releases/download/v0.3.0/istio.yaml
@@ -144,12 +144,13 @@ service mesh. If you install any of the following options, you must install
 ## Installing Knative components
 
 Each Knative component must be installed individually. You can decide which
-components to install based on what you plan to do with Knative.
+components and observability plugins to install based on what you plan to do
+with Knative.
 
 ### Choosing Knative installation files
 
 The following Knative installation files are available:
-- **Serving Component**:
+- **Serving Component and Observability Plugins**:
   - https://github.com/knative/serving/releases/download/v0.3.0/serving.yaml
   - https://github.com/knative/serving/releases/download/v0.3.0/monitoring.yaml
   - https://github.com/knative/serving/releases/download/v0.3.0/monitoring-logs-elasticsearch.yaml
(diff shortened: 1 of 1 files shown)


install/Knative-custom-install.md
```diff
@@ -66,13 +66,13 @@ service mesh. If you install any of the following options, you must install
 
 #### Istio installation options
 
-| Istio Install Filename     | Description                                                            |
-| -------------------------- | ---------------------------------------------------------------------- |
-| **[`istio-crds.yaml`][a]** | Creates CRDs before installing Istio.                                  |
-| **[`istio.yaml`][b]**      | Install Istio with service mesh enabled (automatic sidecar injection). |
-| [`istio-lean.yaml`][c]     | Install Istio and disable the service mesh by default.                 |
+**Install files in _bold italics_ are the recommended standard install files suitable for most use cases.**
```
carieshmarie:
```suggestion
**The files in _bold italics_ are the recommended standard install files suitable for most use cases.**
```

I originally read "Install files" as Verb-noun, so suggest using an article to make it clearer.

install/Knative-custom-install.md
```diff
@@ -178,28 +179,28 @@ files from the Knative repositories:
...
-| [`in-memory-channel.yaml`][4.3]                  | Installs only the in-memory channel provisioner.                                                      |
-| [`kafka.yaml`][4.4]                              | Installs only the Kafka channel provisioner.                                                          |
-| **knative/eventing-sources**                     |                                                                                                       |
-| **[`release.yaml`][5.1]**                        | Installs the following sources: [Kubernetes][6], [GitHub][6.1], [Container image][6.2], [CronJob][6.3]|
-| [`release-gcppubsub.yaml`][5.2]                  | Installs the following sources: [PubSub][6.4]                                                         |
-| [`message-dumper.yaml`][5.3]                     | Installs an Event logging service for debugging.                                                      |
+**Install files in _bold italics_ are the recommended standard install files suitable for most use cases.**
```
carieshmarie:
```suggestion
**The files in _bold italics_ are the recommended standard install files suitable for most use cases.**
```

RichieEscarez:
I just noticed that we removed all the dependency info. For example the way the info is not states that Eventing no longer requires Serving. Is that a new improvement for 0.3?

install/Knative-custom-install.md
```diff
@@ -66,13 +66,13 @@ service mesh. If you install any of the following options, you must install
 
 #### Istio installation options
 
-| Istio Install Filename     | Description                                                            |
-| -------------------------- | ---------------------------------------------------------------------- |
-| **[`istio-crds.yaml`][a]** | Creates CRDs before installing Istio.                                  |
-| **[`istio.yaml`][b]**      | Install Istio with service mesh enabled (automatic sidecar injection). |
-| [`istio-lean.yaml`][c]     | Install Istio and disable the service mesh by default.                 |
+**The files in _bold italics_ are the recommended standard install files suitable for most use cases.**
```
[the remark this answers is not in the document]
samodell: I don't know, I really like how scannable the table is this way. I do wish it was easier in markdown to show a table's legend, but I think it's hard to miss that bold line on top of the table, and as you look at each group of components, the bold italic ones jump out at you. I don't love the idea of repeating "Tip: compatible with most use cases" for each file where that's true.

How would you feel about using some other symbol (maybe †, if we can get that to show up in markdown) to denote this meaning, instead of just the formatting?

samodell:
> I just noticed that we removed all the dependency info. For example the way the info is now states that Eventing no longer requires Serving. Is that a new improvement for 0.3?

I thought so, but I actually think that's not correct. I'll update this to indicate that Eventing still depends on Serving.

install/Knative-custom-install.md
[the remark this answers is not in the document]
RichieEscarez: Right now this line renders like a double title and the table holds 4 conventions (1.code and link for install file, 2.code bold italic and link for "common use case", 3. link to more info, 4. bold table section for component). Honestly, I think the trouble here is that our adv install guide is trying to also act as an install reference document(and help non adv users too). 

Maybe the solution is the maintain our adv. user audience focus. I think the adv user audience is going to know whether or not they want an ELK stack and what that means. To a new user, I dont think "suitable for most use cases"  (-> common use case?) is actually helpful due to the lack of detail (i immediately ask my self 'why?', what are our common use cases and are they not covered in the host specific install guides).

(temp fix) If you really want to combine new users info here too, we could clean up the table by simply adding a different indicator (†) but I really dont like the extra bold (header looking) line and would like that perhaps, moved to the end, under the table.

(long term solution) Or we could have a separate document that is focused on new users and better discuss each "use case" and what files to install.

RichieEscarez:
> > I just noticed that we removed all the dependency info. For example the way the info is now states that Eventing no longer requires Serving. Is that a new improvement for 0.3?
> 
> I thought so, but I actually think that's not correct. I'll update this to indicate that Eventing still depends on Serving.

Yeah, we should validate all the dependency info that was in the original version to make sure they should be removed (ie. it was not just serving, another example is that you must install Serving before an observability plugin, and the new format fails to indicate that too)

samodell:
@RichieEscarez  My force pushed wiped the commits you made to my branch, which I was worried would happen, so I copied the changes you committed into my local copy, so they're still in the final outcome even though they've disappeared from the log here.

samodell:
> > > I just noticed that we removed all the dependency info. For example the way the info is now states that Eventing no longer requires Serving. Is that a new improvement for 0.3?
> > 
> > 
> > I thought so, but I actually think that's not correct. I'll update this to indicate that Eventing still depends on Serving.
> 
> Yeah, we should validate all the dependency info that was in the original version to make sure they should be removed (ie. it was not just serving, another example is that you must install Serving before an observability plugin, and the new format fails to indicate that too)

Added a "dependencies" section to the table; tried to keep it is as simple as possible.

install/Knative-custom-install.md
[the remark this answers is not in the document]
samodell: I don't think it hurts to call out what the "standard" or "common" install files are, even for advanced users. It's good for them to know what baseline they're deviating from. I went with option 1.

I feel like "a separate document focused on new users" already exists; the on-any-K8s install kind of serves that purpose. Or one of the many platform-specific guides.

install/Knative-custom-install.md
```diff
@@ -178,33 +180,33 @@ files from the Knative repositories:
...
-| [`message-dumper.yaml`][5.3]                     | Installs an Event logging service for debugging.                                                      |
+| Knative Install Filename                         | Notes                                                                                                 | Dependencies                                                      |
+| -------------------------------------------------| ----------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------- |
+| **knative/serving**                              |                                                                                                       |                                                                   |
+| **[`serving.yaml`][1.1]_†_**                     | Installs the  Serving component.                                                                      |                                                                   |
```
RichieEscarez:
```suggestion
| [`serving.yaml`][1.1]_†_                     | Installs the  Serving component.                                                                      |                                                                   |
```
remove the extra bold for all the † items in the tables and it works for me

install/Knative-custom-install.md
```diff
@@ -178,33 +180,33 @@ files from the Knative repositories:
...
+| **[`release.yaml`][3.1]_†_**                     | Installs the Build component.                                                                         |                                                                   |
+| **knative/eventing**                             |                                                                                                       |                                                                   |
+| **[`release.yaml`][4.1]_†_**                     | Installs the Eventing component. Includes the in-memory channel provisioner.                          | Serving component                                                 |
+| [`eventing.yaml`][4.2]                           | Installs the Eventing component. Does not include the in-memory channel provisioner.                  | Serving component                                                 |
+| [`in-memory-channel.yaml`][4.3]                  | Installs only the in-memory channel provisioner.                                                      | Eventing component                                                |
```
RichieEscarez: Nit: the dependency for these channel install files and any of the event sources below is both Eventing and Serving 

(just another one of those unnecessary cognitive load things where we force them to read the whole eventing row to mentally map more dependencies vs. all in one place)

RichieEscarez:
> @RichieEscarez My force pushed wiped the commits you made to my branch, which I was worried would happen, so I copied the changes you committed into my local copy, so they're still in the final outcome even though they've disappeared from the log here.

cool, no worries (thanks for mentioning it)

install/Knative-custom-install.md
samodell: I added in the additional/implied dependencies.

install/Knative-custom-install.md
samodell: I ended up removing all the formatting from the new symbol; I don't think it needs it.

samodell:
Ready for another look, @RichieEscarez

PR #781, opened 2019-01-16
Format markdown

mattmoor-sockpuppet:


samodell:
\hold

It looks like I can't do that through the GitHub UI; @mattmoor  can you revert those two files? I'm not sure why, but Prettier's suggestions are causing markdown rendering errors in:
serving/installing-logging-metrics-traces.md
and
serving/samples/helloworld-elixir/README.md

PR #788, opened 2019-01-18
Format markdown

mattmoor-sockpuppet:


samodell:
Same mardown rendering errors as #781  -- 

serving/installing-logging-metrics-traces.md
and
serving/samples/helloworld-elixir/README.md

\hold

PR #808, opened 2019-01-25
fix: update nodejs hello world example

soldair:
update nodejs hello world example such that the correct dependencies end up in the target image.

Fixes #(issue-number)

## Proposed Changes

-
-
-

### DIFF
--- a/serving/samples/helloworld-nodejs/README.md
+++ b/serving/samples/helloworld-nodejs/README.md
@@ -92,16 +92,12 @@ recreate the source files from this folder.
    # Create and change to the app directory.
    WORKDIR /usr/src/app
 
-   # Copy application dependency manifests to the container image.
-   # A wildcard is used to ensure both package.json AND package-lock.json are copied.
-   # Copying this separately prevents re-running npm install on every code change.
-   COPY package*.json ./
-
-   # Install production dependencies.
-   RUN npm install --only=production
-
    # Copy local code to the container image.
    COPY . .
+
+   # Install production dependencies but discard the node_modules 
+   # copied from your dev environment
+   RUN npm ci --only=production
 
    # Service must listen to $PORT environment variable.
    # This default value facilitates local development.
--- a/serving/samples/helloworld-nodejs/README.md
+++ b/serving/samples/helloworld-nodejs/README.md
@@ -97,7 +97,8 @@ recreate the source files from this folder.
 
    # Install production dependencies but discard the node_modules 
    # copied from your dev environment
-   RUN npm ci --only=production
+   RUN rm node_modules
+   RUN npm install --only=production
 
    # Service must listen to $PORT environment variable.
    # This default value facilitates local development.


serving/samples/helloworld-nodejs/README.md
```diff
@@ -92,17 +92,13 @@ recreate the source files from this folder.
...
-
-   # Install production dependencies.
-   RUN npm install --only=production
-
    # Copy local code to the container image.
    COPY . .
 
+   # Install production dependencies but discard the node_modules 
+   # copied from your dev environment
+   RUN npm ci --only=production
```
steren: One problem is that now `npm ci` will be run anytime a source file changes. In the previous version, `npm install` was only ran if a change in `package*.json` happened.

Do you think there is any way to preserve the original optimization while also avoiding to copy local node_modules?

soldair: there is not a simple way that i can tell to generate a correct installation without the user code.

the previous optimization could only every apply to a project without local install/build scripts so was really a false flag of comfort only applicable to hello world.

steren: The problem I see by forcing to use `npm ci` is that it would fail if there is no `package-lock.json`. This mmight not be desired for a Hello world, since users might just want to copy paste package-lock.json and deploy.

Maybe use an explicit `rm` command  to get riud of `node_modules` and do `npm install`?

soldair: you could consider changing it to
WORKDIR /usr/src/tmp
COPY package*.json ./
RUN npm ci --only-production --ignore-scripts
WORKDIR /usr/src/app
COPY . .
RUN cp -fr /usr/src/tmp/node_modules ./node_modules
RUN npm install --only=production --prefer-offline

this would cache all of the network IO but still run scripts when user code is in place

soldair: @steren please let me know if you'ed prefer this solution ill update the pr

soldair: good catch

soldair:
With a .dockerignore we still have a problem where if a user has install scripts npm install will likely fail but at least that's a much more obvious problem and one that happens at build time rather than crashing at runtime

soldair:
If we do support a dockerignore file at the gcloud level we'll have to make sure it doesn't get copied to the workspace otherwise images created wont have the node_modules we created.
On the other side if we dont get support for dockerignore in gcloud we can use .docker ignore in builds to ignore build deps.

steren:
This sample's README particular uses `docker`, so we should be OK using `.dockerignore`.

Regarding custom scripts. I think that this goes beyond the Hello world use case and would require a customization of the Dockerfile. For example by running `npm run build` after copying the app's code.

soldair:
I dont think that'll help because the first npm command we want to cache will fail. But I agree that we aren't going to get to a place where we can cache npm install and make it work for everyone.  hello world isn't the place for ultimate solutions :)

soldair:
FWIW my vote is to make it slow for everyone but work and be easy to understand. And linking to other example docker files with optimizations like cached npm install.

soldair:
Yeah. Faster user experience is super valuable once you get started.

If my project has a postinstall script npm runs it automatically on npm install. If that script depends on files in my project I haven't copied into the image yet it'll fail.

steren:
I think we can live with postinstall scripts failing, in particular since they are not use in the Hello World.

Can I let you add a .dockerignore?

soldair:
Also keep in mind that install scripts are an essential part of deploying a server that has any kind of frontend application. I think this will make any idea of reusing this docker file beyond the hello world end in user discomfort. Maybe I can add a note that if they have install scripts to move the install command after the copy

Yeah happy to add it :)

steren:
This is the `Dockerfile` that is officially recommended bu Node.js.

As I said, users should rather add `RUN npm run build` if they want to manipulate their source code, using a postinstall is not a good idea IMO.

serving/samples/helloworld-nodejs/README.md
```diff
@@ -92,17 +92,14 @@ recreate the source files from this folder.
    # Create and change to the app directory.
    WORKDIR /usr/src/app
 
-   # Copy application dependency manifests to the container image.
-   # A wildcard is used to ensure both package.json AND package-lock.json are copied.
-   # Copying this separately prevents re-running npm install on every code change.
-   COPY package*.json ./
```
steren: You are removing a build optimization.

soldair: Let's just close this. We left this off with the realization that there is no way to make a docker file that solves edge cases and is easy to consume in the context of hello world. This docker file will produce broken images if the node_modules tree has a native dependency. It needs an ignore file to omit the users local node_modules from the `copy . .` But ignore files are complex as we've discussed before.

soldair: It npm installs then copies the local files into the image including local node_modules like it's doing all the work then undoing it.

grayside:
Why not use `--skip-scripts` in your Dockerfile for a first install, then run your web asset step as a followup step.

Sticking with the workflows optimized for ensuring consistent local development builds comes at the cost of routinely longer docker builds because you are invalidating the cached layer containing node_modules.

It is a trade-off, and this sample is aimed at surfacing the recommended practice from the containers + nodejs community.

PR #814, opened 2019-01-28
Fix cronjob and github sample

### LINKED ISSUE #813: eventing cronjob does not work
When I follow steps at https://github.com/knative/docs/blob/master/eventing/samples/cronjob-source/README.md to run cronjob sample, found that the image `github.com/knative/eventing-sources/cmd/message_dumper:latest` cannot be found, am I missing anything?

https://github.com/knative/docs/blob/master/eventing/samples/cronjob-source/service.yaml#L13

FYI @zxDiscovery @whynowy

whynowy:
Fixes #813

Used `ko` image path in the spec however the installation command in README.rd was "kubectl apply ..."

## Proposed Changes

- Use a real message_dumper image instead of `ko` image path, so that user can try out the samples without `ko`.

### DIFF
--- a/eventing/samples/cronjob-source/README.md
+++ b/eventing/samples/cronjob-source/README.md
@@ -27,7 +27,7 @@ spec:
       revisionTemplate:
         spec:
           container:
-            image: github.com/knative/eventing-sources/cmd/message_dumper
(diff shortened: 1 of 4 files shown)


whynowy:
@gyliu513

whynowy:
Realized github-source sample is actually working ok, only need to fix cronjob sample. Closed this one and will file another one.

PR #815, opened 2019-01-28
Fix cronjob sample bug

### LINKED ISSUE #813: eventing cronjob does not work

whynowy:
Use real message_dumper image instead of ko image path.

Fixes #813 

Used `ko` image path in the spec however the installation command in README.rd was "kubectl apply ..."

## Proposed Changes

-  Use a real message_dumper image instead of `ko` image path, so that user can try out the samples without `ko`.

### DIFF
--- a/eventing/samples/cronjob-source/README.md
+++ b/eventing/samples/cronjob-source/README.md
@@ -27,7 +27,7 @@ spec:
       revisionTemplate:
         spec:
           container:
-            image: github.com/knative/eventing-sources/cmd/message_dumper
+            image: gcr.io/knative-releases/github.com/knative/eventing-sources/cmd/message_dumper
 ```
 
 Use following command to create the service from `service.yaml`:
--- a/eventing/samples/cronjob-source/service.yaml
+++ b/eventing/samples/cronjob-source/service.yaml
@@ -10,4 +10,4 @@ spec:
       revisionTemplate:
         spec:
           container:
-            image: github.com/knative/eventing-sources/cmd/message_dumper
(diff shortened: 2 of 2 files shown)


gyliu513:
lgtm, thanks @whynowy

ahmetb:
I see you just used a message_dumper image without any tags here @whynowy. 

Looks like in `gcp-pubsub-source/subscriber.yaml` we have this image's tag automatically provided:

https://github.com/knative/docs/blob/9689f68901fdee5fea4e9c09343e2f29ca8e2c56/eventing/samples/gcp-pubsub-source/subscriber.yaml#L11-L15

Does anyone know the cronjob sample doesn't have this comment and the sha256 tag of the same image? Is there an automation at play here? If so, both YAMLs should use the same mechanism?

/cc @FuriKuri 
(last person to edit that line from "git blame")

whynowy:
@ahmetb Not sure if they explicitly added the sha256 tag or did it with an automation. Currently CronJob is doing the way as github (https://github.com/knative/docs/blob/master/eventing/samples/github-source/service.yaml#L12), uses the `lastest` tag.

ahmetb:
I opened #830 which explains the situation better.

PR #853, opened 2019-02-01
site: add PR#851 - channels

RichieEscarez:
merge in and add Hugo build and front matter to recent changes from master

### DIFF
--- a/docs/eventing/channels/README.md
+++ b/docs/eventing/channels/README.md
@@ -10,8 +10,6 @@ procedure:
     ```
 -->
 
-# Knative Channels
-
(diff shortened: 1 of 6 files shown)


docs/eventing/channels/README.md
```diff
@@ -10,8 +10,6 @@ procedure:
     ```
 -->
 
-# Knative Channels
```
carieshmarie: Should Channels be capitalized here? I'm seeing a mix of capitalization and non-capitalization - not sure I have strong opinions either way, but I do think it should be consistent.

RichieEscarez: no i dont think so, fixed it below

PR #804, opened 2019-01-24
#789: serving/samples/helloworld-csharp: update to dotnet sdk 2.2

grayside:
Fixes #789 

## Proposed Changes

- Update Dockerfile to use dotnet sdk 2.2
- Update code for dotnet 2.2 compatibility

## Steps Taken

1. Switch Dockerfile to extend from microsoft/dotnet:2.2-sdk

2. Update application code for dotnet 2.2:
```sh
cd serving/samples/helloworld-csharp
docker run --rm --workdir /app -v $PWD:/app/helloworld-csharp microsoft/dotnet:2.2-sdk dotnet new web --force -o helloworld-csharp
```

3. Confirm application code does not need refactor

4. Restore custom changes (e.g., listen to `$PORT` and `$TARGET`)

```sh
git checkout Startup.cs
git checkout Program.cs
```

5. Build & Test container image locally

PR #883, opened 2019-02-08
Updated instructions to use latest SDK, cleaned up unnecessary import…

### LINKED ISSUE #789: helloworld-csharp broken with dotnet sdk 2.2
If I go install dotnet SDK today, I'll get 2.2.

But the dotnet SDK provided in https://github.com/knative/docs/blob/b850378d5c17e5540d7bb0387133233485c3d662/serving/samples/helloworld-csharp/Dockerfile uses 2.1

The docs also say 2.1 but not sure how realistic it is to expect people will install/have the older version.

meteatamel:
…s and added a verification step

Fixes #789


grayside:
This is a dup of #804, but I'm happy to see this one proceed as I'm not a C# developer.

samodell:
@meteatamel  Can you take a look at your CLA issue?

meteatamel:
I'm not sure what I'm supposed to do in regard to "Need a CLA" error?

meteatamel:
@samodell CLA issue should be fixed now

meteatamel:
@grayside Before, port binding was done in code. I simply moved it out of the code to an env variable. I don't see how this is any different than before, other than the 5 lines of extra code that we don't need anymore.

grayside:
The difference is that at build time, $ASPNETCORE_URLS will be immediately set to the current value of $PORT (`8080` by default in the Dockerfile). However, the [runtime contract](https://github.com/knative/serving/blob/master/docs/runtime-contract.md#protocols-and-ports) specifies to bind to the value of $PORT at runtime.

We can set the $ASPNETCORE_URLS variable in the Dockerfile but only if it's done in a way where the interpolation of the $PORT variable happens as part of the entrypoint.

meteatamel:
Maybe I'm missing something obvious but what I don't fully understand is that before my commit, the PORT was also set to 8080 in Dockerfile and that value was used in the code. To change the port that was being bound to, you had to change PORT in Dockerfile and rebuild, right? How was this runtime and not build time?

grayside:
The `ENV PORT 8080` line in the code is a default, but not necessarily what knative does. Let me give you a couple local docker usage examples:

**Relying on the PORT var in the Dockerfile**

```
docker run --rm -p 9090:8080 helloworld-dotnet:latest
```

**Just the Container Runtime Contract Please**

```
docker run --rm -p 9090:9091 -e PORT=9091 helloworld-dotnet:latest
```

The PORT environment variable can be assigned any value at runtime, and the container image will use it to override the default provided by the ENV statement.

meteatamel:
Thanks for the concrete example, it makes sense now.  I've been trying to make this work with this:
```
ENV PORT 8080
CMD ["dotnet", "out/helloworld-csharp.dll", "--urls", "http://0.0.0.0:${PORT}"]
```
But for some reason, dotnet keeps ignoring ${PORT} while running in Docker.  If I hardcode the port like this, it works:
```
CMD ["dotnet", "out/helloworld-csharp.dll", "--urls", "http://0.0.0.0:9090"]
```

Shell has no problems in reading $PORT:
```
CMD ["sh", "-c", "echo Port: ${PORT}"]
```

Any ideas why $PORT is ignored in dotnet?

grayside:
Python shows one way to do this: https://github.com/knative/docs/blob/master/serving/samples/helloworld-python/Dockerfile

Note with your change, we need to propagate the runtime value of $PORT to $ASPNETCORE_URLS in the entrypoint, not just set $PORT

meteatamel:
I was thinking of not defining $ASPNETCORE_URLS at all and just rely on $PORT in dotnet command to bind to the right port. However, as I said, I couldn't make dotnet recognize $PORT for some odd reason.

samodell:
@meteatamel  Thanks for resolving the CLA issue! Ping me for a review when you've resolved @grayside 's concern.

meteatamel:
@samodell I reverted my commit that @grayside was concerned about. We can go ahead and merge this pull request. If I find a way to address @grayside's concern, I'll submit a separate PL later.

samodell:
Thanks, @meteatamel !

PR #886, opened 2019-02-08
Add gloo to docs

ilackarms:
[Gloo, the Envoy-based Kubernetes Ingress / API Gateway](https://gloo.solo.io/) can now [serve as a clusteringress controller for knative-serving](https://gloo.solo.io/getting_started/kubernetes/gloo_with_knative/). 

## Proposed Changes

- Add an alternate installation option for OpenShift, Minishift, Minikube to use `glooctl` to install Knative without Istio, using Gloo as the Gateway controller instead.

### DIFF
--- a/install/Knative-custom-install.md
+++ b/install/Knative-custom-install.md
@@ -30,6 +30,11 @@ to run multiple installation commands.
 
 ## Installing Istio
 
+> Note: [Gloo](https://gloo.solo.io/) now provides an alternative installation option
+> for those wishing to install Knative without Istio. Note that Knative-Eventing
+> is not supported by Gloo at this time. [Click here](Knative-with-Gloo.md) to follow the
+> Gloo-based install instructions for Knative.
+
 Knative depends on [Istio](https://istio.io/docs/concepts/what-is-istio/) for
 traffic routing and ingress. You have the option of injecting Istio sidecars and
 enabling the Istio service mesh, but it's not required for all Knative
--- a/install/Knative-custom-install.md
+++ b/install/Knative-custom-install.md
@@ -32,8 +32,7 @@ to run multiple installation commands.
 
 > Note: [Gloo](https://gloo.solo.io/) is available as an alternative to Istio.
 > Gloo is not currently compatible with the Knative Eventing component.
-> [Click here](Knative-with-Gloo.md) to follow the
-> Gloo-based install instructions for Knative.
+> [Click here](Knative-with-Gloo.md) to install Knative with Gloo.
 
 Knative depends on [Istio](https://istio.io/docs/concepts/what-is-istio/) for
 traffic routing and ingress. You have the option of injecting Istio sidecars and
--- a/install/Knative-custom-install.md
+++ b/install/Knative-custom-install.md
@@ -30,9 +30,9 @@ to run multiple installation commands.
 
 ## Installing Istio
 
-> Note: [Gloo](https://gloo.solo.io/) now provides an alternative installation option
-> for those wishing to install Knative without Istio. Note that Knative-Eventing
-> is not supported by Gloo at this time. [Click here](Knative-with-Gloo.md) to follow the
+> Note: [Gloo](https://gloo.solo.io/) is available as an alternative to Istio.
+> Gloo is not currently compatible with the Knative Eventing component.
+> [Click here](Knative-with-Gloo.md) to follow the
 > Gloo-based install instructions for Knative.
 
 Knative depends on [Istio](https://istio.io/docs/concepts/what-is-istio/) for
--- a/install/README.md
+++ b/install/README.md
@@ -25,11 +25,11 @@ Knative depends on an Ingress/Gateway which is capable of routing requests to Kn
 Currently, two options exist which provide this functionality: 
 [Istio](https://istio.io/), the Envoy-based Service Mesh, and [Gloo](https://gloo.solo.io/), the Envoy-based API Gateway.
 
-[Install with Gloo](Knative-with-Gloo.md): Gloo functions as a lightweight gateway for Knative. Choose this option if you don't require service  mesh in your cluster and want a lighter-weight alternative to Istio. Note that *Knative-Eventing* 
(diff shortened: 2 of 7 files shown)


samodell:
Hey @ilackarms , thanks for your PR! Question -- why have you chosen to add Gloo as an option for only some of the install guides? Does it only work with these types of clusters at the moment?

My other thought is that if Gloo is going to be presented in the docs as an equal option to Istio, (and especially if it is just the first option that may be added) we may need to move the Istio/Gloo instructions out of the install guides altogether and put them in their own files that we link out to, to avoid these install guides becoming truly unwieldy. Hold for the moment on this.

/hold

ilackarms:
hi @samodell thanks for the quick reply! 

The reason Gloo is only added as an option for some of the install guides is because we provide an install option for Knative-Serving that doesn't include Istio. That works for the install guides where Serving is the only Knative component being installed. However Eventing has a harder dependency on Istio than Serving does, so we're unable to provide an install path for that component without Istio. That means the install doc would instruct people to install *both* Istio and Gloo, which doesn't seem to make much sense. We'd be very interested in adding the Gloo instructions to *all* of the install guides, but I don't see how it will work at the moment given the issue with Eventing. I'm open to any suggestions on this.

And yes, I agree that a better structure for the docs would be to separate the "Install Istio/Gloo" (e.g. `Gateway Install`) section out from the "Install Knative" section. The idea in this PR was to introduce a minimal set of changes, but I'm open to restructuring the docs in this way if the maintainers would be open to reviewing such a change.

Thanks again for the thorough response!

samodell:
@ilackarms  Ahh, I see. Well, the other install guides only instruct users to install all components as a form of expediency -- "If you want to be able to try out everything, do this!" What I'm thinking is we should pull the instructions for using Gloo into its own guide, and then we can link out to the Gloo steps from the Istio section of every install guide as an alternative. In the Gloo guide, you can specify the component limitations for Gloo; IE, it only works with Serving right now.

For this PR, why don't you tackle moving the Gloo steps into one stand-along page? You can then add in the links to Gloo from all the install guides in a follow-up PR. (Or in this one, if you prefer.)

I'm imagining a note like this:

```
> Note: You can use [Gloo](link) instead of Istio if you only install the Knative Serving component.
> Gloo is not currently compatible with the Knative Eventing component.
> See [Gloo install guide](link) for instructions.
``` 

We can revisit the idea of doing something similar for Istio in the future if it ever becomes possible to sub in Gloo for every component, but right now I think we should leave that in the standard install guides, since it has no compatibility issues with other Knative components.

ilackarms:
@samodell what do you think of the following?
1 - separate the Gloo steps into one stand-alone page
2 - restructure the "install landing page" to refer to 2 install paths: one using istio, one using gloo.
   * the istio install path will then link to the existing docs
   * the gloo install path will link to the standalone gloo install page

how does this approach sound?

ilackarms:
i've updated the PR with the suggested changes

install/Knative-custom-install.md
```diff
@@ -30,6 +30,11 @@ to run multiple installation commands.
 
 ## Installing Istio
 
+> Note: [Gloo](https://gloo.solo.io/) now provides an alternative installation option
+> for those wishing to install Knative without Istio. Note that Knative-Eventing
```
samodell: Global comment: "Knative-Eventing" --> "the Knative Eventing component".

install/README.md
```diff
@@ -20,6 +20,19 @@ clusters.
 
 ## Installing Knative
 
+Knative depends on an Ingress/Gateway which is capable of routing requests to Knative Services.
+
+Currently, two options exist which provide this functionality: 
+[Istio](https://istio.io/), the Envoy-based Service Mesh, and [Gloo](https://gloo.solo.io/), the Envoy-based API Gateway created by [Solo.io](www.solo.io).
```
samodell:
```suggestion
[Istio](https://istio.io/), the Envoy-based Service Mesh, and [Gloo](https://gloo.solo.io/), the Envoy-based API Gateway.
```

samodell: To keep it parallel, taking this extra info out.

install/Knative-custom-install.md
```diff
@@ -30,6 +30,11 @@ to run multiple installation commands.
 
 ## Installing Istio
 
+> Note: [Gloo](https://gloo.solo.io/) now provides an alternative installation option
```
samodell:
```suggestion
> Note: [Gloo](https://gloo.solo.io/) is available as an alternative to Istio.
```

install/Knative-custom-install.md
samodell:
```suggestion
> Gloo is not currently compatible with the Knative Eventing component.
```

install/Knative-custom-install.md
```diff
@@ -30,6 +30,11 @@ to run multiple installation commands.
...
+> for those wishing to install Knative without Istio. Note that Knative-Eventing
+> is not supported by Gloo at this time. [Click here](Knative-with-Gloo.md) to follow the
```
samodell:
```suggestion
> See [Knative Install using Gloo](Knative-with-Gloo.md) to use Gloo instead of Istio.
```

install/Knative-custom-install.md
```diff
@@ -30,6 +30,11 @@ to run multiple installation commands.
...
+> Gloo-based install instructions for Knative.
```
samodell:
```suggestion
```

samodell: Copy changes to this note globally.

install/README.md
```diff
@@ -20,6 +20,19 @@ clusters.
...
+
+[Install with Gloo](Knative-with-Gloo.md): Gloo functions as a lightweight gateway for Knative. Choose this option if you don't require service  mesh in your cluster and want a lighter-weight alternative to Istio. Note that *Knative-Eventing* 
+is not supported by Gloo at this time.
+
+[Install with Istio](#Installing-Knative-with-Istio) Istio is a popular service mesh that includes a Knative-compatiblle
```
samodell:
```suggestion
[Install with Istio](#Installing-Knative-with-Istio) Istio is a popular service mesh that includes a Knative-compatible
```

samodell: Create a heading, ## Installing Knative with Gloo, and put the info about Gloo there along with the link to the guide.

Move the information about Istio below, into the "Installing Knative with Istio" section. Remove the link.

PR #890, opened 2019-02-13
Update command and formatting in README.md

poy:
Setup Knative Eventing's command was not in code block. This makes is easy to not quite see that there is a step to perform. This PR fixes that.


gyliu513:
Thanks @poy , but this was fixed and merged in https://github.com/knative/docs/pull/895

poy:
@gyliu513 Great! It looks like there are still some typos. I've submitted #910 to fix that.

PR #906, opened 2019-02-18
Unify the doc command.

gyliu513:
The document is using different options, some using long name while
some using short name, this PR is trying to update all long names
to short names. Including updating --namespace to -n, --filename
to -f, --output to -o, --selector to -l.

### DIFF
--- a/build/auth.md
+++ b/build/auth.md
@@ -80,7 +80,7 @@ into their respective files in `$HOME`.
 1.  Execute the build:
 
     ```shell
-    kubectl apply --filename secret.yaml serviceaccount.yaml build.yaml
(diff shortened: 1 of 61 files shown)


samodell:
Hey @gyliu513 , when the docs team last discussed this, we decided to use long names instead of short names for clarity. We felt this would avoid causing confusion for users who haven't seen the short names before. It's possible some short names have been re-added over time, or that we missed some in our initial pass, but we don't want to make the change in this direction.

If you wanted to check the code base for short names and replace them with long names, I could approve that.

samodell:
Closing. Feel free to open a new PR if you wanted to reverse the direction, as mentioned above.

PR #915, opened 2019-02-20
Updating install docs to v0.4

samodell:
v0.4 release has been cut for all components; updating install files to match.

Adds information about clusterrole.yaml, which is now necessary if using the Build and Serving components together.


gyliu513:
lgtm, can you rebase @samodell

Issue #882, opened 2019-02-07
Clarify Istio install steps


samodell:
See https://github.com/knative/serving/issues/3108

NorikDavtian:
Thanks for making this ticket. It is easy to skip the notes right underneath the command markup. Clarifying the installation steps will help.

moeyui1:
It seems to be fixed by adding 
```bash
curl -L https://github.com/knative/serving/releases/download/v0.3.0/istio-crds.yaml \
  | kubectl apply --filename -
```
before applying `istio.yaml`. May I submit a PR to fix this bug?

samodell:
Absolutely, @moeyui1 ! We'd appreciate it.

digikin:
The docs are still incorrect.  Here is exactly what the instructions commands are:
```
kubectl apply --filename https://raw.githubusercontent.com/knative/serving/v0.7.0/third_party/istio-1.1.7/istio-crds.yaml &&
curl -L https://raw.githubusercontent.com/knative/serving/v0.7.0/third_party/istio-1.1.7/istio.yaml \
  | sed 's/LoadBalancer/NodePort/' \
  | kubectl apply --filename -

# Label the default namespace with istio-injection=enabled.
kubectl label namespace default istio-injection=enabled
```
As @moeyui1 pointed out, they need to be:
```
curl -L https://github.com/knative/serving/releases/download/v0.3.0/istio-crds.yaml   | kubectl apply --filename -
```
Then
```
kubectl apply --filename https://raw.githubusercontent.com/knative/serving/v0.7.0/third_party/istio-1.1.7/istio-crds.yaml && curl -L https://raw.githubusercontent.com/knative/serving/v0.7.0/third_party/istio-1.1.7/istio.yaml   | sed 's/LoadBalancer/NodePort/'   | kubectl apply --filename -
```

PR #917, opened 2019-02-21
Correct Istio install steps for minikube (#882)

moeyui1:
Fixes #(882)

## Proposed Changes
- The `istio-crds.yaml` should be apply before the `istio.yaml`.


samodell:
fixes #882

PR #925, opened 2019-02-22
Format markdown

mattmoor-sockpuppet:


samodell:
Putting these on hold for a bit, since they create terrible merge conflicts for Richie to deal with as he syncs master with release-0.3 (which is the branch that has the updates/file moves necessary for the website).

PR #935, opened 2019-02-25
Fix docs

vagababov:
## Proposed Changes

- make java and kotlin examples to use $IP_ADDRESS env var
- add a snippet to automate its setting.
- add some `{}` around the env vars in the examples.

/cc @dgerd

### DIFF
--- a/serving/samples/helloworld-java/README.md
+++ b/serving/samples/helloworld-java/README.md
@@ -208,9 +208,9 @@ folder) you're ready to build and deploy the sample app.
    helloworld-java     helloworld-java.default.example.com
    ```
 
-1. Now you can make a request to your app to see the result. Presuming the
-   the address you see returned in the previous step is in `IP_ADDRESS` env
-   variable:
+1. Now you can make a request to your app to see the result. Presuming,
+   the IP address you got in the step above is in the `${IP_ADDRESS}`
+   env variable:
 
    ```shell
    curl -H "Host: helloworld-java.default.example.com" http://${IP_ADDRESS}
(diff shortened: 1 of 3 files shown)


serving/samples/helloworld-java/README.md
```diff
@@ -200,11 +208,12 @@ folder) you're ready to build and deploy the sample app.
...
 
-1. Now you can make a request to your app to see the result. Replace
-   `{IP_ADDRESS}` with the address you see returned in the previous step.
+1. Now you can make a request to your app to see the result. Presuming,
+   the IP address you got in the step above is in the `${IP_ADDRESS}`
+   env variable:
 
    ```shell
-   curl -H "Host: helloworld-java.default.example.com" http://{IP_ADDRESS}
+   curl -H "Host: helloworld-java.default.example.com" http://${IP_ADDRESS}
```
dgerd: nit: The domain may not be example.com

It would be better to get it. See https://github.com/knative/docs/tree/master/serving/samples/rest-api-go#access-the-service

vagababov:
> All of the samples have different versions of the same things to get the IP, get the domain, curl the service. It makes me wonder if we should do:
> 
> 1. Send users through a single getting started walkthrough before going to the samples, and assume knowledge of how to query (with links back in case the walkthrough is missed).
> 2. Use something like markdown_helper to encapsulate duplicate content in a single file: https://github.com/BurdetteLamar/markdown_helper/blob/master/markdown/use_cases/include_files/reuse_text/use_case.md#reuse-text
> 
> @samodell, do you have any thoughts on this?

I do concur, that some _basic standard path_ should be outlined somewhere and then referenced from all the examples.

Though this change is mostly tactical.

dgerd:
> Though this change is mostly tactical.

I agree. Lets not block this change, but I did want to take the opportunity to raise the issue and I would like to get a github issue open to do one of those two things (or a third to be proposed option).

serving/samples/helloworld-java/README.md
vagababov: And done.

samodell:
/lgtm
/approve

@vagababov & @dgerd  Totally agree with your point on standardizing how we document frequently repeated tasks. Please open an issue and reference this PR.

PR #945, opened 2019-02-27
undo adding explicit install command for in-memory channel provisioner

samodell:
Undo #944

PR #944, opened 2019-02-27
Add in-memory-channel to enable eventing release

mchmarny:
Eventing release was failing for me due to missing `ClusterChannelProvisioner`
```
unable to recognize "https://github.com/knative/eventing/releases/download/v0.4.0/release.yaml": no matches for kind "ClusterChannelProvisioner" in version "eventing.knative.dev/v1alpha1"
```
Adding `https://github.com/knative/eventing/releases/download/v0.4.0/in-memory-channel.yaml` fixes this

Fixes #(issue-number)

## Proposed Changes

-
-
-

PR #960, opened 2019-03-06
Fix knative installation for Gardener

vlerenc:
Missing manifest, cc @vasu1124.


vlerenc:
/assign @mattmoor

Could you please take a look (the command was suggested automatically).

samodell:
This is the same issue going on with #944 and #945 -- in theory, you shouldn't need to explicitly install the in-memory channel provisioner if you install eventing from https://github.com/knative/eventing/releases/download/v0.4.0/release.yaml, but in practice people are getting errors without it. We're investigating this.

/hold

samodell:
Verified this morning that you don't have to explicitly install the in-memory channel provisioner. If install fails on the first attempt, just re-rerun the install commands and it will succeed on second attempt.

PR #996, opened 2019-03-13
Update the monitoring file version

Justin2997:
## Proposed Changes
To reflect version v0.4.0 we need to update file version on this file.


Justin2997:
I signed it!

PR #997, opened 2019-03-14
update ref docs - v0.3 api

RichieEscarez:
Update the ref docs to v0.3 (initial build was based on 0.2).

Add some "how to" instructions in README.

Also small fix to build script (only way I could get it to run on my linux box).


hack/gen-api-reference-docs.sh
```diff
@@ -160,9 +160,10 @@ main() {
         "${knative_eventing_sources_root}"
 
     log "Generated files written to ${clone_root}/out/."
-    log "Copy the files in reference/ directory to knative/docs."
-    if command -v open >/dev/null; then
```
ahmetb: it'd be cool to still keep `command -v open` :) 

but I guess you're not merging this to master so it's cool.

hack/gen-api-reference-docs.sh
```diff
@@ -25,20 +25,20 @@ REFDOCS_REPO="https://${REFDOCS_PKG}.git"
 REFDOCS_VER="5c208a6"
 
 KNATIVE_SERVING_REPO="github.com/knative/serving"
-KNATIVE_SERVING_COMMIT="v0.2.3"
-KNATIVE_SERVING_OUT_FILE="serving.md"
+KNATIVE_SERVING_COMMIT="v0.3.0"
```
ahmetb: ah I guess people will edit these numbers at some point :)

we can totally make this configurable with a default value, like 

```
KNATIVE_SERVING_COMMIT="${KNATIVE_SERVING_COMMIT:-v0.3.0}"
```

or make it required like:

```
KNATIVE_SERVING_COMMIT="${KNATIVE_SERVING_COMMIT?:specify KNATIVE_SERVING_COMMIT variable}"
```

either way is cool with me.

docs/reference/README.md
```diff
@@ -4,3 +4,15 @@
 - [Build API](build.md)
 - [Eventing API](eventing/eventing.md)
 - [Event Sources API](eventing/eventing-sources.md)
+
+Note: The Knative API reference documentation is manually generated using the
```
ahmetb: I'd probably start a new section here like

    ## Updating API Reference docs (for Knative maintainers)

hack/gen-api-reference-docs.sh
RichieEscarez: I could not get past this error: https://stackoverflow.com/questions/42463929/couldnt-find-a-file-descriptor-referring-to-the-console-on-ubuntu-bash-on-win

And the solution suggested worked. Not sure if you know of an alternate?

Also, while im thinking about these troubleshooting efforts, it reminded me that I ran into "no tmp space" storage size errors and had to go in an manually delete all the cloned content for each failed attempt of the command. I realized that the docs got generated just fine but since I reran the command until the "open command" error resolved, I ended up with 10-15 copies of all the clones. 

Perhaps an option to delete on error? Or some other magic you might have?

ahmetb: > I could not get past this error: [stackoverflow.com/questions/42463929/couldnt-find-a-file-descriptor-referring-to-the-console-on-ubuntu-bash-on-win](https://stackoverflow.com/questions/42463929/couldnt-find-a-file-descriptor-referring-to-the-console-on-ubuntu-bash-on-win)

`open` is referring to the macOS tool here actually.  :) didn't realize linux has it too. dont worry about it here. in my new PR xdg comes first.

> it reminded me that I ran into "no tmp space" storage size errors and had to go in an manually delete all the cloned content for each failed attempt of the command.

this sounds odd and shouldn't happen (and doesn't happen).

> I realized that the docs got generated just fine but since I reran the command until the "open command" error resolved, I ended up with 10-15 copies of all the clones.

yeah that's a legit bug. but temp dirs are supposed to be cleaned up on every reboot. so I assumed people would not run low on disk. knative repos arent that big?

> Or some other magic you might have?

I think ideally we'd add a cleanup step running on errors etc. Opened https://github.com/knative/docs/issues/1000 to track.

hack/gen-api-reference-docs.sh
ahmetb: I fixed this in my PR #998

RichieEscarez:
Thanks @ahmetb !

The "auto approved" (since im an admin) told Prow to merge but I will go in and apply all of your comments.

ahmetb:
I don't think you need to change anything for this PR. We should probably make script fixes in the `master`

PR #998, opened 2019-03-14
hack: configurable repo refs for api docs gen

ahmetb:
The API docs generation script can now be called with non-hardcoded commit/tag
refs.

	KNATIVE_SERVING_COMMIT=v0.4.0 \
	  KNATIVE_BUILD_COMMIT=v0.4.0 \
	  KNATIVE_EVENTING_COMMIT=v0.2.1 \
	  KNATIVE_EVENTING_SOURCES_COMMIT=v0.2.1 ./gen-api-reference-docs.sh

This should ease docs releases a tiny bit.

### DIFF
--- a/hack/gen-api-reference-docs.sh
+++ b/hack/gen-api-reference-docs.sh
@@ -15,6 +15,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 set -euo pipefail
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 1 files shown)


hack/gen-api-reference-docs.sh
```diff
@@ -163,6 +164,8 @@ main() {
     log "Copy the files in reference/ directory to knative/docs."
     if command -v open >/dev/null; then
         open "${clone_root}/out/"
+    elif command -v xdg-open >/dev/null; then
```
RichieEscarez: This failed for me on Linux. I searched and found this on stackoverflow: 
[On linux] "The open command is a link to the openvt command and opens a binary in a new virtual console." which is not what we want and why the conditional does not fail.

I tried swapping the commands so that 'xdg-open' is tried/run first, before 'open', and it worked. But since I dont have my Mac setup right now to test the script, i've only just tired running 'xdg-open', which results in "-bash: xdg-open: command not found" (so I think we should be okay?).

hack/gen-api-reference-docs.sh
```diff
@@ -163,6 +164,8 @@ main() {
     log "Copy the files in reference/ directory to knative/docs."
     if command -v open >/dev/null; then
```
RichieEscarez:
```suggestion
    if command -v xdg-open >/dev/null; then
```

hack/gen-api-reference-docs.sh
```diff
@@ -163,6 +164,8 @@ main() {
...
         open "${clone_root}/out/"
```
RichieEscarez:
```suggestion
        xdg-open "${clone_root}/out/"
```

hack/gen-api-reference-docs.sh
RichieEscarez:
```suggestion
    elif command -v open >/dev/null; then
```

hack/gen-api-reference-docs.sh
```diff
@@ -163,6 +164,8 @@ main() {
...
+    elif command -v xdg-open >/dev/null; then
+        xdg-open "${clone_root}/out/"
```
RichieEscarez:
```suggestion
        open "${clone_root}/out/"
```

RichieEscarez:
sorry, I expected the "blocked review" to prevent Prow from merging

PR #1015, opened 2019-03-18
fix serving directory path

RPing:
## Proposed Changes

- fix serving directory path in setting-up-a-logging-plugin.md


RPing:
I signed it!

PR #1016, opened 2019-03-18
Fix(docs): Fix broken links in docs/build/README.md

navidshaikh:
Reference all markdown documents relatively.

### DIFF
--- a/docs/build/README.md
+++ b/docs/build/README.md
@@ -36,10 +36,10 @@ More information about this use case is demonstrated in
 See the following reference topics for information about each of the build
 components:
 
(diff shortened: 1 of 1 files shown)


samodell:
Thanks for your PR, @navidshaikh ! I appreciate you fixing the broken links, but we actually prefer using relative links whenever possible -- not sure how the full link made their way into the docs here. Would you mind rewriting this PR to use relative links instead?

samodell:
It looks like this has already been updated in master, so closing this out. Thanks for your willingness to contribute, @navidshaikh ! Sorry for our slowness.

PR #1026, opened 2019-03-19
Update README.md

cycgp:
Update URL path of Knative Serving and Knative Eventing


cycgp:
I signed it!

samodell:
Thanks for your PR, @cycgp , but we want to keep these links pointing into the docs, not to the component repos on GitHub.

PR #1030, opened 2019-03-19
 serving/samples/helloworld-kotlin: pin gradle to known working version

grayside:
Pin the version of gradle (to 4.10).

Note that gradle:5.2 is also available, but I've found it does not build on Cloud Build.


saturnism:
consider using gradle wrapper in the future.

grayside:
Hi @saturnism, if the gradle wrapper makes sense in a "helloworld" Dockerfile for Kotlin, it would be great to get a PR or an issue with more details on the suggestion. I am not a Kotlin developer and am just looking to stabilize this sample.

ahmetb:
/lgtm
thanks for pinning to a MAJOR.MINOR.

PR #1025, opened 2019-03-19
Merge website files into Master

RichieEscarez:
Merging all the changes from the website testing and enablement in the `release-03` branch, back into `master`.

This is the first step to enabling the `master` branch to render in the website (will immediately cut 0.4 (`release-0.4`) and then add all versions to the doc build so they all appear in the website: knative.dev/docs).

fixes #724 , fixes #725 , fixes #726 , fixes #727 , fixes #728 , fixes #729 , fixes #730  

## Staged content

These files have been staged here:
https://deploy-preview-12--knative.netlify.com/docs/ 
**Note:** *View staged content only under ../docs/ (there are still some todos for the staging site so these have been manually built only in that part of the site). Also, keep in mind that main menu links are hard codes and will bring you out of staging, to knative.dev*

## Next steps

Website open action items: https://github.com/knative/docs/projects/5

PR #1031, opened 2019-03-19
Fix broken links to knative install instruction

chaodaiG:
all links to knative installation instruction still thinks `install` is on top level, and broke a bunch of them


RichieEscarez:
Thanks for your PR. There is a lot of Website related work that is in progress and these are all fixed already: https://github.com/knative/docs/pull/1025

Also (fyi), going forward, all links within the knative/docs repo, must be relative links.

PR #1035, opened 2019-03-20
serving/sample/helloworld-php: fix container build error from upstream

grayside:
PHP 7.2.6-apache has somehow lost the built-in php.ini-* configuration files which we use to set a container default in the helloworld-php sample.

On top of that, 7.2.6 is an ancient and unsupported minor release. To track updates to PHP (and fix this build error) let's loosen the pinning to php:7.2-apache so backwards compatible/supported updates to the PHP version and container image are pulled in.

## Proposed Changes

- Change base image from `php:7.2.6-apache` to `php:7.2-apache`

/cc @averikitsch


ahmetb:
+1 all pinnings must use :MAJOR.MINOR (or where possible, just :MAJOR. Otherwise you'd be missing out on security fixes to the language.)

PR #954, opened 2019-03-05
Added Hello World Java using Spark Java Framework

hpandeycodeit:
Fixes #832 

## Proposed Changes

- Added Hello World Java using Spark Java Framework
- Renamed helloworld-java to helloworld-java-spring
- Other java examples will be added in seperate PR

PR #1037, opened 2019-03-20
Follow up on #954 - Added Hello World Java using Spark Java Framework

### LINKED ISSUE #832: Proposal: update helloworld-java with simpler code
Our [simple Java hello world](https://github.com/knative/docs/tree/master/serving/samples/helloworld-java) uses Spring Boot. This adds a lot of complexity for this simple use case and Spring is not a framework that works great in serverless environment.

Proposal: 

* Replace `helloworld-java` with simpler code. For example, see this existing [minimal Java 11 sample](https://github.com/ludoch/samples/tree/master/java11/smallest-sample) or use a mote lightweight framework, like http://sparkjava.com/.

* Delete the existing sample or move it to `helloworld-java-spring` (I can understand trhat we would still want to showcase Spring Boot.

CC @ludoch @grayside

saturnism:
Fixes #832 #954


matzew:
IMO this is a better location for that example:

https://github.com/knative/docs/tree/master/community/samples/serving

PR #1039, opened 2019-03-20
Add resource cleanup instructions for several event-source examples

lberk:
### Proposed Changes

- cronjob-source: cleanup event source, service
- iot-core: cleanup gcpubsub, channel, gcpubsub-source and subscription
- kubernetes-event-source: cleanup channel, service account, k8s event, and subscription

### DIFF
--- a/docs/eventing/samples/iot-core/README.md
+++ b/docs/eventing/samples/iot-core/README.md
@@ -201,3 +201,35 @@ see them in the subscriber.
         -src "iot-core demo" \
         -events 10
     ```
+
+### Cleanup
+
+To cleanup the knative resources:
+
+1.  Remove the gcppubsub controller:
+    ```shell
+    pushd $HOME/go/src/github.com/knative/eventing-sources
+    kubectl delete --filename config/default-gcppubsub.yaml
+    popd
+    ```
+
+2.  Remove the channel:
+    ```shell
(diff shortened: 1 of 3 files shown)


docs/eventing/samples/iot-core/README.md
```diff
@@ -201,3 +201,35 @@ see them in the subscriber.
...
         -events 10
     ```
+
+### Cleanup
+
+To cleanup the knative resources:
+
+1.  Remove the gcppubsub controller:
+    ```shell
+    pushd $HOME/go/src/github.com/knative/eventing-sources
```
gyliu513: Instead of using `pushd`, how about use `cd` instead?

lberk: We can do that, but I wanted to be consistent with the setup instructions.  I think it'd make sense to use the same (whichever that is) in both places.  What would you prefer?

gyliu513: +1 to keep consistent, but I'd prefer that we use `cd` instead. I think we can handle this in other PR.

lberk:
@RichieEscarez Hi! I think this is a relatively simple PR, is there anything else you'd like to see before getting an /approved ?

matzew:
@samodell I think it's /approve :)

Issue #1021, opened 2019-03-18
Eventing sample updates for v0.5


samodell:
[/docs/docs/eventing/samples](https://github.com/knative/docs/tree/master/docs/eventing/samples)

- [x] gcp-pubsub-source
- [x] iot-core
- [x] kubernetes-event-source

PR #1049, opened 2019-03-21
Update the GcpPubSubSource example to use Broker, rather than Channel.

Harwayne:
Part of #1021

## Proposed Changes

- Switches the sample from using `Channel` to use `Broker`.

### DIFF
--- a/docs/eventing/samples/gcp-pubsub-source/README.md
+++ b/docs/eventing/samples/gcp-pubsub-source/README.md
@@ -22,7 +22,7 @@ source is most useful as a bridge from other GCP services, such as
    addition, install the GCP PubSub event source from `release-gcppubsub.yaml`:
 
    ```shell
-   kubectl apply --filename https://github.com/knative/eventing-sources/releases/download/v0.3.0/release-gcppubsub.yaml
+   kubectl apply --filename https://github.com/knative/eventing-sources/releases/download/v0.4.1/gcppubsub.yaml
    ```
 
 1. Enable the `Cloud Pub/Sub API` on your project:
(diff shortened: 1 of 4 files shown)


docs/eventing/samples/gcp-pubsub-source/README.md
```diff
@@ -25,6 +25,12 @@ source is most useful as a bridge from other GCP services, such as
    kubectl apply --filename https://github.com/knative/eventing-sources/releases/download/v0.4.1/gcppubsub.yaml
    ```
 
+   Until 0.5.0 is released (and the previous URL is updated to point at it), you will need to have a copy of the `knative/eventing-sources` repo locally and from its root:
```
samodell: @adamharwayne  In theory, this work is for the 0.5 release, so we should be documenting the "state of the world" as it will be in 0.5.

Harwayne: adamharwayne  has been retired in favor of @Harwayne.

Should I give fake 0.5.0 URLs? Two lines above the URL has 0.4.1, should I just replace it with 0.5.0?

samodell: Maybe think of it as not-yet-active URLs, instead of fake. We'll double-check that the URLs match the resources that are actually released when v0.5 is cut.

Harwayne: I have tested that this guide works as-is (fb03dde). Should I now change it to point at not-yet-active 0.5.0 URLs and submit?

Harwayne: Discussed and decided to submit with the expect 0.5.0 URLs.

Harwayne:
Ping.

PR #1059, opened 2019-03-22
Minor change to our upgrade docs

duglin:
People need to delete this one Eventing resource before they can apply
the yaml for the upgrade. W/o this they'll see an error like this:
```
The StatefulSet "controller-manager" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```

Once people are ok with this text I'll then apply it to the other md files.

### DIFF
--- a/docs/install/Knative-custom-install.md
+++ b/docs/install/Knative-custom-install.md
@@ -283,6 +283,15 @@ commands below.
    kubectl delete deploy knative-ingressgateway -n istio-system
    ```
 
+   If you have the Knative Eventing Sources component installed,
+   you will also need to delete the following resource before upgrading:
+   ```
+   kubectl delete statefulset/controller-manager -n knative-sources
+   ```
+   While the deletion of this resource during the upgrade process will
+   not prevent modifications to Eventing Source resources, those changes will
+   not be completed until the upgrade process finishes.
+
 1. To install Knative components or plugins, specify the filenames in the
    `kubectl apply` command:
 
--- a/docs/install/Knative-with-AKS.md
+++ b/docs/install/Knative-with-AKS.md
@@ -181,6 +181,15 @@ your Knative installation, see
    kubectl delete deploy knative-ingressgateway -n istio-system
    ```
 
+   If you have the Knative Eventing Sources component installed,
+   you will also need to delete the following resource before upgrading:
+   ```
+   kubectl delete statefulset/controller-manager -n knative-sources
+   ```
+   While the deletion of this resource during the upgrade process will
+   not prevent modifications to Eventing Source resources, those changes will
+   not be completed until the upgrade process finishes.
+
 1. Run the `kubectl apply` command to install Knative and its dependencies:
    ```bash
    kubectl apply --filename https://github.com/knative/serving/releases/download/v0.4.0/serving.yaml \
--- a/docs/install/Knative-with-GKE.md
+++ b/docs/install/Knative-with-GKE.md
@@ -182,6 +182,15 @@ see [Performing a Custom Knative Installation](./Knative-custom-install.md).
    kubectl delete deploy knative-ingressgateway -n istio-system
    ```
 
+   If you have the Knative Eventing Sources component installed,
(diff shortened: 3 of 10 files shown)


docs/install/Knative-custom-install.md
```diff
@@ -283,6 +283,15 @@ commands below.
    kubectl delete deploy knative-ingressgateway -n istio-system
    ```
 
+   You will also need to delete the following resource:
+   ```
+   kubectl delete statefulset/controller-manager -n knative-sources
+   ```
+   While the deletion of this resource during the upgrade process will
+   not prevent your users from successfully requesting updates to the Knative
+   Eventing resources, those changes will not be acted upon, or realized,
```
grantr: It's actually the CRDs in eventing-sources that are affected:

- ContainerSource
- CronJobSource
- GitHubSource
- KubernetesEventSource
- AwsSqsSource
- GcpPubSubSource

Eventing CRDs that are not affected:
- Channel
- ClusterChannelProvisioner
- Subscription

Think it's worth listing the resources that are affected? It's a long list so it may just be noise.

docs/install/Knative-custom-install.md
```diff
@@ -283,6 +283,15 @@ commands below.
    kubectl delete deploy knative-ingressgateway -n istio-system
    ```
 
+   You will also need to delete the following resource:
```
grantr: What do you think about qualifying this paragraph with "If you have installed Knative Eventing Sources, ..." since it only applies to that project?

samodell: This is exclusively a problem for people upgrading Eventing? I would call that out here, since not everyone uses all the components.

samodell: Could just start the sentence with, "If you want to upgrade the Knative Eventing component..."

docs/install/Knative-custom-install.md
```diff
@@ -283,6 +283,15 @@ commands below.
...
+   ```
+   kubectl delete statefulset/controller-manager -n knative-sources
+   ```
+   While the deletion of this resource during the upgrade process will
```
samodell: Does this paragraph (lines 290-294) mean that deleting this resource won't break current eventing implementations in the time between deleting this resources and upgrading to the next version? If so, let's just state that.

If not, can you try to simplify + clarify this? I'm not sure I understand.

docs/install/Knative-custom-install.md
samodell: I think listing those out is probably too noisy.

docs/install/Knative-custom-install.md
samodell: Oh, Grant beat me to this suggestion above.

docs/install/Knative-custom-install.md
duglin: Yeah - listing them seems  like a lot and I'm not sure what the person would do with that info.

docs/install/Knative-custom-install.md
duglin: Well, it won't break anything but it will put any changes made "on hold". Meaning, your `kuebctl apply <some-eventing-sources-resource>` will be accepted but not actually take effect until the rest of the install finishes. That's what I was trying to say...

docs/install/Knative-custom-install.md
```diff
@@ -283,6 +283,15 @@ commands below.
    kubectl delete deploy knative-ingressgateway -n istio-system
    ```
 
+   If you have the Knative Eventing Sources component already installed then
```
samodell:
```suggestion
   If you have the Knative Eventing Sources component installed,
```

docs/install/Knative-custom-install.md
```diff
@@ -283,6 +283,15 @@ commands below.
...
+   you will also need to delete the following resource:
```
samodell:
```suggestion
   you will also need to delete the following resource before upgrading:
```

PR #1080, opened 2019-03-26
Clarify installation steps with Docker For Mac

chtison:
Keep istio gateway service to `LoadBalancer` type.
Remove command replacing `LoadBalancer` by `NodePort` in knative serving yaml installation step because `grep LoadBalancer serving.yaml` returns nothing.


sebgoa:
@chtison could you sign the Google CLA before we test and review this ? thank you

chtison:
I signed it!

sebgoa:
Could you send the PR towards master instead of the release-0.4 branch. We release 0.5.0 already and need to make sure this gets to master.

(interestingly we don't need to NodePort either in minikube since LoadBalancer will work as well).

/lgtm

/ok-to-test

samodell:
Yep, @chtison  this PR should be made against the master branch.

PR #1083, opened 2019-03-26
Add e2e tests for kotlin, nodejs, php, python, ruby, and shell sample apps

chaodaiG:
Add more e2e tests for kotlin, nodejs, php, python, ruby, and shell sample apps


test/e2e/config.yaml
```diff
@@ -18,3 +18,44 @@ languages:
...
+      - "Dockerfile"
+  - language: "nodejs"
+    expectedOutput: "Hello Node.js Sample v1!"
+    copies:
+      - "index.js"
+      - "package.json"
+      - "service.yaml"
+      - "Dockerfile"
+  - language: "php"
+    expectedOutput: "Hello PHP Sample v1!\\n"
```
adrcunha: I suggest your comparison ignores trailing blanks, LF or CR.

chaodaiG: The output is actually not a LF, it's a string `\n`, that's why I used double `\` here.

adrcunha: This doesn't seem right. Looks like an error in the sample. If so, please fix it.

chaodaiG: I do have a plan for consolidating all these outputs, will address this as well

PR #1090, opened 2019-03-26
Adding Apache Kafka source ref.

matzew:
as per title.

steps:
* updated yaml
* run the generator, and check in the MD file changes 

/cc @evankanderson @dubee


matzew:
@mattmoor It looks like the genrator is not really 100% friend with your format bot :-)

I ran the `go run eventing/sources/generator/main.go` as discussed... but the MD file has a few more changes...

Perhaps worth to file a GH-issue for that?

PR #1092, opened 2019-03-26
Consolidate helloworld sample apps output and make sure instruction consistent with source code

chaodaiG:
Consolidating helloworld sample apps output, so that they are using the same pattern, also makes sure README.md files are consistent with source code

### DIFF
--- a/docs/serving/samples/hello-world/helloworld-go/README.md
+++ b/docs/serving/samples/hello-world/helloworld-go/README.md
@@ -66,7 +66,7 @@ recreate the source files from this folder.
    WORKDIR /go/src/github.com/knative/docs/helloworld
    COPY . .
 
-   # Build the helloworld command inside the container.
+   # Build the command inside the container.
    # (You may fetch or manage dependencies here,
    # either manually or with a tool like "godep".)
    RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld
--- a/docs/serving/samples/hello-world/helloworld-go/README.md
+++ b/docs/serving/samples/hello-world/helloworld-go/README.md
@@ -188,7 +188,7 @@ folder) you're ready to build and deploy the sample app.
 
    ```shell
    curl -H "Host: helloworld-go.default.example.com" http://35.203.155.229
-   Hello World: Go Sample v1!
+   Hello Go Sample v1!
    ```
 
    > Note: Add `-v` option to get more detail if the `curl` command failed.
--- a/docs/serving/samples/hello-world/helloworld-go/README.md
+++ b/docs/serving/samples/hello-world/helloworld-go/README.md
@@ -60,7 +60,7 @@ recreate the source files from this folder.
    # Use the offical Golang image to create a build artifact.
    # This is based on Debian and sets the GOPATH to /go.
    # https://hub.docker.com/_/golang
-   FROM golang as builder
+   FROM golang:1.12 as builder
 
    # Copy local code to the container image.
    WORKDIR /go/src/github.com/knative/docs/helloworld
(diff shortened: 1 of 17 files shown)


chaodaiG:
/cc @adrcunha 

Reference to what we discussed in PR https://github.com/knative/docs/pull/1083

docs/serving/samples/hello-world/helloworld-go/README.md
```diff
@@ -60,13 +60,13 @@ recreate the source files from this folder.
    # Use the offical Golang image to create a build artifact.
    # This is based on Debian and sets the GOPATH to /go.
    # https://hub.docker.com/_/golang
-   FROM golang as builder
+   FROM golang:1.12 as builder
```
adrcunha: Why 1.12 explicitly?

docs/serving/samples/hello-world/helloworld-go/README.md
```diff
@@ -60,13 +60,13 @@ recreate the source files from this folder.
...
 
    # Copy local code to the container image.
    WORKDIR /go/src/github.com/knative/docs/helloworld
    COPY . .
 
-   # Build the helloworld command inside the container.
```
adrcunha: That's actually a "command". I would write "Build the helloworld binary inside the container" though.

docs/serving/samples/hello-world/helloworld-kotlin/README.md
```diff
@@ -109,7 +109,7 @@ recreate the source files from this folder.
    ```docker
    # Use the official gradle image to create a build artifact.
    # https://hub.docker.com/_/gradle
-   FROM gradle as builder
+   FROM gradle:4.10 as builder
```
adrcunha: Same question about a specific version of gradle here.

docs/serving/samples/hello-world/helloworld-nodejs/README.md
```diff
@@ -93,7 +93,7 @@ recreate the source files from this folder.
    # Copy application dependency manifests to the container image.
    # A wildcard is used to ensure both package.json AND package-lock.json are copied.
    # Copying this separately prevents re-running npm install on every code change.
-   COPY package*.json ./
+   COPY package.json package*.json ./
```
adrcunha: Really? COPY doesn't recognized package.json when using the wildcard? If so, please update the comment.

docs/serving/samples/hello-world/helloworld-go/README.md
chaodaiG: The purpose of this PR is consolidating README files, all these changes are due to inconsistency between source file and README. Preferably I would pick the one comes later, but it's not doable as they all lost history in github UI due to recent movement. So I just blindly picked one that I thought being later

chaodaiG: for example this line comes from this file:
https://github.com/knative/docs/blob/52837ad26ad4cd2e3e93a5c17e8a32941d1e6a23/docs/serving/samples/hello-world/helloworld-go/Dockerfile#L4

adrcunha: Thanks for clarifying.

docs/serving/samples/hello-world/helloworld-go/README.md
chaodaiG: same here

docs/serving/samples/hello-world/helloworld-kotlin/README.md
chaodaiG: same here

docs/serving/samples/hello-world/helloworld-nodejs/README.md
chaodaiG: Nice catch. I just tested out, wildcard works for package.json. Updated

Issue #1088, opened 2019-03-26
Small Error in Auto Scaling Sample


ciaranRoche:
## Expected Behaviour
Complete the [Auto Scaling Sample](https://github.com/knative/docs/tree/master/docs/serving/samples/autoscale-go) as described in the repo.

## Actual Behavior
 When you get to the step where you send 30 seconds of traffic using the following  command 
```
hey -z 30s -c 50 \
  -host "autoscale-go.default.example.com" \
  "http://${IP_ADDRESS?}?sleep=100&prime=10000&bloat=5" \
  && kubectl get pods
```
An error is thrown, this is due to a previous step where there is a small typo in where you export the `IP_ADDRESS`
This is what it is currently: 
```
export IP_ADDRESS=`kubectl get svc $INGRESSGATEWAY --namespace istio-system --output jsonpath=" .  {.status.loadBalancer.ingress[*].ip}"`
```
When it should be:
```
export IP_ADDRESS=`kubectl get svc $INGRESSGATEWAY --namespace istio-system --output jsonpath="{.status.loadBalancer.ingress[*].ip}"`
```
## Steps to Reproduce the Problem

1. Follow the current [Auto Scaling Sample](https://github.com/knative/docs/tree/master/docs/serving/samples/autoscale-go) as described

## Additional Info
**Additional context**
I would be happy to submit a PR to fix this, I decided to open an Issue on it first as I am unsure of the format for fixing such mistakes in the documentation. Felt it a safe bet to open an issue first.

**Install information:**
 - Platform (GKE, IKS, AKS, etc.): GKE
 - Knative Version: 0.4.x

abrennan89:
Hi @ciaranRoche :) are you currently working on this issue? I have found another issue for the autoscaling samples so might make sense to group them if there are a bunch of small changes?

@samodell how do we go about getting issues assigned? I can add PRs but not assign an issue to myself :)

ciaranRoche:
@abrennan89 I am finished working on this issue, just waiting on one more approval is all. I do not mind adding those changes to my PR [1] if you want to leave a comment on the issue you found.

[1] - https://github.com/knative/docs/pull/1094

ciaranRoche:
This issue can be closed, now that https://github.com/knative/docs/pull/1094 has been merged.

PR #1094, opened 2019-03-27
fix export IP Address step

ciaranRoche:
Fixes #(issue-number)
https://github.com/knative/docs/issues/1088
## Proposed Changes
Fixes small error in `export IP_ADDRESS` step which causes an error later on in the sample.


ciaranRoche:
I signed it!

sebgoa:
@samodell I think this can be merged.

PR #1097, opened 2019-03-27
add a link to ftp source

vaikas-google:
## Proposed Changes

- Add a link to proof of concept FTP/SFTP source.
-
-

### DIFF
--- a/docs/eventing/sources/README.md
+++ b/docs/eventing/sources/README.md
@@ -2,7 +2,7 @@
 This is a generated file and should not be changed manually. All changes should follow the
 procedure:
 
-1. Update the information in [`sources.yaml`](sources.yaml).
+1. Update the information in [`sources.yaml`](./sources.yaml).
 
 2. Run the generator tool:
     ```shell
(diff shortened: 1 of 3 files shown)


docs/eventing/sources/README.md
```diff
@@ -2,62 +2,69 @@
 This is a generated file and should not be changed manually. All changes should follow the
 procedure:
 
-1. Update the information in [`sources.yaml`](./sources.yaml).
+1. Update the information in [`sources.yaml`](sources.yaml).
```
RichieEscarez: i see this is only a comment and does not render in GitHub nor the website (so it wont break anything). 
But FYI, due to the Hugo build, all links should be relative and any link to files within the same folder require the " ./ " prefix. 
(i will capture all these requirements in the docs contributor guide and push that out soon)

vaikas-google:
Awesome, thanks, we should then modify our generator script. I'll do that.

docs/eventing/sources/README.md
vaikas-google: ok, I fixed the ReadmeTemplate and reran.

vaikas-google: ok, I fixed the template the generator uses.
@RichieEscarez PTAL

PR #1111, opened 2019-03-29
Improve Java Hello World

### LINKED ISSUE #970: Pin Java examples base docker image
Currently the Java example uses `openjdk:8-jre-alpine`. However, some JVM version will support container support out of the box (8u191 and on), while older version of may require additional flags.

If a wrong version of openjdk:8-jre-alpine was used (because it was cached), then the app may see OOMKilled, since it won't respect cgroup memory limit.

Moreover, openjdk8 official image don't have the latest updates (8u202, for example).

## Recommendation
Change base Java image to AdoptOpenJDK & pin version: `adoptopenjdk/openjdk8:jdk8u202-b08-alpine-slim`

Or, since AdoptOpenJDK images always configures the container support flags, any image versions will work: `adoptopenjdk/openjdk8:alpine-slim`

saturnism:
Fixes #970

## Proposed Changes

- Use AdoptOpenJDK for newest version of JDK that has proper container support.
- Upgraded Spring Boot version.


docs/serving/samples/hello-world/helloworld-java/pom.xml
```diff
@@ -14,7 +14,7 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>2.0.3.RELEASE</version>
+		<version>2.1.3.RELEASE</version>
```
gyliu513: Why upgrade to 2.1.3 directly? What is the versioning guidance for this?

docs/serving/samples/hello-world/helloworld-java/.dockerignore
```diff
@@ -0,0 +1 @@
+target/
```
gyliu513: Does target will be generated during docker build?

docs/serving/samples/hello-world/helloworld-java/pom.xml
saturnism: it's the latest GA version.

docs/serving/samples/hello-world/helloworld-java/.dockerignore
saturnism: this ignores locally generated target/ dir, so we avoid copying unnecessary files to docker daemon during build

saturnism:
not sure why the test failed; is it flaky test? this pr shouldn't affect the tests.

RichieEscarez:
> not sure why the test failed; is it flaky test? this pr shouldn't affect the tests.

the integration tests for samples have only been running for a few weeks now so it might be possible that either the test is failing on something outside of your set of changes or the test itself is broken?

it might be the latter because its not finding the build artifacts:
`W0409 20:33:58.432] Remote dir gs://knative-prow/pr-logs/pull/knative_docs/1111/pull-knative-docs-integration-tests/1115714356525404160/artifacts not exist yet`

@chaodaiG can you please help us determine how to get past the build failure?

chaodaiG:
@RichieEscarez , there was an issue for all knative testing on Tuesday, it was resolved. I just requeued the test, it should pass now

RichieEscarez:
> @RichieEscarez , there was an issue for all knative testing on Tuesday, it was resolved. I just requeued the test, it should pass now

Thanks!

PR #1112, opened 2019-03-29
Fix 401 exception when push image to repository with Kaniko

vincent-pli:
Fixes #(issue-number)
I try to create "builder" make "kaniko" as a step to push image to docker.hub,
then hit exception 401, seems there is no authority to do that.
As we know, the "builder" will launch a "init container" to do that for us after set the "service account".

After some search, I find in "kaniko", it use a env parameter: DOCKER_CONFIG which point to ".docker/config.json". by default, the value is "/kaniko/.docker/".
As we know, the "build" will put the docker authentication file at "/builder/home/.docker/"

So, to make the example work, need add additional config to rewrite the " DOCKER_CONFIG"
## Proposed Changes
/docs/docs/build/builds.md
-
-
-

### DIFF
--- a/docs/build/builds.md
+++ b/docs/build/builds.md
@@ -283,6 +283,9 @@ spec:
       args:
         - --dockerfile=${DOCKERFILE}
         - --destination=${IMAGE}
+      env:
+        - name: "DOCKER_CONFIG"
+          value: "/builder/home/.docker/"
 ```
 
 #### Using a `ServiceAccount`


vincent-pli:
I signed it!

samodell:
@vincent-pli  Can you double-check that the email used to sign your commit is the same as the email you used to sign your CLA?

vincent-pli:
@samodell 
Thanks, I confirmed my mail, actually I guess I sign with a wrong mail.
Now, I correct it, please take a look, thanks

vincent-pli:
I signed it!

PR #964, opened 2019-03-06
Update traffic splitting example to use Release mode

dgerd:
## Proposed Changes

This change updates the traffic splitting example to use release mode
introducing the concepts of the mode in three steps. This change also
simplifies running the commands and removes the need to mutate files on
disk to run through the example.

PR #1116, opened 2019-03-29
Resolving bad merge with cherry-pick (#964)

RichieEscarez:
Update traffic splitting example to use Release mode (#964) - Also applies to v0.3, v0.4

* Update traffic splitting example to use Release mode

This change updates the traffic splitting example to use release mode
introducing the concepts of the mode in three steps. This change also
simplifies running the commands and removes the need to mutate files on
disk to run through the example.

* Use envsubst instead of perl. Simplify image path

* Use long names for arguments

* Update diff statement

* Add envsubst install instructions


docs/serving/samples/traffic-splitting/updated_configuration.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 The Knative Authors
```
dgerd: This file should be deleted. It is the same as `release_sample.yaml` above

RichieEscarez:
duplicate file removed, ptal

PR #1117, opened 2019-03-29
Resolving bad merge in v0.3, with cherry-pick 

RichieEscarez:
Update traffic splitting example to use Release mode (#964)

* Update traffic splitting example to use Release mode

This change updates the traffic splitting example to use release mode
introducing the concepts of the mode in three steps. This change also
simplifies running the commands and removes the need to mutate files on
disk to run through the example.

* Use envsubst instead of perl. Simplify image path

* Use long names for arguments

* Update diff statement

* Add envsubst install instructions

Fixes #(issue-number)

## Proposed Changes

-
-
-


RichieEscarez:
duplicate file removed, ptal

PR #1128, opened 2019-04-02
Update PR template with branch info

samodell:
## Proposed Changes

- Gives info about what branch to open your PR against.
- Explains cherry-picking process

### DIFF
--- a/.github/pull-request-template.md
+++ b/.github/pull-request-template.md
@@ -1,4 +1,11 @@
-Fixes #(issue-number)
+Fixes #issue-number
+
+<!-- Most PRs should be opened against the master branch.
+
+If the change should also be in the most recent release, add the
+cherrypick-0.X label, for example, cherrypick-0.5. If you can, open a PR for the
+cherrypick yourself after your PR has been merged into the master branch. Once
+the cherry-pick PR has merged, remove the cherry-pick label. -->
 
 ## Proposed Changes
 
--- a/.github/pull-request-template.md
+++ b/.github/pull-request-template.md
@@ -1,11 +1,27 @@
+<!-- General PR guidelines:
+
+New contributors:
+
+If you are new to Git/GitHub and want to make a quick fix to the docs,
+open your PR against the release branch where you found the error, such as
+"release-0.5".
+
+Regular contributors:
+
+Most PRs should be opened against the master branch.
+
+If the change should also be in the most recent numbered release, add the
+corresponding "cherrypick-0.X" label; for example, "cherrypick-0.5", to the
+original PR. Best practice is to open a PR for the cherry-pick yourself after
+your original PR has been merged into the master branch. Once the cherry-pick PR
+has merged, remove the cherry-pick label from the original PR.
+
+For more information on contributing to the Knative Docs, see:
+https://www.knative.dev/contributing/docs-contributing/
+
+ -->
+
 Fixes #issue-number
-
-<!-- Most PRs should be opened against the master branch.
-
-If the change should also be in the most recent release, add the
-cherrypick-0.X label, for example, cherrypick-0.5. If you can, open a PR for the
-cherrypick yourself after your PR has been merged into the master branch. Once
-the cherry-pick PR has merged, remove the cherry-pick label. -->
 
 ## Proposed Changes
 


.github/pull-request-template.md
```diff
@@ -1,4 +1,11 @@
-Fixes #(issue-number)
+Fixes #issue-number
+
+<!-- Most PRs should be opened against the master branch.
+
+If the change should also be in the most recent release, add the
```
RichieEscarez:
```suggestion
If the change should also be in the most recent release (master branch), you must add the
```

.github/pull-request-template.md
```diff
@@ -1,4 +1,11 @@
...
+cherrypick-0.X label, for example, cherrypick-0.5. If you can, open a PR for the
```
RichieEscarez:
```suggestion
corresponding "cherrypick-0.X" labels to the PR and indicate which branch(es) your change applies to, for example, "cherrypick-0.5". Best practice is to open a PR for the
```

.github/pull-request-template.md
```diff
@@ -1,4 +1,11 @@
...
+cherrypick yourself after your PR has been merged into the master branch. Once
```
RichieEscarez:
```suggestion
cherrypick yourself after your PR has been merged into the master branch. Make sure to
reference the original change by adding the PR number to each "cherry-pick PR". Once
```

.github/pull-request-template.md
samodell: By the "most recent release," I meant the most recent numbered release, not the master branch.

.github/pull-request-template.md
samodell: I added the "corresponding" wording and the quotes around the specific label names. I would rather keep the "If you can" wording, because I don't want to discourage people who don't feel comfortable with git from contributing.

RichieEscarez: I feel the same too which is why i used "best practice" but I also was to prevent us from having to do the maintenance work (especially if it ends up being P3 and we should probably not spend time on P3 given the backlog of P1). Just thinking long term here (today we definitely want everything shiny and perfect, but I dont think we can keep that up for long).

samodell: Okay, I switched the wording to "best practice."

.github/pull-request-template.md
RichieEscarez: Gotcha, sorry, i see that now.

The "use case" in my mind was a non contributor (so read this too quickly and from a different perspective). For example, I was thinking that this should focus on helping someone who has not read the guidelines and is just trying to fix something in their release (most recent release).

Maybe we should reconsider the guidance around "open the PR against Master and the cherrypick it into a branched release"? If someones goal is to fix something in a release they are using, and they never get to cherrypicking a change from master into a branch-release, then the overhead to "contributing a quick fix" might be more than what "today's usual contributor" is (ie today we mostly get PRs from developers working on code for v.Next). 

Thoughts?

samodell: I think the reasoning here is that, if you commit into a numbered release instead, and it's a change that should be carried forward, the user then needs to cherry-pick the change into master. In theory, most changes made probably should be carried forward, so I think we'd be doing a lot more cherry-picking overall if we directed people to make commits into a release branch, since not all changes need to be carried back into the last release.

I also think the current workflow is aligned with what the other repos in Knative do, and I get the feeling it's a pretty common Git workflow.

.github/pull-request-template.md
```diff
@@ -1,4 +1,22 @@
-Fixes #(issue-number)
+<!-- PR tips
```
RichieEscarez:
```suggestion
<!-- General PR guidelines:
```

.github/pull-request-template.md
```diff
@@ -1,4 +1,22 @@
...
+
+New contributors:
+
+If you are new to Git/GitHub and want to make a quick fix to the docs,
+open your PR against the release branch where you found the error, such as
+"release-0.5".
+
+Experienced contributors:
```
RichieEscarez:
```suggestion
Experienced/regular contributors:
```
Maybe something along these lines to qualify that you dont need to have a "level of experience" but rather this is for someone who regularly contributes to knative?

.github/pull-request-template.md
```diff
@@ -1,4 +1,22 @@
...
+
+Most PRs should be opened against the master branch.
+
+If the change should also be in the most recent numbered release, add the
+corresponding "cherrypick-0.X" label; for example, "cherrypick-0.5." Best
```
RichieEscarez:
```suggestion
corresponding "cherrypick-0.X" label; for example, add "cherrypick-0.5" to PR-original. Best
```
Something to add clarity between all the references to PR?

.github/pull-request-template.md
```diff
@@ -1,4 +1,22 @@
...
+practice is to open a PR for the cherry-pick yourself after your PR has been
```
RichieEscarez:
```suggestion
practice is to open a subsequent pull request, PR-cherrypick, using the merged commit number of PR-original.
```

.github/pull-request-template.md
```diff
@@ -1,4 +1,22 @@
...
+merged into the master branch. Once the cherry-pick PR has merged, remove the
```
RichieEscarez:
```suggestion
Once PR-cherrypick is merged, you remove the "cherrypick-0.X" label from PR-original.
```

.github/pull-request-template.md
```diff
@@ -1,4 +1,22 @@
...
+cherry-pick label. -->
```
RichieEscarez:
```suggestion
Learn more: https://www.knative.dev/contributing/docs-contributing/ -->
```

.github/pull-request-template.md
samodell: Added.

.github/pull-request-template.md
samodell: Changed to just "regular"

.github/pull-request-template.md
samodell: I added some wording to try to accomplish this.

.github/pull-request-template.md
samodell: I don't think this level of detail belongs in the note. I opened issue #1136 to add information about cherry-picking PRs back into releases that includes where to get the commit number from.

.github/pull-request-template.md
samodell: Added this link.

samodell:
See what you think, @RichieEscarez ! Rewrote with the suggestions you made in mind.

PR #1131, opened 2019-04-02
Mention IKS's one-click install process

duglin:

### DIFF
--- a/docs/install/Knative-with-IKS.md
+++ b/docs/install/Knative-with-IKS.md
@@ -6,13 +6,7 @@ type: "docs"
 ---
 
 This guide walks you through the installation of the latest version of Knative
-using pre-built images.
-
-You may also have it all installed for you by clicking the button below:
-[![Deploy to IBM Cloud](https://bluemix.net/deploy/button_x2.png)](https://console.bluemix.net/devops/setup/deploy?repository=https://git.ng.bluemix.net/start-with-knative/toolchain.git)
-
-More
-[instructions on the deploy button here](https://git.ng.bluemix.net/start-with-knative/toolchain/blob/master/README.md).
+on an IBM Cloud Kubernetes Service (IKS) cluster.
 
 You can find [guides for other platforms here](./README.md).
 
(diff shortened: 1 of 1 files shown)


RichieEscarez:
Thanks for the contribution!

From just a quick scan of the change, i noticed a typo but more importantly, this documentation should probably have two sections to make clear that there are two separate paths.

For example use an H2 title, "Installing Knative using an IKS managed add-on" and another H2 title: "Manually installing Knative on IKS" (where both have their own Before you begin sections, in which is where you can explain how to get the add-on before the steps on how to use it).

sebgoa:
I agree with Richie.

I would modify the intro a tidbit to have a bullet list with the two install method (add-on and manual) and then make the add-on install on-par with the manual using the same header...

otherwise /lgtm

We should really think about extracting the common install steps that are in all the install guides.

sebgoa:
Oh...and you probably want to move to 0.5.0

duglin:
sorry for the delay - see what you think about the updates.

docs/install/Knative-with-IKS.md
```diff
@@ -131,7 +125,28 @@ components, the recommended configuration for a cluster is:
...
+
+The easiest way to install it is using the Managed Knative add-on facility.
+This one-click install process will install Knative, and Istio if not already
+installed, and provide automatic updates and lifecycle management of your
+Knative control plane.
+
+You can get the add-on via the "Add-ons" tab of your Kubernetes cluster's
+console page, or via the command line:
+```bash
+bx ks cluster-addon-enable -y knative $CLUSTER_NAME
```
sebgoa: how do we get `bx` CLI ?

docs/install/Knative-with-IKS.md
```diff
@@ -131,7 +125,28 @@ components, the recommended configuration for a cluster is:
...
+```
+
+For more information about the add-on see
+[here](https://cloud.ibm.com/docs/containers?topic=containers-knative_tutorial#knative_tutorial).
+
+## Manually installing Knative on IKS 
+
```
sebgoa: I would add a little intro sentence here :)

docs/install/Knative-with-IKS.md
duglin: thanks for this!  it should be `ibmcloud` not `bx` (that's old).

docs/install/Knative-with-IKS.md
duglin: Done! see what you think.

PR #1187, opened 2019-04-16
Update READMEs to make community samples discoverable.

mhamwala:
Fixes #issue-number

## Proposed Changes

- This PR updates the README docs/docs/serving/samples/ to make community samples discoverable.
- Adds a note to the README in docs/community/samples/

PR #1197, opened 2019-04-18
cherrypick: community samples discoverable. (#1187)

mhamwala:
Fixes #issue-number

## Proposed Changes

- This PR updates the README docs/docs/serving/samples/ to make community samples discoverable.
- Adds a note to the README in docs/community/samples/


samodell:
Thanks for helping out with the cherry-picking, @mhamwala ! We are just now working on adding this guidance to the contributor guidelines, but it's best practice to only update the most recent numbered release, so in this case, just 0.5. We want the older releases to reflect the documentation at that point in time, and for the most part we probably won't update older releases unless an update is super critical.

mhamwala:
ahh okay that makes sense! no worries!

PR #1206, opened 2019-04-18
docs: Remove `--save` from `npm install`

grant:
## Proposed Changes

- Remove `--save` from `npm install` script.

See https://stackoverflow.com/a/19578808


samyhm2000:
Please unsubscribe me from this thread.


Sameer Madamaiah

PR #1207, opened 2019-04-19
update sample.yaml to fix the build-template name to buildpacks-cnb

paassdc:
Fixes #issue-number

according the buildpacks build-templates name in cnb.yaml updated to 'buildpacks-cnb', the build-template name in buildpacks sample.yaml should update to 'buildpacks-cnb' as well

## Proposed Changes
sample.yaml
-
-
-


RichieEscarez:
thanks! 

https://github.com/knative/build-templates/blob/master/buildpacks/cnb.yaml

PR #1200, opened 2019-04-18
Fix some spelling mistakes in docs

SataQiu:
In the process of reading, I try to fix these spelling mistakes.

PR #1209, opened 2019-04-19
Cherrypick: fix some spelling mistakes in docs (#1200)

SataQiu:
Cherrypick: fix some spelling mistakes in docs (#1200)


ZhiminXiang:
we probably don't need to cherrypick it. We can let the fix go to release 0.6.

PR #1243, opened 2019-04-28
Remove references to Istio yamls attached to releases.

mattmoor:
In 0.6 we will stop shipping Istio yamls, as we are not in the business of
redistributing Istio (incl. supporting these yamls).  In preparation for that,
this change makes a few changes:
1. Update things to point to the 0.5.2 release
1. Point to our in-tree Istio files (which is what we were attaching)
1. For GKE, direct users to the Istio addon instead.
1. For GKE, add a pointer to CR-GKE (if folks wanted hosted).

This is a half-measure to avoid things breaking.  We need a page that points
folks to the Istio docs, and calls out any specialized instructions for
what Knative needs.

Fixes: https://github.com/knative/serving/issues/3670

### DIFF
--- a/docs/install/Knative-custom-install.md
+++ b/docs/install/Knative-custom-install.md
@@ -83,9 +83,9 @@ service mesh. If you install any of the following options, you must install
 
 † These are the recommended standard install files suitable for most use cases.
 
-[a]: https://github.com/knative/serving/releases/download/v0.5.0/istio-crds.yaml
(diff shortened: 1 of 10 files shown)


mattmoor:
/hold

I was too aggressive about `0.5.2` rewrites, will revert the build/eventing paths

ZhiminXiang:
This PR has merge conflicts. Please rebase.

mattmoor:
Rebased.

PR #1248, opened 2019-04-29
Add validation token support to autoscaler-go

rgregg:
## Proposed Changes

We wanted to use the autoscaler-go sample to do some load testing. However many load testing services require the target to return a specific string to verify that they should be load tested (vs. a DDoS attack).

This update uses the **VALIDATION** env-var to define the string that should be returned from a URL that includes that token, and appears to be a common format for validating a URL from these types of sites.

This PR also fixes that the sample listens only on port 8080, which is not the desired experience in the container runtime contract.

### DIFF
--- a/docs/serving/samples/autoscale-go/autoscale.go
+++ b/docs/serving/samples/autoscale-go/autoscale.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"sync"
 	"time"
+	"os"
 )
 
 // Algorithm from https://stackoverflow.com/a/21854246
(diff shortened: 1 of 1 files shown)


rgregg:
Overriding CLA, since I messed up my github author settings.

grayside:
Is this token considered a secret? Might be worth clarifying that in a comment.

rgregg:
I don't think they're considered a secret since they are frequently stored in DNS TXT entries. I'll add a comment to that effect.

PR #1250, opened 2019-04-30
Correct traffic splitting in version 0.5

vincent-pli:
Fixes #issue-number

## Proposed Changes
To update the original ksvc from `runLatest` to `release` should use `kubectl replace` rather than `kubectl apply`, if not, the command will raise exception: `expected exactly one, got both: runLatest release`


vincent-pli:
I signed it!

PR #1251, opened 2019-04-30
Correct traffic splitting

vincent-pli:
Fixes #issue-number

## Proposed Changes
Update ksvc from `runLatest` to `release` should use `kubectl replace` rather than `kubectl apply`, 
or will get exception: `expected exactly one, got both: runLatest, release`


vincent-pli:
I believe I have already signed the cla, that's weird

PR #1260, opened 2019-05-01
Document serving tag resolution

jonjohnsonjr:
Our controller does resolution of tags to digests, which has been a
source of confusion. This documents the fact that we do it, why we do
it, and how to configure the controller to work around common issues.

### DIFF
--- a/docs/serving/tag-resolution.md
+++ b/docs/serving/tag-resolution.md
@@ -1,0 +1,95 @@
+---
+title: "Enabling tag to digest resolution"
+linkTitle: "Tag resolution"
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
cc @mattmoor anything else you can think of for this?

mattmoor:
/lgtm

@samodell Should we link this from somewhere?

samodell:
/approve

We can definitely add links to this from relevant content, but I do want to get the content merged in now. It will be automatically linked from the table of contents on the side of the pages, at a minimum.

Issue #299, opened 2018-07-30
Could Istio maybe be installed through official channels for that project?


krancour:
Is there a reason we "repackage" Istio in [this manifest](https://storage.googleapis.com/knative-releases/serving/latest/istio.yaml)?

imo, unless a very specific Istio configuration is required, it should be installable via the official channels for that project instead-- e.g. Istio's Helm chart.

Even if Istio _does_ require a very specific configuration, I would suggest that documenting that is preferable to re-packaging Istio, which leaves users to speculate how that Istio installation differs from the norm.

sslavic:
From last week's TGIK session:
- one can find two `PATCH`es labeled in https://storage.googleapis.com/knative-releases/serving/latest/istio.yaml
  - first patch adds istio-system namespace, to minimize number of installation steps because compared to Helm chart it's possible with single plain yaml to install namespace and reference it too
  - second patch is for https://github.com/knative/serving/issues/1370
- Knative's Istio plain yaml for the most part (everything but two patches) is generated from upstream Istio Helm chart but Helm is not used for deployment e.g. Tiller is hard to install in a secure way

tcnghia:
Agreed.  Being pluggable means we should allow users to customize their own installation of Istio, and if a customer already has their own way of installing Istio they don't need to reinstall based on what we provide.

That said, we can see the Istio we provide as an example setting for a quick start experience, so our users don't need to first install Helm and read many Istio getting started doc to figure out what options need to be passed to Helm.

In the `repackaging` we documented all the Helm options we used to generate the YAML, I think we could bring similar docs in the `knative/docs` repo (as you suggested).  We should fix knative/serving#1370 so the entire thing could be done within a single Helm command.

krancour:
> Tiller is hard to install in a secure way

Absolutely no disagreement from me on that point.

I'm more worried about drift, and as you've mentioned, pluggability. byo-Istio should not only be possible (it is-- I've done it), but also considered a likely scenario. Istio's about to go 1.0.0 and Knative is still very young. It won't be long before we start encountering clusters in the wild that already have Istio installed to their liking.

Maybe after #1370 is fixed, could the possibility of drift be mitigated for now (while avoiding Tiller security issues) by doing this instead of repackaging?

```
$ kubectl create ns istio-system
$ helm template install/kubernetes/helm/istio --name istio --namespace istio-system --set | kubectl apply -f -
```

krancour:
Also see https://github.com/istio/istio/issues/7505

[...]

Issue #1244, opened 2019-04-28
Istio install page


mattmoor:
In 0.6 we will stop shipping Istio (what we use is customized for development in ways that are unsuitable for production).

Vendors with hosted Istio offerings should prefer those in their docs (e.g. [GKE docs here](https://github.com/knative/docs/pull/1243)), but we should have a page that describes:
1. Where to find Istio's installation guide
2. How to customize Istio for use with Knative.

Assigning to @ZhiminXiang as the TLS work is going to affect this (and we'll probably need a similar page for cert-manager).

sebgoa:
Hum, please provide detailed instructions not just a redirect.

Users should not have to be bounced to another project and have to read the documentation of another project to get a working knative.

I understand the production setup argument, but for first installs/testing, there should be as few `kubectl apply` as possible and as few doc hunting as possible.

PR #1272, opened 2019-05-05
Doc about installing Istio from official source

ZhiminXiang:
Fixes #issue-number
https://github.com/knative/docs/issues/1244

## Proposed Changes
A new doc about installing custom Istio.

More context: we will not ship Istio with Knative in Knative 0.6 and after. Therefore, we need a doc for users to install custom Istio for the use with Knative

### DIFF
--- a/docs/serving/installing-istio.md
+++ b/docs/serving/installing-istio.md
@@ -1,0 +1,139 @@
+---
+title: "Performing a Custom Istio Installation"
+weight: 15
+type: "docs"
+---
+
+Use this guide to perform a custom installation of Istio for use with Knative.
+
+## Download Istio
+
+Run below command to download Istio.
+```shell
+# Download and unpack Istio
+ISTIO_VERSION=1.1.3
+DOWNLOAD_URL=https://github.com/istio/istio/releases/download/${ISTIO_VERSION}/istio-${ISTIO_VERSION}-linux.tar.gz
+
+wget $DOWNLOAD_URL
+tar xzf istio-${ISTIO_VERSION}-linux.tar.gz
+
+cd istio-${ISTIO_VERSION}
+```
+
+## Install Istio CRDs 
+Default Istio Installation
+Run below command to install Istio CRDs first.
+```shell
+for i in install/kubernetes/helm/istio-init/files/crd*yaml; do kubectl apply -f $i; done
+```
+wait a few seconds for the CRDs to be committed in the Kubernetes API-server
+
+## Custom Installation
+
+### Istio with Sidecar Injector
+If you need sidecar injection in your cluster, sidecar injector and related 
+configurations are needed in your Istio. Run below command to install the 
+custom Istio.
+```shell
+# A template with sidecar injection enabled.
+helm template --namespace=istio-system \
+  --set sidecarInjectorWebhook.enabled=true \
+  --set sidecarInjectorWebhook.enableNamespacesByDefault=true \
+  --set global.proxy.autoInject=disabled \
+  --set global.disablePolicyChecks=true \
+  --set prometheus.enabled=false \
+  `# Disable mixer prometheus adapter to remove istio default metrics.` \
+  --set mixer.adapters.prometheus.enabled=false \
+  `# Disable mixer policy check, since in our template we set no policy.` \
+  --set global.disablePolicyChecks=true \
+  `# Set gateway pods to 1 to sidestep eventual consistency / readiness problems.` \
+  --set gateways.istio-ingressgateway.autoscaleMin=1 \
+  --set gateways.istio-ingressgateway.autoscaleMax=1 \
+  --set gateways.istio-ingressgateway.resources.requests.cpu=500m \
+  --set gateways.istio-ingressgateway.resources.requests.memory=256Mi \
+  `# Enable SDS in the gateway to allow dynamically configuring TLS of gateway.` \
+  --set gateways.istio-ingressgateway.sds.enabled=true \
+  `# More pilot replicas for better scale` \
+  --set pilot.autoscaleMin=2 \
+  `# Set pilot trace sampling to 100%` \
+  --set pilot.traceSampling=100 \
+  install/kubernetes/helm/istio \
+  `# Removing trailing whitespaces to make automation happy` \
+  | sed 's/[ \t]*$//' \
+  > ./istio.yaml
+
+  kubectl apply -f istio.yaml
+  ```
+
+### Istio with no Sidecar Injector
+If you don't need sidecar injection in your cluster, you can install an 
+Istio without sidecar injector. Run below command to install the custom Istio.
+```shell
+# A lighter template, with no sidecar injection.
+helm template --namespace=istio-system \
+  --set sidecarInjectorWebhook.enabled=false \
+  --set global.proxy.autoInject=disabled \
+  --set global.omitSidecarInjectorConfigMap=true \
+  --set global.disablePolicyChecks=true \
+  --set prometheus.enabled=false \
+  `# Disable mixer prometheus adapter to remove istio default metrics.` \
+  --set mixer.adapters.prometheus.enabled=false \
+  `# Disable mixer policy check, since in our template we set no policy.` \
+  --set global.disablePolicyChecks=true \
+  `# Set gateway pods to 1 to sidestep eventual consistency / readiness problems.` \
+  --set gateways.istio-ingressgateway.autoscaleMin=1 \
+  --set gateways.istio-ingressgateway.autoscaleMax=1 \
+  `# Set pilot trace sampling to 100%` \
+  --set pilot.traceSampling=100 \
+  install/kubernetes/helm/istio \
+  `# Removing trailing whitespaces to make automation happy` \
+  | sed 's/[ \t]*$//' \
+  > ./istio-lean.yaml
+
+kubectl apply -f istio-lean.yaml
+```
+
+### Install Cluster Local Gateway
+If you want to use the feature of [cluster local route](https://github.com/knative/docs/blob/master/docs/serving/cluster-local-route.md), an extra Istio
+cluster local gateway needs to be added into your cluster. Run below command
+to install the cluster local gateway.
+```shell
+# Generate the extra gateway.
+helm template --namespace=istio-system \
+  --set gateways.custom-gateway.autoscaleMin=1 \
+  --set gateways.custom-gateway.autoscaleMax=1 \
+  --set gateways.custom-gateway.cpu.targetAverageUtilization=60 \
+  --set gateways.custom-gateway.labels.app='cluster-local-gateway' \
+  --set gateways.custom-gateway.labels.istio='cluster-local-gateway' \
+  --set gateways.custom-gateway.type='ClusterIP' \
+  --set gateways.istio-ingressgateway.enabled=false \
+  --set gateways.istio-egressgateway.enabled=false \
+  --set gateways.istio-ilbgateway.enabled=false \
+  install/kubernetes/helm/istio \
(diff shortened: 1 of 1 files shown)


samodell:
Thanks @ZhiminXiang ! I'm taking a look at this now.

docs/serving/installing-istio.md
```diff
@@ -0,0 +1,139 @@
...
+  install/kubernetes/helm/istio \
+  `# Removing trailing whitespaces to make automation happy` \
+  | sed 's/[ \t]*$//' \
+  > ./istio-lean.yaml
+
+kubectl apply -f istio-lean.yaml
+```
+
+### Install Cluster Local Gateway
+If you want to use the feature of [cluster local route](https://github.com/knative/docs/blob/master/docs/serving/cluster-local-route.md), an extra Istio
```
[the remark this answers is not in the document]
samodell: Can you add in a quick note about why someone might want to use cluster local routes?

docs/serving/installing-istio.md
```diff
@@ -0,0 +1,139 @@
...
+ISTIO_VERSION=1.1.3
+DOWNLOAD_URL=https://github.com/istio/istio/releases/download/${ISTIO_VERSION}/istio-${ISTIO_VERSION}-linux.tar.gz
+
+wget $DOWNLOAD_URL
+tar xzf istio-${ISTIO_VERSION}-linux.tar.gz
+
+cd istio-${ISTIO_VERSION}
+```
+
+## Install Istio CRDs
```
samodell: Just to make sure I understand, after this step for installing the CRDs, you then have to go and follow the instructions for either "Istio with Sidecar Injector" or "Istio with no Sidecar Injector," right? You're definitely not done at this point?

docs/serving/installing-istio.md
```diff
@@ -0,0 +1,139 @@
...
+  | sed "s/[[:space:]]*$//" \
+  > ./istio-local-gateway.yaml
+
+kubectl apply -f istio-local-gateway.yaml
+```
+
+### Istio with Secret Discovery Service
+[Secret Discovery Service](https://istio.io/docs/tasks/traffic-management/secure-ingress/sds/) is needed if you want to dyanmically update your Gateway 
+with multiple TLS certificates to terminate TLS connection. The below`helm` flag is needed in your `helm` command to enable SDS.
+```
```
samodell: Can you show the full `helm` command? Showing just this snippet is a bit confusing.

docs/serving/installing-istio.md
```diff
@@ -0,0 +1,139 @@
+---
+title: "Performing a Custom Istio Installation"
+weight: 15
+type: "docs"
+---
+
+Use this guide to perform a custom installation of Istio for use with Knative.
+
+## Download Istio
```
samodell: Before this section, can we add a ## Before you begin section like this:
```
## Before you begin

You need:
- A Kubernetes cluster created
- `helm` installed
```

sebgoa:
I am really not a big fan of this.

We should discuss this at the serving wg or toc cc/ @mattmoor 

Unless istio is now 100% optional.

If we still need a bit of istio to get all Knative components to work, we should bundle a batteries including release in our own release of knative.

This doc is forcing people to use Helm, to clone an istio release and also learn a few startup options. this is a big ask for getting started.

I do understand we cannot provide a production ready install in our bundle but we need to be conscious of getting started efforts.

docs/serving/installing-istio.md
```diff
@@ -0,0 +1,139 @@
...
+
+Run below command to download Istio.
+```shell
+# Download and unpack Istio
```
samodell: This all seems a little complicated; is there a reason not to use the command istio suggests:

```
curl -L https://git.io/getLatestIstio | ISTIO_VERSION=1.1.5 sh -
```
From: https://istio.io/docs/setup/kubernetes/download/

docs/serving/installing-istio.md
```diff
@@ -0,0 +1,139 @@
...
+Run below command to install Istio CRDs first.
+```shell
+for i in install/kubernetes/helm/istio-init/files/crd*yaml; do kubectl apply -f $i; done
+```
+wait a few seconds for the CRDs to be committed in the Kubernetes API-server
+
+## Custom Installation
+
+### Istio with Sidecar Injector
+If you need sidecar injection in your cluster, sidecar injector and related
```
samodell: How will people know if they need sidecar injection or not? Let's add some guidance here around that.

docs/serving/installing-istio.md
```diff
@@ -0,0 +1,139 @@
...
+Default Istio Installation
+Run below command to install Istio CRDs first.
+```shell
+for i in install/kubernetes/helm/istio-init/files/crd*yaml; do kubectl apply -f $i; done
+```
+wait a few seconds for the CRDs to be committed in the Kubernetes API-server
+
+## Custom Installation
+
+### Istio with Sidecar Injector
```
samodell: To @sebgoa 's point, can we offer a recommend install (without using helm) that will be appropriate for most use cases? Having to make a choice about what Istio to install before even attempting to install Knative could really discourage people; we should provide a clear, easy path to get started.

samodell:
Fixes #299

mattmoor:
@sebgoa Ack, this is intended as the option that works basically anywhere as a worst-case fallback.  Ultimately it is a non-goal of Knative to fix installing Istio.

That said, the hope for Istio is the same as for K8s: we should bias folks towards managed Istio products wherever we can, as we do at the K8s level for GKE, IKS, AKS, EKS, ...

If you look at the story for GKE (now), you basically just add `--addons=Istio` and you get a managed form of this that's actually intended for production use and has a support story when vulnerabilities are found.

What we have in knative/serving isn't intended for production, quite the opposite, it's designed for development and has some knobs turned that make it dangerous to use for production (e.g. disabling HPA on the gateways).

---
re: istio being 100% optional

In 0.6, you can install Knative with:
```
kubectl apply -f serving.yaml -l networking.knative.dev/ingress-provider!=istio
```

The label selector will complete strip out Istio and you can replace it with whatever you want.  There is Gloo, and IIRC @bbrowning has an Ambassador prototype.  I think @tcnghia is going to demo one of these at the Serverless Practitioners Summit before Kubecon.

docs/serving/installing-istio.md
ZhiminXiang: Thanks for the suggestion! Added more context about this.

docs/serving/installing-istio.md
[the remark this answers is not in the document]
ZhiminXiang: Done

docs/serving/installing-istio.md
ZhiminXiang: Added an example of installing Istio with ingress `SDS` and sidecar injector here.

samodell:
Hm, I missed those updates to the GKE install page getting merged in. That does work really well for people on GKE, but how many of the other platforms have similarly easy to set up managed Istio offerings? 

For now, I think we'll have to pull out the old instructions directing users to install our provided version of Istio and replace them with a link to this guide about installing Istio. (For example, the PKS guide still points people to install our Istio files: https://knative.dev/development/install/knative-with-pks/#installing-istio

FYI @mattmoor  and @ZhiminXiang

docs/serving/installing-istio.md
```diff
@@ -0,0 +1,139 @@
+---
+title: "Performing a Custom Istio Installation"
```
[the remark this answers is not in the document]
samodell: But this guide will also have to work for the other install guides, for folks not on GKE -- pretty much all the install guides had steps for installing Istio using our provided files. For example, PKS: https://knative.dev/development/install/knative-with-pks/#installing-istio

Right now, I'm thinking we'll link to this guide for installing Istio from the other main install guides that were installing Isito that way, meaning for a lot of folks, they might not be "customizing Istio," they'll just be installing it for use with Knative. I'll go through and suggest some changes that might work.

docs/serving/installing-istio.md
samodell: Hm, even in that IKS doc, the "manual instructions" point to installing Istio using our provided files, which are going away. From that place in the guide, I'd like to point users who want to perform a manual install to this page to install Istio.

ZhiminXiang:
> Hm, I missed those updates to the GKE install page getting merged in. That does work really well for people on GKE, but how many of the other platforms have similarly easy to set up managed Istio offerings?
> 
> For now, I think we'll have to pull out the old instructions directing users to install our provided version of Istio and replace them with a link to this guide about installing Istio. (For example, the PKS guide still points people to install our Istio files: https://knative.dev/development/install/knative-with-pks/#installing-istio
> 
> FYI @mattmoor and @ZhiminXiang

Thanks for the suggestion @samodell ! How about keeping the old instructions and also providing a link of this doc in each Knative installation page.
1. The old instruction is used as a standard and easy installation for users who just want to have a quick setup. And this part is maintained by the contributors who are familiar with the platform. They could replace it with a platform-specific instruction, or any instruction working best with the platform.
2. The instructions we provide in this PR can be used as a detailed/fallback guide for users who have some knowledge about Istio and want to use some custom/advanced features of Istio with Knative.

WDYT?

docs/serving/installing-istio.md
ZhiminXiang: The Istio file (https://raw.githubusercontent.com/knative/serving/v0.5.2/third_party/istio-1.0.7/istio.yaml) won't go away as it is a part of v0.5.2 which has been released. So the instruction in IKS will still be valid.
But we won't have any Istio yaml in our 0.6.0 release.

samodell:
@ZhiminXiang  -- Since this doc is going to be merged into master (which will be v0.6 when the cut is made next Tuesday) I think it makes sense to go ahead and pull out the references to our own releases of Istio here. I'll make some suggestions to that effect.

docs/serving/installing-istio.md
```diff
@@ -0,0 +1,139 @@
...
+---
+
+Use this guide to perform a custom installation of Istio for use with Knative.
+
+## Download Istio
+
+Run below command to download Istio.
+```shell
+# Download and unpack Istio
+ISTIO_VERSION=1.1.3
```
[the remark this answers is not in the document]
sebgoa: @samodell you can define a var without export, see: https://stackoverflow.com/questions/1158091/defining-a-variable-with-or-without-export

I think that's what you "asked"

sebgoa:
While managed products have big advantages, the big advantage of open source tech is that you can install it wherever you want and manage it the way you want.

Some vendors definitely want to attract people to their managed offering ( : ) ) but as an open source project I disagree with Knative trying to bias people one way or another.

I would be (would have been) in favor to keep the `istio.yaml` for development with big caveats so that users understand this is not for production, then link to a more complete install.

Right now, this will tell people to either 1-adopt a managed istio solution or 2-use Helm (yet another tool).

At a minimum I would like to see:

`helm template ... | kubectl apply -f -` so that we don't tell people run tiller.

sebgoa:
I will take that:

> At a minimum I would like to see:

> helm template ... | kubectl apply -f - so that we don't tell people run tiller.

back.

That's what the docs tells people to do. So that.s good.

docs/serving/installing-istio.md
samodell: Yes, and since we won't have Istio .yaml in the 0.6 release, we need to update all the docs for 0.6 to point to these manual instructions for install, from the various install guides where no managed option exists. We won't point back to the 0.5.2 istio .yamls from the 0.6 docs. That's something I can take care of after we merge this page in, though.

samodell:
Merging this in so that I can open my own PR and do a few edits. Thanks for the great start, @ZhiminXiang !

/approve
/lgtm

PR #1273, opened 2019-05-06
address wrong manifest in gcppubsub sample

### LINKED ISSUE #1264: bad links/missing files in pubsub eventing example
The 'verify' section of https://github.com/knative/docs/tree/release-0.5/docs/eventing/samples/gcp-pubsub-source#verify as well as https://github.com/knative/docs/tree/master/docs/eventing/samples/gcp-pubsub-source has some issues. 

In the 0.5 version, the readme says: 

> We will verify that the published message was sent into the Knative eventing system by looking at what is downstream of the GcpPubSubSource. If you deployed the Subscriber, then continue using this section. If not, then you will need to look downstream yourself.

However, the link is bad: there's no Subscriber section.

In the master version, the text is:

> The function and the subscription were created by applying the subscriber.yaml manifest in the deployment section above.

However, there's no subscriber.yaml (as far as I can tell ..)

[This example appeared to deploy okay for me, but AFAICT is not actually working in the sense of responding to new pubsub data...  haven't dug into it yet but given the above I'm wondering if stuff is missing].

sebgoa:
This fixes #1264 issue in master.

Do we want to fix the release-0.5 as well ?

### DIFF
--- a/docs/eventing/samples/gcp-pubsub-source/README.md
+++ b/docs/eventing/samples/gcp-pubsub-source/README.md
@@ -76,9 +76,9 @@ source is most useful as a bridge from other GCP services, such as
    namespace `default`, feel free to change to any other namespace you would
    like to use instead:
 
-```shell
-kubectl label namespace default knative-eventing-injection=enabled
-```
+    ```shell
+    kubectl label namespace default knative-eventing-injection=enabled
+    ```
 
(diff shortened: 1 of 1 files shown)


samodell:
/lgtm
/approve

I'd go with "yes" for the 0.5 question. Added the cherry-pick label.

PR #1277, opened 2019-05-06
fix broken demo

matzew:
@mattmoor  the service is call `event-display`


matzew:
@RichieEscarez mind checking here ?

PR #1289, opened 2019-05-08
Correct syntax of build example

johscheuer:
This PR corrects the build example which defines two build steps with the previous version only one step will be executed (this is actually an bug in knative build, I will create another issue in the according repo).


johscheuer:
I will fix the other build docs too

sebgoa:
that's correct, good catch

/lgtm

/ok-to-test

johscheuer:
I finally reviewed and corrected all build examples

PR #1296, opened 2019-05-09
eventing/README: add KafkaSource net attribute

danp:
Followup to https://github.com/knative/eventing-sources/pull/351

## Proposed Changes

- Mention the KafkaSource net attribute in the eventing README


csantanapr:
thank you @danp for this 🎉

PR #1297, opened 2019-05-10
Simplify cURL invocation

JRBANCEL:
Is there an intent behind setting the Host header manually and targeting the IP address instead of targeting the host directly?
If not it can be simplified.

### DIFF
--- a/docs/install/getting-started-knative-app.md
+++ b/docs/install/getting-started-knative-app.md
@@ -146,20 +146,19 @@ assigned an external IP address.
    you entered.
 
 1. Now you can make a request to your app and see the results. Replace
-   `IP_ADDRESS` with the `EXTERNAL-IP` you wrote down, and replace
    `helloworld-go.default.example.com` with the domain returned in the previous
    step.
 
    ```shell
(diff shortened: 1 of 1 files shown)


samodell:
Hm, interesting -- I don't know if there was a reason why we went with the original command, but hopefully someone on the eng side can chime in on this.

sebgoa:
`example.com` does not resolve

You hit the ingress on its IP address and pass the actual hostname in the Host header.

I don't see how your changes would actually work.

JRBANCEL:
Knative sets up a xip.com domain by default now: https://github.com/knative/serving/issues/2968

Therefore, the xip.com based hostname will resolve to the cluster IP properly.
Is there a scenario where this would not be true?

sebgoa:
afaic tell, the PR is https://github.com/knative/serving/pull/3559 which adds a post-install job to setup xip.io in the `config-domain` configmap. This gets used in development with `ko apply config/post-install`

I am not quite sure that this will be part of the released bundle for generic install.

JRBANCEL:
You are correct. This is not the default behavior.
I am closing this PR for now.

PR #1281, opened 2019-05-07
Add the doc about installing Cert-Manager and enabling Auto TLS

ZhiminXiang:
## Proposed Changes
Add the doc about installing Cert-Manager and enabling Auto TLS

PR #1314, opened 2019-05-14
Istio updates

### LINKED ISSUE #1244: Istio install page

samodell:
Fixes #1244 

## Proposed Changes

- Adds a set of "default" instructions
- Adds numbered steps
- Moves installing-istio.md file from /serving/ to /install/


docs/install/installing-istio.md
```diff
@@ -0,0 +1,281 @@
...
+   ```
+
+1. Enter the following command to install the Istio CRDs first:
+   ```shell
+   for i in install/kubernetes/helm/istio-init/files/crd*yaml; do kubectl apply -f $i; done
+   ```
+   Wait a few seconds for the CRDs to be committed in the Kubernetes API-server,
+   then continue with these instructions.
+
+1. Enter the following command to install Istio:
```
samodell: @ZhiminXiang 

Is this the basic Istio installation we should recommend?

Bonus points -- is there a way to do our "basic" installation without using `helm`?

docs/install/installing-istio.md
```diff
@@ -0,0 +1,281 @@
...
+This default installation enables [automatic sidecar injection][1].
+
+## Customizing your installation
+
+You can easily customize your Istio installation with `helm`. The below sections
+cover a few useful customizations and their purpose.
+
+### Installing Istio with sidecar injection
+
+If you need Istio service mesh, and want to enable it by
```
samodell: @ZhiminXiang 
can we add some info on why someone might need Istio service mesh? Would be helpful to give an example use case for doing so.

docs/install/installing-istio.md
```diff
@@ -0,0 +1,281 @@
...
+    `# Removing trailing whitespaces to make automation happy` \
+    | sed 's/[ \t]*$//' \
+    > ./istio-lean.yaml
+
+  kubectl apply -f istio-lean.yaml
+  ```
+
+### Installing Istio with Secret Discovery Service
+
+[Secret Discovery Service](https://istio.io/docs/tasks/traffic-management/secure-ingress/sds/)
```
samodell: @ZhiminXiang 
Similar point to above; can we give a use case example for why someone might want to use SDS?

docs/install/installing-istio.md
```diff
@@ -0,0 +1,281 @@
...
+  | sed 's/[ \t]*$//' \
+  > ./istio.yaml
+
+  kubectl apply -f istio.yaml
+
+```
+
+### Updating your install to use cluster local gateway
+
+If you want your Routes to be visible only inside the cluster, you may
```
samodell: @ZhiminXiang 
Is this an "update"? IE, if you run this command when you already have Istio installed, will it update your install to add a new gateway?

docs/install/installing-istio.md
```diff
@@ -0,0 +1,281 @@
...
+type: "docs"
+---
+
+This guide walks you through manually installing and customizing Istio for use
+with Knative.
+
+If your cloud platform offers a managed Istio installation, we recommend
+installing Istio that way, unless you need the ability to customize your
+installation. If your cloud platform offers a managed Istio installation,
+the [install guide](./) for your specific platform will have those instructions.
```
samodell: @RichieEscarez  -- so these links 404 when browsing on GitHub; but I think they'd work on the site. I forget, should I be using the GitHub-appropriate URLs, and then the script you made a while back makes the links appropriate for the website at build time?

docs/install/installing-istio.md
ZhiminXiang: 1. The default command is exactly the same as the command in the section `Installing Istio with sidecar injection`. We should probably keep only one of them.
2. We probably want to recommend users to use the instruction `installing Istio with no sidecar injection` as the default instructions. @tcnghia @mattmoor WDYT?

docs/install/installing-istio.md
ZhiminXiang: Istio service mesh provides below benefits for users:
1. [mTLS](https://istio.io/docs/tasks/security/mutual-tls/): with service mesh enabled, users can turn on mTLS, which will make service-to-service traffic within cluster secure.
2. [Authorization policy](https://istio.io/docs/tasks/security/authz-http/): with service mesh enabled, users can leverage Istio authorization policy, which controls the access of each service within service mesh based on Istio service role.

docs/install/installing-istio.md
```diff
@@ -0,0 +1,281 @@
...
+    --set pilot.traceSampling=100 \
+    install/kubernetes/helm/istio \
+    `# Removing trailing whitespaces to make automation happy` \
+    | sed 's/[ \t]*$//' \
+    > ./istio-lean.yaml
+
+  kubectl apply -f istio-lean.yaml
+  ```
+
+### Installing Istio with Secret Discovery Service
```
ZhiminXiang: I just realized that the title I used here is not accurate enough because `Secret Discovery Service` is too broad. What we need is is `Secret Discovery Service` in ingress gateway.
So I would change the title to `Installing Istio with Secret Discovery Service in Ingress Gateway`

docs/install/installing-istio.md
ZhiminXiang: Thanks for mentioning this!
With SDS enabled in ingress gateway, users can configure Gateway TLS more flexibly.
1. Users can dynamically update the Gateway TLS with multiple TLS certificates to terminate TLS connection. See [example](https://istio.io/docs/tasks/traffic-management/secure-ingress/sds/)
2. If users want to use [Auto TLS feature] of Knative, they need to install Istio with SDS in the gateway. See [instructions of auto TLS](https://github.com/knative/docs/pull/1281/files#diff-1ac0a9c578d517c34bf1c688503c8733R14)

docs/install/installing-istio.md
RichieEscarez: Use GitHub-appropriate links (sorry ive not gotten back to finishing the authors guide). So `[install guide](./README.md)` is what you should use and the site build script fixes it at build time for knative.dev.

samodell: Cool, that's what I thought. Thanks!

docs/install/installing-istio.md
samodell: Updated to recommend no sidecar injection.

docs/install/installing-istio.md
```diff
@@ -0,0 +1,280 @@
...
+
+- _automatic sidecar injection_: Enables the Istio service mesh by
+  [automatically injecting the Istio sidecars][1]. The sidecars are injected
+  into each pod of your cluster as they are created.
+
+- _manual sidecar injection_: Provides your Knative installation with traffic
+  routing and ingress, without the Istio service mesh. You do have the option of
+  later enabling the service mesh if you [manually inject the Istio sidecars][2].
+
+If you are just getting started with Knative, you should choose automatic
```
ZhiminXiang: This statement seems conflict with https://github.com/knative/docs/pull/1314/files#diff-1116bcf887cf6b5e21e5e4f96fa026b1R28 because installing Istio withouth sidecar injection means no automatic sidecar injection.

samodell:
Opened #1319  to track additional work needed to update a few of the Install files which were doing unique things during the Istio install.

Otherwise, I think this is pretty good to go.

docs/install/installing-istio.md
samodell: Whoops, good catch! I switched it back and forth too many times :)

docs/install/installing-istio.md
samodell: And updated to go back to recommending automatic sidecar injection as the default, since I found info supporting that in the Custom Install guide, which I believe was reviewed by Matt and Nghia.

docs/install/installing-istio.md
```diff
@@ -0,0 +1,280 @@
...
+sidecar injection and enable the Istio service mesh.
+
+Due to current dependencies, some installable Knative options require the Istio
```
samodell: Are lines 51-53 still true? If they are, I would lean towards installing Istio with automatic sidecar injection as the default, since the majority of the install instructions right now tell people to install all the components as well as the observability plug-ins. I don't want to recommend a default that doesn't work for a lot of use cases...

samodell: @ZhiminXiang  @mattmoor

ZhiminXiang: + @grantr 
I guess in eventing side we are working on getting rid of mesh. I think Grant probably knows better about this part.

samodell:
Okay, @ZhiminXiang  I switched the recommendation back to installing without sidecar injection. 

Please let me know what you think and LGTM if you're happy with it!

docs/install/installing-istio.md
grantr: Eventing 0.6 no longer requires Istio. :tada: @akashrv worked on this, so he knows the most about it.

docs/install/installing-istio.md
```diff
@@ -0,0 +1,272 @@
...
+
+You need:
+- A Kubernetes cluster created.
+- [`helm`](https://helm.sh/) installed.
+
+## Installing Istio
+
+When you install Istio, there are a few options depending on your goals. For a
+basic Istio installation suitable for most Knative use cases, follow the
+[Installing Istio with sidecar injection](#installing-istio-with-sidecar-injection)
```
ZhiminXiang: nit: I think here it should be `Installing Istio without sidecar injection`

ZhiminXiang:
Thanks @samodell . Generally it looks good to me. Just one nit.

ZhiminXiang:
/lgtm
Thanks @samodell . It looks really good to me.

docs/install/installing-istio.md
akashrv: That is correct. Eventing doesn't require Istio, but would continue to work with Istio mesh.

PR #1315, opened 2019-05-14
Correct the path in the Dockerfile

vincent-pli:
## Proposed Changes
Cannot build the image follow the doc: 
https://github.com/knative/docs/tree/master/docs/serving/samples/telemetry-go
will hit $GOPATH exception, it's caused by a incorrect path in the Dockfile

### DIFF
--- a/docs/serving/samples/telemetry-go/Dockerfile
+++ b/docs/serving/samples/telemetry-go/Dockerfile
@@ -14,7 +14,7 @@
 
 FROM golang AS builder
 
-WORKDIR /go/src/github.com/knative/docs/
+WORKDIR /go/src/github.com/knative/docs/docs/
 ADD . /go/src/github.com/knative/docs/
 
 RUN CGO_ENABLED=0 go build ./serving/samples/telemetry-go/
@@ -22,6 +22,6 @@ RUN CGO_ENABLED=0 go build ./serving/samples/telemetry-go/
 FROM gcr.io/distroless/base
 
 EXPOSE 8080
-COPY --from=builder /go/src/github.com/knative/docs/telemetry-go /sample
+COPY --from=builder /go/src/github.com/knative/docs/docs/telemetry-go /sample
 
 ENTRYPOINT ["/sample"]


RichieEscarez:
@grayside can you please validate these changes for us? It looks like several Dockerfiles across all samples might be missing these filepath changes (from docs repo restructure to support Hugo site builds) and Id like your input here before applying changes globally.

vincent-pli:
@RichieEscarez 
Sure, I'm glad to check all the Dock files in the samples. but a bit late, since it's night in China 😪

docs/serving/samples/telemetry-go/Dockerfile
```diff
@@ -14,14 +14,14 @@
 
 FROM golang AS builder
 
-WORKDIR /go/src/github.com/knative/docs/
+WORKDIR /go/src/github.com/knative/docs/docs/
 ADD . /go/src/github.com/knative/docs/
```
grayside: Given this is run at the sample root, the relationship between the directory inside the builder stage and the "real" directory outside it is nice to facilitate alignment with local development. However, this shouldn't throw errors.

grayside: Not in scope of this PR, but the way the builder section is set up now has a number of things that seem odd.

* Copying more than the sample into the container
* The current `ADD` instead of `COPY . .`
* Running build from a higher-level directory path

Also, this would be even more different in switching to use go modules.

vincent-pli: I guess the sample try to use the third part package in the `vendor` for compiler.
Fo the `ADD` and the other thing, I think maybe it's individual preference 😄

vincent-pli:
@grayside 
I tried this: `go get -d github.com/knative/docs/docs/serving/samples/telemetry-go`, it could work for me, could you please retry it and supply some more details?

sebgoa:
I tested the changes in the PR, it does build properly. Looks like an artefacts of the previous doc structure.

/lgtm

/ok-to-test

grayside:
Will address my issues in follow-up, as they are not related to this specific PR. I have no objection to this moving forward if it's passing tests and works for others :)

PR #1322, opened 2019-05-14
Change the configuration for local Kube cluster in Minikube and Docker desktop

houshengbo:
Fixes https://github.com/knative/serving/issues/4057

## Proposed Changes

- Change the configuration of cluster from allocating 4 CPUs to 6 CPUs for local Kubernetes


sebgoa:
/lgtm
/ok-to-test

That's a bit insane :(, way too much resources needed.

thanks for the fix, I also read the upstream issue in serving.

PR #1323, opened 2019-05-14
eventing-sources API reference update

nachocano:
## Proposed Changes

- Re-generating eventing-sources API references with following env variables:
KNATIVE_EVENTING_SOURCES_COMMIT=b6cdf390669a28bc0d6808476851e2a34aa94305
These commits are the latest HEAD as of 05/14/2019 @12:00pm PST.

- As most things in eventing-sources have moved to the contrib package, I'm manually concatenating files for now. The right solution would be to actually change the binary that is being called, besides the script, to properly include identifiers. That may take more time, thus before the 0.6, this seems appropriate. We can later on cherrypick an update for 0.6.


nachocano:
/assign @samodell 
we need this other one before the 0.6 cut. 

we need to properly update the script and the binary that generates this, but that may take more time... @grantr started looking at it. We can later on cherry pick the more complicated change back to 0.6, and that will fix other stuff in the references.

samodell:
/lgtm
/approve

Sounds good to me!

PR #1326, opened 2019-05-15
Use proper revision names in blue green sample

### LINKED ISSUE #1316: Blue/Green Sample is wrong
the Blue/Green sample:
https://knative.dev/docs/serving/samples/blue-green-deployment/

Makes reference of revisions `0001` and `0002` which don't exist anymore.

I am pretty sure the naming changed and now we get a hash appended to the revision name.

sebgoa:
Fixes #1316 

## Proposed Changes

Since ~0.5 revision names are not `00001` and subsequent increment. There have a generated name.

Thus the blue green sample, which is our second guide in the landing page is Broken.

this PR fixes that by adding a `kubectl command` to get the revision names.

### DIFF
--- a/docs/serving/samples/blue-green-deployment.md
+++ b/docs/serving/samples/blue-green-deployment.md
@@ -58,9 +58,17 @@ kubectl apply --filename blue-green-demo-config.yaml
 configuration "blue-green-demo" configured
 ```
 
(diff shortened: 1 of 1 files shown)


samodell:
Great catch, @sebgoa -- thanks!

/lgtm
/approve

PR #1343, opened 2019-05-17
Update check install version docs

grantr:
## Proposed Changes

- Add section about the new Knative Serving release version label (v0.4.0+), while
keeping the old way in a different section.
- Add an Eventing section. The Eventing components have the release
version label, but it's set to "devel" in 0.6.0, so we can't use it this time :(


samodell:
Thanks, @grantr !

PR #1356, opened 2019-05-21
Fix formatting on auto tls

kevinsimper:
There is a small formatting error that makes the page look weird

image


RichieEscarez:
Thanks @kevinsimper! I also added a fix for PR#1202

/lgtm
/approve

PR #1183, opened 2019-04-16
typo

vijaySamanuri:
Fixes #issue-number

## Proposed Changes

-
-
-

PR #1370, opened 2019-05-24
typo in build readme

sebgoa:
typo


sebgoa:
this is #1183  which we have not merged because of lack of CLA.

we can merge this and then close 1183

PR #1373, opened 2019-05-26
remove an unnecessary comma

toshi0607:
no issue

## Proposed Changes

- remove an unnecessary comma on the doc


sebgoa:
It would be nice to fix some more than just the comma, but thanks for engaging.

/ok-to-test

PR #1374, opened 2019-05-27
Format markdown

mattmoor-sockpuppet:


sebgoa:
dupe of #1368 

not sure why but this should be closed in favor of #1368

PR #1380, opened 2019-05-29
Fix the wrong information in grpc example

### LINKED ISSUE #1381: grpc example has an incorrect version for knative and incorrect tag to push
## Expected Behavior
Successful execution of this example.

## Actual Behavior
Docker push command uses a wrong tag name, and knative should pin to 0.6 or later.

1. Follow the instruction at https://github.com/knative/docs/tree/master/docs/serving/samples/grpc-ping-go

## Additional Info
**Additional context**

**Install information:**
 - Platform (GKE, IKS, AKS, etc.):
 - Knative Version:

houshengbo:
Fixes #1381

## Proposed Changes

- Based on the syntax in sample.yaml, it should be version 0.6 or later.
- The command "docker push" uses a wrong tag name for the image. This PR corrects it.

### DIFF
--- a/docs/serving/samples/grpc-ping-go/README.md
+++ b/docs/serving/samples/grpc-ping-go/README.md
@@ -2,7 +2,7 @@ A simple gRPC server written in Go that you can use for testing.
 
 ## Prerequisites
 
-- [Install the Knative Serving version 0.4 or later](../../../install/README.md).
+- [Install the latest version of Knative Serving](../../../install/README.md).
 
 - Install [docker](https://www.docker.com/).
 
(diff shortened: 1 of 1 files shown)


docs/serving/samples/grpc-ping-go/README.md
```diff
@@ -2,7 +2,7 @@ A simple gRPC server written in Go that you can use for testing.
 
 ## Prerequisites
 
-- [Install the Knative Serving version 0.4 or later](../../../install/README.md).
+- [Install the Knative Serving version 0.6 or later](../../../install/README.md).
```
sebgoa: we could remove the version entirely that way it won't get out of date and always reference the latest install doc.

sebgoa:
/ok-to-test

Aside from my comment about removing the version entirely, this is

/lgtm

docs/serving/samples/grpc-ping-go/README.md
houshengbo: 👌

sebgoa:
thanks, I think it is better like this.

/lgtm

docs/serving/samples/grpc-ping-go/README.md
RichieEscarez: my only concern here is if someone with an existing older version of Serving (pre 0.6) tries to run this

what if we say the "latest version?

```suggestion
- [Install the latest version of Knative Serving](../../../install/README.md).
```

houshengbo:
@RichieEscarez changed into latest version.

PR #1390, opened 2019-05-29
Use local sources for TestSampleApp/java-spring

### LINKED ISSUE #1389: test/e2e.TestSampleApp/java-spring is failing
https://testgrid.knative.dev/knative-docs#continuous

adrcunha:
It's not uncommon that spring.io is down and this test fails because it can't fetch the sources from start.spring.io. Use a local zipfile instead.

Fixes #1389


adrcunha:
/hold

Synced up with Chao, this will render the documentation about the sample out-of-date.

adrcunha:
> Synced up with Chao, this will render the documentation about the sample out-of-date.

This is not true. `helloworld.zip` is a project, and the documentation instructs the reader to modify its main app (which results in https://github.com/knative/docs/blob/master/docs/serving/samples/hello-world/helloworld-java-spring/src/main/java/com/example/helloworld/HelloworldApplication.java).

In case the local zip file gets outdated to the point that the sample won't run, the test will fail (as expected). If the sample source code in the repo or the README file are out of sync, the sample consistency test will fail (as expected).

Thus this change is OK to be merged.

/hold cancel

sebgoa:
Seems fine except that I am not a fan of having zip files in version control.

Can't we put the content of the zip in version control and instruct the reader to zip it ?

adrcunha:
> Can't we put the content of the zip in version control and instruct the reader to zip it ?

This is only for automated tests. No markdown file was touched, so user instructions won't change.
The idea for the sample validation tests is to mimic the sample instructions as close as possible, thus my preference for using an original zipfile from spring.io.

PR #1391, opened 2019-05-29
Fixing Broker manual event emit curl example

viktorbenei:
Upgrading the `curl` example in Broker's manual event emit example as the CloudEvents v0.1 format and no longer works, v0.2 does.

Tested with Knative Eventing 0.6

### DIFF
--- a/docs/eventing/broker-trigger.md
+++ b/docs/eventing/broker-trigger.md
@@ -250,10 +250,10 @@ While SSHed into a `Pod` and run:
 curl -v "http://default-broker.default.svc.cluster.local/" \
   -X POST \
   -H "X-B3-Flags: 1" \
-  -H "CE-CloudEventsVersion: 0.1" \
-  -H "CE-EventType: dev.knative.foo.bar" \
-  -H "CE-EventTime: 2018-04-05T03:56:24Z" \
-  -H "CE-EventID: 45a8b444-3213-4758-be3f-540bf93f85ff" \
+  -H 'CE-SpecVersion: 0.2' \
+  -H "CE-Type: dev.knative.foo.bar" \
+  -H "CE-Time: 2018-04-05T03:56:24Z" \
+  -H "CE-ID: 45a8b444-3213-4758-be3f-540bf93f85ff" \
   -H "CE-Source: dev.knative.example" \
   -H 'Content-Type: application/json' \
   -d '{ "much": "wow" }'


viktorbenei:
I signed it!

sebgoa:
/ok-to-test

@n3wscott could you have a quick look since you are a cloud event guru :)

docs/eventing/broker-trigger.md
```diff
@@ -250,10 +250,10 @@ While SSHed into a `Pod` and run:
 curl -v "http://default-broker.default.svc.cluster.local/" \
   -X POST \
   -H "X-B3-Flags: 1" \
-  -H "CE-CloudEventsVersion: 0.1" \
-  -H "CE-EventType: dev.knative.foo.bar" \
-  -H "CE-EventTime: 2018-04-05T03:56:24Z" \
-  -H "CE-EventID: 45a8b444-3213-4758-be3f-540bf93f85ff" \
+  -H 'CE-SpecVersion: 0.2' \
```
grantr: Headers are case-insensitive in 0.2, but the CloudEvents examples all use lowercase header names like `ce-specversion`. Probably a good idea to be consistent with those. https://github.com/cloudevents/spec/blob/v0.2/http-transport-binding.md#314-examples

Also, the 0.2 spec itself mandates double quotes around all header values (except JSON objects and arrays), e.g. `ce-specversion: "0.2"`. This will change in spec 0.3, so I would be fine with this example NOT adding double quotes. In practice most (all?) implementations of CloudEvents accept unquoted header values.

samodell: @viktorbenei  To Grant's first point, let's change the header names to lower case here.

To your second point, @grantr , I'm not sure I understand what you mean -- this PR changes one instance of double-quotes to single quotes. Should we leave it as-is (double-quotes)?

And thanks for the review!

grantr: Sorry, that was an unclear explanation of a subtle issue. :smile:

This is not referring to the shell quotes around `foo` in `-H "foo"`. The CloudEvents 0.2 spec mandates that its HTTP header **values** be quoted because it wants them to be valid JSON. To do that here looks like:


```
-H 'CE-SpecVersion: "0.2"' \
```

Note the double quotes around `0.2` in addition to the single quotes around `CE-SpecVersion: "0.2"`.

The CloudEvents 0.3 spec (wip) drops this requirement, such that headers are specified exactly as in this PR, e.g.

```
-H 'CE-SpecVersion: 0.2'
```

Note the lack of double quotes around `0.2`. Because this spec change is happening in 0.3, I think we should just ignore what the 0.2 spec says and **not** add double quotes around `0.2`, but there may be disagreement on how closely we should align our examples with the CloudEvents spec as written.

At this point I'm kind of sorry I brought it up because it's been so complicated to explain something that I recommend ignoring completely. @n3wscott can probably commiserate - this particular aspect of the spec has wasted so much time already. 😬

samodell: Ahhh, okay. Got it! Well, I do appreciate the thorough explanation :) I also agree with leaving it as-is.

ryandawsonuk:
Looks like this fixes https://github.com/knative/eventing/issues/1392

samodell:
@viktorbenei  Can you make the change Grant mentioned re: lowercasing? Also, it would be great if you could open this PR against the 0.7 branch instead.

viktorbenei:
Sure @samodell - opened a new PR targeting release-0.7 https://github.com/knative/docs/pull/1570

PR #1393, opened 2019-05-29
Fix broker rolebinding for master

whynowy:
Fixes

It should be `ServiceAccount` instead of `User` for broker `RoleBinding` subject.

## Proposed Changes

-
-
-


whynowy:
@Harwayne

RichieEscarez:
@whynowy the title of this PR says "for master" so I just want to double check and make sure that this should not be cherry picked into 0.6?

Harwayne:
> @whynowy the title of this PR says "for master" so I just want to double check and make sure that this should not be cherry picked into 0.6?

#1394 is making a similar change in 0.6.

PR #1348, opened 2019-05-17
Fix the broken Istio installation command

ZhiminXiang:
Fixes #1338

## Proposed Changes

- Add a step to create istio-system namespace
- Update the command of installing istio-lean.yaml according to the newest [download-istio.sh](https://github.com/knative/serving/blob/master/third_party/istio-1.1.3/download-istio.sh#L68)

PR #1400, opened 2019-05-30
Fix the broken Istio installation command

srvaroa:
Following the documentation produces:

    Error from server (NotFound): error when creating "istio.yaml": namespaces "istio-system" not found

Create instead the istio-system namespace before the actual installation steps start, using the same method as in the current master.


ZhiminXiang:
I had a PR https://github.com/knative/docs/pull/1348 to fix this issue, but forgot to cherry pick it into release-0.6. Could you please directly cherrypick it so that we can make `release-0.6` branch consistent with `master`? Thanks!

srvaroa:
Thanks for the pointer @ZhiminXiang, done

Issue #1239, opened 2019-04-25
How to build the docs website (knative.dev) locally?


aliok:
**Describe the change you'd like to see**
Instructions on how to build the docs locally.

**Additional context**
I noticed some formatting errors and wanted to debug them but I can't debug without being able to build the docs locally.

mhamwala:
@aliok It's built into some code editors, so if your using (assuming your using a mac)
`Atom`:
Just open the Markdown file and hit Ctrl+Shift+M to toggle the Markdown preview panel next to it. It handles HTML and images also.
`vscode`:
I think its Ctrl+Shift+P then hit Markdown: Open Preview.
if your not using one of these editors i suggest a quick google and hopefully your editor supports this :)

samodell:
Hey @aliok  -- do you mean building the docs to see how they'll appear on the knative.dev website? If that's the case, @RichieEscarez  can help with that.

It's true that the markdown rendering on the website isn't perfectly in sync with GitHub's own markdown rendering; I have seen some formatting issues. We'd love for you to look into those.

aliok:
> Hey @aliok -- do you mean building the docs to see how they'll appear on the knative.dev website? If that's the case, @RichieEscarez can help with that.

@samodell  yes, exactly. Some Markdown content is rendered strange and I wanted to debug them locally. But I don't know what kind of set up is used here.

aliok:
Ok, now I found it. https://github.com/knative/website is building it.

I will send a PR to note this in the docs repo.

RichieEscarez:
Status: I will open a PR for this some time tomorrow (sorry, our team has been out of office the first part of the week at a workshop so playing catch up now).

@sebgoa @aliok 
For the interim, the local build steps are generally:

1. Install and setup Hugo local server: https://gohugo.io/getting-started/quick-start/
2. Meet Docsy requirements: https://github.com/google/docsy-example/blob/master/content/en/docs/Getting%20started/_index.md
3. Clone knative/website
4. Meet Hugo's expectation that all content lives under a single root directory. Either:
    - clone knative/docs into the `content/en` folder
    - clone knative/docs elsewhere and then create a symlink to there from the `content/en` folder (of knative/website)
5. Run `hugo server' from the root of knative/website and open your localhost

There is a little more to this (small details) but Ill consolidate all these steps and the missing detail into a single page.

RichieEscarez:
Instructions and the new local build script are here: https://github.com/knative/website/pull/72/files

[...]

PR #1409, opened 2019-05-31
Switch from literal includes to {{readFile}} invocations.

evankanderson:
## Proposed Changes

- Use `{{readFile "local-filename"}} instead of repeating text content.

I need instructions on how to test if this worked correctly -- any tips or a writeup I can use?

Note that filenames and `pubspec.yaml` had actually drifted slightly from the content in GitHub, which is why it would be nice to do includes like this.

I also have a question out to @jonasfj about the Dockerfile contents, which don't seem to COPY or build the dart application.

### DIFF
--- a/community/samples/serving/helloworld-dart/README.md
+++ b/community/samples/serving/helloworld-dart/README.md
@@ -22,15 +22,7 @@ be created using the following instructions.
 1. Create a new directory and write `pubspec.yaml` as follows:
 
    ```yaml
-   name: hello_world_dart
-   publish_to: none # let's not accidentally publish this to pub.dartlang.org
-   description: Hello world server example in Dart
-
-   environment:
-     sdk: ">=2.1.0 <3.0.0"
-
-   dependencies:
-     shelf: ^0.7.3
+   {{readFile "pubspec.yaml"}}
    ```
 
 2. If you want to run locally, install dependencies. If you only want to run in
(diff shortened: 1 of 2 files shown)


RichieEscarez:
Sorry, i need to block some time to complete the docs maintainer info but a really short version of the local testing instructions are here: https://github.com/knative/docs/issues/1239#issuecomment-488764565

RichieEscarez:
I think the "out of the box" version of that Hugo command requires file paths relative to the site root. But, i created an override version to use relative files ([readfile.md](https://github.com/knative/website/blob/master/layouts/shortcodes/readfile.md)) - that way content will work as it gets "versioned" (moved into an archive folder not at root).

For example ([_index.md](https://github.com/knative/docs/blob/master/docs/_index.md)):

`{{% readfile file="README.md" relative="true" markdown="true" %}}`

RichieEscarez:
A few things need to get merged (and then tested) but the support for including code is here: https://github.com/knative/website/pull/63

rgregg:
I'm generally not a fan of this. While it makes our website stay up to day, it makes the markdown not readable without manually parsing these tags. I know it's a pain to keep the content up to date in two places, but I think that pain is worth the ability to read these markdown files.

evankanderson:
The markdown seems fairly straightforward, and helps to ensure that the file contents actually align:

https://github.com/knative/docs/blob/cb26c02fed762540c9e4a28aa87e71434c03b603/community/samples/serving/helloworld-dart/README.md

I still need to merge and figure out how to validate the hugo/netlify templates, but I've been buried under other Google work.

PR #1410, opened 2019-06-03
Explain upstream ONBUILD instructions for Dart sample

jonasfj:


jonasfj:
@evankanderson

jonasfj:
@evankanderson, this was for your benefit, feel free to close it if you don't think it's a good fit..
It was mostly intended to illustrate how we could make this seem less magic :)

evankanderson:
TIL about [`ONBUILD`](https://docs.docker.com/engine/reference/builder/#onbuild).

Thanks for the documentation!

/lgtm
/approve

PR #1416, opened 2019-06-04
latest of eventing...

matzew:
@samodell  looks like the docs were not fully covering the latest intsall for eventing


matzew:
@RichieEscarez @samodell   I've rebased - would be good to land this.

Generally, not sure what's the best to keep all these different docs up to date ... 

/cc @abrennan89

PR #1437, opened 2019-06-08
Fix e2e test breakage by using url instead of domain

chaodaiG:
This PR fixes e2e test error by using url instead of domain for finding domain URL. This PR unblocks PR from being merged in docs, but will makes e2e tests diverge from instructions. subsequent update is needed from #1438

### DIFF
--- a/test/e2e/sampleapptestbase.go
+++ b/test/e2e/sampleapptestbase.go
@@ -69,7 +69,7 @@ func cleanup(yamlFilePath, workDir string) {
 }
 
 func serviceHostname(appName string) string {
-	return noStderrShell("kubectl", "get", "rt", appName, "-o", "jsonpath={.status.domain}", "-n", servingNamespace)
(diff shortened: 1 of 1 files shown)


test/e2e/sampleapptestbase.go
```diff
@@ -157,7 +157,8 @@ func checkDeployment(t *testing.T, appName, expectedOutput string) {
 		t.Fatalf("Ingress not found (ingress='%s', host='%s')", ingressAddr, serviceHost)
 	}
 	t.Logf("Curling %s/%s", ingressAddr, serviceHost)
-
+	
+	serviceHost = strings.Replace(serviceHost, "http://", "", 1)
```
yt3liu: Since this used to take domain name, maybe also need to strip out path from the url?

mattmoor: In general: you can parse it as a `net/url.URL` and extract the `.Host` for the host header below, and replace the `Host` with the `ingressAddr` to curl to.

This is good enough for anything we'd emit today.

mattmoor:
/lgtm

to unblock the continuous build.

Issue #1438, opened 2019-06-08
Use url instead of domain for finding domain URL of service


chaodaiG:
**Describe the change you'd like to see**
All instructions uses `.status.url` instead of `.status.domain` for finding domain URL

**Additional context**
Use `url` in favor of `domain`, as the latter was deprecated since this PR: https://github.com/knative/serving/pull/4197

PR #1439, opened 2019-06-08
Migrate serving samples away from features disabled in 0.7.

mattmoor:
Eliminate several samples using JIT Builds.

Stop using `.status.domain` everywhere.

Fixes: #1438

cc @chaodaiG


chaodaiG:
/approve

status.domain part looks good to me, waiting on someone else who is more familiar with JIT part apply lgtm

PR #1442, opened 2019-06-10
Fix the helloworld-go sample.

ScrapCodes:
On running the sample service.yaml 
Following error is produced.
```
Error from server (InternalError): error when creating "service.yaml": Internal error occurred: admission webhook "webhook.serving.knative.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "template"

```

## Proposed Changes

- Fixed the sample service.yaml.
- Fixed the doc to reflect the correct service.yaml contents.

### DIFF
--- a/docs/install/getting-started-knative-app.md
+++ b/docs/install/getting-started-knative-app.md
@@ -53,13 +53,22 @@ metadata:
   name: helloworld-go # The name of the app
   namespace: default # The namespace the app will use
 spec:
-  template:
-    spec:
-      containers:
-        - image: gcr.io/knative-samples/helloworld-go # The URL to the image of the app
-          env:
+apiVersion: serving.knative.dev/v1alpha1 # Current version of Knative
+kind: Service
+metadata:
+  name: helloworld-go # The name of the app
+  namespace: default # The namespace the app will use
+spec:
+  runLatest:
+    configuration:
+      revisionTemplate:
+        spec:
+          container:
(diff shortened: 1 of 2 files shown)


ScrapCodes:
It might be that I am running a previous release of knative, but in that case other samples use the previous code still, so they shall be updated?
Actually, now the reason that this change will be helpful, for someone new, it becomes a stumbling block, because, less they may realise that it is due to difference in the releases. At least, error message does not make it evident.
Since this sample is the first thing a new user lands on, it might just be beneficial to keep it compatible with wider set of releases. I would be grateful to be corrected, if someone thinks otherwise.

RichieEscarez:
Hi @ScrapCodes, previous version of code samples are located under either the corresponding branch in GitHub ([for example here is 0.5](https://github.com/knative/docs/blob/release-0.5/docs/install/getting-started-knative-app.md)), or in the corresponding section in the website (use the version drop down at the top of the site (ie. the [v0.5-docs](https://knative.dev/v0.5-docs/install/getting-started-knative-app/) section)

Also, for future reference, please sign the CLA (details in a comment above) if you'd like us to review and merge your changes.

Im going to close this since this code sample was correctly [updated for 0.6](https://github.com/knative/docs/commit/8e59e4144fd5eca81cad23d20775aa447593f773#diff-3726a6a18860a070c867af0a388773f1) but feel free to reopen (and re-target the branch that corresponds with the version of Knative you have) if there is a needed fix.

Issue #1367, opened 2019-05-22
Knative Build code samples(docs/build/samples.md)


abhip:
## Expected Behavior
Whenever I click the hyper link "Learn how to use Knative to build apps from source code and then run those containers." under Complex Code Samples , it should navigate to the link.

## Actual Behavior
Instead I get the following message 

"Page Not Found
Looks like you've followed a broken link or entered a URL that doesn't exist on this site.

 Back to our site"

## Steps to Reproduce the Problem

1. Goto https://knative.dev/docs/build/samples/
2. Click on Learn how to use Knative to build apps from source code and then run those containers. under Complex Code Samples

moficodes:
Looks like the source-to-url folder was removed on knative version 0.0.6

Because the folder is present in version 0.0.5.

I think the team is trying to move the examples out to the community and only hold docs in the main repo.

zhangtbj:
Hi all,
I created a PR to rollback the `source-to-url` folder remove:
https://github.com/knative/docs/pull/1445

I think there is no any error in this `source-to-url` sample, it is better to bring it back and let people can continue learning and try the end-2-end build and deploy process.

An error page is not good user experience.

Hope this is helpful.

abrennan89:
@samodell build is gone now, isn't it? Is there a plan to remove it in the docs?

PR #1445, opened 2019-06-10
Add source-to-url-go samples back for build Complex code samples

zhangtbj:
New contributors:
release: "release-0.6"

Fixes #issue-number
https://github.com/knative/docs/issues/1367
## Proposed Changes
Hi all,
I am reading the knative build document, and when I try to follow the end-2-end process to build and deploy app by using the this page:
https://knative.dev/docs/build/samples/#complex-code-samples

I found it doesn't work. I think it is better t add the `source-to-url` sample back first. Then remove and fix it after the team confirms the solution.

### DIFF
--- a/docs/serving/samples/source-to-url-go/README.md
+++ b/docs/serving/samples/source-to-url-go/README.md
@@ -1,0 +1,263 @@
+A Go sample that shows how to use Knative to go from source code in a git
+repository to a running application with a URL.
+
+This sample uses the [Build](../../../build/README.md) and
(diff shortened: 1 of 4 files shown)


samodell:
As of 7.0, Serving and Build no longer work together. These samples were deleted intentionally, and it looks like some of the other Build samples might need to be removed too.

samodell:
But thank you for trying to fix the problem! We appreciate folks opening PRs :)

PR #1463, opened 2019-06-13
Update cert manager on GCP example to upgrade to http by default

mikehelmick:
## Proposed Changes

The current documentation only has httpsRedirect set on the https endpoint itself. In order for the ingress to redirect all traffic to use https, the tls section with httpsRedirect: true needs to be on the http section as well.

### DIFF
--- a/docs/serving/using-cert-manager-on-gcp.md
+++ b/docs/serving/using-cert-manager-on-gcp.md
@@ -259,6 +259,10 @@ spec:
       protocol: HTTP
     hosts:
     - "*"
+    tls:
+      # Sends 301 redirect for all http requests.
+      # Omit to allow http and https.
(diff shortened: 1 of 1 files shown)


ZhiminXiang:
/lgtm

Besides, I don't think we need `httpRedirect` for `https` port: https://github.com/knative/docs/pull/1463/files#diff-d1a99181387b20140f29a92d28d9b5eaL269
Could you please delete it? :)

mikehelmick:
@ZhiminXiang - I removed the redirect from the https section.

mikehelmick:
Also - rebased the PR so that it only shows the actual commits

samodell:
Thanks @mikehelmick  and @ZhiminXiang !

PR #1490, opened 2019-06-19
Add new RoleBindings to broker manual setup

grantr:
These were added to the automatic setup in https://github.com/knative/eventing/pull/1426. They're not required for the Broker to work, but they enable observability features like consolidated logging, metrics, and tracing.

## Proposed Changes

- Document additional RoleBindings to be created when manually preparing a namespace for a Broker.

/cc @Fredy-Z

### DIFF
--- a/docs/eventing/broker-trigger.md
+++ b/docs/eventing/broker-trigger.md
@@ -124,16 +124,19 @@ required once per namespace. These instructions will use the `default`
 namespace, but you can replace it with any namespace you want to install a
 `Broker` into.
 
-Create the `ServiceAccount`.
+Create the `ServiceAccount` objects.
 
 ```shell
 kubectl -n default create serviceaccount eventing-broker-ingress
 kubectl -n default create serviceaccount eventing-broker-filter
 ```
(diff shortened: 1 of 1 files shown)


docs/eventing/broker-trigger.md
```diff
@@ -124,16 +124,19 @@ required once per namespace. These instructions will use the `default`
...
 ```shell
 kubectl -n default create serviceaccount eventing-broker-ingress
 kubectl -n default create serviceaccount eventing-broker-filter
 ```
 
-Then give it the needed RBAC permissions:
+Then give them the needed RBAC permissions:
 
 ```shell
+kubectl -n default create rolebinding eventing-broker-ingress \
```
Fredy-Z: 137-139 are exactly the same as 140-142, added by mistake?

grantr: Good catch! This command was added by mistake.

samodell:
Nice, @grantr ! And thanks for the review, @Fredy-Z !

/approve

PR #1493, opened 2019-06-19
Update to use `tag` instead of `name` in traffic

kwyn:
Fixes #issue-number

## Proposed Changes

- Changes use of `name` to `tag` to follow the new API instead of the deprecated API.


kwyn:
I signed it!

PR #1497, opened 2019-06-21
Add label kservice instead of route

nak3:
`kubectl label route` does not makes sense, as change for `route` is
reconciled.

This patch changes to add label to kservice instead of route.

## Proposed Changes

- Add label kservice instead of route

### DIFF
--- a/docs/serving/cluster-local-route.md
+++ b/docs/serving/cluster-local-route.md
@@ -18,5 +18,5 @@ For example, if you didn't set a label when you created the Route
 `helloworld-go` and you want to make it local to the cluster, run:
 
 ```shell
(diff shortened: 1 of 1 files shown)


samodell:
@tcnghia  Can you or someone else verify this change?

sebgoa:
This is correct.

The page is a bit confusing, it referes to `helloworld-go` which is here: https://github.com/knative/docs/blob/master/docs/serving/samples/hello-world/helloworld-go/README.md and this howto creates a Service. Hence the route is created by the service, thus we need to label the service and not the route.

I do think this is mentioned as an example. In the general case where a route is created without a ksvc labeling the route will have the correct effect.

I would add a link to the helloworld-go example.

RichieEscarez:
#1617 merged and fixes this

PR #1525, opened 2019-06-25
Install updates -- v0.7.0

samodell:
- Updates locations for install resources and `git clone` commands to the 0.7 release. 
- Removes references to clusterrole.yaml, since that is not necessary with Build being spun out to Tekton Pipelines


samodell:
Ready for review! @carieshmarie or @RichieEscarez  -- this needs to be merged before 0.7 is cut.

PR #1546, opened 2019-06-27
Update Knative-with-GKE.md

meteatamel:
When you install the default Knative 0.7, nothing gets installed under knative-sources namespace. So, removing the check for pods under this namespace.

Fixes #issue-number

## Proposed Changes

-
-
-


samodell:
@meteatamel  Good catch! That line exists in a number of other install files; would you be willing to delete it from those as well?

samodell:
I wanted to get this change in quickly, so I opened PR #1556 . Thanks for pointing this out!

PR #1548, opened 2019-06-27
[RELEASE-0.7]Update test-infra

chaodaiG:
Bringing changes that fixes `0.7.x` serving release yaml not exist error

/cc @mattmoor

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -116,14 +116,14 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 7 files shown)


RichieEscarez:
@chaodaiG is this going to be a required task for each knative/docs release? Do we need to coordinate prior to a docs release to avoid breaking tests?

chaodaiG:
@RichieEscarez , it's just for 0.7, since it so happened that a critical patch was pushed to test-infra right after 0.7 creation. This shouldn't happen in future releases

PR #1562, opened 2019-06-30
change the worng serial number

tx19980520:
change the worng installation procedure serial number

## Proposed Changes

- change the serial number of Install on a Kubernetes cluster's doc

### DIFF
--- a/docs/install/Knative-with-any-k8s.md
+++ b/docs/install/Knative-with-any-k8s.md
@@ -74,7 +74,7 @@ your Knative installation, see
    --filename https://github.com/knative/serving/releases/download/v0.7.0/monitoring.yaml
    ```
 
-1. To complete the install of Knative and its dependencies, run the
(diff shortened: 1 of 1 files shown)


docs/install/Knative-with-any-k8s.md
```diff
@@ -74,7 +74,7 @@ your Knative installation, see
    --filename https://github.com/knative/serving/releases/download/v0.7.0/monitoring.yaml
    ```
 
-1. To complete the install of Knative and its dependencies, run the
+2. To complete the install of Knative and its dependencies, run the
```
gyliu513: There is no need to update this and MarkDown will handle this.

samodell:
Yep, these are automatic numbers. It's true that the rendering isn't working perfectly on the website right now, but that will be fixed soon. Thanks for the PR, though!

PR #1566, opened 2019-07-01
Added new section for custom-metrics API details

### LINKED ISSUE #1519: Add caveat about custom-metrics API
**Doc requirement from Markus**

Registering the custom-metrics API can clash if the user already has a custom-metric API implementation hooked up in their cluster.

We'll need to document that caveat and document that one will not be able to use custom-metrics (and HPA on concurrency metrics for that matter) on those clusters for now.

**Additional context**

We'll need to document a caveat, basically saying that if you already have a custom-metrics API implementation in your cluster (TODO: provide command to find that out) you cannot use this and you can deactivate the installation by (TODO: provide command to do that).

abrennan89:
Fixes #1519 

## Proposed Changes

- Added details about checking for existing custom-metrics API and deleting an existing implementation if needed.
- new `using-custom-metrics-API.md` file.

### DIFF
--- a/docs/serving/using-custom-metrics-API.md
+++ b/docs/serving/using-custom-metrics-API.md
@@ -1,0 +1,29 @@
+---
+title: "Using the custom-metrics API"
+weight: 10
+type: "docs"
+---
+
+The `custom-metrics` API is installed by default when you install Knative Serving.
+
+>**IMPORTANT:** If you already have an existing `custom-metrics` API implementation in your cluster, you must disable this so that Knative Serving can use the default Knative Serving installation of the `custom-metrics` API. Enabling both implementations can cause issues.
+>
+>If you intend to use an existing `custom-metrics` API implementation in your cluster, you will not be able to use custom metrics or HPA concurrency metrics on that cluster. These features require the Knative Serving `custom-metrics` API implementation.
+
+# Checking for existing custom-metrics API implementations
+
+To check if your cluster has an existing `custom-metrics` API implementation, use the following command.
+
+```
+kubectl get apiservice | grep v1beta1.custom.metrics.k8s.io
+```
+
+# Deleting existing custom-metrics API installations
+
+If you have an existing `custom-metrics` API implementation in your cluster, you must delete this using the following command.
+
+```
+kubectl delete apiservice v1beta1.custom.metrics.k8s.io
+```
+
+**IMPORTANT:** This step is not required for inexperienced users, as it is possible to break consumers of a custom-metrics API by deleting this. Only delete API implementations if you are sure that it is safe to do so.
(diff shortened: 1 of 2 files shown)


abrennan89:
@samodell any idea why these tests are failing? The errors don't really give any info.

docs/serving/using-custom-metrics-API.md
```diff
@@ -0,0 +1,29 @@
+---
+title: "Using the custom-metrics API"
+weight: 10
+type: "docs"
+---
+
+The `custom-metrics` API is installed by default when you install Knative Serving.
+
+>**IMPORTANT:** If you already have an existing `custom-metrics` API implementation in your cluster, you must delete this so that Knative Serving can use the default Knative Serving installation of the `custom-metrics` API. Enabling both implementations can cause issues.
```
samodell: How is this scenario (having an existing API implementation) likely to come about? Might be nice to signal how users could know if this was likely to apply to them, if possible.

docs/serving/using-custom-metrics-API.md
```diff
@@ -0,0 +1,29 @@
+---
+title: "Using the custom-metrics API"
+weight: 10
+type: "docs"
+---
+
+The `custom-metrics` API is installed by default when you install Knative Serving.
```
samodell: What does the custom metrics API allow users to do?

docs/serving/using-custom-metrics-API.md
```diff
@@ -0,0 +1,29 @@
+---
+title: "Using the custom-metrics API"
```
samodell: Hmm, it doesn't seem like this guide covers how to use the custom-metrics API -- if documentation for accessing the metrics already exists, we should point to it. If it doesn't, maybe we could add it to this guide?

abrennan89:
@markusthoemmes can you please advise for Sam's comments, thanks.

docs/serving/using-custom-metrics-API.md
markusthoemmes: It allows users to make use of concurrency based scaling in the **HPA**.

docs/serving/using-custom-metrics-API.md
markusthoemmes: True. We don't really need a usage guide at this point though. What we need is documentation about the caveats that the implementation of the custom-metrics API in Knative Serving can bring.

Enabling Serving's custom-metric API implementation is needed to enable a set of features in Knative Serving. Today, this feature is "concurrency based autoscaling by the HPA". In the future this will also be "access to existing metrics from custom autoscaler implementations".

docs/serving/using-custom-metrics-API.md
samodell: @abrennan89  Could you add this info to this page?

docs/serving/using-custom-metrics-API.md
samodell: Ahh, I see. Thanks!

samodell:
/lgtm
/approve

Sorry for the wait on this one!

PR #1570, opened 2019-07-02
Fixing Broker manual event emit curl example

viktorbenei:
Upgrading the curl example in Broker's manual event emit example as the CloudEvents v0.1 format and no longer works, v0.2 does.

Tested with Knative Eventing 0.6


viktorbenei:
Updated, fixed version of https://github.com/knative/docs/pull/1391

samodell:
/ok-to-test
/approve
/lgtm

Thanks, @viktorbenei !

PR #1602, opened 2019-07-16
Fixed the numbering of steps

rranjan-us:
Numbering of steps were 1. 1. 1., fixed to incremental steps under "Install kubectl and Minikube" and "Installing Knative". If that was intention, feel free to reject the changes.

Fixes #issue-number

## Proposed Changes

-
-
-

### DIFF
--- a/docs/install/Knative-with-Minikube.md
+++ b/docs/install/Knative-with-Minikube.md
@@ -24,9 +24,9 @@ v1.12 or newer.
    version. You need v1.10 or newer. If your `kubectl` is older, follow the next
    step to install a newer version.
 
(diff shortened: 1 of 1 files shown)


samodell:
Hi @rranjan-us -- thanks for your PR! So, markdown will take lists that are all numbered as "1" and correctly render the numbers in sequence. This reduces maintenance burden (if you need to add/move/remove an item in the list, you don't have to renumber). However, this is not rendering correctly on the website at the moment. We are planning to fix this issue, so we are going to leave things numbered this way so that we will be able to benefit from automatic numbering in the future.

PR #1612, opened 2019-07-23
fix broken links.

lionelvillard:
## Proposed Changes

- fix broken links to various eventing sources.

### DIFF
--- a/docs/eventing/sources/README.md
+++ b/docs/eventing/sources/README.md
@@ -10,58 +10,64 @@ procedure:
     ```
 -->
 
-Event Sources are Kubernetes Custom Resources which provide a mechanism for
-registering interest in a class of events from a particular software system.
-Since different event sources may be described by different Custom Resources,
-this page provides an index of the available source resource types as well as
-links to installation instructions.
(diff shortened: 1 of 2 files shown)


rgregg:
This PR really changes the formatting of the document. Do we need to make those changes, or could we just fix the broken links?

samodell:
@rgregg  When I preview the rendered changes, this looks fine to me -- just seems like some words have changed within the table, and the links have been updated. Can you point to what formatting changes you were concerned about?

lionelvillard:
the diff looks weird but the rendering is ok.

rgregg:
The diff is what I was looking at, where the formatting of the table has changed dramatically (looks like maybe an auto-table formatter was used). My big concern is that the semantics of the table format changed (e.g. rows don't start and end with |) and that might cause issues for rendering engine used on the website vs. how GitHub renders the content.

lionelvillard:
readme.md is generated, so maybe the generator has changed?

rgregg:
Good point, let's not block on my feedback :)

PR #1617, opened 2019-07-24
Rewrite the cluster-local route topic

rgregg:
## Proposed Changes

- Rewrite the cluster-local route topic to be more descriptive and share more about why you would do it and the options available for how.
-
-


rgregg:
Reading through #1497 made me realize this topic was lacking a lot of context. I tried to fix that up.

samodell:
Awesome! Thanks for this, @rgregg !

/lgtm
/approve

PR #1620, opened 2019-07-26
removing annotation does not remove broker

vaikas-google:
Fixes: https://github.com/knative/eventing/issues/1504

## Proposed Changes

- Add a note about removing of annotation on a namespace does not delete broker.

### DIFF
--- a/docs/eventing/broker-trigger.md
+++ b/docs/eventing/broker-trigger.md
@@ -116,6 +116,12 @@ namespace.
 kubectl -n default get broker default
 ```
 
+*NOTE* `Broker`s created due to annotation will not be removed if you
+remove the annotation. For example, if you annotate the namespace,
+which will then create the `Broker` as described above. If you now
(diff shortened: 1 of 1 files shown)


docs/eventing/broker-trigger.md
```diff
@@ -116,6 +116,12 @@ namespace.
 kubectl -n default get broker default
 ```
 
+*NOTE* `Broker`s created due to annotation will not be removed if you
+remove the annotation. For example, if you annotate the namespace,
+which will then create the `Broker` as described above. If you now
+remove the annotation, the `Broker` will not be removed, you have
+to manually delete it.
```
Harwayne: Add the command to delete it:

For example, to delete the injected Broker from the `foo` namespace:

```shell
kubectl -n foo delete broker default
```

vaikas-google: done, PTAL.

PR #1625, opened 2019-07-30
serving/samples: update helloworld-php to php 7.3

grayside:
The current main PHP version is 7.3.

## Proposed Changes

- Update Dockerfile to pull from php:7.3-apache instead of php:7.2-apache


samodell:
Thanks, @grayside ! Can you confirm that you tested this out and it still works?

grayside:
I tested this locally with `docker build` and `docker run`, and it worked fine. This change does not have knative-level implications.

samodell:
Excellent! 
/approve
/lgtm

Issue #1661, opened 2019-08-07
v0.8/v0.9 Knative API build errors


RichieEscarez:
Details here: https://github.com/ahmetb/gen-crd-api-reference-docs/issues/10

RichieEscarez:
For 0.8, this broke just the Serving API.
For 0.9, both Serving and Eventing APIs error out during the build:
`F0919 09:52:05.470892   19869 main.go:444] type invalid type has kind=Unsupported which is unhandled`

RichieEscarez:
@ahmetb Has there been any progress here?

Now that both the Serving and Eventing API doc builds fail, I'm wondering if this issue will remain as a blocking item through the next releases, or if Knative needs to investigate an alternative?

/cc @rgregg

RichieEscarez:
Fixed for 0.11 and master. Closing.

PR #1669, opened 2019-08-08
Allow for non-percent routing

duglin:

## Proposed Changes

- Add text to allow for other types of routing, aside from percentage based

Related to: https://github.com/knative/serving/pull/5036

ping @dgerd

### DIFF
--- a/docs/serving/spec/knative-api-specification-1.0.md
+++ b/docs/serving/spec/knative-api-specification-1.0.md
@@ -148,9 +148,38 @@ Service object to reduce the cognitive overhead for application developers – t
 Service object should provide sufficient controls to cover most of application
 deployment scenarios (by frequency).
 
-Unless otherwise noted, implementations of this specification MAY define
-additional fields for these resources, however, those extensions MUST NOT
-contradict the semantics defined within this specification.
+## Extentions
+
+Extending the Knative resource model allows for custom semantics to be
+offered by implementions of the specification. Unless otherwise noted,
+implementations of this specification MAY define extensions but those
+extensions MUST NOT contradict the semantics defined within this specification.
+
+There are several ways in which implementations can extend the model:
+- Annotations and Labels
+  Allowing end users to include annotations or labels on the Knative resources
+  allows for them to indicate that they would like some additional semantics
+  applied to those resources. When defining annotations, or labels, it
+  is STRONGLY RECOMMENDED that they have some vendor-specific prefix to
+  avoid any potential naming conflict with other extensions or future
+  annotations defined by the specification.
+
+  Note that this would allow for new controllers to be added to the system,
+  to detect these annotations/labels and act upon them, without requiring
+  code changes to the core Knative components. As such, when possible, this
+  mechanism is preferred.
+
+- Additional Properties
+  There might be times when annotations and labels can not be used to
+  properly (or easily) allow end users to convey their desired semantics,
+  in which case additional well-defined properties might need to be
+  defined by implementations.
+
+  In these cases vendor-specific properties MAY be defined and it is
+  STRONGLY RECOMMENDED that they be named, or prefixed, in such a way
+  to clearly indicate their scope and purpose. Choosing a name that
+  is too generic might lead to conflicts with other vendor extensions
+  or future changes to the specification.
 
 ## Service
 
--- a/docs/serving/spec/knative-api-specification-1.0.md
+++ b/docs/serving/spec/knative-api-specification-1.0.md
@@ -157,6 +157,10 @@ extensions MUST NOT contradict the semantics defined within this specification.
 
 There are several ways in which implementations can extend the model:
 * Annotations and Labels<br>
+  _Note_: Because this mechanism allows new controllers to be added to the
+  system without requiring code changes to the core Knative components, it is
+  the preferred mechanism for extending the Knative interface.
+
   Allowing end users to include annotations or labels on the Knative resources
   allows for them to indicate that they would like some additional semantics
   applied to those resources. When defining annotations, or labels, it
@@ -165,11 +169,6 @@ There are several ways in which implementations can extend the model:
   annotations defined by the specification. For more information on
   annotations and labels, see
   [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions).
-
-  Note that this would allow for new controllers to be added to the system,
-  to detect these annotations/labels and act upon them, without requiring
-  code changes to the core Knative components. As such, when possible, this
-  mechanism is preferred.
 
 * Additional Properties<br>
   There might be times when annotations and labels can not be used to
@@ -182,6 +181,27 @@ There are several ways in which implementations can extend the model:
   to clearly indicate their scope and purpose. Choosing a name that
   is too generic might lead to conflicts with other vendor extensions
    or future changes to the specification.
+
+  For example, adding authentication on a per-tag basis via annotations
+  might look like:
+  ```
+  annotations:
+    knative.vendor.com/per-tag-auth: "{'cannary': true, 'latest': true}"
+  ```
+  but, that is not as user-friendly as extending the `traffic` section itself:
+  ```
+  spec:
+    traffic:
+    - revisionName: a
+      tag: foo
+      knative.vendor.com/auth: true
+    - revisonName: b
+      percent: 100
+      tag: stable
+    - configurationName: this
+      tag: latest
+      knative.vendor.com/auth: true
+  ```
 
 ## Service
 
@@ -1420,14 +1440,9 @@ Min: 0
 <br>
 Max: 100
    </td>
-   <td>The
-
-<a href="#request-routing">Percent is optionally used to specify the percentage of requests which should be allocated from
-the main Route domain name</a> to the specified <code>revisionName</code> or
-<code>configurationName</code>.
-
+   <td>The <code>percent</code> is optionally used to specify the percentage of requests which should be allocated from the main Route domain name to the specified <code>revisionName</code> or <code>configurationName</code>.
 <p>
-When percentage based routing is being used, all <code>traffic</code> sections for a single Service MUST have a <code>percent</code> value, and all values MUST sum to 100.
+To indicate that percetage based routing is to be used, all <code>traffic</code> sections for a single Service MUST have a <code>percent</code> value, and all values MUST sum to 100.
    </td>
    <td>OPTIONAL
    </td>
--- a/docs/serving/spec/knative-api-specification-1.0.md
+++ b/docs/serving/spec/knative-api-specification-1.0.md
@@ -148,6 +148,10 @@ Service object to reduce the cognitive overhead for application developers – t
 Service object should provide sufficient controls to cover most of application
 deployment scenarios (by frequency).
 
+Unless otherwise noted, implementations of this specification MAY define
+additonal fields for these resources, however, those extensions MUST NOT
+contradict the semantics defined within this specification.
+
 ## Service
 
 The Knative Service represents an instantiation of a single serverless container
@@ -1309,6 +1313,9 @@ Restrictions to the values of the field are noted in the Description column.
 
 ## TrafficTarget
 
+This resource specifies the how the network traffic for a particular
+Revision or Configuration is to be configured.
+
 <table>
   <tr>
    <td><strong>FieldName</strong>
@@ -1376,7 +1383,7 @@ Restrictions to the values of the field are noted in the Description column.
    </td>
    <td>int
 <br>
-(Required)
+(Optional)
 <br>
 Min: 0
 <br>
@@ -1384,14 +1391,14 @@ Max: 100
    </td>
    <td>The
 
-<a href="#request-routing">percentage of requests which should be allocated from
+<a href="#request-routing">Percent is optionally used to specify the percentage of requests which should be allocated from
 the main Route domain name</a> to the specified <code>revisionName</code> or
 <code>configurationName</code>.
 
 <p>
-All <code>percent</code> values in <code>traffic</code> MUST sum to 100.
+When percentage based routing is being used, all <code>traffic</code> sections for a single Service MUST have a <code>percent</code> value, and all values MUST sum to 100.
    </td>
-   <td>REQUIRED
+   <td>OPTIONAL
    </td>
   </tr>
   <tr>
--- a/docs/serving/spec/knative-api-specification-1.0.md
+++ b/docs/serving/spec/knative-api-specification-1.0.md
@@ -1442,7 +1442,7 @@ Max: 100
    </td>
    <td>The <code>percent</code> is optionally used to specify the percentage of requests which should be allocated from the main Route domain name to the specified <code>revisionName</code> or <code>configurationName</code>.
 <p>
-To indicate that percetage based routing is to be used, at least one <code>traffic</code> section have a non-zero <code>percent</code> value, and all values MUST sum to 100. Note, a missing <code>precent</code> value implies zero.
+To indicate that percentage based routing is to be used, at least one <code>traffic</code> section MUST have a non-zero <code>percent</code> value, and all values MUST sum to 100. Note, a missing <code>precent</code> value implies zero.
    </td>
    <td>OPTIONAL
    </td>
--- a/docs/serving/spec/knative-api-specification-1.0.md
+++ b/docs/serving/spec/knative-api-specification-1.0.md
@@ -180,7 +180,7 @@ There are several ways in which implementations can extend the model:
   STRONGLY RECOMMENDED that they be named, or prefixed, in such a way
   to clearly indicate their scope and purpose. Choosing a name that
   is too generic might lead to conflicts with other vendor extensions
-   or future changes to the specification.
+  or future changes to the specification.
 
   For example, adding authentication on a per-tag basis via annotations
   might look like:
@@ -1364,7 +1364,7 @@ Restrictions to the values of the field are noted in the Description column.
 
 ## TrafficTarget
 
-This resource specifies the how the network traffic for a particular
+This resource specifies how the network traffic for a particular
 Revision or Configuration is to be configured.
 
 <table>
--- a/docs/serving/spec/knative-api-specification-1.0.md
+++ b/docs/serving/spec/knative-api-specification-1.0.md
@@ -1442,7 +1442,7 @@ Max: 100
    </td>
    <td>The <code>percent</code> is optionally used to specify the percentage of requests which should be allocated from the main Route domain name to the specified <code>revisionName</code> or <code>configurationName</code>.
 <p>
-To indicate that percetage based routing is to be used, all <code>traffic</code> sections for a single Service MUST have a <code>percent</code> value, and all values MUST sum to 100.
+To indicate that percetage based routing is to be used, at least one <code>traffic</code> section have a non-zero <code>percent</code> value, and all values MUST sum to 100. Note, a missing <code>precent</code> value implies zero.
    </td>
    <td>OPTIONAL
    </td>
--- a/docs/serving/spec/knative-api-specification-1.0.md
+++ b/docs/serving/spec/knative-api-specification-1.0.md
@@ -149,7 +149,7 @@ Service object should provide sufficient controls to cover most of application
 deployment scenarios (by frequency).
 
 Unless otherwise noted, implementations of this specification MAY define
-additonal fields for these resources, however, those extensions MUST NOT
+additional fields for these resources, however, those extensions MUST NOT
 contradict the semantics defined within this specification.
 
 ## Service
--- a/docs/serving/spec/knative-api-specification-1.0.md
+++ b/docs/serving/spec/knative-api-specification-1.0.md
@@ -30,7 +30,7 @@ APPROVED</p>
 
    </td>
    <td><p style="text-align: right">
-2019-11-04</p>
+2020-12-16</p>
 
    </td>
   </tr>
@@ -38,7 +38,7 @@ APPROVED</p>
    <td><p style="text-align: right">
 <strong>Version</strong></p>
    </td>
-   <td> 1.0.1  </td>
+   <td> 1.0.2  </td>
   </tr>
 </table>
 
@@ -193,7 +193,7 @@ There are several ways in which implementations can extend the model:
   spec:
     traffic:
     - revisionName: a
-      tag: foo
+      tag: cannary
       knative.vendor.com/auth: true
     - revisonName: b
       percent: 100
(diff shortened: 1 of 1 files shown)


RichieEscarez:
Hi Doug and @mattmoor ,

How does this [knative-api-specification-1.0.md](https://github.com/knative/docs/pull/1669/files#diff-1da8094b70dd26b8a97b11eeb57e225f) file differ from our generated Serving API reference at https://knative.dev/v0.7-docs/reference/serving-api/ (linking 0.7 due to [known issue for 0.8](#1661))? 

Im wondering if some the content in this hand crafted "knative-api-specification-1.0.md" file can be moved into Go source code so that it gets built in the currently generated Serving API ref docs (for example ["Service"](https://knative.dev/v0.7-docs/reference/serving-api/#serving.knative.dev/v1beta1.Service) - is this a wrong or incomplete version that gets generated from our Go source and then published on the site)? 

Im concern that we will have two versions of "Serving API docs" (come v1). Our current API ref doc build info is [here](https://github.com/knative/docs/blob/master/docs/reference/README.md). If these documents are different, what language do we need to add to clarify their differences (when is one used over the other)? 

I also noticed that this doc is not published on knative.dev. When or if it does get added to the site, we should move it into the docs/reference/ folder (maybe in a new `/serving/` sub directory: `/docs/reference/serving/knative-api-specification-1.0.md`) so that it renders in the proper location on knative.dev.

dgerd:
> How does this knative-api-specification-1.0.md file differ from our generated Serving API reference at https://knative.dev/v0.7-docs/reference/serving-api/ (linking 0.7 due to known issue for 0.8)?

The API Specification describes the behavior that is expected for any Knative implementation. The API reference documentation describes the behavior of the Knative Serving implementation of that specification.

> Im wondering if some the content in this hand crafted "knative-api-specification-1.0.md" file can be moved into Go source code so that it gets built in the currently generated Serving API ref docs

There is likely a better way to maintain the details. The OpenAPI format is likely better than Go source code as it then can be used to generate clients as well as documentation.

> for example "Service" - is this a wrong or incomplete version that gets generated from our Go source and then published on the site

The Go source generation is better than nothing, but has some drawbacks. It pulls from the internal representation of the types rather than the API itself. This means that some go details that are irrelevant to an API leak into it (like inlined types) and some types that we use internally like Kubernetes core/v1.PodSpec are actually not fully exposed in the API.

RichieEscarez:
Thanks for the clarification!

Perhaps (to avoid potential confusion in the future?) these types of Specification documents should live outside of the same folder that contains the behavior-related-documentation about our implementation?

For example in a new folder at the root of the knative/docs repo? Open to other suggestions too but due to its location im concerned that I wont be the only one mixes up these "API" docs?

`spec/serving/knative-api-specification-1.0.md`

duglin:
ok I tried to add some more guidance text around extensions, see what you think.

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -148,6 +148,39 @@ Service object to reduce the cognitive overhead for application developers – t
 Service object should provide sufficient controls to cover most of application
 deployment scenarios (by frequency).
 
+## Extentions
```
evankanderson:
```suggestion
## Extensions
```

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -148,6 +148,39 @@ Service object to reduce the cognitive overhead for application developers – t
...
+
+Extending the Knative resource model allows for custom semantics to be
+offered by implementions of the specification. Unless otherwise noted,
+implementations of this specification MAY define extensions but those
+extensions MUST NOT contradict the semantics defined within this specification.
+
+There are several ways in which implementations can extend the model:
+- Annotations and Labels
```
evankanderson: I think you want `**` and maybe a `<br>` here for this to render the way you wrote:

https://github.com/knative/docs/blob/f6adfe6791816ed60c35ff6d8c0a3035457d70b4/docs/serving/spec/knative-api-specification-1.0.md#extentions

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -148,6 +148,39 @@ Service object to reduce the cognitive overhead for application developers – t
...
+  allows for them to indicate that they would like some additional semantics
+  applied to those resources. When defining annotations, or labels, it
+  is STRONGLY RECOMMENDED that they have some vendor-specific prefix to
+  avoid any potential naming conflict with other extensions or future
+  annotations defined by the specification.
+
+  Note that this would allow for new controllers to be added to the system,
+  to detect these annotations/labels and act upon them, without requiring
+  code changes to the core Knative components. As such, when possible, this
+  mechanism is preferred.
```
evankanderson: Maybe put this first, so it's clear what the recommendation is?

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -148,6 +148,39 @@ Service object to reduce the cognitive overhead for application developers – t
...
+Extending the Knative resource model allows for custom semantics to be
+offered by implementions of the specification. Unless otherwise noted,
+implementations of this specification MAY define extensions but those
+extensions MUST NOT contradict the semantics defined within this specification.
+
+There are several ways in which implementations can extend the model:
+- Annotations and Labels
+  Allowing end users to include annotations or labels on the Knative resources
+  allows for them to indicate that they would like some additional semantics
+  applied to those resources. When defining annotations, or labels, it
```
evankanderson: Reference https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions ?

In particular:

> Annotations have very different intended usage from labels. They are primarily generated and consumed by tooling and system extensions, or are used by end-users to engage non-standard behavior of components. For example, an annotation might be used to indicate that an instance of a resource expects additional handling by non-kubernetes controllers. Annotations may carry arbitrary payloads, including JSON documents. **Like labels, annotation keys can be prefixed with a governing domain (e.g. example.com/key-name). Unprefixed keys (e.g. key-name) are reserved for end-users. Third-party components must use prefixed keys.** Key prefixes under the "kubernetes.io" and "k8s.io" domains are reserved for use by the kubernetes project and must not be used by third-parties.

(emphasis mine)

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -148,6 +148,39 @@ Service object to reduce the cognitive overhead for application developers – t
...
+  Note that this would allow for new controllers to be added to the system,
+  to detect these annotations/labels and act upon them, without requiring
+  code changes to the core Knative components. As such, when possible, this
+  mechanism is preferred.
+
+- Additional Properties
+  There might be times when annotations and labels can not be used to
+  properly (or easily) allow end users to convey their desired semantics,
+  in which case additional well-defined properties might need to be
+  defined by implementations.
```
evankanderson: Worth mentioning that customizing information in a repeated field or map is an example of this?

docs/serving/spec/knative-api-specification-1.0.md
duglin: done

docs/serving/spec/knative-api-specification-1.0.md
duglin: done

docs/serving/spec/knative-api-specification-1.0.md
duglin: done

docs/serving/spec/knative-api-specification-1.0.md
duglin: Not following, this *is* the first one.  Can you elaborate?

docs/serving/spec/knative-api-specification-1.0.md
duglin: Can you give an example of what you're thinking of? I'm not sure I grok it.

abrennan89:
@duglin is this PR complete now or still relevant / WIP? :slightly_smiling_face:

duglin:
Still relevant, still needed, still waiting for reviews and answers 🤣

Ping @evankanderson

evankanderson:
Sorry, I thought Dan was handling this, but of course he's no longer doing so.

/assign @dprotaso 
/unassign @dgerd 

I'm looking again now.

docs/serving/spec/knative-api-specification-1.0.md
evankanderson: I meant move this paragraph before most of the other one.

```md
   _NOTE_: Because this mechanism allows new controllers to be added to the
   system without requiring code changes to the core Knative components, it is
   the preferred mechanism for extending the Knative interface.

   Allowing end users to include annotations or labels on the Knative resources
   allows for them to indicate that they would like some additional semantics
   applied to those resources...
```

docs/serving/spec/knative-api-specification-1.0.md
evankanderson: An example would be if you wanted to support auth on a per-tag basis, doing this with an annotation might be cumbersome:

```yaml
    annotations:
      knative.vendor.com/per-tag-auth: "{'cannary': true, 'latest': true}"
```

vs

```yaml
spec:
  traffic:
  - revisionName: a
    tag: foo
    knative.vendor.com/auth: true
  - revisonName: b
    percent: 100
    tag: stable
  - configurationName: this
    tag: latest
    knative.vendor.com/auth: true
```

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -1315,22 +1353,22 @@ Restrictions to the values of the field are noted in the Description column.
...
+(Optional)
 <br>
 Min: 0
 <br>
 Max: 100
    </td>
    <td>The
 
-<a href="#request-routing">percentage of requests which should be allocated from
+<a href="#request-routing">Percent is optionally used to specify the percentage of requests which should be allocated from
```
evankanderson, line 1364: This doesn't make it clear how to select percentage based routing (which is the default, I think). Maybe phrase this as "percentage based routing is the default unless selected by an extension"?

dprotaso:
For my curiosity what other routing strategies are you looking to implement?

> in which case additional well-defined properties might need to be defined by implementations.

A few things I'm thinking about

### Typed schemas
We need to consider how would we accomplish vendor specific extensions when we eventually want typed structural schema for each of our resources (https://github.com/knative/serving/issues/912).  With a schema we'll be able to move most defaulting & validation to the K8s API server. 

The downside is you can't inline vendor specific fields wherever you like. You'll need the explicit escape hatch `x-kubernetes-preserve-unknown-fields` in our schemas. But littering those everywhere seems like an anti-pattern.


### Semantics, Defaulting & Validation

From @dgerd above
> We also have the potential for two extensions to produce conflicting syntax with different semantics. This can both be surprising to a user and hurts ones of the highlights of Knative which is workload portability.

Additionally, I think it's hard to reason about defaults & validation once users start to use vendor extensions.

As an example below - would Knative apply the percentage default (0) to the tag `foo`? 

```yaml
spec:
  traffic:
  - revisionName: a
    tag: foo
    knative.vendor.com/auth: true
  - revisonName: b
    percent: 50
    tag: stable
```

We could say percentage based routing is entirely opt-in or out (within the `traffic` block) regardless if the vendor extension is present. Thus in the example above I would expect an error to be thrown because the total % doesn't equal 100.

But it's not obvious to me if I opt-out of the percentage based routing & don't have a vendor extension what should the behaviour be when I curl the route's URL.

### Alternative Approach

This discussion seems analogous to the K8s ingress v2 [discussions](https://www.youtube.com/watch?v=Ne9UJL6irXY). I don't have deep knowledge but I know it's trying to solve the problem: How do we allow vendors to experiment with APIs and pull in the agreed upon bits into the core ingress type.

Drawing upon what I know I'm proposing we leave our Route API to be completely portable (no vendor extensions) with the current defaulting semantics. We allow percentage based routing to become optional but tighten the language to address the gaps I mentioned above.

We suggest vendors write their own K8s types that conform to the 'Route' duck type. We provide flexibility in our conformance tests to target different API groups/Kinds. 

For end-users this makes it explicit they're opt-ing into a different and non-portable API. 

One downside I see is adding new fields to the core 'Route' could conflict downstream. Maybe we still encourage vendors to have a special prefix?

abrennan89:
@evankanderson @duglin another reminder to follow this one up please :slightly_smiling_face:

docs/serving/spec/knative-api-specification-1.0.md
duglin: Done

docs/serving/spec/knative-api-specification-1.0.md
duglin: ok done

docs/serving/spec/knative-api-specification-1.0.md
duglin: I think I cover that now below.

duglin:
updated please take another look

duglin:
> For my curiosity what other routing strategies are you looking to implement?

We'd like to support routing based on HTTP header values.

> > in which case additional well-defined properties might need to be defined by implementations.
> 
> A few things I'm thinking about
> ### Typed schemas
> 
> We need to consider how would we accomplish vendor specific extensions when we eventually want typed structural schema for each of our resources ([knative/serving#912](https://github.com/knative/serving/issues/912)). With a schema we'll be able to move most defaulting & validation to the K8s API server.
> 
> The downside is you can't inline vendor specific fields wherever you like. You'll need the explicit escape hatch `x-kubernetes-preserve-unknown-fields` in our schemas. But littering those everywhere seems like an anti-pattern.

Maybe, but it seems like we may not have a choice if using things like annotations/labels aren't good enough.

> ### Semantics, Defaulting & Validation
> 
> From @dgerd above
> 
> > We also have the potential for two extensions to produce conflicting syntax with different semantics. This can both be surprising to a user and hurts ones of the highlights of Knative which is workload portability.
> 
> Additionally, I think it's hard to reason about defaults & validation once users start to use vendor extensions.
> 
> As an example below - would Knative apply the percentage default (0) to the tag `foo`?
> 
> ```yaml
> spec:
>   traffic:
>   - revisionName: a
>     tag: foo
>     knative.vendor.com/auth: true
>   - revisonName: b
>     percent: 50
>     tag: stable
> ```
> 
> We could say percentage based routing is entirely opt-in or out (within the `traffic` block) regardless if the vendor extension is present. Thus in the example above I would expect an error to be thrown because the total % doesn't equal 100.
> 
> But it's not obvious to me if I opt-out of the percentage based routing & don't have a vendor extension what should the behaviour be when I curl the route's URL.

I agree, I think it would be an error. Perhaps we should do what you suggested and make it an all or nothing thing w.r.t. routing?  I'm a little nervous about that because it might be possible to define multiple-layers of routing, meaning multiple routing schemes can be used together, and I wouldn't want to preclude the use of those in the future.

> ### Alternative Approach
> 
> This discussion seems analogous to the K8s ingress v2 [discussions](https://www.youtube.com/watch?v=Ne9UJL6irXY). I don't have deep knowledge but I know it's trying to solve the problem: How do we allow vendors to experiment with APIs and pull in the agreed upon bits into the core ingress type.
> 
> Drawing upon what I know I'm proposing we leave our Route API to be completely portable (no vendor extensions) with the current defaulting semantics. We allow percentage based routing to become optional but tighten the language to address the gaps I mentioned above.
> 
> We suggest vendors write their own K8s types that conform to the 'Route' duck type. We provide flexibility in our conformance tests to target different API groups/Kinds.
> 
> For end-users this makes it explicit they're opt-ing into a different and non-portable API.

We don't have a Routing duck type today, right? Just want to make sure I'm not missing something.

The idea of a Routing duck type is an interesting one. How would that look in the yaml?  An annotation/label to indicate which duck type instance to use? I guess similar to pluggable auto-scalers, right?

What does it take to turn Route into a duck type from a docs/UX perspective?

@mattmoor any thoughts on this?

> One downside I see is adding new fields to the core 'Route' could conflict downstream. Maybe we still encourage vendors to have a special prefix?

This is just normal issues with extensions - it's a well known issue, and yep, people deal with it mainly via prefixes.  I would avoid a single well-defined prefix (http got rid of X-), it causes annoyances. See the CloudEvents primer for the long discussions they had in this space.

duglin:
@mattmoor any thoughts on https://github.com/knative/docs/pull/1669#issuecomment-665125551 ?

mattmoor:
I don't completely follow the duck type idea, it would be interesting to see it in a proposal in a better medium w/ example.

> We'd like to support routing based on HTTP header values

What do you have in mind?  There is work in progress to support routing to tagged revisions based on header instead of the host name.  cc @tcnghia

dprotaso:
> I don't completely follow the duck type idea

A vendor could make their own CRD that mimics our Route with their own extensions. We allow our conformance tests to be run with a different GroupVersionKind. I dunno if I like this idea anymore. My earlier issues were with management of the CRD & schema - but I realize now a vendor could just take the OSS schema/definition modify it and maintain their own fork.

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -1309,6 +1364,9 @@ Restrictions to the values of the field are noted in the Description column.
 
 ## TrafficTarget
 
+This resource specifies the how the network traffic for a particular
```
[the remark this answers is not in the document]
duglin: fixed

duglin:
ok I updated this with the latest typos/fixes.

As for @dprotaso's CRD comment... can that be treated as an impl detail/choice? It seems to me that even if CRDs might be a good choice for non-percent routing, the idea of allowing for other properties would still apply.  No?

duglin:
ping @dprotaso @mattmoor - any more updates? See my previous comment

duglin:
oh ping @evankanderson too

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -148,6 +148,61 @@ Service object to reduce the cognitive overhead for application developers – t
...
+  ```
+  annotations:
+    knative.vendor.com/per-tag-auth: "{'cannary': true, 'latest': true}"
+  ```
+  but, that is not as user-friendly as extending the `traffic` section itself:
+  ```
+  spec:
+    traffic:
+    - revisionName: a
+      tag: foo
```
csantanapr, line 196:
```
      tag: cannary
```
to match the example above

csantanapr:
Update the status table at the top of the doc to indicate, date of last update and version bump from `1.0.1` to `1.0.2`

abrennan89:
@duglin can you please try to close this before 2021? :slightly_smiling_face:

docs/serving/spec/knative-api-specification-1.0.md
duglin: fixed

duglin:
@carlos made updates per your comments.

@evankanderson @dprotaso @mattmoor can I get a review of this?

csantanapr:
/lgtm
From my end looks good, thanks @duglin for the updates I didn't see them because `@carlos` not by handle

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -1376,22 +1434,17 @@ Restrictions to the values of the field are noted in the Description column.
...
-   <td>The
-
-<a href="#request-routing">percentage of requests which should be allocated from
-the main Route domain name</a> to the specified <code>revisionName</code> or
-<code>configurationName</code>.
-
+   <td>The <code>percent</code> is optionally used to specify the percentage of requests which should be allocated from the main Route domain name to the specified <code>revisionName</code> or <code>configurationName</code>.
 <p>
-All <code>percent</code> values in <code>traffic</code> MUST sum to 100.
+To indicate that percetage based routing is to be used, all <code>traffic</code> sections for a single Service MUST have a <code>percent</code> value, and all values MUST sum to 100.
```
evankanderson, line 1445:
```suggestion
To indicate that percentage based routing is to be used, all <code>traffic</code> sections for a single Service MUST have a <code>percent</code> value, and all values MUST sum to 100.
```

Hmm... I don't think requiring all sections to have a `percent` value is what we want (see discussion I mentioned above). Today we allow omitting the percentage value when it is zero, like so:

```yaml
  traffic:
    - revisionName: xyz-123
      percent: 100
    - revisionName: xyz-124
      tag: canary
```

I think you want to say:

> If any target within a `traffic` block has a non-zero `percent` value, this indicates that percentage-based routing is to be used. In this case, the the `percent` values MUST sum to 100.

(or the equivalent)

evankanderson (commented):
Sorry, I still have the one nit about unset `percent` values being allowed to default to zero to match current behavior and allow the simple tag-only use case.

Other than that, the rest of this is great.

abrennan89:
@duglin pleeeeeeeeeeeeeeeease? :smile:

docs/serving/spec/knative-api-specification-1.0.md
duglin: @evankanderson ok I modified the text - see what you think

duglin:
> @duglin pleeeeeeeeeeeeeeeease? 😄

Tag - @evankanderson is it

docs/serving/spec/knative-api-specification-1.0.md
```diff
@@ -1376,22 +1434,17 @@ Restrictions to the values of the field are noted in the Description column.
...
-   <td>The
-
-<a href="#request-routing">percentage of requests which should be allocated from
-the main Route domain name</a> to the specified <code>revisionName</code> or
-<code>configurationName</code>.
-
+   <td>The <code>percent</code> is optionally used to specify the percentage of requests which should be allocated from the main Route domain name to the specified <code>revisionName</code> or <code>configurationName</code>.
 <p>
-All <code>percent</code> values in <code>traffic</code> MUST sum to 100.
+To indicate that percetage based routing is to be used, at least one <code>traffic</code> section have a non-zero <code>percent</code> value, and all values MUST sum to 100. Note, a missing <code>precent</code> value implies zero.
```
evankanderson, line 1445:
```suggestion
To indicate that percentage based routing is to be used, at least one <code>traffic</code> section MUST have a non-zero <code>percent</code> value, and all values MUST sum to 100. Note, a missing <code>precent</code> value implies zero.
```

I think you lost a "MUST/MAY/SHOULD". Also, a typo in "percentage".

evankanderson (commented):
Will `/lgtm` this as soon as the "section have a" has the missing word.

docs/serving/spec/knative-api-specification-1.0.md
duglin: doi - fixed

csantanapr:
/lgtm
Typo requested by @evankanderson is now fixed

evankanderson (commented):
/approve
/lgtm

abrennan89:
Thank you so much! :heart_eyes:

PR #1671, opened 2019-08-08
Add port number to route example

edvinerikson:
When having TLS termination in a Gateway, this example stops working because the gateway doesn't know what to do with HTTP traffic on port 443. Fixing the port to 80 solved this for me.

## Proposed Changes

- Added port number to routing example

### DIFF
--- a/docs/serving/samples/knative-routing-go/routing.yaml
+++ b/docs/serving/samples/knative-routing-go/routing.yaml
@@ -54,4 +54,6 @@ spec:
       # eventually be directed to LOgin service.
       - destination:
           host: istio-ingressgateway.istio-system.svc.cluster.local
+          port:
+            number: 80
         weight: 100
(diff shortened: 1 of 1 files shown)


docs/serving/samples/knative-routing-go/routing.yaml
```diff
@@ -54,4 +54,6 @@ spec:
       # eventually be directed to LOgin service.
       - destination:
           host: istio-ingressgateway.istio-system.svc.cluster.local
+          port:
```
ZhiminXiang: could you also add the `port` to line https://github.com/knative/docs/pull/1671/files#diff-f0c3699af8b098b3338421148e699d4bL42 for `/search` ?

edvinerikson: Missed that one! Fixed.

PR #1676, opened 2019-08-09
[docs] Update R sample to use Go

averikitsch:


community/samples/serving/helloworld-r/Dockerfile
```diff
@@ -1,21 +1,16 @@
...
+# Build r-base image
+FROM r-base:3.6.0
 
-# Run the web service on container startup. Here we use the gunicorn
-# webserver, with one worker process and 8 threads.
-# For environments with multiple CPU cores, increase the number of workers
-# to be equal to the cores available.
-CMD exec gunicorn --bind :$PORT --workers 1 --threads 8 app:app
+# Copy Go binary
+COPY --from=builder /go/bin/invoke /go/bin/invoke
```
grayside: `/go/bin/` has no particular meaning in the r-base image.

Do we want the binary location to match the builder stage to support the service developer, or do we want it to match the default working directory of someone starting up a shell in the container for troubleshooting, and placing it at `/invoke`?

community/samples/serving/helloworld-r/Dockerfile
```diff
@@ -1,21 +1,16 @@
...
+COPY HelloWorld.R .
+
+ENTRYPOINT ["/go/bin/invoke"]
```
grayside: Prefer `CMD` for starting up the service. This makes it easier to interact with the container in other ways. For example,

to open a shell in the container with the current config:

```
docker run --rm -it --entrypoint=/bin/bash kn-hello-r
```

to open a shell with a `CMD` entrypoint:

```
docker run --rm -it kn-hello-r /bin/bash
```

The latter is more consistent with the other samples.

community/samples/serving/helloworld-r/HelloWorld.R
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/Rscript
+#!/bin/Rscript
```
grayside: Is the capitalization normal? This is the kind of thing that tends to have OS-level inconsistencies.

averikitsch: Yes, this is standard R.

averikitsch:
@grayside I've changed WORKDIR to be more consistent with the other samples. This looks like go-style. Does this matter a lot or should we simplify it for non-go users?

grayside:
Unless we want to switch the Go code to use go modules, we should use the go-style workdir. I think it's okay to be very golang-centric in the go stage.

My intention is to work on retrofitting to go modules in 1.13, which reminds me to file that as an issue.

PR #1721, opened 2019-08-20
[WIP] second (or third) attempte to update  debugging eventing to 0.8

aslom:


aslom:
The remaining to do is to change example.yaml to use CRD instead of CCP for channel and test it. I would like also to add something that shows how to hide events as API event source produces lot of them. @RichieEscarez @Harwayne

PR #1732, opened 2019-08-24
[Sample] Update C# hello world serving sample according to Microsoft docs, to multistep build

mattwelke:
## Proposed Changes

- Update C# hello world sample to reflect current best practices in Microsoft docs (https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/docker/building-net-docker-images?view=aspnetcore-2.2#build-and-deploy-manually)
- Uses a multi stage build which results in the image being ~200 MB instead of ~2 GB and image can be built without SDK installed on dev machine
- Removes .dockerignore file because copied files are explicit with multi stage build
- Updates README file to reflect new Dockerfile and docs links

Tested by building the image before and after this Dockerfile change and ensuring they both worked the same way (responded to requests on a port specified by PORT env var).


grayside:
Hi Welkie, thanks for making the changes. If Microsoft prefers to direct people to Docker Hub, I'm okay with keeping that link. Can we add a similar link for the second stage? Also, the README needs to be updated to match the Dockerfile.

Sorry for asking for so many minor changes! I've tested the new Dockerfile locally and this works.

welkie:
@grayside No problem. This should be good now. It's got the link at the top of each stage of the build in the Dockerfile, and they're Docker Hub links. The readme is up to date now too.

docs/serving/samples/hello-world/helloworld-csharp/Dockerfile
```diff
@@ -1,18 +1,19 @@
...
 # https://hub.docker.com/r/microsoft/dotnet
-FROM microsoft/dotnet:2.2-sdk
-
-# Install production dependencies.
-# Copy csproj and restore as distinct layers.
+FROM mcr.microsoft.com/dotnet/core/sdk:2.2 AS build
 WORKDIR /app
+
+# copy csproj and restore as distinct layers
 COPY *.csproj .
```
PicardParis: The `COPY *.ext .` pattern will eventually bite non-experts (this works because there's only 1 `csproj`).
→ I'd suggest using `COPY *.csproj ./` instead.

welkie: Nice catch. I never knew this.

> If multiple <src> resources are specified, either directly or due to the use of a wildcard, then <dest> must be a directory, and it must end with a slash /

welkie:
@grayside Ready for a review. Made the changes you requested and included a fix for an issue @PicardParis caught.

mattwelke:
Do you need me to integrate the latest changes on the base branch before this can be merged?

samodell:
/approved

Sorry for the wait, @mattwelke ! Thanks for your contribution.

samodell:
I meant --

/approve

PR #1742, opened 2019-08-27
Add/fix some documentation for the autoscaler configuration.

vagababov:
We definitely need more data there and document with more examples.
But this is for another PR.

/assign @samodell @RichieEscarez


vagababov:
PIng?

PR #1753, opened 2019-08-30
samples/grpc-ping-go: align with knative sample and README practices

grayside:
## Proposed Changes

- Server listens on `0.0.0.0` instead of `localhost`.
- Dockerfile and README should run and document all commands from the context of the sample as working directory.
- Comment cleanup in Dockerfile.
- README cleanup to improve clarity and alignment with other sample READMEs

## Testing Status

* Container builds & runs locally, but I could not adapt instructions for local testing.
* Locally-build container deploys to Cloud Run on GKE via kubectl apply
* Running gRPC service successfully tested with README instructions.


docs/serving/samples/grpc-ping-go/README.md
```diff
@@ -13,47 +13,101 @@ A simple gRPC server written in Go that you can use for testing.
...
    ```shell
-   docker run -ti --entrypoint=/client docker.io/{username}/grpc-ping-go \
+   docker run --rm -it {username}/grpc-ping-go \
+     /client \
      -server_addr="${SERVICE_IP}:80" \
      -server_host_override="grpc-ping.default.example.com" \
      -insecure
    ```
+
+  The arguments after the container tag `{username}/grpc-ping-go` replace the container's default entrypoint which starts the server with a command that starts the gRPC client.
```
averikitsch: This hard to understand.

docs/serving/samples/grpc-ping-go/Dockerfile
```diff
@@ -12,21 +12,31 @@
...
+# Copy local code to the container image.
+WORKDIR /go/src/github.com/knative/docs/docs/serving/samples/grpc-ping-go
+COPY . ./
 
-RUN CGO_ENABLED=0 go build -tags=grpcping ./docs/serving/samples/grpc-ping-go/
-RUN CGO_ENABLED=0 go build -tags=grpcping ./docs/serving/samples/grpc-ping-go/client
+# Build the command inside the container.
+# To facilitate gRPC testing, this container includes a client command.
+RUN CGO_ENABLED=0 go build -tags=grpcping -o ./ping-server
+RUN CGO_ENABLED=0 go build -tags=grpcping -o ./ping-client ./client
```
averikitsch: Where do the folders, ping-server and ping-client, come from?

grayside: Those are the binary names generated by the build command. The first is building main in the current directory, the second is building the client directory.

docs/serving/samples/grpc-ping-go/README.md
grayside: Which aspect is hard to understand (and it can be all of them):

* Swapping out the default entrypoint command
* Starting the gRPC client in the gRPC server container
* Something else?

averikitsch: If "starts the server with a command that starts the gRPC client." == "Starting the gRPC client in the gRPC server container", then the new wording is better.

docs/serving/samples/grpc-ping-go/Dockerfile
averikitsch: Thanks

averikitsch:
If you are willing, it would be nice to update the description of this sample: "A simple gRPC server written in Go that you can use for testing."
It might be nice to give a blurb about grpc and the different components of this sample (server, client). Also why do we say "that you can use for testing."?

docs/serving/samples/grpc-ping-go/README.md
grayside: I think this needs to be clearer, what do you think of this:

---

## Testing the service

Testing the gRPC service requires using a gRPC client built from the same protobuf definition
used by the server.

1. Fetch the created ingress hostname and IP.

   ```shell
   # Put the ingress IP into an environment variable.
   export SERVICE_IP=$(kubectl get svc istio-ingressgateway --namespace istio-system --output jsonpath="{.status.loadBalancer.ingress[*].ip}")
   ```

1. Use the gRPC client to send message streams to the gRPC server.

   The Dockerfile builds the client binary. To run the client you will use the
   same container image deployed for the server with an override to the
   entrypoint command to use the client binary instead of the server binary.

   Replace `{username}` with your Docker Hub user name and run the command:

   ```shell
   docker run --rm -it {username}/grpc-ping-go \
     /client \
     -server_addr="${SERVICE_IP}:80" \
     -server_host_override="grpc-ping.default.example.com" \
     -insecure
   ```

  The arguments after the container tag `{username}/grpc-ping-go` are used
  instead of the entrypoint command defined in the Dockerfile `CMD` statement.

averikitsch: That is great!

grayside:
Updates address additional README improvements.

Issue #1686, opened 2019-08-12
Samples: Upgrade go to 1.13 and use go modules


grayside:
Go samples specify golang, golang:1.11, or golang:1.12.

## Describe the change you'd like to see

Following the stable release of Go 1.13 with a container image on [Docker Hub](https://hub.docker.com/_/golang), samples using Go should go through the following:

* Upgrade samples to use golang:1.13
* Switch to use modules by default, [matching the behavior of the go toolchain](https://tip.golang.org/doc/go1.13#modules). This includes adding a go.mod and go.sum.
* Case by case evaluation if other language changes should be applied.

grayside:
go 1.13 is now available.

PR #1774, opened 2019-09-05
serving/samples: update helloworld-go to go:1.13 and go modules

grayside:
First step toward #1686 

## Proposed Changes

- Upgrades from go 1.12 to go 1.13
- Upgrades to use go modules
- Pins alpine image to 3.10
- Tweaks working directory to /app
- Rename binary in container from `helloworld` to `server`
- Use `go mod download` to facilitate a layer cache of go deps.
- Use `-mod=readonly` flag on builds for strictness of container builds
- Some minor style changes around syntax and comments

### DIFF
--- a/docs/serving/samples/hello-world/helloworld-go/Dockerfile
+++ b/docs/serving/samples/hello-world/helloworld-go/Dockerfile
@@ -3,12 +3,19 @@
 # https://hub.docker.com/_/golang
 FROM golang:1.13 as builder
 
-# Copy local code to the container image.
+# Create and change to the app directory.
 WORKDIR /app
+
+# Retrieve application dependencies.
+# This allows the container build to reuse cached dependencies.
+COPY go.* ./
+RUN go mod download
+
+# Copy local code to the container image.
 COPY . ./
 
 # Build the binary.
-RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld-server
+RUN CGO_ENABLED=0 GOOS=linux go build -mod=readonly -v -o server
 
 # Use the official Alpine image for a lean production container.
 # https://hub.docker.com/_/alpine
@@ -17,7 +24,7 @@ FROM alpine:3
 RUN apk add --no-cache ca-certificates
 
 # Copy the binary to the production image from the builder stage.
-COPY --from=builder /app/helloworld-server /helloworld-server
+COPY --from=builder /app/server /server
 
 # Run the web service on container startup.
-CMD ["/helloworld-server"]
+CMD ["/server"]
--- a/docs/serving/samples/hello-world/helloworld-go/Dockerfile
+++ b/docs/serving/samples/hello-world/helloworld-go/Dockerfile
@@ -15,7 +15,7 @@ RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld
 # Use the official Alpine image for a lean production container.
 # https://hub.docker.com/_/alpine
 # https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds
-FROM alpine:3.10
+FROM alpine:3
 RUN apk add --no-cache ca-certificates
 
 # Copy the binary to the production image from the builder stage.
--- a/docs/serving/samples/hello-world/helloworld-go/Dockerfile
+++ b/docs/serving/samples/hello-world/helloworld-go/Dockerfile
@@ -10,7 +10,7 @@ COPY . ./
 # Build the command inside the container.
 # (You may fetch or manage dependencies here,
 # either manually or with a tool like "godep".)
-RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld
+RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld-server
 
 # Use the official Alpine image for a lean production container.
 # https://hub.docker.com/_/alpine
@@ -19,7 +19,7 @@ FROM alpine:3
 RUN apk add --no-cache ca-certificates
 
 # Copy the binary to the production image from the builder stage.
-COPY --from=builder /app/helloworld /helloworld
+COPY --from=builder /app/helloworld-server /helloworld-server
 
 # Run the web service on container startup.
-CMD ["/helloworld"]
+CMD ["/helloworld-server"]
--- a/docs/serving/samples/hello-world/helloworld-go/Dockerfile
+++ b/docs/serving/samples/hello-world/helloworld-go/Dockerfile
@@ -1,24 +1,25 @@
 # Use the offical Golang image to create a build artifact.
 # This is based on Debian and sets the GOPATH to /go.
 # https://hub.docker.com/_/golang
-FROM golang:1.12 as builder
+FROM golang:1.13 as builder
 
 # Copy local code to the container image.
-WORKDIR /go/src/github.com/knative/docs/helloworld
-COPY . .
+WORKDIR /app
+COPY . ./
 
 # Build the command inside the container.
 # (You may fetch or manage dependencies here,
 # either manually or with a tool like "godep".)
 RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld
 
-# Use a Docker multi-stage build to create a lean production image.
+# Use the official Alpine image for a lean production container.
+# https://hub.docker.com/_/alpine
 # https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds
-FROM alpine
+FROM alpine:3.10
 RUN apk add --no-cache ca-certificates
 
 # Copy the binary to the production image from the builder stage.
-COPY --from=builder /go/src/github.com/knative/docs/helloworld/helloworld /helloworld
+COPY --from=builder /app/helloworld /helloworld
 
 # Run the web service on container startup.
 CMD ["/helloworld"]
--- a/docs/serving/samples/hello-world/helloworld-go/Dockerfile
+++ b/docs/serving/samples/hello-world/helloworld-go/Dockerfile
@@ -7,9 +7,7 @@ FROM golang:1.13 as builder
 WORKDIR /app
 COPY . ./
 
-# Build the command inside the container.
-# (You may fetch or manage dependencies here,
-# either manually or with a tool like "godep".)
+# Build the binary.
 RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld-server
 
 # Use the official Alpine image for a lean production container.
--- a/docs/serving/samples/hello-world/helloworld-go/README.md
+++ b/docs/serving/samples/hello-world/helloworld-go/README.md
@@ -121,6 +121,13 @@ cd knative-docs/docs/serving/samples/hello-world/helloworld-go
                  value: "Go Sample v1"
    ```
 
+1. Use the go tool to create a
+   [`go.mod`](https://github.com/golang/go/wiki/Modules#gomod) manifest.
+
+   ```shell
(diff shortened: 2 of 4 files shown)


docs/serving/samples/hello-world/helloworld-go/Dockerfile
```diff
@@ -1,24 +1,25 @@
 # Use the offical Golang image to create a build artifact.
 # This is based on Debian and sets the GOPATH to /go.
 # https://hub.docker.com/_/golang
-FROM golang:1.12 as builder
+FROM golang:1.13 as builder
```
ahmetb: Any particular reason not to use golang:1?

We have to keep updating this every 6 months otherwise.

docs/serving/samples/hello-world/helloworld-go/Dockerfile
```diff
@@ -1,24 +1,25 @@
...
 # (You may fetch or manage dependencies here,
 # either manually or with a tool like "godep".)
 RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld
 
-# Use a Docker multi-stage build to create a lean production image.
+# Use the official Alpine image for a lean production container.
+# https://hub.docker.com/_/alpine
 # https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds
-FROM alpine
+FROM alpine:3.10
```
ahmetb: Ditto. https://twitter.com/ahmetb/status/1164211673210609665?s=21

docs/serving/samples/hello-world/helloworld-go/Dockerfile
```diff
@@ -1,24 +1,25 @@
...
 # Copy local code to the container image.
-WORKDIR /go/src/github.com/knative/docs/helloworld
-COPY . .
+WORKDIR /app
+COPY . ./
 
 # Build the command inside the container.
 # (You may fetch or manage dependencies here,
 # either manually or with a tool like "godep".)
 RUN CGO_ENABLED=0 GOOS=linux go build -v -o helloworld
```
ahmetb: Imo “-o hello-server” is more descriptive. Or “-o app”

In such tutorials we end up with naming everything hello, so it’s often unclear what they refer to.

docs/serving/samples/hello-world/helloworld-go/Dockerfile
grayside: Given that sometimes the minor version has significant change to best practices (e.g., modules) I think the clarity is worthwhile for the reader and motivation to update.

docs/serving/samples/hello-world/helloworld-go/Dockerfile
grayside:
```suggestion
FROM alpine:3
```

docs/serving/samples/hello-world/helloworld-go/Dockerfile
grayside: Agreed.

docs/serving/samples/hello-world/helloworld-go/README.md
```diff
@@ -72,27 +72,28 @@ cd knative-docs/docs/serving/samples/hello-world/helloworld-go
    # Use the offical Golang image to create a build artifact.
    # This is based on Debian and sets the GOPATH to /go.
    # https://hub.docker.com/_/golang
-   FROM golang:1.12 as builder
-
+   FROM golang:1.13 as builder
+
```
tbpg: Extra whitespace (here and in a few other places).

docs/serving/samples/hello-world/helloworld-go/README.md
```diff
@@ -72,27 +72,28 @@ cd knative-docs/docs/serving/samples/hello-world/helloworld-go
...
+   
    # Copy local code to the container image.
-   WORKDIR /go/src/github.com/knative/docs/helloworld
-   COPY . .
-
+   WORKDIR /app
+   COPY . ./
+   
    # Build the command inside the container.
    # (You may fetch or manage dependencies here,
```
tbpg: Delete this comment?

docs/serving/samples/hello-world/helloworld-go/Dockerfile
```diff
@@ -1,24 +1,25 @@
...
+FROM golang:1.13 as builder
 
 # Copy local code to the container image.
-WORKDIR /go/src/github.com/knative/docs/helloworld
-COPY . .
+WORKDIR /app
+COPY . ./
 
 # Build the command inside the container.
 # (You may fetch or manage dependencies here,
```
tbpg: Delete this comment?

docs/serving/samples/hello-world/helloworld-go/Dockerfile
```diff
@@ -1,24 +1,25 @@
...
 # This is based on Debian and sets the GOPATH to /go.
 # https://hub.docker.com/_/golang
-FROM golang:1.12 as builder
+FROM golang:1.13 as builder
 
 # Copy local code to the container image.
-WORKDIR /go/src/github.com/knative/docs/helloworld
-COPY . .
+WORKDIR /app
+COPY . ./
```
tbpg: I wonder if this should do the `go mod download` trick, since many people will copy this? I know it's been decided on in the past, but consider adding it.

(I'm not attached to adding it - definitely makes things more complex, but it might be worth it.)

docs/serving/samples/hello-world/helloworld-go/Dockerfile
grayside: Simplified the comment. The Dockerfile spacing and comments are designed to emphasize a pipeline of steps and provide context on how the container is assembled.

docs/serving/samples/hello-world/helloworld-go/README.md
grayside: The Dockerfile spacing and comments are designed to emphasize a pipeline of steps and provide context on how the container is assembled.

grayside: Ah, README whitespace. Addressing.

docs/serving/samples/hello-world/helloworld-go/Dockerfile
grayside: Added go mod download and modified go build to use -mod=readonly for strictness of the build.

chaodaiG:
let me TAL

Fredy-Z:
> Looking at the error, I see:
> 
> ```
>             Step 3/10 : COPY go.* ./
>             COPY failed: no source files were specified
> ```
> 
> However, there is plainly a go.mod file in this sample and it builds locally just fine. Is it possible the test rig itself is skipping a file or missing the addition of go.mod somehow?
> 
> @chaodaiG could you take a look?

@grayside you'll need to add `go.mod` in https://github.com/knative/docs/blob/533e6b7bf310bcce3ccfb628e0e736f826235844/test/sampleapp/config.yaml#L14.
Also as I tried in https://github.com/knative/docs/pull/1787, you'll also need to remove `-mod=readonly` from the `go build` command.

chaodaiG:
Thank you @Fredy-Z for looking into this!

docs/serving/samples/hello-world/helloworld-go/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/knative/docs/docs/serving/samples/hello-world/helloworld-go
+
+go 1.13
```
Fredy-Z: Missed an empty line after this line. Otherwise `go build` will not work with `-mod=readonly`.

Fredy-Z:
> Thanks @Fredy-Z,
> 
> I want to make sure I understand what's happening here.
> 
> 1. Why can't I use -mod=readonly? That seems like it should be the best practice in a container, especially where the deps are downloaded via `go mod download`? In this case, the test framework appears to be constraining the implementation of the app in unexpected ways.
> 2. Why doesn't the test process copy all the files, perhaps excluding just .dockerignore? In this case, the test framework seems to be defining unexpected conventions.

@grayside sorry for the late reply.

For 1, I double checked your PR, this reason that you cannot have `-mod=readonly` is because `go.mod` file needs to be ended with an empty line, see https://github.com/knative/docs/pull/1774#discussion_r324506203.
For 2, I'm not sure what's the historic reason for only copying specific files, to me it's totally fine to just copy all files. @chaodaiG can you take a look?

chaodaiG:
> > Thanks @Fredy-Z,
> > I want to make sure I understand what's happening here.
> > 
> > 1. Why can't I use -mod=readonly? That seems like it should be the best practice in a container, especially where the deps are downloaded via `go mod download`? In this case, the test framework appears to be constraining the implementation of the app in unexpected ways.
> > 2. Why doesn't the test process copy all the files, perhaps excluding just .dockerignore? In this case, the test framework seems to be defining unexpected conventions.
> 
> @grayside sorry for the late reply.
> 
> For 1, I double checked your PR, this reason that you cannot have `-mod=readonly` is because the `go.mod` file is not ended with an empty line, see [#1774 (comment)](https://github.com/knative/docs/pull/1774#discussion_r324506203).
> For 2, I'm not sure what's the historic reason for only copying specific files, to me it's totally fine to just copy all files. @chaodaiG can you take a look?

A piece of background, I have observed that all files needed for tests are already inlined in `README.md` files, so there is a unit test verifying the consistency between committed code and README, and the unit test goes through all files needs to be copied and verify if they are identical with `README.md`. See test here: https://github.com/knative/docs/blob/533e6b7bf310bcce3ccfb628e0e736f826235844/test/sampleconsistency/sampleconsistency_test.go#L96

grayside:
For `-mod=readonly`, the container builds fine for me locally as-is, not sure what I'd change there. Why is go build in the container different in CI here?

For file-copy: It makes sense to me that we test that files we include in the README are consistent with source code, but that should not determine what is copied. For example, there are lockfiles such as `package-lock.json` for Node.js which need to be included in the build.

grayside:
Updated to add the go.mod to the test config, and updated the README to add an instruction on how to generate go.mod. Note that I am not copying the go.mod file into the README.

chaodaiG:
> For `-mod=readonly`, the container builds fine for me locally as-is, not sure what I'd change there. Why is go build in the container different in CI here?
> 
> For file-copy: It makes sense to me that we test that files we include in the README are consistent with source code, but that should not determine what is copied. For example, there are lockfiles such as `package-lock.json` for Node.js which need to be included in the build.

For file-copy, I'm fine with any change to make it smarter. One thing to keep in mind is, currently it's very tightly enforced to make sure all files needed for compilation are inspected in unit+e2e tests, if we keep separated list of what's copied and what's verified, there is a risk of these 2 lists diverged. So for any example I would recommend inline the content in README, unless it doesn't make sense at all. And in the latter case there is a way to get around it(not highly encouraged though) by running custom `cp` command, see example for Spring here: https://github.com/knative/docs/blob/533e6b7bf310bcce3ccfb628e0e736f826235844/test/sampleapp/config.yaml#L4

grayside:
It looks like the test is now passing, which tells me that at least some of what we were discussing was a misunderstanding that the tests require inlining all files.

chaodaiG:
that would be a bug if that's the case. We can have this PR pass and I'll TAL

grayside:
@chaodaiG there are files that are required for the build but are not and should not be inlined in the READMEs, for example, `package-lock.json` for Node.js. If the build leaves these out, they may pass but the result will be non-deterministic.

chaodaiG:
I agree that there are certain exceptions like in `package.json` case, which as I mentioned earlier can get exempted by adding a manual `cp` step in `preCommand`.

PR #1825, opened 2019-09-24
Update install link

dgerd:


samodell:
This link is working correctly on Knative.dev (https://knative.dev/docs/serving/getting-started-knative-app/) but not on GitHub.

@RichieEscarez  Can you investigate?

Thanks for pointing this out, @dgerd ! We have a script that runs during the website build process that is supposed to make the links work for both GitHub and the site, but it seems like something funny might be going on. (The site and GitHub use different relative linking paths, which is annoying.)

RichieEscarez:
This fix was committed to the release-0.9 branch (-> fixed under /docs/) while this is for `master` (-> /development/)

PR #1829, opened 2019-09-26
Add note about foundation status

lindydonna:


docs/README.md
```diff
@@ -55,6 +55,8 @@ With a clear project scope, lightweight governance model, and clean lines of
 separation between pluggable components, the Knative project establishes an
 efficient contributor workflow.
 
+While not part of a software foundation, Knative provides contributors with a clear project scope, open governance with multiple companies represented, and clean lines of separation between pluggable components. The Knative project strives for an efficient contributor workflow.
```
samodell: @lindydonna  There's an extra space in the front of the paragraph. Also, what do you think of this edited version:

"While not a part of a software foundation, Knative provides contributors with a clear project scope and an open governance model where multiple companies are represented. The Knative project strives for clean lines of separation between pluggable components and an efficient contributor workflow."

RichieEscarez: There is some repetition in the paragraphs of this section. We should try to combine these statements? Also, should our formal "status" be explained in the "knative/community" repo? Maybe in this topic: https://github.com/knative/community/blob/master/CONTRIBUTING.md

jdumars: I like the change, but would add a comma after components (or perhaps an 'as well as') so it doesn't seem like a comparison between components and workflow:

"While not a part of a software foundation, Knative provides contributors with a clear project scope and an open governance model where multiple companies are represented. The Knative project strives for clean lines of separation between pluggable components as well as an efficient contributor workflow."

lindydonna: Good catch, the previous paragraph should be deleted. I'll do that now.

lindydonna: @jdumars I added "both" to separate the two phrases. A comma isn't grammatical there. :)

jdumars: Thanks!

docs/README.md
```diff
@@ -51,9 +51,7 @@ systems and pass the benefits along to their customers.
 
 ### Contributors
 
-With a clear project scope, lightweight governance model, and clean lines of
-separation between pluggable components, the Knative project establishes an
-efficient contributor workflow.
+While not a part of a software foundation, Knative provides contributors with a clear project scope and an open governance model where multiple companies are represented. The Knative project strives for both clean lines of separation between pluggable components and an efficient contributor workflow.
```
duglin: Not LGTM
This is just asking for public disagreement and it would suggest that this is best solved by the SC, so please revert the “open governance” change

PR #1836, opened 2019-09-29
fix typo (offical -> official)

sqs:


vaikas-google:
@sqs looks like the unit tests are indeed failing and not flaky. Would you mind fixing that?

RichieEscarez:
> @sqs looks like the unit tests are indeed failing and not flaky. Would you mind fixing that?

Sounds like you need to sync the README too:

Error:
`--- FAIL: TestDocSrc/go (0.00s)
        sampleconsistency_test.go:97: README.md file is missing line 0 ('# Use the official Golang image to create a build artifact.') from file '../../docs/serving/samples/hello-world/helloworld-go/Dockerfile'
            Additional info:
            All files required for running sample apps are checked against README.md, the content of source files should be identical with what's in README.md file`

PR #1840, opened 2019-09-30
Document xip.io with Gloo.

mattmoor:
/assign @tcnghia 
/assign @ilackarms 

This is a precursor to cleaning up all of the `curl -H 'Host: ...'` stuff in samples, which is very Istio specific.

### DIFF
--- a/docs/install/Knative-with-Gloo.md
+++ b/docs/install/Knative-with-Gloo.md
@@ -79,6 +79,45 @@ rerun the command to see the current status.
 
 Now you can deploy an app using your freshly installed Knative environment.
 
+## Configuring DNS
+
+Knative dispatches to different services based on their hostname, so it greatly
+simplifies things to have DNS properly configured. For this, we must look up the
+external IP address that Gloo received. This can be done with the following command:
+
+```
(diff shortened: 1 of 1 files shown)


mattmoor:
@ilackarms it is notable that the instructions don't work at HEAD due to the version of things installed by `glooctl install knative`.  Do you plan to update to 0.9 soon and/or include the v1 types (if the cluster is 1.14+)?

ilackarms:
this all looks good to me @mattmoor 
i can run some manual tests to confirm if you'd like

ilackarms:
> @ilackarms it is notable that the instructions don't work at HEAD due to the version of things installed by `glooctl install knative`. Do you plan to update to 0.9 soon and/or include the v1 types (if the cluster is 1.14+)?

let me open an issue for that. currently gloo should still support 0.9 via `glooctl install knative --install-knative-version 0.9.0`

will the omission of the v1 types be problematic? we want to switch to them, just want to gauge the current impact

mattmoor:
@ilackarms it isn't a problem yet, but when we cut 0.10 (including the docs for 0.10) the docs will start referencing v1 types

PR #1844, opened 2019-10-01
update to k8s bump (for 0.10.+)

matzew:
Added a similar suggestion to the GDoc that's backing this file:

https://docs.google.com/document/d/1UHT3W61tbX1qnEPA0qtoNUACki0d8FkXJ10JHOQu9uU/edit#

@dgerd @mattmoor Feel free to ignore - ran into this, while starting similar doc for eventing


dgerd:
This is the right place to update it. To avoid forking the Google doc is now deprecated in favor of this. I added a note at the top now.

As for the change I am fine with it, but we should (1) Validate that no new fields were added to K8s objects we reference (2) Update all of the links out to the Kubernetes docs in the `Detailed Resource Types` to 1.14.

RichieEscarez:
Assigning back to @matzew since it sounds like more work is necessary before this can go in

PR #1871, opened 2019-10-08
Remove comment about limitation that doesn't exist anymore

JRBANCEL:
We used to recommend deploying only a single instance of Envoy per ingress Gateway because we had no visibility on when they would be ready. This is not true anymore since we probe them now.
Also, set the max to 2 to show it is possible.

A user got confused and asked about it in this [issue](https://github.com/knative/serving/issues/5766).


samodell:
/lgtm
/approve

Thanks for submitting this update!

PR #1873, opened 2019-10-08
Update a few invalid links and a couple tips for prerequisite

yolocs:
Update a few invalid links and a couple tips for prerequisite


docs/eventing/getting-started.md
```diff
@@ -24,15 +24,17 @@ To complete this guide, you will need the following installed and running:
...
 ```sh
 kubectl get pods --namespace knative-eventing
 ```
 
 If the `knative-eventing` namespace or the `in-memory-channel-controller-*` does not exist, use the following steps to install Knative Eventing with the in-memory channel provisioner:
 
-1. Make sure that you have a functioning Kubernetes cluster. See the [Comprehensive Install guide](../install/_index.md) for more information.
+1. Make sure that you have a functioning Kubernetes cluster. See the [Comprehensive Install guide](../install) for more information.
+   - Old versions of Knative Serving doesn't necessarily work well with latest Knative Eventing, so try to install the latest version of Knative Serving.
+   - If your Kubernetes cluster comes with pre-installed Istio, make sure it has `cluster-local-gateway` [deployed](https://github.com/knative/serving/blob/master/DEVELOPMENT.md#deploy-istio).
```
yolocs: @vaikas-google this is what I meant to add. I can drop it if it's mentioned somewhere that makes more sense.

vaikas-google: Yeah, so this is fine for now, but I'd like to see us better document this in the existing docs repo so we don't have to leave the repo to fix this.
So, it might be better to link here:
https://knative.dev/docs/install/installing-istio/

However, these instructions require the use of helm, which for this is not necessary. Also it's a little more complicated because depending on which version they might need to install a different version. Maybe add something about that (see the entry for workaround here):
https://github.com/knative/eventing/issues/1973

yolocs: I do think referencing the istio doc might be too "heavy" in this case. So I added a little bit more context from the workaround in knative/eventing#1973

PR #1889, opened 2019-10-10
fix incorrect link in-memory-channel-crd.yaml => in-memory-channel.yaml

vaikas-google:
Fixes  https://github.com/knative/eventing/issues/2024

## Proposed Changes

- change incorrect filename from in-memory-channel-crd.yaml to in-memory-channel.yaml
-
-


grantr:
We should also backport this to 0.9

PR #1893, opened 2019-10-11
Fix Ambassador install instructions to avoid panic

concaf:
Fixes https://github.com/datawire/ambassador/issues/1871

## Proposed Changes

This commit installs Ambassador using ambassador-rbac.yaml instead
of ambassador-knative.yaml which will be deprecated soon.

ambassador-rbac.yaml contains the latest Ambassador install
manifests which have sufficient RBAC to update ingress resources
when required. ambassador-knative.yaml does not have these
permissions and hence Ambassador panics for anyone who follows
the instructions.

### DIFF
--- a/docs/install/Knative-with-Ambassador.md
+++ b/docs/install/Knative-with-Ambassador.md
@@ -57,8 +57,13 @@ You can install Ambassador with `kubectl`:
 
 ```
 kubectl apply \
-  --filename https://getambassador.io/yaml/ambassador/ambassador-knative.yaml \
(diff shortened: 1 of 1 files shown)


containscafeine:
I've also added instructions to install Ambassador in a separate namespace - this fixes https://github.com/datawire/ambassador/issues/1872

romachalm:
Hi, 

line 94 : it should be `$ kubectl get svc ambassador -n ambassador` instead of `kubectl get svc ambassador`

concaf:
@romachalm thanks, fixed it.

PR #1954, opened 2019-11-01
Add a script to check for dead links in markdown files

adrcunha:
Also update test-infra for the latest goodness.

/assign @RichieEscarez 
/cc @eallred-google

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -456,14 +456,14 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:68050b4ce26531655bac0f6e109d04eb0a4550ead73e320119ed3c628ced4a0a"
+  digest = "1:5299d75a2b08a91c54ffb8b76afe21eb5f1ecf7bc4d67b859fc081f9415452bc"
   name = "knative.dev/test-infra"
   packages = [
     "scripts",
     "tools/dep-collector",
   ]
   pruneopts = "UT"
-  revision = "17f2331e80ad0d3e170ea2bae45c3922744f83af"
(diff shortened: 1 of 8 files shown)


samodell:
@adrcunha  Thanks for working on this! Before we merge this in, could you add some information about how this will work?

My main questions: 

Does it block merges on failures?

Have we checked its output for the Docs repo to see if it seems reasonable?

Have we checked it for the URLs that caused false positives with the last URL validator we tried? That one failed on links to  http://localhost , for example. (Context for the problems with the old link checker are in #750 )

adrcunha:
> @adrcunha Thanks for working on this! Before we merge this in, could you add some information about how this will work?

This was discussed with Richie, thus I added him as a reviewer.

> Does it block merges on failures?

No. It does nothing right now because it's not enabled. The idea is to enable it as a non blocker and make it informative so we can properly evaluate its usefulness, fine-tune and eventually, if proven good enough, enable as a blocker.

> Have we checked its output for the Docs repo to see if it seems reasonable?

Yes. It found several dead links in the existing docs, and I'm about to send the report to Richie, hope it helps cleaning up the docs.

> Have we checked it for the URLs that caused false positives with the last URL validator we tried? That one failed on links to http://localhost , for example. (Context for the problems with the old link checker are in #750 )

750 is about the previous tool used. The current tool (liche) has whitelisting capabilities. And now build failures are properly exposed in the build logs through Spyglass.

eallred-google:
I shouldn't block merges on failures.  It's providing a warning.  Turning this on should give us an idea of where / how to tweak this if we wanted to block merge on failures.

samodell:
Awesome! I'll leave this to Richie to sign off on, then.

PR #1965, opened 2019-11-05
Fix sample README typo/word order

cji:
## Proposed Changes

While going through the sample docs I noticed a small typo that I then noticed had propagated to other READMEs as well. 

- change `the with URL` to `with the URL`


cji:
@googlebot I signed it!

RichieEscarez:
Thanks for the PR! Unfortunately, it looks like you have picked up a bunch of extra commits. Can you please try rebasing so this PR lists only a commit from you (and excludes all the changes from those other commits)?

cji:
@RichieEscarez Whoops - I see what happened. I created me branch off master and then opened the PR against the release version branch. I've fixed this in #1967 since it was easier to just redo this than fix my git mess. :)

PR #1967, opened 2019-11-06
fix small typo in sample READMEs

cji:
## Proposed Changes

While going through the sample docs I noticed a small typo that I then noticed had propagated to other READMEs as well. 

- change `the with URL` to `with the URL`


samodell:
/ok-to-test
/lgtm
/approve

Thanks so much for finding this and fixing it globally, @cji !

PR #1970, opened 2019-11-06
fix links to repos in temp API refdocs files

### LINKED ISSUE #1962: Broken link for Serving API documentation
Right now https://knative.dev/docs/reference/serving-api/

has a link to `https://github.com/knative/serving/tree/%7B%7B%3C%20branch%20%3E%7D%7D/pkg/apis`

which seems to be broken.

/kind bug

RichieEscarez:
fixes #1962

Added the dynamic variables in these temp files while the "API ref docs tool" is still not working but forgot that these reference files are "ignored" (to avoid build errors with the HTML thats normally here when the "API ref docs tool" is working.)


RichieEscarez:
@samodell this is ready to go (you can ignore the link error because thats due to the lack of the script in this branch)

PR #1972, opened 2019-11-08
Created new Knative Services section

### LINKED ISSUE #978: Can't find docs on how to create cluster internal-only Knative service
I searched for:

- internal only knative app
- knative services without external endpoint
- knative do not expose

but didn't land anywhere in the docs repo explaining how this can be done. Are we lacking docs on this topic?

/kind bug
/area serving
/area networking

abrennan89:
Fixes #978 

## Proposed Changes

- Added new section for 'Using Knative Services'
- Removed getting started with app deployment section and section on cluster-local services and pulled all of this information into one section


docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
+---
+title: "User-created Knative Services"
+weight: 01
+linkTitle: "User-created Knative Services"
+type: "docs"
+---
+
+Knative services are Kubernetes Services that a user creates to deploy an application. Each Knative Service is defined by a Route and a Configuration, which have the same name as the Service, contained in a YAML file.
```
dgerd: This first sentence is not quite right. Knative Services are backed by multiple Kubernetes "Services". This is because the Kubernetes Service are essentially load balancers over a Deployment. See https://kubernetes.io/docs/concepts/services-networking/service/ 

The Knative Service attempts to model a more logical definition of a Service or Application. To do so it has the concept of Revisions which are each backed by a deployment with 2 K8s Services.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Before you begin
+
+To use Knative Services, you will need:
+* A Kubernetes cluster with [Knative installed](https://knative.dev/docs/install/index.html).
+
+## Creating a Knative Service
+To create an application, you need to create a YAML file that defines a Knative Service.
+This YAML file specifies metadata about the application, points to the hosted image of the app for deployment, and allows the deployment to be configured.
```
dgerd: The word "deployment" here might be confusing as it is 
```suggestion
This YAML file specifies metadata about the application, points to the hosted image of the app, and allows the service to be configured.
```

The word deployment here might be confusing as it is overloaded with K8s deployments.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+
+This guide uses the [Hello World sample app in Go](https://knative.dev/docs/serving/samples/hello-world/helloworld-go) to demonstrate the structure of a Service YAML file and the basic workflow for deploying an app, but these steps can be adapted for your own application if you have an image of it available on Docker Hub, Google Container Registry, or another container image registry.
```
dgerd:
```suggestion
This guide uses the [Hello World sample app in Go](https://knative.dev/docs/serving/samples/hello-world/helloworld-go) to demonstrate the structure of a Service YAML file and the basic workflow for deploying an app. These steps can be adapted for your own application if you have an image of it available on Docker Hub, Google Container Registry, or another container image registry.
```

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+
+The Hello World sample app does the following:
+1. reads in an env variable, TARGET, from the configuration .yaml file
```
dgerd:
```suggestion
1. reads in an environment variable, TARGET, from the configuration YAML file
```

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+2. prints “Hello World: \${TARGET}!“. If TARGET isn’t defined, it will print “NOT SPECIFIED”.
```
dgerd:
```suggestion
1. prints `Hello World: \${TARGET}!`. If TARGET isn’t defined, it will print `NOT SPECIFIED`.
```

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Procedure
+
+1. Create a new file named `service.yaml` containing the following information.
```
dgerd: To make it easier to follow you could make this a [HereDoc](https://en.wikipedia.org/wiki/Here_document) that outputs to service.yaml so all I need to do is copy the command and press enter.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+
+```yml
+apiVersion: serving.knative.dev/v1 (1)
```
dgerd:
```suggestion
apiVersion: serving.knative.dev/v1 # (1)
```

Make this and below comments so it copy/paste works.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+**(3)** The namespace the app will use   
+**(4)** The URL to the image of the app   
+**(5)** The environment variable printed out by the sample app
+
+**NOTE:** If you’re deploying an image of your own app, update the name of the app and the URL of the image accordingly.
+
+## Creating a private cluster-local Service
+By default services deployed through Knative are published to an external IP
+address, making them public services on a public IP address and with a
+[public URL](./using-a-custom-domain.md).
```
dgerd: Probably worth noting here that the public URL out of the box is "example.com" which is non-functioning as it does not point to the public IP address.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+
+While this is useful for services that need to be accessible from outside of the
+cluster, frequently you may be building a backend service which should not be
+available off-cluster.
+
+Knative provides two ways to enable private services which are only available
+inside the cluster:
+
+1. To make all services only cluster-local, change the default domain to
```
dgerd:
```suggestion
1. To make all services available only from within the cluster, change the default domain to
```

RichieEscarez:
Awesome! Thanks @abrennan89! I dont have time to get to this today but will soon.

Initial items ive noticed: 

- the markdown indentation issues for nested content (ie. need to indent code samples (use spaces, not tabs). [see section below about the auto-one-off staging builds]

- our audience for these tasks are not operators so the meta about what underlies a Knative service is unnecessary and might distract users (ie developers who want to deploys services to Knative)
  - we should move the "Knative services = K8s service" details down into a separate concept topic (our [Nav proposal](https://docs.google.com/spreadsheets/d/1L3EAeTX7RRdZhcQZeNmxwN4DCncuImCXA5A1_KQvwSk/edit?ts=5d826691#gid=413198505&range=116:116) currently has that at the bottom of the Serving section)
     
image

  - The title should also refocus towards that developer audience (ie. they dont need to know about the system, just how to get their services running in it). For example, from their point of view, they will always be "creating [user-created] Knative services" so we should avoid qualifying it because they might distract them from the task and cause them to ask "do I need to know what a non-user-created service is?".

**A hard to use but convenient staging option**

Also, we have in progress "docs-contributor" option of viewing a staged version of the PR. Today, the build runs and creates a unique "staged-version" of a PR (ie every commit will trigger and run a new build -> therefore the URL to this changes each time).

That said, you can preview that unique instance of your content if you look at the build page and find that build (usually a recent build near the top of the list - You click on one of those builds and then searching by your branch name works well to determine if its your content. Then when the build is successful, you click the big green [ Preview ] button and navigate to your topic): https://app.netlify.com/sites/knative/deploys

Here is the latests (remember that by committing more builds those changes to a different URL):
https://5dc5b121252e1800072510b0--knative.netlify.com/docs/serving/using-knative-services/

(sorry its so difficult right now but i will eventually get back to refining this process and documenting how it works)

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+   `svc.cluster.local` by
+   [editing the `config-domain` config map](./using-a-custom-domain.md). This
+   will change all services deployed through Knative to only be published to the
+   cluster, none will be available off-cluster.
```
dgerd: Might be too much information for this guide, but you can also add this domain as part of a selector statement to apply this to some of the services.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+1. To make an individual service cluster-local, the service or route can be
+   labeled in such a way to prevent it from getting published to the external
+   gateway.
+
+### Label a service to be cluster-local
+
+To configure a Service to only be available on the cluster-local network (and
+not on the public Internet), you can apply the
+`serving.knative.dev/visibility=cluster-local` label to the Service or Route
```
dgerd: Since this is "using knative services" you will want to apply it to the Service. The Service will propagate the label to the Route. Any label added to a Route that is owned by a Service will be overwritten.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+To see if your application has been deployed successfully, you need the host URL and IP
+address created by Knative.
+
+**NOTE:** If your cluster is new, it can take some time before the Service is
+assigned an external IP address.
+
+1. To find the IP address for your Service, enter:
+
+   ```shell
+   # In Knative 0.2.x and prior versions, the `knative-ingressgateway` service was used instead of `istio-ingressgateway`.
```
dgerd: 0.2.x is so old now that I don't think it is worth the confusion here.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+   INGRESSGATEWAY=knative-ingressgateway
+
+   # The use of `knative-ingressgateway` is deprecated in Knative v0.3.x.
+   # Use `istio-ingressgateway` instead, since `knative-ingressgateway`
+   # will be removed in Knative v0.4.
```
dgerd: Again these versions are old enough now that I would remove this comment..

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,233 @@
...
+   the `.yaml` file, replace `helloworld-go` in the above commands with the name
+   you entered.
+
+1. Now you can make a request to your app and see the results. Replace
+   `IP_ADDRESS` with the `EXTERNAL-IP` you wrote down, and replace
+   `helloworld-go.default.example.com` with the domain returned in the previous
+   step.
+
+   ```shell
+   curl -H "Host: helloworld-go.default.example.com" http://${IP_ADDRESS}
```
dgerd: It is probably worth explaining why we have to inject the host header here and that this is not necessary if you configure a domain associated with your IP.

abrennan89:
>     * our audience for these tasks are not operators so the meta about what underlies a Knative service is unnecessary and might distract users (ie developers who want to deploys services to Knative)

@RichieEscarez do you mean the section _Modifying Knative Services_? I think this came from the links you provided initially about Services so I had included it, but maybe it's not required? Let me know if you want me to remove it or if you meant a different part :slightly_smiling_face:

docs/serving/using-knative-services.md
abrennan89: @RichieEscarez wdyt with regards to document scope?

abrennan89:
@dgerd @RichieEscarez thanks so much for all the feedback and input!
I've implemented most comments now but there are a couple of comments @RichieEscarez for you just to clarify.
Lmk if there are any additional changes required.

docs/serving/using-knative-services.md
abrennan89: First time I've heard of this :slightly_smiling_face: @RichieEscarez do you know if this is something we're using and if so, do we have any contributor docs for how to do this? Thanks.

docs/serving/using-knative-services.md
abrennan89: I think this is covered in / under the scope of https://knative.dev/docs/serving/using-a-custom-domain/ maybe?

samodell:
Hey @abrennan89 ! Richie and I were discussing this, and we're not sure this is the direction we want to take the getting started content. Are you able to come to tomorrow or next week's Documentation WG to discuss? Tomorrow is the 4:30pm PT meeting, and next week is the 9:30am PT one.

docs/serving/using-knative-services.md
RichieEscarez: @dgerd is this an example of what your are proposing (HereDoc), here in one of our other topics: https://knative.dev/docs/eventing/samples/kafka/channel/#creating-a-kafkachannel-channel-crd ?

ie. Where you wrap the [YAML] with with the following so that users can edit, apply, and deploy in a single command?

```
cat <<-EOF | kubectl apply -f -
---
[ALL_THE_YAML_GETS_COPIED_HERE?]
EOF
```

Question: This syntax does not create a copy of that file locally for the user, right? Is there a way to also create a local file?

@abrennan89 today, we have the [`readfile` shortcode](https://github.com/knative/website/blob/master/layouts/shortcodes/readfile.md) that will copy in an actual code file (at markdown build time) so that you dont have to manually copy and hardcode them into the markdown. Example here https://knative.dev/smoketest/

`{{< readfile file="./docs/serving/samples/hello-world/helloworld-go/service.yaml" code="true" lang="yaml" >}}`

Per @dgred's feedback, maybe I could open a PR to add an option to `readfile` (another flag: `heredoc="true"`) so that it automatically adds the "HereDoc" syntax around that code?

dgerd: > This syntax does not create a copy of that file locally for the user, right? Is there a way to also create a local file?

This syntax does not. `tee` is a common command that can be inserted between to make a local copy. Alternatively the file creation through a heredoc and the application through kubectl could be broken into two steps if storing the file locally is desired.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,232 @@
...
+Knative services are used to deploy an application. Knative Services attempt to model a more logical definition of a Service or Application. To achieve this, Services use the concept of Revisions which are each backed by a deployment with 2 Kubernetes Services.   
+
+For more information about Kubernetes Services, see the [Kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/service/).   
+
+Each Knative Service is defined by a Route and a Configuration, which have the same name as the Service, contained in a YAML file.
+
+## Before you begin
+
+To use Knative Services, you will need:
+* A Kubernetes cluster with [Knative installed](https://knative.dev/docs/install/index.html).
```
RichieEscarez: I think it would be beneficial to add a conditional statement here with a pointer to "configuring custom domains" so that its clear that there is additional work to just deploying a service before you can access it externally.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,232 @@
...
+spec:
+ template:
+  spec:
+   containers:
+    - image: gcr.io/knative-samples/helloworld-go # (4)
+     env:
+      - name: TARGET # (5)
+       value: "Go Sample v1"
+```
+**(1)** Current version of Knative
```
RichieEscarez: For long term maintenance, can we avoid qualifying the lines of code and just use a list of some sort? [Example here](https://knative.dev/development/eventing/accessing-traces/#configuration-options) (a definition list would work great too).
That way if the YAML changes (ie new or removed lines) then we can avoid updating numbers in the codes and the list. [I like to avoid and reserve numbered lists for only items that require some type of linear order (due to the limits of markdown, the numbers below end up looking like lists (rather than key value mappings)].

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,232 @@
...
+**(2)** The name of the app   
+**(3)** The namespace the app will use   
+**(4)** The URL to the image of the app   
+**(5)** The environment variable printed out by the sample app
+
+> Note: If you’re deploying an image of your own app, update the name of the app and the URL of the image accordingly.
+
+## Creating a private cluster-local Service
+By default services deployed through Knative are published to an external IP
+address, making them public services on a public IP address and with a
```
RichieEscarez: I think that if we use active voice here, we can address @dgerd's point too, for example:

```suggestion
address. After you configure your [custom domain](./using-a-custom-domain.md),  you can publicly access your services.
```

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,232 @@
...
+[public URL](./using-a-custom-domain.md).
```
RichieEscarez:
```suggestion
```

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,232 @@
...
+
+While this is useful for services that need to be accessible from outside of the
+cluster, frequently you may be building a backend service which should not be
+available off-cluster.
+
+Knative provides two ways to enable private services which are only available
+inside the cluster:
+
+1. To make all services only available only from within the cluster, change the default domain to
```
RichieEscarez: Similar to the comment above. Since these are two options and not a set of linear steps, 'bulleted lists' work better here.

docs/serving/using-knative-services.md
RichieEscarez: I think it can work here if we remove/revert combining all that "deploying an application" content in this file.

@dgerd Question: is creating "cluster-local services" not a common task? the idea here is that we teach users how to create a Knative service with this task and the provide the option for configuring that task to be "private"? If its not going to be a typical task, we could instead use "progressive disclosure" and only provide a "statement and link" to point users to a separate sub-topic (ie revert the "cluster-local-route" topic).

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,232 @@
...
+   `svc.cluster.local` by
+   [editing the `config-domain` config map](./using-a-custom-domain.md). This
+   will change all services deployed through Knative to only be published to the
+   cluster, none will be available off-cluster.
+1. To make an individual service cluster-local, the service or route can be
```
RichieEscarez: "cluster-local" appears to be a visibility label that is Knative specific but Kubernetes refers to "both "cluster-private-IP" and "cluster-internal IP" terms". For consistency we should try to use the same terminology (versus saying "to be cluster-local")

For example: You can configure a Knative service to use a "cluster-private-IP"/"cluster-internal IP" if you add the `cluster-local` label to that service.

docs/serving/using-knative-services.md
```diff
@@ -0,0 +1,232 @@
...
+kubectl get ksvc helloworld-go
+
+NAME            URL                                              LATESTCREATED         LATESTREADY           READY   REASON
+helloworld-go   http://helloworld-go.default.svc.cluster.local   helloworld-go-2bz5l   helloworld-go-2bz5l   True
+```
+
+The service returns the a URL with the `svc.cluster.local` domain, indicating
+the service is only available in the cluster local network.
+
+## Deploying an application using Knative Services
```
RichieEscarez: From here down i think ALL but "Modifying Knative Services", is out of scope for this "Creating a Knative service" task and should move back into the "Getting Started" guide.

Issue #1750, opened 2019-08-29
Broken links on knative.dev


eallred-google:
List of broken links / 404 / timeout on knative.dev

- [ ] 404 Error navigating to; https://github.com/knative/docs/tree/release-0.8/docs/serving/samples/hello-world/README.md

Source Page: https://knative.dev/docs/serving/samples/hello-world/ -> click on View/Edit this page 

- [x] 404 Error navigating to; 

https://github.com/knative/docs/blob/master/install/README.md
Source Page: https://knative.dev/docs/serving/samples/hello-world/helloworld-java-spark/ -> click on installation instructions

- [x] Timeout navigating to;

http://blog.arungupta.me/smaller-java-image-docker-multi-stage-build/
SourcePage: https://knative.dev/docs/serving/samples/hello-world/helloworld-java-spark/ -> click on Creating Smaller Java Image using Docker Multi-Stage Build.

- [ ] 404 Error navigating to; 

https://github.com/knative/eventing/blob/master/pkg/apis/eventing/v1alpha1/channel_types.go#L36
https://github.com/knative/eventing/blob/master/pkg/apis/eventing/v1alpha1/cluster_channel_provisioner_types.go
https://github.com/knative/eventing/blob/master/pkg/apis/eventing/v1alpha1/channel_types.go
https://github.com/knative/eventing/blob/master/pkg/apis/eventing/v1alpha1/subscription_types.go
https://github.com/knative/eventing/blob/master/pkg/apis/eventing/v1alpha1/subscription_types.
Source Page: https://knative.dev/docs/eventing/

- [ ] 404 Error navigating to; 

https://github.com/knative/eventing-contrib/blob/master/contrib/awssqs/pkg/apis/sources/v1alpha1/aws_sqs_types.go
https://github.com/knative/eventing-contrib/blob/master/contrib/gcppubsub/pkg/apis/sources/v1alpha1/gcp_pubsub_types.go
https://github.com/knative/eventing-contrib/blob/master/contrib/github/pkg/apis/sources/v1alpha1/githubsource_types.go
Source Page: https://knative.dev/docs/eventing/sources/

- [ ] 404 Error navigating to; 

https://github.com/knative/eventing/tree/master/contrib/gcppubsub/config
https://github.com/knative/eventing/tree/master/contrib/natss/config
https://github.com/knative/eventing/tree/master/contrib/natss/config/provisioner
Source Page: https://knative.dev/docs/eventing/channels/channels-crds/

[...]

PR #1974, opened 2019-11-09
fix broken links (from link checking report tool testing)

RichieEscarez:
contributes to fixing items in #1750

/cc @adrcunha  @eallred-google

### DIFF
--- a/community/samples/README.md
+++ b/community/samples/README.md
@@ -6,7 +6,7 @@ original author is unable to maintain their contribution. If you find that
 something isn't working, lend a helping hand and fix it in a PR.
 [Learn more about the lifespan of samples](../../contributing/DOCS-CONTRIBUTING.md).
 
-[**See all Knative code samples**](../../docs/samples/)
+[**See all Knative code samples**](../../../docs/samples/)
 
 ### Serving samples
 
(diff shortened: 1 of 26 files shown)


adrcunha:
The link checker failed because the script isn't available. Looks like this PR is based on an old commit. Any chance you can rebase, Richie?

RichieEscarez:
> The link checker failed because the script isn't available. Looks like this PR is based on an old commit. Any chance you can rebase, Richie?

I rebased and it failed again. This is against the "release-0.10" branch (vs. "master") and I don't think we checked in that script to this branch?

adrcunha:
Good catch, I didn't realize that this is not the master branch. Indeed, unless we check in the presubmit script (do we want to?) the job will fail.

RichieEscarez:
> Good catch, I didn't realize that this is not the master branch. Indeed, unless we check in the presubmit script (do we want to?) the job will fail.

Is there additional work associated with adding that script to the `release-0.10` branch? I would be nice to have it but I dont think it warrants any extra effort on top of just opening a PR to copy that file there.

On that note, will there be any additional "release" work needed when we cut/create a release branch for an archived docs version (ie. for the "release-0.11" branch)?

adrcunha:
> > Good catch, I didn't realize that this is not the master branch. Indeed, unless we check in the presubmit script (do we want to?) the job will fail.
> 
> Is there additional work associated with adding that script to the `release-0.10` branch? It would be nice to have it but I dont think it warrants any extra effort on top of just opening a PR to copy that file there.

No, it's just a cherrypick PR.

> On that note, will there be any additional "release" work needed when we cut/create a release branch for an archived docs version (ie. for the "release-0.11" branch)?

No, at that point the script will be already in the branch.

PR #1984, opened 2019-11-14
Cleanup Knative install versions page

dgerd:
1. Change to a command that returns just the release version
2. Move current methods of querying to the top of the page
3. Combine old methods into a single section
  * Describe old steps rather than walk through as we expect very few
  people to be running these older versions.

### DIFF
--- a/docs/install/check-install-version.md
+++ b/docs/install/check-install-version.md
@@ -5,39 +5,33 @@ weight: 20
 type: "docs"
 ---
 
-## Knative Serving
 
-To query the version of Knative Serving running on the cluster enter the
-following command:
+To obtain the version of the Knative component that you have running on your cluster, you query for the 
+`[component].knative.dev/release` label with the following commands:
 
-```bash
-kubectl get namespace knative-serving -o 'go-template={{index .metadata.labels
-"serving.knative.dev/release"}}'
-```
+* Knative Serving
+   ```bash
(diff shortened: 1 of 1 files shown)


docs/install/check-install-version.md
```diff
@@ -5,97 +5,33 @@ weight: 20
...
-If you aren't already signed in, you'll need to sign in to a Google account
-before you can view the container details.
-
-On the container details page, you'll see a section titled "Container
-classification," and in that section is a list of tags. The versions of Knative
-you have installed will appear in the list as `v0.1.1`, or whatever version you
-have installed:
-
-![Shows list of tags on container details page; v0.1.1 is the Knative version and is the first tag.](../images/knative-version.png)
+1. Query your Knative Controller to receive a URL.
```
RichieEscarez:
```suggestion
1. Query your Knative Controller to receive the `Image` URL. For example, you can run `kubectl describe deploy controller --namespace knative-serving` or `kubectl describe deploy eventing-controller --namespace knative-eventing`
```

RichieEscarez: I did a quick edit. 
Also, Im okay with leaving this out too if you feel strongly about expecting nobody to need those details. 
Otherwise this looks great! Thanks @dgerd (LMK and ill merge)

dgerd: I like the edit. Lets leave them in few a little while longer, but we should consider removing this section after a few more releases.

PR #1987, opened 2019-11-15
This adds a simple cloudevents-go sample.

mattmoor:
This example is intended to showcase how to start an event receiver as a Knative
Service, and then have that service either send a new event to a `$SINK` or
respond with it (e.g. for working with a Broker).

Once we're happy with this, we should broaden this to cover more languages.

/assign @duglin @n3wscott

### DIFF
--- a/docs/serving/samples/cloudevents/cloudevents-go/README.md
+++ b/docs/serving/samples/cloudevents/cloudevents-go/README.md
@@ -1,0 +1,120 @@
+---
+title: "Cloud Events - Go"
+linkTitle: "Go"
+weight: 1
+type: "docs"
+---
+
+A simple web app written in Go that can receive and send Cloud Events that you
+can use for testing.  It supports running in two modes:
+1. The default mode has the app reply to your input events with the output
+  event, which is simplest for demonstrating things working in isolation,
+  but is also the model for working for the Knative Eventing `Broker` concept.
+
+2. `K_SINK` mode has the app send events to the destination encoded in `$K_SINK`,
+  which is useful to demonstrate how folks can synthesize events to send to
+  a Service or Broker when not initiated by a Broker invocation (e.g.
+  implementing an event source)
+
+The application will use `$K_SINK`-mode whenever the environment variable is specified.
+
+Follow the steps below to create the sample code and then deploy the app to your
+cluster. You can also download a working copy of the sample, by running the
+following commands:
+
+```shell
+git clone -b "{{< branch >}}" https://github.com/knative/docs knative-docs
+cd knative-docs/docs/serving/samples/cloudevents/cloudevents-go
+```
+
+## Before you begin
+
+- A Kubernetes cluster with Knative installed and DNS configured. Follow the
+  [installation instructions](../../../../install/README.md) if you need to
+  create one.
+- [Docker](https://www.docker.com) installed and running on your local machine,
+  and a Docker Hub account configured (we'll use it for a container registry).
+
+## The sample code.
+
+1. If you look in `cloudevents.go`, you will see two key functions for the
+   different modes of operation:
+
+   ```go
+   func (recv *Receiver) ReceiveAndSend(ctx context.Context, event cloudevents.Event) error {
+     // This is called whenever an event is received if $K_SINK is set, and sends a new event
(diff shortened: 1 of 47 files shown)


docs/serving/samples/cloudevents/cloudevents-go/cloudevents.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	log.Printf("Got an event from: %q", req.Name)
+
+	resp := handle(req)
+	log.Printf("Sending event: %q", resp.Message)
+
+	r := cloudevents.NewEvent(cloudevents.VersionV1)
+	r.SetType("dev.knative.docs.sample")
+	r.SetSource("https://github.com/knative/docs/docs/serving/samples/cloudevents/cloudevents-go")
+	r.SetData(resp)
```
matzew: `SetDataContentType` ?

docs/serving/samples/cloudevents/cloudevents-go/README.md
```diff
@@ -0,0 +1,120 @@
...
+type: "docs"
+---
+
+A simple web app written in Go that can receive and send Cloud Events that you
+can use for testing.  It supports running in two modes:
+1. The default mode has the app reply to your input events with the output
+  event, which is simplest for demonstrating things working in isolation,
+  but is also the model for working for the Knative Eventing `Broker` concept.
+
+2. `SINK` mode has the app send events to the destination encoded in `$SINK`,
```
n3wscott: the source contract us using `K_SINK`. We need to change it here or change the spec for sources.

n3wscott:
Other than the question around `SINK` vs `K_SINK`, LGTM

docs/serving/samples/cloudevents/cloudevents-go/README.md
mattmoor: Done

docs/serving/samples/cloudevents/cloudevents-go/cloudevents.go
mattmoor: Done

docs/serving/samples/cloudevents/cloudevents-go/cloudevents.go
```diff
@@ -0,0 +1,105 @@
...
+	log.Printf("Sending event: %q", resp.Message)
+
+	r := cloudevents.NewEvent(cloudevents.VersionV1)
+	r.SetType("dev.knative.docs.sample")
+	r.SetSource("https://github.com/knative/docs/docs/serving/samples/cloudevents/cloudevents-go")
+	r.SetDataContentType("application/json")
+	r.SetData(resp)
+
+	ctx = cloudevents.ContextWithTarget(ctx, recv.Target)
+	_, _, err := recv.client.Send(ctx, event)
```
duglin: Would it make sense to return what the sink might send back?

mattmoor: You tell me :)

cc @n3wscott

mattmoor: What I mean is: I understand the Source use-case (Send) and the Broker use-case (Recv/Return), but I'm not 100% clear on the use case for what you describe.  So I'd defer you you all who sit in the cloudevents meetings.

cc @markpeek too

duglin: well, we have flows in eventing where the sink can return an event, so I just thought it might be nice to show what that even is (if it's there) instead of just always returning a 202.  I assume this returns a 202 (or 200) w/o a body, right?

mattmoor: > we have flows in eventing where the sink can return an event,

Right, this is why the sample also has ReceiveAndReply (below).  I am just wondering what the case is where we receive an event, send an event elsewhere, and then reply with the event the recipient responded to us with 😅

duglin: just to show the result of the flow, otherwise we're just bit-bucketing the data.  I kind of view this like showing the result of a KnEventing Sequence - except just one step is used.

PR #1991, opened 2019-11-18
specify main Go file to support locally running golang helloworld

syedashrafulla:
Within this directory, `go build -v -o server && chmod +x server && ./server` fails with "syntax error near unexpected token `newline'" because there is no main Go file in the build. Adding the main Go file like `go build -v -o server helloworld.go && chmod +x server && ./server` works, which is the proposed file change. With this file change, we can then deploy the helloworld golang container locally via:
```
docker build . --tag gcr.io/monitoring-api-test/helloworld' && docker run --publish 8080:8080 --env PORT=8080 --name local gcr.io/monitoring-api-test/helloworld
```

Fixes #issue-number

## Proposed Changes

-
-
-

### DIFF
--- a/docs/serving/samples/hello-world/helloworld-go/Dockerfile
+++ b/docs/serving/samples/hello-world/helloworld-go/Dockerfile
@@ -16,7 +16,7 @@ COPY . ./
 
 # Build the binary.
 # -mod=readonly ensures immutable go.mod and go.sum in container builds.
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=readonly -v -o server
+RUN CGO_ENABLED=0 GOOS=linux go build -mod=readonly -v -o server helloworld.go
 
 # Use the official Alpine image for a lean production container.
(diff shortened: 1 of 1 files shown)


RichieEscarez:
The unit test needs to be resolved too:

sampleconsistency_test.go:97: README.md file is missing line 18 ('RUN CGO_ENABLED=0 GOOS=linux go build -mod=readonly -v -o server helloworld.go') from file '../../docs/serving/samples/hello-world/helloworld-go/Dockerfile'
    Additional info:
    All files required for running sample apps are checked against README.md, the content of source files should be identical with what's in README.md file, the list of the files to be verified is the same set of files used for running sample apps, they are configured in `/test/sampleapp/config.yaml`. If an exception is needed the file can be configured to be copied as a separate step in `PreCommand` such as: https://github.com/knative/docs/blob/65f7b402fee7f94dfbd9e4512ef3beed7b85de66/test/sampleapp/config.yaml#L4

PR #1992, opened 2019-11-18
Remove namespace definition from hello-world service.yaml

sshariff01:
## Proposed Changes

Remove the namespace definition from the service.yaml in each of the helloworld-*
samples.

### Why

Specifying the namespace in service.yaml expects there to exist a namespace with
the specified name, in this case 'default'.  When namespace is unspecified, the
service.yaml will default to using the namespace specified in the default $KUBECONFIG.

### DIFF
--- a/docs/serving/samples/hello-world/helloworld-csharp/README.md
+++ b/docs/serving/samples/hello-world/helloworld-csharp/README.md
@@ -126,7 +126,6 @@ cd knative-docs/docs/serving/samples/hello-world/helloworld-csharp
    kind: Service
    metadata:
      name: helloworld-csharp
-     namespace: default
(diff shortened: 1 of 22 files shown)


sshariff01:
@googlebot I signed it!

sshariff01:
Link checks as part of `pull-knative-docs-markdown-link-check` fail with the following output:

```
Activated service account credentials for: [prow-job@knative-tests.iam.gserviceaccount.com]
+ ./test/presubmit-link-check.sh
/usr/local/bin/runner.sh: line 119: ./test/presubmit-link-check.sh: No such file or directory
+ EXIT_VALUE=127
+ set +o xtrace
```

PR #1993, opened 2019-11-20
Updating PubSub example due to change in Destination

nachocano:
## Proposed Changes

- Updating PubSub example due to change in Destination


nachocano:
@samodell friendly ping

PR #2004, opened 2019-11-26
Update installing-istio.md

l31g:
Adding the download Istio step to the manual sidecar injection section.

Fixes #issue-number

## Proposed Changes

-
-
-


l31g:
@googlebot I signed it!

PR #2019, opened 2019-12-04
reflecting 1.0 CE payload

matzew:
doc updates for https://github.com/knative/eventing/pull/2259

/assign @nachocano


matzew:
@evankanderson for the `ApiServerSource` we have a subject:

* https://github.com/knative/eventing/blob/master/pkg/adapter/apiserver/events/events.go#L102

For `CronjobSource`, I do not really think a subject is needed :thinking:

PR #2022, opened 2019-12-05
Update C# sample to .Net Core 3.1 

### LINKED ISSUE #2021: Update C# sample to .Net Core 3.1
https://github.com/knative/docs/tree/master/docs/serving/samples/hello-world/helloworld-csharp

.Net Core 3.1 is available.

meteatamel:
Fixes #2021 

## Proposed Changes

- Updated C# sample to use .NET Core 3.1


EatonZ:
It might be worth updating your Dockerfile to look more like the [Microsoft default](https://docs.microsoft.com/en-us/visualstudio/containers/container-build?view=vs-2019).
Mainly in regard to the order of image layers/which images are used.
Their example is slightly out of date in regard to the versions, but the documentation still makes sense. [Here's what a fresh 3.1 Dockerfile looks like](https://pastebin.com/raw/MBLwZqQV).

I have been experimenting with Cloud Run lately and am currently unsure whether to follow your Dockerfile or Microsoft's. Let me know if your example is specifically designed for optimal Cloud Run compatibility/performance.

meteatamel:
@EatonZ Thanks for the reference. No, the Dockerfile I had is not specifically designed for Knative/CloudRun. I'd have to compare and analyze the two Dockerfiles to see what the difference is. I'll update here once I get a chance to do that.

grayside:
Do we want to proceed with this as an interim state to get to 3.1 or wait on a revised approach in this PR?

meteatamel:
@grayside My vote is to merge this as it is now. The suggestion from @EatonZ can be considered an optimization that I can investigate when I get a chance.

grayside:
Let's see if I have this power:

/approved

samodell:
/approve

@meteatamel  feel free to ping me if you see things sit around too long!

PR #2036, opened 2019-12-11
Documentation fix for .NET core sample on Docker for Windows

rahulrai-in:
I had to make the following modifications to get the sample working on Docker for Windows.

1. Specify the port type and value.
2. Specify the Hostname in the header of cURL request. 

The URL received by executing the `kubectl get ksvc` command is also not correct. I can update the output of the command as well if the reviewers request that change as well.

Fixes #issue-number

## Proposed Changes

- Update the documentation

### DIFF
--- a/docs/serving/samples/hello-world/helloworld-csharp/README.md
+++ b/docs/serving/samples/hello-world/helloworld-csharp/README.md
@@ -134,6 +134,9 @@ cd knative-docs/docs/serving/samples/hello-world/helloworld-csharp
              env:
                - name: TARGET
                  value: "C# Sample v1"
+             ports:
+               - name: http1
+                 containerPort: 80
    ```
 
 ## Building and deploying the sample
@@ -181,7 +184,7 @@ folder) you're ready to build and deploy the sample app.
    the URL below with the URL returned in the previous command.
 
    ```shell
-   curl http://helloworld-csharp.default.1.2.3.4.xip.io
+   curl -H "Host: helloworld-csharp.default.1.2.3.4.xip.io" http://localhost
    Hello C# Sample v1!
    ```
 


rahulrai-in:
@googlebot I signed it!

grayside:
Hi @rahulrai-in thanks for your PR! Looking it over, I hadn't noticed before that we're suggesting these docs can guide use of the sample for a local docker instance. Cool! I suspect the problems here are not C#-specific, but systemic and related to local Docker or Docker for Windows.

* Overriding the port looks like an end-run around a networking challenge that's not normally needed with a cloud-based cluster.
* Using `localhost` in the curl command would limit it to working locally.

Ideally, the port mapping would not be needed, and the "xip.io" URL has dedicated steps to point back to your localhost.

@RichieEscarez this looks like a broader problem than this PR's current scope.

RichieEscarez:
Thanks @grayside!

@mattmoor  @dgerd  any insight here about running Docker on Windows?

/cc @evankanderson

rahulrai-in:
> Hi @rahulrai-in thanks for your PR! Looking it over, I hadn't noticed before that we're suggesting these docs can guide use of the sample for a local docker instance. Cool! I suspect the problems here are not C#-specific, but systemic and related to local Docker or Docker for Windows.
> 
> * Overriding the port looks like an end-run around a networking challenge that's not normally needed with a cloud-based cluster.
> * Using `localhost` in the curl command would limit it to working locally.
> 
> Ideally, the port mapping would not be needed, and the "xip.io" URL has dedicated steps to point back to your localhost.
> 
> @RichieEscarez this looks like a broader problem than this PR's current scope.

Hi @grayside, can you point me to the source code where the hostname `xip.io` maps to `localhost`? I might be able to locate the issue.

dgerd:
You can read how xip works on their site: http://xip.io/ 

xip will just return back the IP address within the domain name. So in this example your response will be `1.2.3.4`. This assumes that you have the ability to egress and reach xip for DNS.

I don't have any first hand experience with Docker for Windows and think @grayside covered all my feedback.

rahulrai-in:
@dgerd @mattmoor I was able to successfully run Knative on Docker for Windows. Here is a brief writeup if you are interested. https://thecloudblog.net/post/durable-cloud-event-production-from-knative-container-source-with-azure-durable-task-framework/

dgerd:
@rahulrai-in Nice writeup!

Regarding this PR I am still not sure that the Host changes or port changes are what we want for the C# example as they will not generically work in most cloud environments. These seem more like platform specific changes to work in Docker for Windows rather than language specific changes within the example. 

I would rather see this called out or explained separate from the example.

docs/serving/samples/hello-world/helloworld-csharp/README.md
```diff
@@ -181,7 +184,7 @@ folder) you're ready to build and deploy the sample app.
    the URL below with the URL returned in the previous command.
 
    ```shell
-   curl http://helloworld-csharp.default.1.2.3.4.xip.io
+   curl -H "Host: helloworld-csharp.default.1.2.3.4.xip.io" http://localhost
```
evankanderson: This shouldn't necessarily point to localhost -- it depends on where your loadbalancer lives. I'd suggest instead editing `config-domain` to have xip.io reflect the LoadBalancer or NodePort's address.

docs/serving/samples/hello-world/helloworld-csharp/README.md
```diff
@@ -134,6 +134,9 @@ cd knative-docs/docs/serving/samples/hello-world/helloworld-csharp
              env:
                - name: TARGET
                  value: "C# Sample v1"
+             ports:
+               - name: http1
+                 containerPort: 80
```
evankanderson: On line 52 of this file, it says:

```csharp
       string port = Environment.GetEnvironmentVariable("PORT") ?? "8080";
       string url = String.Concat("http://0.0.0.0:", port);
```

So I don't think this should say 80?

abrennan89:
@evankanderson @rahulrai-in this looks like it was raised against version 0.10, so I recommend just closing it at this point.
Please raise an issue in the repo if there's still work to do on this for the latest Knative version.

PR #2045, opened 2019-12-12
Added instructions for installing kn

### LINKED ISSUE #2044: Create `Installing `kn` CLI` docs
Add documentation in the new Client section for how to download and install `kn` CLI.

abrennan89:
Fixes #2044 

## Proposed Changes

- Added nightly executable binary method
- Added container images
- Added github / Go method


samodell:
I think we can at least use a link from Serving/Eventing installs to these instructions, vs. copying them into those topics as well. I'm imagining we'll add an optional step to the installation guides that points to these steps for installing kN.

I've been talking with Richie, and I think what we're thinking is that the /client/ folder should be moved into the /reference/ section, but then this page on installing the client should live in the /install/ folder. I think that's all captured in the nav proposal: https://docs.google.com/spreadsheets/d/1L3EAeTX7RRdZhcQZeNmxwN4DCncuImCXA5A1_KQvwSk/edit#gid=413198505

I apologize for not cross-referencing your first PR with the nav proposal; I should have! Does that seem reasonable, @abrennan89 ?

abrennan89:
Adding a link to these docs from Serving, Eventing and the install section sounds like a good plan @samodell - I was sort of waiting to see if we could include the content instead, which is why I was asking about the re-use mechanism for .md files, but you can let me know if you'd prefer to include the file or just link out to this doc.

If we move the Client section to the Reference section, where would the user stories that we document as tasks live? Would these just be split out and included under Serving and Eventing?

That makes sense to me, but the suggestion to have a separate Client section at the same level as Serving and Eventing also came from @rhuss , so I'd just like to check with you Roland that there are no `kn` use cases that don't relate directly to Serving or Eventing and would need to be documented in a Client specific section?

Thanks folks for all the input and your patience!

docs/install/install-kn.md
```diff
@@ -0,0 +1,26 @@
+---
+title: "Installing the `kn` Client"
+linkTitle: "Installing the `kn` Client"
+weight: 10
+type: "docs"
+---
+
+## Install `kn` using the nightly executable binary
```
samodell: Shouldn't we also offer installing the latest release of the client? I'm not sure if most users want to download and work with the nightly release.

https://github.com/knative/client/blob/master/docs/README.md has instructions on downloading the most recent release.

samodell: We should put an intro sentence on the page, so that the title doesn't sit right on top of this h2.

docs/install/install-kn.md
```diff
@@ -0,0 +1,26 @@
...
+To install the `kn` Client, you must download the executable binary for your system. Links to the latest nightly executable binary are available here:
```
samodell: Insert a new line after "available here:"

docs/install/install-kn.md
```diff
@@ -0,0 +1,26 @@
...
+  - [macOS](https://storage.googleapis.com/knative-nightly/client/latest/kn-darwin-amd64)
+  - [Linux](https://storage.googleapis.com/knative-nightly/client/latest/kn-linux-amd64)
+  - [Windows](https://storage.googleapis.com/knative-nightly/client/latest/kn-windows-amd64.exe)
+
+You must place the executable binary in your system path, and make sure that it is executable.
+
+## Install `kn` using Go
```
samodell: I don't think we should assume that folks have Go installed. Or, if we are going to assume that, we should mention it in a "Before you begin" section.

docs/install/install-kn.md
samodell: I also think we should offer some information about who should install the nightly executable, and probably a warning that it may not be super stable?

docs/install/install-kn.md
```diff
@@ -0,0 +1,26 @@
...
+1. Check out the [Client repository](https://github.com/knative/client).
+1. Run the command:
+  ```
+  go install ./cmd/kn
+  ```
+
+## `kn` container images
```
samodell: What are these container images for? Can we explain what user might do with these?

docs/install/install-kn.md
```diff
@@ -0,0 +1,26 @@
...
+The `kn` container images are available here:
+- [Nightly container image](gcr.io/knative-nightly/knative.dev/client/cmd/kn)
+- [Latest release](gcr.io/knative-releases/knative.dev/client/cmd/kn)
```
samodell: This PR should also delete the /client/ directory we added -- sorry about the confusion on that!

samodell: Oops, strike that. Right, we're moving the whole section into reference.

docs/install/install-kn.md
abrennan89: The 'latest' release is also covered under the *kn container images* section of the docs: https://knative.dev/development/install/install-kn/

I will check in with the CLI team about what to put for a warning/audience note about the nightly executable.

docs/install/install-kn.md
abrennan89: I added this to cover installing 'latest' - I think it's just how the CLI is installed if you're not using the binaries or Go methods.
The information came from: https://github.com/knative/client/blob/master/docs/README.md
i.e.
To use the kn container image:

    Nightly: gcr.io/knative-nightly/knative.dev/client/cmd/kn
    Latest release: gcr.io/knative-releases/knative.dev/client/cmd/kn

PR #2048, opened 2019-12-13
404 Fixes and Cherry-picks from 0.10

RichieEscarez:
- Fixed some missed 404s
- Update the syntax for the readfile (prevent others from copying the outdated syntax)
- Add all the back-log of cherry picks that were add to only 0.10 and never pushed into master. 
   See commits for details.


samodell:
CLA overridden because all these commits have already been submitted to Knative in a different branch.

PR #2052, opened 2019-12-13
[WIP] Initial commit to restructure reference section

### LINKED ISSUE #1883: Restructure the "Reference" section
Revise the "Reference and API" section:

- [x] Rename the section to "Reference"
- [x] Move the API docs down into a sub-section "API"
   - [x] Position Serving above Eventing
- [x] Add a new "Release notes" section (initially act as a landing page that links to all the repo "Releases" pages)
- [x] Move and rename "Concepts > Resources" topic as a "Related resources" sub-topic
     - [x] Add `aliases` to frontmatter to set redirect from past filename/url
- [ ] Move "install files and dependencies" matrix into new ref topic https://github.com/knative/docs/issues/1896

[View proposed structure](https://docs.google.com/spreadsheets/d/1L3EAeTX7RRdZhcQZeNmxwN4DCncuImCXA5A1_KQvwSk/edit?ts=5d826691#gid=413198505&range=187:187)

abrennan89:
Fixes #1883 

## Proposed Changes

- Renamed to Reference section
- Added Serving and Eventing sub-folders, re-organized existing files
- Added API sub-folder, moved README.md

Note: This PR is still WIP; this is just the notes for the initial commit. Will follow up in future commits/comments.


RichieEscarez:
Sorry, ive been out sick most of this week @abrennan89. I havent dug in but a few high level comments:

- These files have dependencies on some of the scripts so all changes should align.

- Fixes for the hard coded paths of the moved files, in all the other files in the knative/docs repo should be added here too.

- There should also be a sweep and likely subsequent PRs for all the new files paths in peer repos (ie if knative/serving links to one of these files that is being moved, we should update those paths in that repo too)

- Lots of introductory info is missing. For example, we need to clarify why there is a list of links to ensure that users dont assume they are navigating to something else (and avoid user frustration).
  - Another point to make is that this reference section will include the API and CLI docs (ie, the main README should cover these details). 
  - The empty `/[client](https://github.com/knative/docs/tree/master/docs/client)/` folder that was recently created should be move down into here. Remember that the `kn` client is just a CLI tool, its not a product feature and therefore should live down in the Reference section with all other ref docs (kn is just one of the available tools/methods that you can use to accomplish a task).

- Question: Do you plan to add more content in to the new `/serving/` folder that you created? I dont think pushing the serving API down deeper into another folder is necessary unless it is to contain more Serving API specific content (and even then, we should avoid creating a folder until that additional content is ready to be placed within it).

We can chat more about this next week when I back to 100%.

Thanks for all the hard work!!!!

abrennan89:
Hi @RichieEscarez :slightly_smiling_face: thanks for the comments, I hope you're feeling better and that I didn't disturb you while you were sick!

A few questions/comments:

> These files have dependencies on some of the scripts so all changes should align.

I'm not sure what you mean by this?

> Fixes for the hard coded paths of the moved files, in all the other files in the knative/docs repo should be added here too.

Should we maybe start adding a comment to files with a list at the beginning of all the files that link to them (i.e. this file is linked to in the following files: )? This would make it easier to know all of the places where links need to be updated when a file is moved. If it's just a case of adding aliases now that I know the format I can do this :slightly_smiling_face: 

> Lots of introductory info is missing. For example, we need to clarify why there is a list of links to ensure that users dont assume they are navigating to something else (and avoid user frustration).

I plan to flesh out the context info as I can in this PR. The initial commits were just to get the structure right but I'll add some stuff for this for sure.

> Question: Do you plan to add more content in to the new /serving/ folder that you created? I dont think pushing the serving API down deeper into another folder is necessary unless it is to contain more Serving API specific content (and even then, we should avoid creating a folder until that additional content is ready to be placed within it).

I had only done this to make it consistent with the structure for Eventing, but I see now why they're different. I will undo this change unless we have more information to add for Serving under this PR.


Thanks so much for your help!

abrennan89:
> The empty /client/ folder that was recently created should be move down into here. Remember that the kn client is just a CLI tool, its not a product feature and therefore should live down in the Reference section with all other ref docs (kn is just one of the available tools/methods that you can use to accomplish a task).

I still think we need a Client section at the component level since there are installation instructions etc for `kn` that are user tasks rather than reference docs? I can copy the empty folder to the reference section too though for `kn` reference docs. Wdyt?
CC @rhuss

RichieEscarez:
> > The empty /client/ folder that was recently created should be move down into here. Remember that the kn client is just a CLI tool, its not a product feature and therefore should live down in the Reference section with all other ref docs (kn is just one of the available tools/methods that you can use to accomplish a task).
> 
> I still think we need a Client section at the component level since there are installation instructions etc for `kn` that are user tasks rather than reference docs? I can copy the empty folder to the reference section too though for `kn` reference docs. Wdyt?
> CC @rhuss

A CLI tool doesnt belong at the top level, its not a feature of Knative and should just exist throughout the docs as an integral piece. To keep it brief, an example of what we are going for similar to the kubectl, its just a K8s tool and the overview shows only in the reference area (otherwise, where you can use the tool (ie a k8s task) is where you see mention of kubectl). We want that same model for kn, therefore kn gets installed as a default step with the installation of a Knative component and thereafter, if there is a Knative task where you can use kn, in that task is where we mention the command.

RichieEscarez:
> I'm not sure what you mean by this?
> 
Sorry about not being clear. The API doc build script outputs the files in the same folder structure with the same filenames as the structure today. Since you plan to revert the extra serving folder, updating the script is no longer necessary.

> Should we maybe start adding a comment to files with a list at the beginning of all the files that link to them (i.e. this file is linked to in the following files: )? This would make it easier to know all of the places where links need to be updated when a file is moved. If it's just a case of adding aliases now that I know the format I can do this 
> 
I don't think thats necessary nor effective because that would just add a lot of maintenance overhead to keep all those lists up to date, let alone, it would be impossible to know when someone in another repo creates a new or changes a link.

Instead, you can search at the Knative Org level, which will give you a list of all the files and in which repos those files reside. [Here is an example](https://github.com/search?q=org%3Aknative+%22reference%2Fserving%22&type=Code).

> I plan to flesh out the context info as I can in this PR. The initial commits were just to get the structure right but I'll add some stuff for this for sure.

Okay, cool. FYI. You can set your PR to "draft" to indicate that its not ready for review. You can/should also add the hold/in-progress label too.

samodell:
Re: finding what links need to be updated, I find it easiest to do a grep search of the whole repo on the command line to find the relevant files:

`grep -Ril URL/to/search/for location/to/search`

Then I can open all the files up in Sublime (I actually wrap the grep command with sublime, so instead of getting the output on the command line they automatically pop up in sublime), which has a nifty "find in files" feature that lets you find/replace across multiple files. (You could also do the find/replace from the command line too, but I prefer Sublime for that part.) Not sure what your toolchain looks like for working on these docs, though.

docs/reference/API/README.md
```diff
@@ -0,0 +1,111 @@
...
+use a commit for Serving v0.3.0, you would use
+[4d198d](https://github.com/knative/serving/commit/4d198db8756db2f8a3c228302a97fb3a216a9475).
+
+Using tags from each repo, you would run the following command to generate the
+v0.3.0 API source files:
+
+```
+KNATIVE_SERVING_COMMIT=v0.3.0 \
+KNATIVE_EVENTING_COMMIT=v0.3.0 \
+KNATIVE_EVENTING_CONTRIB_COMMIT=v0.3.0 \
```
mattmoor: Hmm, I don't think we can use shortcodes here, but that might be better if we can manage so that it isn't pointing to something ancient.

abrennan89:
Closing in favor of a new PR once we figure out exactly the plan for this section.

PR #2054, opened 2019-12-13
Adjust API reference docs scripts for knative.dev

bbrowning:
The API reference docs were failing to generate because they didn't
account for the change from github.com/knative/serving to
knative.dev/serving and similar for the eventing and eventing-contrib
repos.

### DIFF
--- a/hack/gen-api-reference-docs.sh
+++ b/hack/gen-api-reference-docs.sh
@@ -26,14 +26,17 @@ REFDOCS_REPO="https://${REFDOCS_PKG}.git"
 REFDOCS_VER="v0.1.5"
 
 KNATIVE_SERVING_REPO="github.com/knative/serving"
+KNATIVE_SERVING_LOCAL_REPO="knative.dev/serving"
 KNATIVE_SERVING_COMMIT="${KNATIVE_SERVING_COMMIT:?specify the \$KNATIVE_SERVING_COMMIT variable}"
 KNATIVE_SERVING_OUT_FILE="serving.md"
 
 KNATIVE_EVENTING_REPO="github.com/knative/eventing"
+KNATIVE_EVENTING_LOCAL_REPO="knative.dev/eventing"
 KNATIVE_EVENTING_COMMIT="${KNATIVE_EVENTING_COMMIT:?specify the \$KNATIVE_EVENTING_COMMIT variable}"
 KNATIVE_EVENTING_OUT_FILE="eventing/eventing.md"
 
 KNATIVE_EVENTING_CONTRIB_REPO="github.com/knative/eventing-contrib"
+KNATIVE_EVENTING_CONTRIB_LOCAL_REPO="knative.dev/eventing-contrib"
 KNATIVE_EVENTING_CONTRIB_COMMIT="${KNATIVE_EVENTING_CONTRIB_COMMIT:?specify the \$KNATIVE_EVENTING_CONTRIB_COMMIT variable}"
 KNATIVE_EVENTING_CONTRIB_OUT_FILE="eventing/eventing-contrib.md"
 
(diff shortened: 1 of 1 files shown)


bbrowning:
I believe this fixes #1661

RichieEscarez:
@bbrowning does that mean you are able to build the 0.11 ref docs successfully (ie. complete the steps [here](https://github.com/knative/docs/blob/master/docs/reference/README.md))?

ahmetb:
Not sure how this fixes that.
What does the local_repo variable do? Did the import path for Knative go pkgs change?

bbrowning:
@RichieEscarez Yes - the script works to generate the 0.11 docs, with one minor change due to me having Go 1.13 locally that requires GOPATH to be set for `go mod` and thus I have to uncomment the 3 lines in gen-api-reference-docs.sh that check to see if it's unset.

That's an unrelated, different issue either due to something in my environment or Go 1.13. But, either way, the reference API docs do generate successfully for recent Knative releases now.

bbrowning:
@ahmetb yes, the Go import path for all the Knative repos changed to use the vanity knative.dev domain. Verbose output from the docs generation showed me that it was failing to resolve the Knative imports and thus leading to all the invalid type errors.

ahmetb:
@bbrowning can you upload the resulting html somewhere like a gist? let's see if it fixes the issues I saw around type parsing.

bbrowning:
I went ahead and opened a PR with the updated v0.11.0 reference docs in #2055. That PR is against master, but perhaps should be against the release-0.11 branch? Either way you should be able to see the diff as well as the new rendered markdown from that PR to check for anything else that still looks wrong.

RichieEscarez:
Awesome! Thanks @bbrowning! I just ran your script and it worked for me too. Ill wait for the green light from @ahmetb before moving forward.

ahmetb:
@RichieEscarez were you able to get serving generated correctly as well?
if so, let's go ahead with it.

hack/gen-api-reference-docs.sh
```diff
@@ -26,14 +26,17 @@ REFDOCS_REPO="https://${REFDOCS_PKG}.git"
 REFDOCS_VER="v0.1.5"
 
 KNATIVE_SERVING_REPO="github.com/knative/serving"
+KNATIVE_SERVING_LOCAL_REPO="knative.dev/serving"
 KNATIVE_SERVING_COMMIT="${KNATIVE_SERVING_COMMIT:?specify the \$KNATIVE_SERVING_COMMIT variable}"
 KNATIVE_SERVING_OUT_FILE="serving.md"
 
 KNATIVE_EVENTING_REPO="github.com/knative/eventing"
+KNATIVE_EVENTING_LOCAL_REPO="knative.dev/eventing"
```
ahmetb: let's rename `_LOCAL_REPO` to `_IMPORT_PATH` if you don't mind.

ahmetb:
I'm just surprised it works without fixing anything but that's cool.

RichieEscarez:
Here are the staging preview build:

https://5df41e57527ac4000777c447--knative.netlify.com/docs/reference/serving-api/
https://5df41e57527ac4000777c447--knative.netlify.com/docs/reference/eventing/

@ahmetb yeah, i get the same output

hack/gen-api-reference-docs.sh
RichieEscarez: @ahmetb done, PTAL

bbrowning:
@RichieEscarez Those broken links need updating in Knative Serving itself. There were erroneously replaced from their github.com URL in https://github.com/knative/serving/pull/4521

bbrowning:
knative/serving#6225 fixes the root issue in Serving so we won't have these broken links in our 0.12 release. I added a 2nd commit to #2055 to fix the links directly in the docs for 0.11 - the commit message for it includes the sed command I ran to automatically fix all of these, similar to the sed command used to fix this in Serving itself.

RichieEscarez:
Zero broken links in the new preview build: https://5df42f87a20218f204084c22--knative.netlify.com/docs/reference/

Thanks @bbrowning!

@ahmetb any concerns with these changes? Can we merge?

PR #2074, opened 2019-12-20
Fix wording to address bug #2023 plus small formatting updates

### LINKED ISSUE #2023: Eventing install docs suggest the command will installs the wrong thing
Page: https://knative.dev/docs/eventing/getting-started/

This claims it installs "eventing sources" but actually this installs the entire Eventing release and NO "eventing sources".

> image

AFAICT eventing sources install isn't documented.

/priority P1 
if that works

samodell:
Fixes #2023 

## Proposed Changes

- Update wording to reflect what's actually happening during install (the guide doesn't actually need to install Eventing sources; the install is just split into two steps so that the CRDs are installed first and then everything else is installed on the second pass)

- Update formatting to add whitespace and use automatic numbering

### DIFF
--- a/docs/eventing/getting-started.md
+++ b/docs/eventing/getting-started.md
@@ -58,7 +58,10 @@ If the `knative-eventing` namespace or the `imc-controller-*` does not exist, us
     kubectl apply --filename https://github.com/knative/eventing/releases/download/{{< version >}}/release.yaml
     ```
 
-   Installing the CRDs first can prevent race conditions and failures during the install.
+   Installing the CRDs first can prevent race conditions, which may cause failures during the install.
+   Race conditions exist during install because a CRD must be defined before an instance of that
+   resource can be created. Defining all CRDs before continuing with the rest of the installation
+   prevents this problem.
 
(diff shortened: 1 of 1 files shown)


docs/eventing/getting-started.md
```diff
@@ -33,22 +40,27 @@ kubectl get pods --namespace knative-eventing
...
 
-3. Install the Eventing sources by running the following command:
+1. Finish installing Eventing resources by running the following command:
 
     ```sh
     kubectl apply --filename https://github.com/knative/eventing/releases/download/{{< version >}}/release.yaml
     ```
 
-4. Confirm that Knative Eventing is correctly installed by running the following command:
+   Installing the CRDs first can prevent race conditions and failures during the install.
```
carieshmarie: Would it make sense to describe briefly what race conditions are, or link to somewhere?

samodell: Done! I couldn't find anything to link to, but added some more info.

carieshmarie: Thanks!

PR #2077, opened 2019-12-23
Set global.mtls.auto to false for istio-lean and clsuter-local-gateway

nak3:
The docs uses default template (`install/kubernetes/helm/istio/values.yaml`) which enables mtls by
default on Istio-1.4+. It causes some problem like https://github.com/knative/serving/issues/6254.

So, this patch sets `global.mtls.auto` to `false` for istio-lean and
clsuter-local-gateway installation.

Fixes https://github.com/knative/serving/issues/6254

cc @ahmetb @tcnghia

### DIFF
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -124,6 +124,7 @@ helm template --namespace=istio-system \
   --set gateways.istio-ingressgateway.autoscaleMax=2 \
   `# Set pilot trace sampling to 100%` \
   --set pilot.traceSampling=100 \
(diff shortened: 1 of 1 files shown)


nak3:
Although we can use minimal template instead default by setting
`--values install/kubernetes/helm/istio/values-istio-minimal.yaml`,
I think this `--set global.mtls.auto=false` is better to avoid unknown 
difference between the templates.

samodell:
/approval

Waiting for an LGTM from @tcnghia  or someone else with networking expertise to confirm this change.

PR #2079, opened 2019-12-30
call version with using '--'

renatocaval:
It seems that the `glooctl` changed and doesn't require prefixing the command with `--`.


renatocaval:
@googlebot I signed it!

PR #2082, opened 2020-01-04
Update and simplify a few install docs

evankanderson:
In other news, I was able to install Istio + Knative using Docker-for-Mac within a fairly small footprint. I think the actual size is closer to 2GB of RAM, but I left it at 6GB to have a margin for error.

## Proposed Changes

- Reduce resource requirements for Docker-for-Mac
- Don't suggest that Docker-for-Mac use the cloud provider's Istio.
- Don't use a NodePort with Docker-for-Mac, because it's not needed.
- Istio was referred to as "Gloo" at one point.


samodell:
Thanks @evankanderson ! 

/approve
/lgtm

PR #2122, opened 2020-01-22
adding in default config map required values

mjschmidt:
Fixes #issue-number

## Proposed Changes

- Fixing the config map for default channel to include partitions and replication factor. Failing to include these values in the config map for default channels will cause an error for default channels about partitions not being allowed to be set to 0.


matzew:
/close 
in fav of https://github.com/knative/docs/pull/2135

PR #2129, opened 2020-01-23
[WIP] expander test, do not merge

samodell:
Idea from @mattmoor , who noticed expanders being used in these docs:

https://github.com/containerbuilding/cbi#push-to-a-registry

Seems to work fine on GitHub; and now we'll see what it looks like on Knative.dev.


samodell:
It looks not great: https://5e2a1d97fbd32400085b4f25--knative.netlify.com/docs/install/

We'll have to table using this in the docs until someone has time to investigate what kind of dev work would be needed to make this render okay on the website.

PR #2135, opened 2020-01-27
fix default template so it works, use event display that works

### LINKED ISSUE #2134: Kafka broker example does not work. missing numPartitions replcationFactor
## Expected Behavior
Following the example in the Kafka to work correctly here:
https://knative.dev/docs/eventing/samples/kafka/channel/

## Actual Behavior
Broker fails to come up correctly with errors about missing numPartitions.

1. Follow the steps up to the 'Configuring Knative Broker...'
1. Broker does not come up and there are errors that it's missing numPartitions
1. That's because the default channel does not specify those.

## Additional Info
**Additional context**

**Install information:**
 - Platform (GKE, IKS, AKS, etc.): GKE
 - Knative Version: Head (after 0.12 cut)

vaikas:
Fixes #2134 

## Proposed Changes

- Set numPartitions and replicationFactor in the default Channel template spec. 
- Use a newer version of Event Display. The one in the instructions fails to decode the incoming message.
-

### DIFF
--- a/docs/eventing/samples/kafka/channel/000-ksvc.yaml
+++ b/docs/eventing/samples/kafka/channel/000-ksvc.yaml
@@ -6,4 +6,4 @@ spec:
   template:
     spec:
       containers:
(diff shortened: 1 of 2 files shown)


matzew:
/lgtm
/approve

replaces https://github.com/knative/docs/pull/2122

PR #2144, opened 2020-01-30
Use PHP built-in web server for "hello world PHP" serving example

SchulteMarkus:
PHP ships with a built-in web server. For a simplified serving example, the built-in web server is used instead of an Apache web server. This way, the technology stack of the example is reduced and the example is more easy to understand.

### DIFF
--- a/docs/serving/samples/hello-world/helloworld-php/Dockerfile
+++ b/docs/serving/samples/hello-world/helloworld-php/Dockerfile
@@ -1,16 +1,10 @@
-# Use the official PHP 7.3 image.
+# Use the official PHP image
 # https://hub.docker.com/_/php
-FROM php:7.3-apache
+FROM php:cli-alpine
 
-# Copy local code to the container image.
-COPY index.php /var/www/html/
+COPY index.php /
+EXPOSE 80
 
-# Use the PORT environment variable in Apache configuration files.
-RUN sed -i 's/80/${PORT}/g' /etc/apache2/sites-available/000-default.conf /etc/apache2/ports.conf
-
-# Configure PHP for development.
-# Switch to the production php.ini for production operations.
-# RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
-# https://hub.docker.com/_/php#configuration
-RUN mv "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"
-
+# Use PHP built-in web server for serving index.php
+# https://www.php.net/manual/en/features.commandline.webserver.php
+ENTRYPOINT ["php", "-S", "0.0.0.0:80", "/index.php"]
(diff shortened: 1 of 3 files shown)


SchulteMarkus:
@googlebot I signed it!

SchulteMarkus:
@googlebot I signed it!

docs/serving/samples/hello-world/helloworld-php/Dockerfile
```diff
@@ -1,16 +1,10 @@
...
-# Switch to the production php.ini for production operations.
-# RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
-# https://hub.docker.com/_/php#configuration
-RUN mv "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"
+COPY index.php /
+EXPOSE 80
 
+# Use PHP built-in web server for serving index.php
+# https://www.php.net/manual/en/features.commandline.webserver.php
+ENTRYPOINT ["php", "-S", "0.0.0.0:80", "/index.php"]
```
evankanderson: This should use `$PORT` in the argument to `-S`, rather than hard-coding to 80, which won't work in certain environments.

SchulteMarkus: Hmm, I am confused. This is the port for the running docker image / the docker container. Port 80 should always be ok from the images point of view?! 
I have searched for another "hello world" knative serving example applying this $PORT pattern, but did not find one. Can you me an URL?

SchulteMarkus:
Hi Evan,

no worries!
I have given feedback on your comment for the changed Dockerfile.
But, maybe, as this PR is an improvement already, it is worth being merged now? In my opinion, the $PORT requirement is a task on it's own (this PR is neither introducing nor downgrading the port 80 behavior).

grayside:
Last I heard PHP's built-in server was single-threaded and not recommended for production use. Unlike built-in serving options in other languages, this is originally aimed at local development with the expectation production continues to use apache or nginx. Has the recommended practice with the built-in web server been revised?

SchulteMarkus:
> Last I heard PHP's built-in server was single-threaded and not recommended for production use. 

From https://www.php.net/manual/en/features.commandline.webserver.php
> **Warning**
This web server was designed to aid application development. It may also be useful for testing purposes or for application demonstrations that are run in controlled environments. It is not intended to be a full-featured web server. It should not be used on a public network.

grayside:
It's a pretty fine balance on this point, because while the examples are not necessarily for production, people copy and use them in production.

Even the [official PHP docker images](https://hub.docker.com/_/php) are not universally regarded as production-ready, I think tracking that as a baseline is preferable.

bshaffer:
I tend to agree with @grayside, and think it's better to keep Apache in the `Dockerfile`, rather than use the PHP built-in web server. The tradeoff for simplicity isn't worth the lack of a legitimate web server.

SchulteMarkus:
"Having a production ready example" is indeed a point with good cause. Closing here.

There are other minor changes in this PR, too ("A simple web app written in Go that", README.md), but not worth updating/misusing this PR.

PR #2156, opened 2020-01-31
Adding better Kafka install doc

matzew:
As per @vaikas request on the Channels WG, here is a simple script for installing Apache Kafka.

This is much easier to follow, instead of sending folks to different websites and have them execute commands from there

/assing @vaikas

### DIFF
--- a/docs/eventing/samples/kafka/README.md
+++ b/docs/eventing/samples/kafka/README.md
@@ -11,11 +11,56 @@ All examples require:
 
 ### Setting up Apache Kafka
 
-If you want to run the Apache Kafka cluster on Kubernetes, the simplest option is to install it by using [Strimzi](https://strimzi.io),
-using the following script:
+If you want to run the Apache Kafka cluster on Kubernetes, the simplest option is to install it by using [Strimzi](https://strimzi.io).
 
+1. Create a namespace for your Apache Kafka installation, like `kafka`:
    ```shell
-   $ ./kafka_setup.sh
+   kubectl create namespace kafka
+   ```
+1. Install the Strimzi operator, like:
+   ```shell
+   curl -L "https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.16.2/strimzi-cluster-operator-0.16.2.yaml" \
+     | sed 's/namespace: .*/namespace: kafka/' \
+     | kubectl -n kafka apply -f -
+   ```
+1. Describe the size of your Apache Kafka installation, like:
+   ```yaml
+   apiVersion: kafka.strimzi.io/v1beta1
+   kind: Kafka
+   metadata:
+     name: my-cluster
+   spec:
+     kafka:
+       version: 2.4.0
+       replicas: 1
+       listeners:
+         plain: {}
+         tls: {}
+       config:
+         offsets.topic.replication.factor: 1
+         transaction.state.log.replication.factor: 1
+         transaction.state.log.min.isr: 1
+         log.message.format.version: "2.4"
+       storage:
+         type: jbod
+         volumes:
+         - id: 0
+           type: persistent-claim
+           size: 100Gi
+           deleteClaim: false
+     zookeeper:
+       replicas: 1
+       storage:
+         type: persistent-claim
+         size: 100Gi
+         deleteClaim: false
+     entityOperator:
+       topicOperator: {}
+       userOperator: {}
+   ```
+1. Deploy the Apache Kafka cluster
+   ```
+   $ kubectl apply -n kafka -f kafka.yaml
    ```
 
 This will install a small, non-production, cluster of Apache Kafka, using one node for Apache Zookeeper and one node for Apache Kafka.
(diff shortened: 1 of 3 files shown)


matzew:
@vaikas @lionelvillard  

I've updated this now.

I added all the yaml (or pure) steps, and I have kept my script, which does all of that, just behind one execution ...

I hope this all makes sense, and we can update the doc, so that - as was reported - users do not struggle with the installation.

docs/eventing/samples/kafka/README.md
```diff
@@ -9,9 +9,81 @@ All examples require:
...
+         type: persistent-claim
+         size: 100Gi
+         deleteClaim: false
+     entityOperator:
+       topicOperator: {}
+       userOperator: {}
+   ```
+1. Deploy the Apache Kafka cluster
+   ```
+   $ kubectl apply -f kafka.yaml
```
aliok:
```suggestion
   $ kubectl apply -n kafka -f kafka.yaml
```

Need the namespace here

docs/eventing/samples/kafka/README.md
```diff
@@ -9,9 +9,81 @@ All examples require:
...
+   NAME                                          READY   STATUS    RESTARTS   AGE
+   my-cluster-entity-operator-65995cf856-ld2zp   3/3     Running   0          102s
+   my-cluster-kafka-0                            2/2     Running   0          2m8s
+   my-cluster-zookeeper-0                        2/2     Running   0          2m39s
+   strimzi-cluster-operator-77555d4b69-sbrt4     1/1     Running   0          3m14s
+   ```
+
+### Installation script
+
+If you want to install the latest version of Strimzi, in just one step, we have a script for your convenience, which does exactly the same steps that are listed above:
```
aliok:
```suggestion
If you want to install the latest version of Strimzi, in just one step, we have a [script](./kafka_setup.sh) for your convenience, which does exactly the same steps that are listed above:
```

Added a link to the script

docs/eventing/samples/kafka/kafka_setup.sh
```diff
@@ -0,0 +1,29 @@
...
+header_text "Using Strimzi Version:                  ${strimzi_version}"
+header_text "Strimzi install"
+kubectl create namespace kafka
+curl -L "https://github.com/strimzi/strimzi-kafka-operator/releases/download/${strimzi_version}/strimzi-cluster-operator-${strimzi_version}.yaml" \
+  | sed 's/namespace: .*/namespace: kafka/' \
+  | kubectl -n kafka apply -f -
+header_text "Applying Strimzi Cluster file"
+kubectl -n kafka apply -f "https://raw.githubusercontent.com/strimzi/strimzi-kafka-operator/${strimzi_version}/examples/kafka/kafka-persistent-single.yaml"
+header_text "Waiting for Strimzi to become ready"
+sleep 5; while echo && kubectl get pods -n kafka | grep -v -E "(Running|Completed|STATUS)"; do sleep 5; done
```
aliok: Not a super big deal, but maybe we can remove these 2 lines.
The loop is prematurely ended because the operator becomes in running state but it hasn't created any other pods yet.

Perhaps we increase the initial sleep to something like 30 seconds.

lionelvillard:
I tried it in `kind`. Does not work, I'm guessing because of the VolumeClaim.

Is this script for testing purpose? If yes then why not using an ephemeral storage?

matzew:
>  If yes then why not using an ephemeral storage?

we can do that, but that does not work great w/ ZK and just 1 node.

This is basically the same installation guide, that we did point to. But apparently that was to complicated, to go to strimzi.io etc. 

So, I tried to add some yamls, and some scripts.  But looks like: it's all wrong 


Let me know if I should just close this PR, and make better use of my time

matzew:
@lionelvillard if we want `ephemeral`

try this:
```
apiVersion: kafka.strimzi.io/v1beta1
kind: Kafka
metadata:
  name: my-cluster
spec:
  kafka:
    version: 2.4.0
    replicas: 1
    listeners:
      plain: {}
      tls: {}
    config:
      offsets.topic.replication.factor: 1
      transaction.state.log.replication.factor: 1
      transaction.state.log.min.isr: 1
      log.message.format.version: "2.4"
    storage:
      type: ephemeral
  zookeeper:
    replicas: 3
    storage:
      type: ephemeral
  entityOperator:
    topicOperator: {}
    userOperator: {}
```


But, I think than folks will say: why three nodes - well, b/c ZK and one node is not great experience ... 

I am out of options here - I just wanted to help and improve, but this seems like wasting time already

matzew:
If we really want `ephemeral ` and just 1 ZK note, than, pleaes let's think aboiut https://github.com/strimzi/strimzi-kafka-operator/pull/2487#issuecomment-580642935  for a second, IMO that's worse - but, I can be wrong here as well

matzew:
> I tried it in kind. Does not work, I'm guessing because of the VolumeClaim.

> Is this script for testing purpose? If yes then why not using an ephemeral storage?

Ok, I updated for a 3 node ZK and having the entire cluster being `ephemeral`

PR #2171, opened 2020-02-04
Remove the reference to VPA

vagababov:
We're not really using VPA, so fix the docs, since someone might set it to true and expect something to happen

/assign @dgerd @samodell @markusthoemmes


docs/serving/configuring-autoscaling.md
```diff
@@ -33,7 +33,6 @@ data:
  container-concurrency-target-default: 100
  container-concurrency-target-percentage: 1.0
  enable-scale-to-zero: true
```
dgerd: This seems to be missing a few other options that are in https://github.com/knative/serving/blob/master/config/core/configmaps/autoscaler.yaml

Like scale-to-zero-grace-period, enable-graceful-shutdown, and requests-per-second-target-default. Can you add those as well with this change?

vagababov: Hm, I didn't want this to be an overhaul of the docs :)
Also `enable-graceful-shutdown` is not working yet.

vagababov: scale-to-zero-grace period is described below.

vagababov:
I did more changes to make it more up to date, but Markus and I need a thorough update for this.

dgerd:
works for me

/lgtm

samodell:
/approved

Thanks for these updates!

PR #2173, opened 2020-02-04
Update Java Spring boot sample to programmatically get PORT

averikitsch:
## Proposed Changes

- Instead of a using a flag to set port, set port in code


averikitsch:
@samodell can you help me get this merged?

PR #2177, opened 2020-02-05
The yaml was not matching the text in the README

matzew:
## Proposed Changes

- the `README` talk about the Broker, but the yaml used `ksvc`, which is wrong


Abd4llA:
good catch!

PR #2185, opened 2020-02-07
Using LATEST tagged release

matzew:
The VERSION string is (unfortunately) today limited to MAJOR.MINOR release (semver terms).

PATCH releases are ignore.

However patch releases do sometimes contain fixes, you wanna pick up.

Proposal:  Where possible (e.g. on all direct KN components) we use the `latest` which gives you, for instance:

* eventing-core.yam in `0.12.0`
* kafka-source.yaml in `0.12.2`

/assign @mattmoor

### DIFF
--- a/docs/install/knative-with-any-k8s.md
+++ b/docs/install/knative-with-any-k8s.md
@@ -45,13 +45,13 @@ The following commands install the Knative Serving component.
 1. Install the [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) (aka CRDs):
 
    ```bash
-   kubectl apply --filename https://github.com/knative/serving/releases/download/{{< version >}}/serving-crds.yaml
+   kubectl apply --filename https://storage.googleapis.com/knative-releases/serving/latest/serving-crds.yaml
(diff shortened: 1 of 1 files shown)


matzew:
for instance in Kafka, we have actually a "backwards compatible" fix in the 0.12.2  ... so that would be not picked up - because the docs are just 0.12.0 - not recognizing the indiviual "latest" tags

mattmoor:
@matzew this doesn't work for older release dropdowns 😬

mattmoor:
Maybe this will make it easier, but not sure: https://github.com/knative/docs/pull/2183

matzew:
hrm - I guess there is work todo on this :-) 

@mattmoor should I close?

I am really OK if we close this, now, since it seems to NOT work w/ older version :-( 

I had hopes :-)

samodell:
/hold

As discussed on Slack, this breaks versioning in the site. Point taken that we should find a way to direct users to the latest patch release, though.

mattmoor:
@matzew yeah, create an issue and we can brainstorm?  I think the bad assumption was that we *always* jointly release, which turns out to be a bad assumption 😅

PR #2187, opened 2020-02-07
move Knative offering under new Getting Started section

lionelvillard:
## Proposed Changes

- move Knative offering under new Getting Started section
-
-

@samodell @mattmoor


samodell:
/hold for discussion

@lionelvillard  Can you expand on why you wanted to make this change? Matt had actually originally placed the Offerings page within install, but I asked him to move it up to the top-level because I think it's fairly important and should be visible above the /install/ directory.

lionelvillard:
My reasoning was if we go with a `Getting Started` section then either you get started by installing Knative or by looking at Knative Offerings.

lionelvillard:
Closing as Knative Offerings is important.

PR #2182, opened 2020-02-07
Reflect the separation of the Channel-based Broker.

mattmoor:
See also: https://github.com/knative/eventing/pull/2519

/assign @vaikas

PR #2190, opened 2020-02-08
Revert "Revert "Reflect the separation of the Channel-based Broker. (#2182)""

mattmoor:
Reverts knative/docs#2184

/hold

Folks seem to be ok with the change, but leaving on hold to make sure.


mattmoor:
I believe this is because my primary email address wasn't one on the CLA, going to try again.

PR #2195, opened 2020-02-10
Cherry-pick 2.10.20

samodell:
Small cherry-pick batch


samodell:
Adding cla:yes since these commits were signed when committed to the master branch.

PR #2203, opened 2020-02-11
chore(resources): Remove link to knctl

rhuss:
knctl doesn't seem to be developed anymore (latest serving dependency: 0.2.1, latest commit: May 2019).
I think its fair to remove it from this page to not mislead potential users
(or at least don't call it 'under active development')

@cppforlife I hope this is ok for you. You did an awesome job to kick off the idea for a Knative client and you paved the way to `kn`. However, I think it would be a bit confusing for people if we give them the choice between `kn` and `knctl`. Please let me know if I'm wrong and you are planning to continue to work on `knctl`.

Thanks !
In general I question the value of the overall page is can be removed completely IMO.

### DIFF
--- a/docs/concepts/resources.md
+++ b/docs/concepts/resources.md
@@ -7,20 +7,6 @@ type: "docs"
 This page contains information about various tools and technologies that are
 useful to anyone developing on Knative.
 
-## Community Resources
-
-This section contains tools and technologies developed by members of the Knative
-community specifically for use with Knative.
-
-### [`knctl`](https://github.com/cppforlife/knctl)
-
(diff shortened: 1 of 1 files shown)


docs/concepts/resources.md
```diff
@@ -7,20 +7,6 @@ type: "docs"
 This page contains information about various tools and technologies that are
 useful to anyone developing on Knative.
 
-## Community Resources
```
abrennan89: I would maybe leave the **Community Resources** heading in because `knctl` isn't the only thing listed here (we still have `ko`).

PR #2210, opened 2020-02-12
A couple little cleanups.

mattmoor:
I should have caught these in code review, but I'll blame github for showing such tight code blocks ;-)
1. Couple missed artifact shortcodes,
2. Alphabetize Channels,
3. Fix some wording around Broker installation that is inaccurate.


n3wscott:
/lgtm

that is much better than the url. nice.

PR #2218, opened 2020-02-14
Fixed webhook sample path

mpetason:
## Proposed Changes

- Fixed the webhook sample path to match the new /docs/ path.


samodell:
Thanks, @mpetason !

PR #2236, opened 2020-02-23
Switch support status for triggermesh external sources

sebgoa:
## Proposed Changes 

- Switch the status of TriggerMesh external sources to "supported" since we now offer support.


sebgoa:
The error does not seem related to this PR.

PR #2244, opened 2020-02-25
Fix doc error about eventing requiring Istio

### LINKED ISSUE #2245: Release 0.12 doc says eventing requires istio
## Expected Behavior
Serving can use Istio as Ingress controller, but eventing does not rely on Ingress.

## Actual Behavior
`Installing Knative with Kourier` section says "Note that Istio is required for the Knative Eventing component."

1. Go to https://knative.dev/docs/install/
1. Scroll down to the `Installing Knative with Kourier` section 
1.

## Additional Info
**Additional context**

**Install information:**
 - Platform (GKE, IKS, AKS, etc.):
 - Knative Version:

houshengbo:
Fixes #2245
## Proposed Changes 

- Change the description about the relation of serving and Istio.

### DIFF
--- a/docs/install/README.md
+++ b/docs/install/README.md
@@ -64,7 +64,7 @@ quickly.
 
 ## Installing Knative with Kourier
 
-[Installing Knative with Kourier](./knative-with-kourier.md): In this guide, Knative will be installed with Kourier to replace the Istio Service Mesh, providing a more lightweight and simple Ingress controller. Note that Istio is required for the Knative Eventing component.
+[Installing Knative with Kourier](./knative-with-kourier.md): In this guide, Knative will be installed with Kourier to replace the Istio Service Mesh, providing a more lightweight and simple Ingress controller.
 
 ### Install guides
 


docs/install/README.md
```diff
@@ -64,7 +64,7 @@ quickly.
 
 ## Installing Knative with Kourier
 
-[Installing Knative with Kourier](./knative-with-kourier.md): In this guide, Knative will be installed with Kourier to replace the Istio Service Mesh, providing a more lightweight and simple Ingress controller. Note that Istio is required for the Knative Eventing component.
+[Installing Knative with Kourier](./knative-with-kourier.md): In this guide, Knative will be installed with Kourier to replace the Istio Service Mesh, providing a more lightweight and simple Ingress controller. Note that Istio can serve as an Ingress controller for the Knative Serving component as well.
```
mpetason: If it isn't required and originally mentioned Eventing, maybe we can just delete the note instead of adding additional info about serving. Above this section is "Installing Knative with Istio" which should cover the content of the note.

evankanderson: +1 to removing text if it's no longer needed. If Kourier can fulfil the cluster-local routes, I don't think we need any note about Istio here.

PR #2249, opened 2020-02-25
link to each channel documentation instead of the code

lionelvillard:
## Proposed Changes 

- link to each channel documentation instead of the code. 
-
-


lionelvillard:
markdownlint should be disabled

n3wscott:
Is it better to point to the README.md? it is more work to get to the src in that case.

lionelvillard:
documentation is not code :-)

PR #2250, opened 2020-02-25
use lowercase title to be consistent with the rest of the doc

lionelvillard:
## Proposed Changes 

- use lowercase title to be consistent with the rest of the doc
-
-


lionelvillard:
markdownlint should be disabled.

Issue #2248, opened 2020-02-25
Document and update examples for v1beta1 for Broker,Trigger, and EventType


vaikas:
## What is changing? (Please include as many details as possible.)
Adding v1beta1 apis for Broker,Trigger, EventType

## How will this impact our users?
Users will have a cleaner API.

## In what release will this take happen (to the best of your knowledge)?
V0.13

## Context
Link to associated PRs or issues from other repos here.

1. Trigger: https://github.com/knative/eventing/pull/2414
1. EventTypes: https://github.com/knative/eventing/pull/2466
1. Broker: https://github.com/knative/eventing/pull/2449

abrennan89:
@vaikas is this issue closed by #2259 ? If not, what else is required here?

PR #2126, opened 2020-01-23
Make documentation for PingSource.

n3wscott:
Dependent on https://github.com/knative/eventing/pull/2427

Document the renamed instructions for PingSource. Formally called CronJobSource.

PR #2259, opened 2020-02-27
Update docs to v1beta1 + add more details about channel based broker

vaikas:
Addresses #2248 

## Proposed Changes 

- Update documents for Broker / Trigger for v1beta1.
- Add a more detailed section for Channel Based Broker and split out the installation from usage.
-

### DIFF
--- a/docs/eventing/broker-trigger.md
+++ b/docs/eventing/broker-trigger.md
@@ -13,13 +13,15 @@ A Broker represents an 'event mesh'. Events are sent to the Broker's ingress and
 are then sent to any subscribers that are interested in that event. Once inside
 a Broker, all metadata other than the CloudEvent is stripped away (e.g. unless
 set as a CloudEvent attribute, there is no concept of how this event entered the
-Broker). There can be different classes of `Broker`s providing different kinds
-of semantics around durability of events, performance, etc. The `Broker` that is
+Broker).
+
+There can be different classes of Brokers providing different kinds
+of semantics around durability of events, performance, etc. The Broker that is
 part of the Knative Eventing repo is used for these examples, it uses Knative
-[Channels](./channels/) for delivering events. Simple example showing a `Broker`
+[Channels](./channels/) for delivering events. You can read more details about
+[Channel Based Broker](./channel-based-broker.md). Simple example showing a `Broker`
 where the configuration is specified in a `ConfigMap` config-br-default-channel,
-which uses `InMemoryChannel`. You can read more details about
-[Channel Based Broker](./channel-based-broker.md)
+which uses `InMemoryChannel`:
 
 Example:
 
@@ -65,8 +67,8 @@ spec:
 
 A Trigger represents a desire to subscribe to events from a specific Broker.
 
-Simple example which will receive all the events from a given (`default`) broker and
-delivers them to Knative Serving service `Service`:
+Simple example which will receive all the events from the given (`default`) broker and
+deliver them to Knative Serving service `my-service`:
 
 ```yaml
 apiVersion: eventing.knative.dev/v1beta1
@@ -116,12 +118,12 @@ have the extension `myextension` with the value `my-extension-value`.
 ### Broker setup
 
 We assume that you have installed a Broker in namespace `default`. If you haven't done that
-yet, [install it from here](./channel-based-broker.md)
+yet, [install it from here](./channel-based-broker.md).
 
 ### Subscriber
 
 Create a function to receive events. This document uses a Knative Service, but
-it could be anything that is [Addressable](https://github.com/knative/eventing/blob/master/docs/spec/interfaces.md).
+it could be anything that is [Callable](https://github.com/knative/eventing/blob/master/docs/spec/interfaces.md).
 
 ```yaml
 apiVersion: serving.knative.dev/v1
--- a/docs/eventing/broker-trigger.md
+++ b/docs/eventing/broker-trigger.md
@@ -18,8 +18,8 @@ of semantics around durability of events, performance, etc. The `Broker` that is
 part of the Knative Eventing repo is used for these examples, it uses Knative
 [Channels](./channels/) for delivering events. Simple example showing a `Broker`
 where the configuration is specified in a `ConfigMap` config-br-default-channel,
-which uses `InMemoryChannel` and delivers failed events to a dlq-service.
-You can read more details about [Channel Based Broker](./channel-based-broker.md)
+which uses `InMemoryChannel`. You can read more details about
+[Channel Based Broker](./channel-based-broker.md)
 
 Example:
 
@@ -33,15 +33,32 @@ spec:
   config:
     apiVersion: v1
     kind: ConfigMap
-	name: config-br-default-channel
+    name: config-br-default-channel
+    namespace: knative-eventing
+```
+
+More complex example, showing the same `Broker` as above
+but with failed events being delivered to Knative Service called `dlq-service`
+
+```yaml
+apiVersion: eventing.knative.dev/v1beta1
+kind: Broker
+metadata:
+  name: default
+spec:
+  # Configuration specific to this broker.
+  config:
+    apiVersion: v1
+    kind: ConfigMap
+    name: config-br-default-channel
     namespace: knative-eventing
   # Where to deliver Events that failed to be processed.
   delivery:
     deadLetterSink:
-	  ref:
+      ref:
         apiVersion: serving.knative.dev/v1
-	    kind: Service
-		name: dlq-service
+        kind: Service
+        name: dlq-service
 ```
 
 ## Trigger
--- a/docs/eventing/broker-trigger.md
+++ b/docs/eventing/broker-trigger.md
@@ -13,29 +13,46 @@ A Broker represents an 'event mesh'. Events are sent to the Broker's ingress and
 are then sent to any subscribers that are interested in that event. Once inside
 a Broker, all metadata other than the CloudEvent is stripped away (e.g. unless
 set as a CloudEvent attribute, there is no concept of how this event entered the
-Broker).
+Broker). There can be different classes of `Broker`s providing different kinds
+of semantics around durability of events, performance, etc. The `Broker` that is
+part of the Knative Eventing repo is used for these examples, it uses Knative
+[Channels](./channels/) for delivering events. Simple example showing a `Broker`
+where the configuration is specified in a `ConfigMap` config-br-default-channel,
+which uses `InMemoryChannel` and delivers failed events to a dlq-service.
+You can read more details about [Channel Based Broker](./channel-based-broker.md)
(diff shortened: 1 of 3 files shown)


vaikas:
I think the missing link should be fixed by: #2126

docs/eventing/broker-trigger.md
```diff
@@ -13,29 +13,63 @@ A Broker represents an 'event mesh'. Events are sent to the Broker's ingress and
...
 a Broker, all metadata other than the CloudEvent is stripped away (e.g. unless
 set as a CloudEvent attribute, there is no concept of how this event entered the
-Broker).
+Broker). There can be different classes of `Broker`s providing different kinds
+of semantics around durability of events, performance, etc. The `Broker` that is
+part of the Knative Eventing repo is used for these examples, it uses Knative
+[Channels](./channels/) for delivering events. Simple example showing a `Broker`
+where the configuration is specified in a `ConfigMap` config-br-default-channel,
+which uses `InMemoryChannel`. You can read more details about
+[Channel Based Broker](./channel-based-broker.md)
```
grantr: typo nit: missing period.

docs/eventing/broker-trigger.md
```diff
@@ -13,29 +13,63 @@ A Broker represents an 'event mesh'. Events are sent to the Broker's ingress and
 are then sent to any subscribers that are interested in that event. Once inside
 a Broker, all metadata other than the CloudEvent is stripped away (e.g. unless
 set as a CloudEvent attribute, there is no concept of how this event entered the
-Broker).
+Broker). There can be different classes of `Broker`s providing different kinds
```
grantr: This should probably be a new paragraph. I'd drop the code tags around `Broker` since the previous uses of Broker don't have them.

docs/eventing/broker-trigger.md
```diff
@@ -13,29 +13,63 @@ A Broker represents an 'event mesh'. Events are sent to the Broker's ingress and
...
+of semantics around durability of events, performance, etc. The `Broker` that is
+part of the Knative Eventing repo is used for these examples, it uses Knative
+[Channels](./channels/) for delivering events. Simple example showing a `Broker`
```
grantr: "Simple example showing..." isn't a complete sentence. You could swap the following sentence "You can read more..." before this one, then add a colon at the end, like

> it uses Knative [Channels](./channels/) for delivering events. You can read more details about
[Channel Based Broker](./channel-based-broker.md). Simple example showing a `Broker` where the configuration is specified in a `ConfigMap` config-br-default-channel, which uses `InMemoryChannel`:

docs/eventing/broker-trigger.md
```diff
@@ -13,29 +13,63 @@ A Broker represents an 'event mesh'. Events are sent to the Broker's ingress and
...
+        kind: Service
+        name: dlq-service
 ```
 
 ## Trigger
 
 A Trigger represents a desire to subscribe to events from a specific Broker.
 
-Example:
+Simple example which will receive all the events from a given (`default`) broker and
```
grantr: Grammar nit: "from **the** given"

docs/eventing/broker-trigger.md
```diff
@@ -13,29 +13,63 @@ A Broker represents an 'event mesh'. Events are sent to the Broker's ingress and
...
+delivers them to Knative Serving service `Service`:
```
grantr: typo nit: "deliver". Also, the name of the service is "my-service" not "Service"

docs/eventing/broker-trigger.md
```diff
@@ -229,25 +134,26 @@ spec:
     spec:
       containers:
       -  # This corresponds to
-         # https://github.com/knative/eventing-contrib/blob/v0.2.1/cmd/message_dumper/dumper.go.
-         image: gcr.io/knative-releases/github.com/knative/eventing-sources/cmd/message_dumper@sha256:ab5391755f11a5821e7263686564b3c3cd5348522f5b31509963afb269ddcd63
+         # https://github.com/knative/eventing-contrib/tree/master/cmd/event_display
+         - image: gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/event_display@sha256:a214514d6ba674d7393ec8448dd272472b2956207acb3f83152d3071f0ab1911
```
grantr: I wonder if we can use the `latest` tag here?

docs/eventing/broker-trigger.md
```diff
@@ -283,62 +189,24 @@ spec:
...
-          env:
-            - name: POD_NAME
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.name
-            - name: POD_NAMESPACE
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.namespace
+  schedule: "*/1 * * * *"
```
grantr: I think the `/1` here isn't needed, it can just be `* * * * *`

docs/eventing/broker-trigger.md
```diff
@@ -283,62 +189,24 @@ spec:
...
+  data: '{"message": "Hello world!"}'
   sink:
-    apiVersion: eventing.knative.dev/v1alpha1
-    kind: Broker
-    name: default
+    ref:
+      # Deliver events to Broker.
+      apiVersion: eventing.knative.dev/v1alpha1
```
grantr: `v1beta1` here right?

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
+---
+title: "Channel Based Broker"
+weight: 30
+type: "docs"
+---
+
+# Channel Based Broker
+
+Knative provides a `Broker` implementation that uses [Channels](./channels/) for
+event routing. You will need to have a Channel provider installed, for example
```
grantr: "Channel CRD" or "Channel implementation" makes more sense to me than "Channel provider" because I'm not sure what provider means here, but I don't have a strong preference.

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
...
+InMemoryChannel (for development purposes), Kafka, Nats, etc. You can choose from
+list of [available channels](https://knative.dev/docs/eventing/channels/channels-crds/)
```
grantr: typo not: missing period.

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
...
+apiVersion: v1
+metadata:
+  name: config-br-defaults
+  namespace: knative-eventing
+data:
+  default-br-config: |
+    clusterDefault:
+      apiVersion: v1
+      kind: ConfigMap
+      name: imc-channel
```
grantr: This should actually be `kafka-channel` as documented

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
...
+      namespace: knative-eventing
+    namespaceDefaults:
+      test-broker-6:
+        apiVersion: v1
+        kind: ConfigMap
+        name: kafka-channel
```
grantr: This should actually be `imc-channel` as documented

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
...
+  namespace: knative-eventing
+  name: imc-channel
+data:
+  channelTemplateSpec.apiVersion: messaging.knative.dev/v1alpha1
+  channelTemplateSpec.kind: InMemoryChannel
+```
+
+```yaml
+# Define how Kafka channels are created. Note we specify
+# extra parameters that are particular to Kakfa Channels, namely
```
grantr: typo nit: Kafka Channels

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
...
+the `Broker` as described above. If you now remove the annotation, the `Broker`
+will not be removed, you have to manually delete it.
+
+For example, to delete the injected Broker from the foo namespace:
+
+```shell
+kubectl -n foo delete broker default
+```
+
+## Installing Broker Manually
```
grantr: Honestly I think this section should be removed. Why would a user ever do this if they can annotate the namespace (and if they can create ServiceAccounts, they can probably annotate the namespace)? Maybe a contributor needs to know, but then it can be part of the eventing repo.

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+Note that these commands each use three different objects, all named
+`eventing-broker-ingress` or `eventing-broker-filter`. The `ClusterRole` is
+installed with Knative Eventing
+[here](https://github.com/knative/eventing/blob/master/config/200-broker-clusterrole.yaml).
+The `ServiceAccount` was created two commands prior. The `RoleBinding` is
+created with this command.
+
+Create RBAC permissions granting access to shared configmaps for logging,
```
grantr: These instructions can be replaced with a ConfigMapPropagation (that's what the namespace controller uses now)

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
...
+  config:
+    apiVersion: v1
+    kind: ConfigMap
+    name: imc-channel
+    namespace: knative-eventing
+EOF
+```
+
+Easiest way to create and control `Broker`s in your cluster is by using an automagic
+injection by labeling a namespace.
```
grantr: This is already documented above. It should either be removed or linked to the previous heading.

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Handling failed event delivery
+
+Broker allows you to specify what to in the case if failed event delivery, say
```
grantr: typo nit: "what to **do** in the case **of**"

docs/eventing/channel-based-broker.md
```diff
@@ -0,0 +1,214 @@
...
+you have a consumer (function for example) that's failing to process the event.
+You can use `DeliverySpec` for it:
```
grantr: It might make more sense to the user to call this `delivery` since `DeliverySpec` is only mentioned in the Go type. The example doesn't have `DeliverySpec` anywhere.

docs/eventing/channel-based-broker.md
vaikas: Yeah, I initially thought about removing it entirely, but for the cases where you might have multiple types of brokers in a given namespace, you can't just label the namespace. You could for example have a Broker foo and bar, and you would need to create one of them manually.

docs/eventing/channel-based-broker.md
vaikas: Is this something that you or @grac3gao could do after we get the initial changes in?

vaikas:
Thanks @grantr PTAL.

PR #2264, opened 2020-03-03
Flow v1beta1

lionelvillard:

## Proposed Changes 

- Update flow doc to use v1beta1 API
-
-


docs/eventing/samples/sequence/sequence-reply-to-event-display/ping-source.yaml
```diff
@@ -1,12 +1,12 @@
-apiVersion: sources.eventing.knative.dev/v1alpha1
-kind: CronJobSource
+apiVersion: sources.knative.dev/v1alpha1
```
vaikas: nit, could also change these to v1alpha2?

PR #2266, opened 2020-03-03
update parallel doc to v1beta1

lionelvillard:

## Proposed Changes 

- update parallel doc to v1beta1
-
-


lionelvillard:
depends on https://github.com/knative/docs/pull/2126

xlight05:
Change in docker image of event_display is crucial because the current doc points to the wrong docker image and its not working. We need this live asap

PR #2284, opened 2020-03-09
Remove DNS pain point for new users

diodesign:
Stepping through the installation guide, I ran into a mild pain point using Minikube and Gloo and/or Istio at the DNS stage. The Magic DNS solution couldn't be used, and the Real DNS method seems perhaps a little heavy-handed - creating DNS entries for what may be dynamically assigned IP addresses - for just getting a Hello, Word application running to see if a cluster is configured correctly.

Judging from issues raised and questions asked online, others were also caught up on this stage, so I suggest the following addition to reduce friction post-installation.

## Proposed Changes 
- The addition of another tab in the DNS section that explains how to use curl to specify the application when connecting to the external IP or CNAME. The rationale being most of the sample applications suggest using curl to communication with the software.
- The tab indicates the Real DNS solution is preferred for production deployments.

NB: The tab title "Temporary DNS" could be improved. The point I wanted to get across was, here's a way to interact immediately with the sample application without having to altering your DNS configuration.

### DIFF
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -261,7 +261,7 @@ We ship a simple Kubernetes Job called "default domain" that will (see caveats) 
 kubectl apply --filename {{< artifact repo="serving" file="serving-default-domain.yaml" >}}
 ```
 
-**Caveat**: This will only work if the cluster LoadBalancer service exposes an IPv4 address, so it will not work with IPv6 clusters, AWS, or local setups like Minikube.  For these, see "Real DNS" or "Temporary DNS".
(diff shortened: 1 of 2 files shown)


docs/install/any-kubernetes-cluster.md
```diff
@@ -293,6 +293,41 @@ kubectl patch configmap/config-domain \
...
+
+    ```bash
+    kubectl get ksvc
+    ```
+
+   The output should be similar to:
+
+   ```bash
+   NAME            URL                                        LATESTCREATED         LATESTREADY           READY   REASON
+   helloworld-go   http://helloworld-go.default.example.com   helloworld-go-vqjlf   helloworld-go-vqjlf   True
```
mattmoor: Remove trailing whitespace:
```suggestion
   helloworld-go   http://helloworld-go.default.example.com   helloworld-go-vqjlf   helloworld-go-vqjlf   True
```

PR #2324, opened 2020-03-19
Removed broken link in autoscale go sample

mpetason:
Fixes an invalid link and removes a Further Reading section.

## Proposed Changes

- Removed further reading from autoscale-go sample as the link doesn't work - https://github.com/knative/serving/blob/master/docs/scaling/DEVELOPMENT.md <-- There doesn't look to be a scaling development.md file for serving in the repository. 

I'm wondering if this should instead point to:

https://github.com/knative/serving/blob/master/docs/scaling/SYSTEM.md

If we want to point to SYSTEM.md I can make the update in the PR, otherwise we probably want to remove the further reading section.

### DIFF
--- a/docs/serving/samples/autoscale-go/README.md
+++ b/docs/serving/samples/autoscale-go/README.md
@@ -270,3 +270,7 @@ kubectl port-forward --namespace knative-monitoring $(kubectl get pods --namespa
 ```
 kubectl delete --filename docs/serving/samples/autoscale-go/service.yaml
 ```
+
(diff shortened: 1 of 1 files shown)


mpetason:
@markusthoemmes Can you take a look at this and let me know if we should remove the further reading section or if I should change the link to point to system.md instead of development.md.

vagababov:
Thanks for the clean up.
Yes, please point to the new file.

PR #2325, opened 2020-03-19
This adds docs for the new mtnamespace work.

mattmoor:
This work is happening here: https://github.com/knative/eventing/pull/2778

### DIFF
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -520,6 +520,13 @@ ConfigMap to specify which configurations are used for which namespaces:
            namespace: knative-eventing
    ```
 
+The following command enables the default Broker on a namespace (here `default`):
+
+   ```bash
+   kubectl label namespace default knative-eventing-injection=enabled
+   ```
+
+
 {{< /tab >}}
 
 {{% tab name="MT-Channel-based" %}}
@@ -564,6 +571,14 @@ ConfigMap to specify which configurations are used for which namespaces:
            namespace: knative-eventing
    ```
 
+This Broker implementation has a small enough incremental overhead that it
+creates a Broker for every namespace by default.  To opt particular namespaces
+out of this behavior you can label them with:
+
+   ```bash
+   kubectl label namespace default knative-eventing-injection=disabled
+   ```
+
 {{< /tab >}}
 
 {{< /tabs >}}
@@ -582,18 +597,6 @@ At this point, you have a basic installation of Knative Eventing!
 
    <!-- This indentation is important for things to render properly. -->
    {{< tabs name="eventing_extensions" >}}
-{{% tab name="Enable Broker" %}}
-
-<!-- Unclear when this feature came in -->
-{{< feature-state version="v0.5" state="alpha" >}}
-
-The following command enables the default Broker on a namespace (here `default`):
-
-   ```bash
-   kubectl label namespace default knative-eventing-injection=enabled
-   ```
-
-{{< /tab >}}
 
 {{% tab name="Github Source" %}}
 


docs/install/any-kubernetes-cluster.md
```diff
@@ -564,6 +571,14 @@ ConfigMap to specify which configurations are used for which namespaces:
            namespace: knative-eventing
    ```
 
+This Broker implementation has a small enough incremental overhead that it
+creates a Broker for every namespace by default.  To opt particular namespaces
+out of this behavior you can label them with:
+
+   ```bash
+   kubectl label namespace default knative-eventing-injection=disabled
```
vaikas: Is this really going to work? if you first create the NS, ns controller will create the broker? perhaps we should say here that you have to create with the label?

mattmoor: You'd have to do it before installing things today.  Honestly, setting this label to "disabled" with EITHER namespace controller should clean up the auto-generated Broker, opened https://github.com/knative/eventing/issues/2796.

matzew:
So, do we really need this, in every namespave ? 

```
k get broker --all-namespaces
NAMESPACE          NAME      READY   REASON   URL                                                                                 AGE
default            default   True             http://broker-ingress.knative-eventing.svc.cluster.local/default/default            11s
knative-eventing   default   True             http://broker-ingress.knative-eventing.svc.cluster.local/knative-eventing/default   11s
knative-serving    default   True             http://broker-ingress.knative-eventing.svc.cluster.local/knative-serving/default    11s
kourier-system     default   True             http://broker-ingress.knative-eventing.svc.cluster.local/kourier-system/default     11s
kube-node-lease    default   True             http://broker-ingress.knative-eventing.svc.cluster.local/kube-node-lease/default    9s
kube-public        default   True             http://broker-ingress.knative-eventing.svc.cluster.local/kube-public/default        11s
kube-system        default   True             http://broker-ingress.knative-eventing.svc.cluster.local/kube-system/default        11s
```

now, what if there is an even bigger cluster, with 100+ namespaces - all not related to Knative at all? they all really need a borker ?

grantr:
> Drop the namespace magic entirely and just have folks create Brokers.

Separately from the question of whether MT brokers should be opt-in or opt-out, the original purpose of the namespace magic was to work around the per-broker requirement of additional deployments, services, RBAC, etc. It seems like this doesn't apply to the MT broker since it doesn't create any additional per-Broker objects.

mattmoor:
For the sake of having an "off" switch should things go bad (as they did for `SinkBinding` 👀  😅 ), I will add some logic to change the default behavior, but I still think that this should be our goal state, and if it's not I think we should probably have a serious convo about the motivation for using a namespace controller.

FWIW, in serving, the `nscert` controller, which creates wildcard certs per namespace is on by default, but an optional installable.  The model here strikes the balance I was mentioning above...   You can create wildcard certs manually, or you can run the NS controller to get them automagically, but opting IN per-namespace makes me wonder why we aren't just having folks create the resource per-namespace (as the mechanism to opt in).

antoineco:
Late to this conversation, but I wanted to link my comment on https://github.com/knative/eventing/pull/2946#issuecomment-611425967.

Regarding the parallel ServiceAccounts, I think there is one major difference here: **all** Pods require a service account to even be scheduled, but **few** need a Broker in every possible namespace. (see comment above)

mattmoor:
Pods and Brokers is apples/oranges.  I cannot create a Trigger without a Broker.

Could Kubernetes have forced folks to pre-create SAs?  Yes.
Do all namespaces have to contain Pods?  No.
So why would they do it?  It's a better experience that's simpler for users.

Frankly that likely costs the system more than what we're doing since it involves storing and rotating auth (used or not).

mattmoor:
I'm going to close this, as it's no longer correct.  As we shift toward MT as the default we should revisit this as it's a strictly better DX in my experience with it thus far.

PR #2329, opened 2020-03-20
Update installing-istio.md

josiemundi:
Cluster local ISTIO_MINOR_VERSION

## Proposed Changes 

The doc currently mentions you need 1.2 or 1.3 and specifies ISTIO_MINOR_VERSION=1.2, however this no longer exists. Have changed to what I believe is now correct.

### DIFF
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -250,8 +250,8 @@ Alternatively, if you want to install the cluster local gateway for **developmen
 without `helm` for an easy installation:
 
 ```shell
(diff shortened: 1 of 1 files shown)


bbrowning:
/ok-to-test

Did we drop the minimum Istio version we support to 1.3? And have we resolved all the problems using 1.5 yet?

evankanderson:
I've seen reports that 1.5 is working.

/approve

We do need to figure out a better strategy for tracking which Istio releases are supported, ideally from the serving tests.

PR #2334, opened 2020-03-24
Updated link to autoscale default

### LINKED ISSUE #2293: Autoscale sample shows wrong number of Pods under load
## Expected Behavior

I expect the autoscale sample:
https://github.com/knative/docs/blob/master/docs/serving/samples/autoscale-go/README.md

To actually depict what really happens.

## Actual Behavior

With the introduction of `container-concurrency-targe-percentage` default to 70%, the autoscaling sample actually ends up with 7 or 8 pods instead of 5 like mentioned in the doc.

1. Run the sample step by step

## Fix

Set `container-concurrency-targe-percentage` to 100% of change the number of pods shown as output.

mpetason:
Fixes #2293 

## Proposed Changes 

- Updated the link to master so that it has the correct config. (not sure if we are avoiding this to not break links later on?)

### DIFF
--- a/docs/serving/samples/autoscale-go/README.md
+++ b/docs/serving/samples/autoscale-go/README.md
@@ -118,7 +118,7 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
 
 Knative Serving autoscaling is based on the average number of in-flight requests
 per pod (concurrency). The system has a default
-[target concurrency of 100](https://github.com/knative/serving/blob/3f00c39e289ed4bfb84019131651c2e4ea660ab5/config/config-autoscaler.yaml#L35-L41)
+[target concurrency of 100](https://github.com/knative/serving/blob/master/config/core/configmaps/autoscaler.yaml#L54-L60)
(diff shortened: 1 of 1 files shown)


evankanderson:
I missed the second comment.

/cancel-lgtm

/assign @vagababov @markusthoemmes

evankanderson:
Do'h, too slow

vagababov:
Yeah link to master won't work.

mpetason:
Where should we be linking? I guess that's why it linked to an older version previously. I can make the update and submit a PR for a fix.

vagababov:
You need to select a particular commit or a branch, which don't really change, e.g. 0.14 when it's cut on tuesday.

mpetason:
Ok, so took the latest update and will use this link I think - https://github.com/knative/serving/blob/8c7134ebfe368ee178229e99166e5bff8167b8ed/config/core/configmaps/autoscaler.yaml#L40-L55

I'll reach out to Victor on slack and see if I should submit another PR with this change.

PR #2341, opened 2020-03-25
add Kong Ingress Controller as a supported network layer

hbagdi:
## Proposed Changes 

Adds Kong as another option for networking layer.

### DIFF
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -219,6 +219,37 @@ The following commands install Istio and enable its Knative integration.
 
 {{< /tab >}}
 
+{{% tab name="Kong" %}}
+
+{{% feature-state version="v0.13" state="" %}}
+
+The following commands install Kong and enable its Knative integration.
+
+1. Install Kong Ingress Controller:
+
+   ```bash
+   kubectl apply --filename https://raw.githubusercontent.com/Kong/kubernetes-ingress-controller/0.9.x/deploy/single/all-in-one-dbless.yaml
+   ```
+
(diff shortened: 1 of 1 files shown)


docs/install/any-kubernetes-cluster.md
```diff
@@ -219,6 +219,37 @@ The following commands install Istio and enable its Knative integration.
...
+   kubectl apply --filename https://bit.ly/k4k8s
+   ```
+
+1. To configure Knative Serving to use Kourier by default:
+
+    ```bash
+    kubectl patch configmap/config-network \
+      --namespace knative-serving \
+      --type merge \
+      --patch '{"data":{"ingress.class":"kong"}}'
```
tcnghia: Use `kong.ingress.networking.knative.dev` to be consistent what other providers currently use?

hbagdi: The Ingress class for Knative Ingress resource and k8s Ingress resource are same in Kong's Ingress controller. This is to avoid users to think of two different ingress classes.

I do understand that this is a deviation from the other controllers that support Knative Ingress.

hbagdi:
ping @tcnghia

evankanderson:
ping @tcnghia 

If you need me to `/approve` once you've finished reviewing, let me know.

tcnghia:
@evankanderson I discussed with Harry on Slack that we probably want to see at least some tests passing before adding doc.

In the meantime @nak3 is working on a proposal on how we display the test results indicator next to the tabs and the criteria to be in Alpha|Beta|Stable.

PR #2347, opened 2020-03-29
Update kafkasource apigroup to match eventing-contrib

lberk:
knative/eventing-contrib#902 updated the apigroup, update the docs to match

this was reported in knative/eventing-contrib#1087


lberk:
@matzew done.

PR #2355, opened 2020-03-31
Added clone of docs to steps and updated paths

mpetason:
Fixes readability and matches serving docs where you will clone the docs first so that you are aware of where the files are. 

## Proposed Changes 
- add doc clone
- update paths

There is some additional cleanup that might be needed since this file doesn't actually exist now - strimzi-topic.yaml. If needed I can take the yaml from the docs and add it as a file in the sample/source folder. 

Also, not sure if we want to move the clone further up. I added it where you start applying files.

### DIFF
--- a/docs/eventing/samples/kafka/source/README.md
+++ b/docs/eventing/samples/kafka/source/README.md
@@ -84,7 +84,7 @@ You must ensure that you meet the [prerequisites listed in the Apache Kafka over
    cd knative-docs/docs/eventing/samples/kafka/source
    ```
 
-2. Build the Event Display Service (`even-display.yaml`)
+2. Build the Event Display Service (`event-display.yaml`)
 
    ```yaml
    apiVersion: serving.knative.dev/v1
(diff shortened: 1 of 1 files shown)


lberk:
/ok-to-test
Thanks for catching this, looks like we were the odd ones out here by specifying a partial path (vs in the same directory).

docs/eventing/samples/kafka/source/README.md
```diff
@@ -77,7 +77,14 @@ You must ensure that you meet the [prerequisites listed in the Apache Kafka over
...
 
-1. Build the Event Display Service (`even-display.yaml`)
+1. Download a copy of the code:
+
+   ```shell
+   git clone -b "{{< branch >}}" https://github.com/knative/docs knative-docs
+   cd knative-docs/docs/eventing/samples/kafka/source
+   ```
+
+2. Build the Event Display Service (`even-display.yaml`)
```
evankanderson, line 87:
```suggestion
2. Build the Event Display Service (`event-display.yaml`)
```

evankanderson: I'm not sure that this is "building" the event-display program anyway.

mpetason: Ah, thanks, missed this, I'll make an update and fix it.

PR #2361, opened 2020-04-02
Deprecate the observability plugin

mattmoor:
/assign @mdemirhan @evankanderson 
/hold

Holding to see that things look properly via stagedocs-bot.


mattmoor:
/unhold

This looks right to me: https://docs-on-the-rocks-2361.default.mink.knobots.io/docs/install/any-kubernetes-cluster/

PR #2377, opened 2020-04-09
Add a test for "index.md" frontmatter and fix the repo to match

evankanderson:
Hugo expects page bundles to have an `_index.md` or `index.md` file, but GitHub expects `README.md`. Until https://github.com/gohugoio/hugo/issues/4691 is fixed (still in flight, negotiating with hugo owner), it seems simplest to simply create `index.md` files for these leaf bundles with appropriate frontmatter and a `{{% readfile file="README.md" %}}` shortcode.

## Proposed Changes 

- Migrates frontmatter in about 50 directories into `index.md` files.
- Adds the `_index.html` frontpage that is in the website, so that we can eventually stop copying the file during website build.
- Adds a test to prevent backsliding


docs/serving/samples/hello-world/helloworld-nodejs/index.md
```diff
@@ -0,0 +1,8 @@
+---
+title: "Hello World - Node.js"
+linkTitle: "Node.js"
+weight: 1
```
mattmoor, line 4: Remove trailing whitespace:
```suggestion
weight: 1
```

evankanderson:
Here is a script **which does not work for some cases** which I used to start the conversion (just for reference):
```shell
#!/bin/bash

echo Working on $1

INDEX="${1%README.md}index.md"

awk '/^---/, /^$/' "$1" > "$INDEX"
echo '{{% readfile file="README.md" %}}' >> "$INDEX"

awk '/^---/, /^$/ {next}; {print}' "$1" > "$1.tmp" && mv "$1.tmp" "$1"

echo Wrote $INDEX and $1
```

evankanderson:
Verified `TestHugoBundles` is run, will fix the markdown issue.

evankanderson:
The test failure is fixed by https://github.com/knative/test-infra/pull/1898

docs/serving/samples/hello-world/helloworld-nodejs/index.md
mattmoor, line 4: Remove trailing whitespace:
```suggestion
weight: 1
```

evankanderson:
/hold

I just found some odd rendering that I need to investigate.

evankanderson:
Example: https://docs-2377.default.docs-on-the-rocks.io/community/samples/serving/helloworld-haskell/ has lost the sidebar.

evankanderson:
The rendering works with https://github.com/knative/website/pull/161 merged and applied on the website side.

I'll still need to update the other branches after this before we can drop the `find -exec if ... ; then mv ...`, but this gets us closer.

PR #2385, opened 2020-04-12
Switch net-kourier to use the artifact shortcode.

mattmoor:
This is blocked on net-kourier actually having a successful nightly, which should be unblocked by: https://github.com/knative/net-kourier/pull/31


evankanderson:
/assign @evankanderson 

Ping me when you have a nightly artifact.

mattmoor:
This should be good to go

Issue #2380, opened 2020-04-10
https://github.com/knative/docs/blob/release-0.13/docs/install/installing-istio.md


PhilCorney:
GitHub URL 404s for the the below Istio version

# Istio minor version should be 1.3, 1.4 or 1.5
export ISTIO_MINOR_VERSION=1.3

export VERSION=$(curl https://raw.githubusercontent.com/knative/serving/master/third_party/istio-${ISTIO_MINOR_VERSION}-latest)

kubectl apply -f https://raw.githubusercontent.com/knative/serving/master/third_party/${VERSION}/istio-knative-extras.yaml

sebgoa:
indeed, 1.3 was removed here https://github.com/knative/serving/commit/0762cfcff12a199c9179a9f2cc700c46d3e99305#diff-15230e5dad94ffaefa7b526c0c1e89e5

PR #2392, opened 2020-04-15
Update istio version for install istio-knative-extras.yaml

MIBc:
Fixes: #2380

## Proposed Changes 

- Update istio version.


sebgoa:
I would lgtm but I don't follow Istio that closely

PR #2403, opened 2020-04-19
Update Dockerfile

haf:
Fixes https://github.com/knative/eventing/issues/3011

## Proposed Changes 

-
-
-

### DIFF
--- a/docs/eventing/samples/helloworld/helloworld-go/Dockerfile
+++ b/docs/eventing/samples/helloworld/helloworld-go/Dockerfile
@@ -1,7 +1,7 @@
 # Use the offical Golang image to create a build artifact.
 # This is based on Debian and sets the GOPATH to /go.
 # https://hub.docker.com/_/golang
-FROM golang:1.12 as builder
+FROM golang:1.14 as builder
 
 # Copy local code to the container image.
 WORKDIR /go/src/github.com/knative/docs/helloworld


docs/eventing/samples/helloworld/helloworld-go/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Use the offical Golang image to create a build artifact.
 # This is based on Debian and sets the GOPATH to /go.
 # https://hub.docker.com/_/golang
-FROM golang:1.12 as builder
+FROM golang:1.14 as builder
```
matzew, line 4: I think we are on 1.13 ?

sebgoa:
in addition to moving to Go 1.13, I think we should actually use `go mod init` instead of installing dep.

if we want to keep dep, we should use `go get` instead of fetching it via curl

docs/eventing/samples/helloworld/helloworld-go/Dockerfile
akashrv: Yes. This should be 1.13. @haf could you please change it and send an update. I'll LGTM.

sebgoa: can we get rid off go dep as well ?

PR #2404, opened 2020-04-20
Change the field name from defaultBrokerClass to default-broker-class

houshengbo:

## Proposed Changes 

-
-
-

### DIFF
--- a/docs/install/operator/configuring-eventing-cr.md
+++ b/docs/install/operator/configuring-eventing-cr.md
@@ -170,7 +170,7 @@ spec:
 
 ## Configuring default broker class
 
-Knative Eventing allows you to define a default broker class when the user does not specify one. The operator ships with two broker classes: `ChannelBasedBroker` and `MTChannelBasedBroker`. The field `defaultBrokerClass` indicates which class to use; if empty, the `ChannelBasedBroker` will be used.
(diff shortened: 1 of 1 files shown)


docs/install/operator/configuring-eventing-cr.md
```diff
@@ -182,5 +180,5 @@ metadata:
   name: knative-eventing
   namespace: knative-eventing
 spec:
-  defaultBrokerClass: MTChannelBasedBroker
+  default-broker-class: MTChannelBasedBroker
```
matzew, line 183: https://github.com/knative/eventing-operator/blob/release-0.14/config/300-eventing-v1alpha1-knativeeventing-crd.yaml#L75

isn't that `defaultBrokerClass` ?

aliok: Yeah we need to change the field name in the CRD file and in the Go code as well: https://github.com/knative/eventing-operator/blob/release-0.14/pkg/apis/eventing/v1alpha1/knativeeventing_types.go#L75

aliok: Ok saw https://github.com/knative/eventing-operator/pull/200 now

houshengbo:
/hold
Let's hold it, since we do not change this name for release 0.14.x.

docs/install/operator/configuring-eventing-cr.md
houshengbo: Let's hold it for now, since we do not change it for 14.

abrennan89:
@houshengbo should this still be on hold? Which version will this be active in?

evankanderson:
I think we wanted to go the other way on names.

Going to close this.

/close.

PR #2409, opened 2020-04-22
Adding myself to reviewers to start reviewing

mpetason:
Adding myself to reviewers to help out with reviews

## Proposed Changes 

- add mpetason to reviewers for docs under OWNER_ALIASES
-
-


mpetason:
@evankanderson Adding myself to reviewers but it looks like I need to be part of the org. Is it invite only after doing some work in the group?

PR #2411, opened 2020-04-22
Updated link and default target from 100 to 70

mpetason:
Fixes an issue with a previously merged PR - https://github.com/knative/docs/pull/2334

## Proposed Changes 
- Updated default from 100 to 70 based on the default in the configs
- Updated link to a static file instead of based on Master

### DIFF
--- a/docs/serving/samples/autoscale-go/README.md
+++ b/docs/serving/samples/autoscale-go/README.md
@@ -118,7 +118,7 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
 
 Knative Serving autoscaling is based on the average number of in-flight requests
 per pod (concurrency). The system has a default
-[target concurrency of 70](https://github.com/knative/serving/blob/8c7134ebfe368ee178229e99166e5bff8167b8ed/config/core/configmaps/autoscaler.yaml#L40-L55)
+[target concurrency of 100](https://github.com/knative/serving/blob/8c7134ebfe368ee178229e99166e5bff8167b8ed/config/core/configmaps/autoscaler.yaml#L74)
 but [we used 10](service.yaml#L25-L26) for our service. We loaded the service
 with 50 concurrent requests so the autoscaler created 5 pods
 (`50 concurrent requests / target of 10 = 5 pods`)
--- a/docs/serving/samples/autoscale-go/README.md
+++ b/docs/serving/samples/autoscale-go/README.md
@@ -118,7 +118,7 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
 
 Knative Serving autoscaling is based on the average number of in-flight requests
 per pod (concurrency). The system has a default
-[target concurrency of 100](https://github.com/knative/serving/blob/8c7134ebfe368ee178229e99166e5bff8167b8ed/config/core/configmaps/autoscaler.yaml#L74)
+[target concurrency of 100](https://github.com/knative/serving/blob/master/config/core/configmaps/autoscaler.yaml)(Search for container-concurrency-target-default)
 but [we used 10](service.yaml#L25-L26) for our service. We loaded the service
 with 50 concurrent requests so the autoscaler created 5 pods
 (`50 concurrent requests / target of 10 = 5 pods`)
--- a/docs/serving/samples/autoscale-go/README.md
+++ b/docs/serving/samples/autoscale-go/README.md
@@ -118,7 +118,7 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
 
 Knative Serving autoscaling is based on the average number of in-flight requests
 per pod (concurrency). The system has a default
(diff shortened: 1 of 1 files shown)


docs/serving/samples/autoscale-go/README.md
```diff
@@ -118,7 +118,7 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
 
 Knative Serving autoscaling is based on the average number of in-flight requests
 per pod (concurrency). The system has a default
-[target concurrency of 100](https://github.com/knative/serving/blob/master/config/core/configmaps/autoscaler.yaml#L54-L60)
+[target concurrency of 70](https://github.com/knative/serving/blob/8c7134ebfe368ee178229e99166e5bff8167b8ed/config/core/configmaps/autoscaler.yaml#L40-L55)
```
markusthoemmes, line 121: We should probably say the the system has a default target concurrency of 100 and a default target utilization of 70%, effectively making it 70 from the autoscaler's PoV.

vagababov, line 121: No target concurrency is 100 by default.
70 is the hotness of revision
When multiplied we get autoscaling target of 70.

mpetason: So change the 70 back to 100? I think I just misread it. Looks like @vagababov said the same thing.

vagababov: Yes, the proper way to phrase is, I think:
keep this hundred and put link to the proper variable (https://github.com/knative/serving/blob/8c7134ebfe368ee178229e99166e5bff8167b8ed/config/core/configmaps/autoscaler.yaml#L74) and then mention the variable your link points to with default of 70% bringing us to the _default effective target_ of 70.

mpetason: I reverted back to 100 and updated the link, but am confused by what we need to get across in this example. Talking about the 70 percentage might confuse people, it confused me because I thought that was the actual target being used instead of default.

mpetason:
So the fix now was just to revert back to what we had in the beginning except for a newer URL showing a more updated version of the autoscaler.

sebgoa:
Why can't this point to the file at master ?

vagababov:
A change of lines will point to nothing on master.

sebgoa:
ok. I don't get it but that's fine, maybe a build concern. https://github.com/knative/serving/blob/master/config/core/configmaps/autoscaler.yaml#L74

mpetason:
Master might change/get updated while the docs are still live, which means that the link could point to the wrong line if there are line edits above the line that we are linking, so pointing at a specific file that doesn't change will ensure that we are always pointing to the correct line number for the configuration option.

docs/serving/samples/autoscale-go/README.md
```diff
@@ -118,7 +118,7 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
 
 Knative Serving autoscaling is based on the average number of in-flight requests
 per pod (concurrency). The system has a default
-[target concurrency of 100](https://github.com/knative/serving/blob/master/config/core/configmaps/autoscaler.yaml#L54-L60)
+[target concurrency of 100](https://github.com/knative/serving/blob/8c7134ebfe368ee178229e99166e5bff8167b8ed/config/core/configmaps/autoscaler.yaml#L74)
```
evankanderson, line 121: I'm nervous about pinning this to something which might be out of date (and therefore, future readers might miss that the default had changed again).

What do you think about either changing this to "master" or the `{{< branch >}}` shortcode?

vagababov: I'd be ok to even to link to just master branch filename and a string to search in it.

mpetason: @vagababov I'm not sure what that looks like, how do I post a URL with the correct search string?

vagababov: I meant, just to the file itself and a text to search, as in `look for X in this file for actual default value`.

docs/serving/samples/autoscale-go/README.md
```diff
@@ -118,7 +118,7 @@ A demonstration of the autoscaling capabilities of a Knative Serving Revision.
 
 Knative Serving autoscaling is based on the average number of in-flight requests
 per pod (concurrency). The system has a default
-[target concurrency of 100](https://github.com/knative/serving/blob/master/config/core/configmaps/autoscaler.yaml#L54-L60)
+[target concurrency of 100](https://github.com/knative/serving/blob/master/config/core/configmaps/autoscaler.yaml)(Search for container-concurrency-target-default)
```
vagababov, line 121:
```suggestion
[target concurrency of 100](https://github.com/knative/serving/blob/master/config/core/configmaps/autoscaler.yaml)(Search for `container-concurrency-target-default`)
```

PR #2417, opened 2020-04-24
[release-0.13] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/assign dprotaso grantr samodell
/cc dprotaso grantr samodell

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -465,14 +465,14 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:8a965ebe2d83033f6a07e926357f7341b6c7f42e165a3e13c7c8113b953a265b"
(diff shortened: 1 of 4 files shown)


matzew:
/hold

do we really need to update the 0.13 release branch ? 

/cc @grantr

Gopkg.lock
```diff
@@ -465,14 +465,14 @@
 
 [[projects]]
   branch = "master"
```
grantr, line 467: Looks like this branch is still tracking test-infra master. We should probably switch this to test-infra release-0.13 to avoid it being updated perpetually. See also https://github.com/knative/test-infra/issues/1899.

PR #2437, opened 2020-05-06
Add default tabs on configuring-autoscaling documentation

### LINKED ISSUE #2436: Autoexpand example section when multiple tabs are given
## Expected Behavior

Examples which have different flavors presented by tabs should already show the default presentation, like in

image

## Actual Behavior

In the current version of the new [autoscaling docs](https://knative.dev/development/serving/configuring-autoscaling/) the examples are not expanded by default, which leads to a view like:

image

For the reader is not clear whether this is already the example or what she should do. It's not obvious that you have to press a link to show the example. Also, if you press a link to reveal one of the example, you get never back to this empty view, so why not already start with the first example shown ? (the borders then make it also clear that this is tabbed view. these borders are missing in the initial view).

1. Got to https://knative.dev/development/serving/configuring-autoscaling/ and check the examples

jonatasbaldin:
Fixes #2436 

## Proposed Changes 
- Adds the default tabs on configuring-autoscaling documentation

### DIFF
--- a/docs/serving/configuring-autoscaling.md
+++ b/docs/serving/configuring-autoscaling.md
@@ -23,7 +23,7 @@ Per-revision settings are done by setting annotations on the revision. When you'
 **Note:** It's important that the annotation is set inside the template key so that it will appear on each revision as they are created. Setting it in the top-level metadata will not propagate them to the revision and thus will not have any effect on autoscaling.
 
 **Example:**
-{{< tabs name="example" >}}
+{{< tabs name="example" default="Per Revision" >}}
 {{% tab name="Per Revision" %}}
 ```yaml
 apiVersion: serving.knative.dev/v1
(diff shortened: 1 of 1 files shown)


evankanderson:
Waiting for #2411 to merge and will approve after that.

vagababov:
We had a discussion about that and kind of agreed on no defaults, but 🤷 
/cc @julz @markusthoemmes

markusthoemmes:
It would've been nice to get feedback via the actual working group here. I left the default tabs out deliberately to have less noise when just scrolling through the doc and searching for the setting you're looking for.

jonatasbaldin:
Hey @markusthoemmes, sorry for that, still understanding the process for the Pull Requests :<

We can revert this if the working group believes it is better 🤔 But I agree with @rhuss that the empty tabs can be misleading if people don't click on it.

markusthoemmes:
Yeah let's leave it as-is. I guess it's debatable anyway. I just would've liked to get the change to review.

julz:
FWIW (I guess this is a feature request :)) I think the core problem is the CSS doesn’t properly support non-expanded tab groups, so they look confusing. The nicest fix would be if we could make it so non-expanded tab groups were less visually confusing, then we wouldn’t need to expand everything by default to work around that.

PR #2441, opened 2020-05-09
CloudEvents sample for ASP.NET Core

mdemirhan:
Sample project showing how to start an event receiver as a Knative Service using ASP.NET Core.

The behavior of this application is slightly different than the Go sample. Instead of switching behavior of the API based on K_SINK, this allows both behaviors to be tested within the same app. 
`api/events` receives an event and responds back with another one.
`api/events/sink` receives an event, forms another event from the input event, sends that output event to the URL specified in K_SINK and replies with the response back to the user.

## Proposed Changes 

-
-
-


mattmoor:
Wow, that's a lot of files 🤣 

Let me try to find someone better versed in .NET to review this, but worst case I'll just try the instructions myself.

thanks for putting this together!

mattmoor:
Reading your PR description, one gotcha is that the way the Broker targets "Addressable" there isn't a way AFAIK to target a specific path (e.g. your `/api/events`).  Ideally at least the request/response path would be aligned with `/` so it can be used with the Broker, but this is why the Go sample did both of these things on `/`.

mdemirhan:
Welcome to .NET world :) It is not that bad though, most is auto-generated stuff (csproj files, setting files, most of Startup.cs, Program.cs etc). Even though they say CE .NET SDK is v1.0, it doesn't seem to have proper support for ASP .NET's latest version, so there is an extra file to convert an HttpResponseMessage to an IActionResult.

Regarding the paths, do you recommend me changing the behavior to match the go sample?

mattmoor:
> Regarding the paths, do you recommend me changing the behavior to match the go sample?

If we can...  I think this is the main use case we want to prepare folks for.

Also FYI the [Sink Binding sample](https://knative.dev/docs/eventing/samples/sinkbinding/) references these samples as "extra credit", so you might try your sample in that context if you are feeling adventurous :D

n3wscott:
> there isn't a way AFAIK to target a specific path (e.g. your `/api/events`).

fyi: You can add a path to something like the trigger or subscription in eventing.

mdemirhan:
> > there isn't a way AFAIK to target a specific path (e.g. your `/api/events`).
> 
> fyi: You can add a path to something like the trigger or subscription in eventing.

So, in that case, should we keep this sample as is? Or still convert it into the go sample format? Either way is fine by me - lmkwyt

evankanderson:
For the simple cases where the container is handling a single event, I'd tend to prefer to handle that event via a `POST / HTTP/1.1`, rather than needing to discover a per-container unique address.

Bundling multiple event handlers into a single container feels like a more-advanced case.

mdemirhan:
Done. PTAL.

docs/serving/samples/cloudevents/cloudevents-dotnet/service.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: cloudevents-dotnet
+  namespace: default
+spec:
+  template:
+    spec:
+      containers:
+        - image: <image>
+          ports:
+          - containerPort: 80
```
mattmoor, line 12: How hard would it be to make this 8080?

docs/serving/samples/cloudevents/cloudevents-dotnet/README.md
```diff
@@ -0,0 +1,108 @@
...
+    -H "ce-type: curl.demo"  \
+    -H "ce-id: 123-abc"  \
+    -d '{"name":"Dave"}' \
+    http://cloudevents-dotnet.default.1.2.3.4.xip.io
+```
+
+You will get back:
+
+```shell
+{"message":"Hello, Dave"}
```
mattmoor, line 99: I'm getting back a structured event here, and the payload looks overly escaped (thanks to @n3wscott for noticing):

```
{
  "specversion": "1.0",
  "type": "dev.knative.docs.sample",
  "source": "https://github.com/knative/docs/docs/serving/samples/cloudevents/cloudevents-dotnet",
  "id": "53b5cd46-7969-4452-95ff-da9563578d84",
  "time": "2020-05-18T16:35:26.0819229Z",
  "datacontenttype": "application/json",
  "data": "{\"message\":\"Hello, Dave\"}"
}
```

docs/serving/samples/cloudevents/cloudevents-dotnet/Startup.cs
```diff
@@ -0,0 +1,42 @@
+using CloudNative.CloudEvents;
```
mattmoor, line 1: We need license headers.

docs/serving/samples/cloudevents/cloudevents-dotnet/service.yaml
mdemirhan: Not hard. Pushed a change to listen on the default 8080 port.

docs/serving/samples/cloudevents/cloudevents-dotnet/README.md
mdemirhan: Great catch. I was doing serialization, but Cloud Events was serializing it one more time. Removed the serialization on my part.

docs/serving/samples/cloudevents/cloudevents-dotnet/Startup.cs
mdemirhan: Will add headers to all files. It seems like most of the samples we have don't have any license headers though, including the go sample I copied from :)

mattmoor: Yeah, I figured 🙈

docs/serving/samples/cloudevents/cloudevents-dotnet/README.md
```diff
@@ -1,10 +1,3 @@
----
-title: "Cloud Events - C#"
-linkTitle: "C#"
-weight: 1
-type: "docs"
----
-
```
mattmoor: Did you mean to drop this?

docs/serving/samples/cloudevents/cloudevents-dotnet/README.md
mattmoor: It was all Scott.  Can you update the sample output to use the "structured" encoding, since that's what we're getting?  Here's what I see now:

```json
{
  "specversion": "1.0",
  "type": "dev.knative.docs.sample",
  "source": "https://github.com/knative/docs/docs/serving/samples/cloudevents/cloudevents-dotnet",
  "id": "d662b6f6-35ff-4b98-bffd-5ae9eee23dab",
  "time": "2020-05-19T01:26:23.3500138Z",
  "datacontenttype": "application/json",
  "data": {
    "message": "Hello, Dave"
  }
}
```

docs/serving/samples/cloudevents/cloudevents-dotnet/README.md
mdemirhan: Yes, it looks like other samples, including the go one now has this information in _index.md instead. So I moved it there.

docs/serving/samples/cloudevents/cloudevents-dotnet/README.md
mdemirhan: Done.

mdemirhan:
Thanks for the review & testing Matt. Updated the sample output in README.md with the correct one.

PR #2449, opened 2020-05-12
Reorder broker docs; add reference to gcp broker

liu-cong:

## Proposed Changes 

- Move broker related docs into a new folder for better organizing.
- Also add a link to GCP broker
-

cc @grantr @vaikas


docs/eventing/broker/alternate/README.md
```diff
@@ -0,0 +1,8 @@
+In the Knative Eventing ecosystem, vendors are welcome to bring their own implementation of Broker
```
vaikas, line 1: While I totes agree that vendors are welcome, so are other communities not strictly restricted to vendors. I'd prefer to rework the wording to be more inclusive.

liu-cong: Good point! Reworded a little bit.

liu-cong:
BTW, this is how it will look like in docs website:

Screen Shot 2020-05-12 at 4 35 22 PM

Issue #2357, opened 2020-04-01
Istio (v1.5.1 or later) should be installed by istioctl command


nak3:
## What is changing? (Please include as many details as possible.)

Istio projects announced that `helm` installation will be deprecated. So `istioctl` command should be used instead of it.

## How will this impact our users?

Users need to use `istioctl` to install istio.

## In what release will this take happen (to the best of your knowledge)?

Istio 1.5.1 or later. Istio 1.4.x and Istio 1.5.0 have a bug so helm should be used.

## Context
Link to associated PRs or issues from other repos here.

Knative Serving's e2e also starts using it.

1. https://github.com/knative/serving/pull/7414

## Additional info

The section https://knative.dev/docs/install/installing-istio/ needs to be updated.

PR #2458, opened 2020-05-13
Use istioctl instead of helm to install istio

nak3:
Fixes https://github.com/knative/docs/issues/2357 https://github.com/knative/serving/issues/7568

## Proposed Changes 

- Use `istioctl` instead of `helm` to install istio.
- Remove SDS related section as Istio 1.5 enables SDS by default. See https://github.com/istio/istio/issues/22306#issuecomment-605641786

NOTE: By this change, Istio 1.5.2 becomes the minimal version. Istio 1.4.x will be [EOL](https://istio.io/news/support/announcing-1.4-eol/) so it should be alright to remove it.

### DIFF
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -18,7 +18,7 @@ instructions.
 You need:
 
 - A Kubernetes cluster created.
-- [`istioctl`](https://istio.io/docs/setup/install/istioctl/) (v1.5.2 or later) installed.
+- [`istioctl`](https://istio.io/docs/setup/install/istioctl/) (v1.5.4 or later) installed.
 
 ## Installing Istio
 
@@ -57,9 +57,6 @@ spec:
     global:
       proxy:
         autoInject: disabled
-      disablePolicyChecks: true
-      defaultPodDisruptionBudget:
-        enabled: false
       useMCP: false
       # The third-party-jwt is not enabled on all k8s.
       # See: https://istio.io/docs/ops/best-practices/security/#configure-third-party-service-account-tokens
@@ -75,27 +72,11 @@ spec:
     ingressGateways:
       - name: istio-ingressgateway
         enabled: true
-        k8s:
-          service:
-            type: LoadBalancer
-            ports:
-            - port: 15020
-              name: status-port
-            - port: 80
-              name: http2
-            - port: 443
-              name: https
-          overlays:
-            # Although we don't need this Gateway, we cannot remove it. Add "dummy" selector for the workaround.
-            # https://github.com/istio/istio/issues/22095
-            - kind: Gateway
-              name: istio-ingressgateway
-              patches:
-                - path: spec.selector.istio
-                  value:
-                    dummy
       - name: cluster-local-gateway
         enabled: true
+        label:
+          istio: cluster-local-gateway
+          app: cluster-local-gateway
         k8s:
           service:
             type: ClusterIP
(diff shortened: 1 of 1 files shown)


docs/install/installing-istio.md
```diff
@@ -29,235 +29,180 @@ instructions. If you're familiar with Istio and know what kind of installation
...
-#### Installing Istio without sidecar injection
+- [Installing Istio with sidecar injection](#installing-istio-with-sidecar-injection)
 
 If you want to get up and running with Knative quickly, we recommend installing
 Istio without automatic sidecar injection. This install is also recommended for
 users who don't need the Istio service mesh, or who want to enable the service
-mesh by [manually injecting the Istio sidecars][2].
+mesh by [manually injecting the Istio sidecars][1].
 
 Enter the following command to install Istio:
```
evankanderson, line 49: This seems really complicated. I was able to get both `default` and `default with SDS` with the following two commands on Istio 1.4.9 (tested yesterday when I rebuilt my cluster):

```shell
istioctl manifest apply
```

```shell
istioctl manifest apply --set values.gateways.istio-ingressgateway.sds.enabled=true
```

I think those both install the sidecar; `--set autoInjection.enabled=false` I think disables the sidecar injection.

(I'm just hoping to have a minimal command line so it's not so intimidating.)

docs/install/installing-istio.md
```diff
@@ -29,235 +29,180 @@ instructions. If you're familiar with Istio and know what kind of installation
...
-Install Istio with [Secret Discovery Service (SDS)][3] to enable a few
-additional configurations for the gateway TLS. This will allow you to:
-
-- Dynamically update the gateway TLS with multiple TLS certificates to terminate
-  TLS connections.
-
-- Use [Auto TLS](../serving/using-auto-tls.md).
-
-The below `helm` flag is needed in your `helm` command to enable `SDS`:
+To automatic sidecar injection, set `autoInject: enabled` in addition to above
```
evankanderson, line 170: I thought this would be `autoInjection.enabled=true` (or false)

docs/install/installing-istio.md
nak3: Thank you @evankanderson
The reason of this complicated definition is that `cluster-local-gateway`. We cannot setup `cluster-local-gateway` with command line and so need to use this IstioOperator definition.

The installation by `istioctl manifest apply` works for external access but we cannot use `.svc.cluster.local` for ksvc.

docs/install/installing-istio.md
nak3: If we use the command line option, yes it is. But as we used the IstioOperator YAML above, I wanted to say "please set `autoInject: enabled` here https://github.com/knative/docs/pull/2458/files#diff-1116bcf887cf6b5e21e5e4f96fa026b1R59"

nak3:
@evankanderson Could you please review this again? I updated the manifest as much as simple.
If we use the command line (`istioctl manifest apply`) without yaml, we need to give up to install cluster-local-gateway. I think it is better to instruct users to install cluster-local-gateway for this way.

docs/install/installing-istio.md
evankanderson: Ah, thanks! I wasn't testing eventing, only external services, so I missed this. It might be worth explaining that the `cluster-local-gateway` is needed for serving+eventing use.

nak3:
> It might be worth explaining that the cluster-local-gateway is needed for serving+eventing use.

Thank you. Yes, I agree.  `cluster-local-gateway` is one of common reported problem in serving. It should be explained tin the docs. (I will try to update the docs.)

PR #2464, opened 2020-05-13
cp issue-2462

vaikas:

## Proposed Changes 

- CP #2463 changes.
-
-


evankanderson:
It looks (both by size and by quick grep/sort/uniq) that `eventing.yaml` has some content that's not in `-core` or `-crds`. `config-imc-event-dispatcher` is one example.

As such, using `eventing.yaml` seems like the maximal choice.

/lgtm
/approve

PR #2476, opened 2020-05-18
fix event_display image name

lionelvillard:
## Proposed Changes 

- fix event_display image name
-
-


mpetason:
/lgtm
Links all work and are accessible

PR #2493, opened 2020-05-22
Rewrite eventing/README.md to reference different usage cases

evankanderson:
…enting#3186

Motivated by https://github.com/knative/eventing/issues/3186, where it wasn't clear that one of the roles of Broker was to be a target for user-generated events.

## Proposed Changes 

- Updates the eventing main document.

### DIFF
--- a/docs/eventing/README.md
+++ b/docs/eventing/README.md
@@ -67,14 +67,22 @@ generic interfaces that can be implemented by multiple Kubernetes resources:
 
 ### Event brokers and triggers
 
-Broker and Trigger objects make it easy to filter events based on event attributes.
+Broker and Trigger objects make it easy to filter events based on event
+attributes.
 
 A Broker provides a bucket of events which can be selected by attribute. It
 receives events and forwards them to subscribers defined by one or more matching
-Triggers.
+Triggers. Since a Broker implements Addressable, event senders can submit events
+to the Broker by POSTing the event to the Broker's `status.address.url`.
 
 A Trigger describes a filter on event attributes which should be delivered to an
 Addressable. You can create as many Triggers as necessary.
+
+For most use cases, a single bucket (Broker) per namespace is sufficient, but
+there are serveral use cases where multiple buckets (Brokers) can simplify
+architecture. For example, separate Brokers for events containing Personally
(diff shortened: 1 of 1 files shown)


cardil:
I do not feel like this PR fixes knative/eventing#3186

knative/eventing#3186 is to address ease of use, security and traceability of sending events. Those can't be resolved by docs change as they are not implemented.

But, I do feel like your changes are welcome. I would like to advice to add your paragraph here as well:

> Broker is intended to be part of the external surface area of eventing. The purpose of having "a bucket of events that you can filter by attribute" is to make it easy for event producers to send events -- they should go into the bucket. Most of the time, there should be just one bucket (per namespace), but there are good cases for allowing multiple buckets, so that's why you can create multiple Brokers (one of the first examples was PII and non-PII Brokers).

evankanderson:
Sorry, when I wrote this, it wasn't clear that https://github.com/knative/eventing/issues/3186 was about access control capabilities for Eventing. I'll remove the "Fixes" reference.

I've also updated the text to include the pointers about how to send events to the Broker (to make clear that this is a public interface), and the example about when to use multiple Brokers.

Abd4llA:
/lgtm
It's awesome that you've updated the README Evan. I was wondering as well what do you think about this section? it seems outdated and blame shows it was added in 0.2 version :D 

> ### Future design goals	### Future design goals
> The focus for the next Eventing release will be to enable easy implementation of	The focus for the next Eventing release will be to enable easy implementation of
> event sources. Sources manage registration and delivery of events from external	event sources. Sources manage registration and delivery of events from external
> systems using Kubernetes	systems using Kubernetes
> [Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/).	[Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/).
> Learn more about Eventing development in the

evankanderson:
I'd be happy to see the whole doc rewritten.

My aim was to make the current interfaces a bit more clear, but it seems like we should also have a "hit list" for docs that need to be updated/rewritten, sorted by impact.

PR #2503, opened 2020-05-23
update ruby and bundle version

xiangshen-dk:
Upgrade Ruby to 2.6 and related libraries. 
It also fixes the build issue on GCP Cloud Build if anybody wants to follow the steps below and deploy the sample to GCP Cloud Run:
https://cloud.google.com/run/docs/quickstarts/build-and-deploy#ruby

## Proposed Changes 

- Upgrade the base image from 2.5-slim to 2.6-slim
- Add 'gem install bundler' in the Dockerfile
- Update the Gemfile.lock


docs/serving/samples/hello-world/helloworld-ruby/Gemfile.lock
```diff
@@ -20,4 +22,4 @@ DEPENDENCIES
   sinatra
 
 BUNDLED WITH
-   1.17.1
+   2.1.4
```
mattmoor, line 25: Add trailing newline:
```suggestion
   2.1.4

```

evankanderson:
It looks like you need to update `README.md` as well:

> sampleconsistency_test.go:97: README.md file is missing line 2 ('FROM ruby:2.6-slim') from file '../../docs/serving/samples/hello-world/helloworld-ruby/Dockerfile'
>     Additional info:
>     All files required for running sample apps are checked against README.md, the content of source files should be identical with what's in README.md file, the list of the files to be verified is the same set of files used for running sample apps, they are configured in `/test/sampleapp/config.yaml`. If an exception is needed the file can be configured to be copied as a separate step in `PreCommand` such as: https://github.com/knative/docs/blob/65f7b402fee7f94dfbd9e4512ef3beed7b85de66/test/sampleapp/config.yaml#L4

PR #2506, opened 2020-05-26
PR to update samples.md with 2 external code samples

meteatamel:

## Proposed Changes 

-
-
-

### DIFF
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -439,13 +439,14 @@ The following commands install the Knative Eventing component.
 1. Install the [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) (aka CRDs):
 
    ```bash
-   kubectl apply --filename {{< artifact repo="eventing" file="eventing-crds.yaml" >}}
+   kubectl apply  --selector knative.dev/crd-install=true \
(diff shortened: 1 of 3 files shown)


docs/install/any-kubernetes-cluster.md
```diff
@@ -439,13 +439,14 @@ The following commands install the Knative Eventing component.
 1. Install the [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) (aka CRDs):
 
    ```bash
-   kubectl apply --filename {{< artifact repo="eventing" file="eventing-crds.yaml" >}}
+   kubectl apply  --selector knative.dev/crd-install=true \
```
evankanderson, line 442: I think we've been preferring the `-crd` and `-core` files rather than using the label filter.

I don't recall why, but @mattmoor or @vaikas may remember what issues we hit.

evankanderson: From https://knative.slack.com/archives/C93E33SN8/p1590510270014300 :

> Kubectl doesn’t apply the selector before blowing up on custom resources

https://github.com/knative/test-infra/issues/1858

evankanderson:
(Also, your git history seems to be confused -- maybe you committed a bad merge?)

meteatamel:
You're right. Let me close and create a new PR.

PR #2514, opened 2020-05-27
Getting started: Added examples with using "kn"

rhuss:


docs/serving/getting-started-knative-app.md
```diff
@@ -36,24 +35,41 @@ docker tag local-image dev.local/local-image
...
+
+- Create a new immutable revision for this version of the app.
+- Perform network programming to create a route, ingress, service, and load
+  balancer for your app.
+- Automatically scale your pods up and down based on traffic, including to zero
+  active pods.
+
+## Creating your Deployment with YAML
+
+Alternatively, to deploy an app using Knative, you can also create the configuration in a `.yaml` file that defines a Service. For more information about the Service object, see the
```
abrennan89, line 65:
```suggestion
Alternatively, to deploy an app using Knative, you can also create the configuration in a YAML file that defines a service.
```

abrennan89:
Added a comment, but it looks like now all the `kubectl` commands have been removed? Instead, we should add examples with tabs that show both `kubectl` and `kn`, since this is optional?

rhuss:
> Added a comment, but it looks like now all the `kubectl` commands have been removed? Instead, we should add examples with tabs that show both `kubectl` and `kn`, since this is optional?

@abrennan89 I think this getting started guide is kind of special as it explain very detailed how to create a first service, including how to use `kubectl`. So just adding a tab is not good enough. Hence I split it up into two sections: How to create with YAML and `kubectl` and how to create a service with `kn`. For other samples later (where there is no explicit reference to `kubectl` in the main text) we can use the tab approach.

An alternative would be to have two "Getting started" guides, for each method. wdyt ?

rhuss:
@abrennan89 Do you know how to solve the link-check build error ? It checks the link in an example text, which can not be (and should not be) resolved, so I consider this to be a false positive.

rhuss:
@rhuss googlebot I fixed it.

rhuss:
@abrennan89 this are the two sections for the example [with](https://github.com/knative/docs/blob/c838c55a4bdcb9cef51e0984b44e6b1e1a8a99c0/docs/serving/getting-started-knative-app.md#creating-your-deployment-with-the-knative-cli) and [without](https://github.com/knative/docs/blob/c838c55a4bdcb9cef51e0984b44e6b1e1a8a99c0/docs/serving/getting-started-knative-app.md#creating-your-deployment-with-yaml) `kn`. I think its best to review not the diff but the whole doc as I did some rearrangements, too.

rhuss:
@googlebot I fixed it.

rhuss:
@googlebot I consent.

rhuss:
@abrennan89 I think the CLA bot thinks that you have to sign the CLA still. Not sure, but maybe a `@googlebot I consent` could be good enough.

abrennan89:
> @abrennan89 Do you know how to solve the link-check build error ? It checks the link in an example text, which can not be (and should not be) resolved, so I consider this to be a false positive.

This isn't actually something I've seen happen before. @mattmoor is there some way to escape this maybe?

abrennan89:
/approve
Looks good!

abrennan89:
@googlebot I consent.

abrennan89:
> @abrennan89 I think the CLA bot thinks that you have to sign the CLA still. Not sure, but maybe a `@googlebot I consent` could be good enough.

I've never had issues with it in any other docs PR though? :thinking:

averikitsch:
Error: `docs/serving/getting-started-knative-app.md\u0026#xA;	ERROR	39s\u0026#xA;		Stat docs/serving/39s: no such file or directory\u0026#xA;	ERROR	http://helloworld-go.default..34.83.80.117.xip.io\u0026#xA;		Lookup helloworld-go.default..34.83.80.117.xip.io: no such host\u0026#xA;`

rhuss:
> I've never had issues with it in any other docs PR though? 🤔

@abrennan89 I think this happens when someone accepts the suggestions provided by a review. Could you try to add a comment with just 

> @googlebot I consent.

This might fix it, but not sure.

rhuss:
I messed up that PR (sorry), replacing it with #2531

PR #2551, opened 2020-06-04
Fix #2415

### LINKED ISSUE #2415: Page needs improvement
## Documentation Unlear
Step 3 makes reference to applying service account when above simply talks of modifying existing.

evankanderson:
Fixes #2415

## Proposed Changes 

- Remove extraneous steps (they were always extraneous, even from version 1 of the doc).


mpetason:
/lgtm

Do we want to wait until the link is fixed before pushing something like this, or since it has been documented this PR can be merged?

PR #2572, opened 2020-06-11
fix get vsvc in debug app issues

runzexia:
Fixes
In fact, each knative ingress corresponds to multiple vsvc, so we should use selector to query.


mpetason:
/lgtm

We'll have to see if anyone else has feedback too. Walking through the steps I do not get a result without using a selector.

Issue #2343, opened 2020-03-27
Refers to knative-serving domain config out of order


jeffgriffith:
'Install istio for knative' gives steps for modifying the config-domain for knative-serving, but we have not yet created knative-serving. Istio is a pre-req. Perhaps suggest doing it after knative-serving is installed or move it there?

nak3:
Yes, I think that we can just remove it. The current doc explains it in another section.

https://knative.dev/docs/install/any-kubernetes-cluster/
> 4. Configure DNS
`Magic DNS (xip.io)`

(Or  add `Magic DNS (xip.io) manual` tab, but I think we don't need it.)

/assign

PR #2573, opened 2020-06-11
Remove "Configuring DNS" from Installing Istio for Knative

nak3:
## Proposed Changes 

This patch removes "Configuring DNS" section from Installing Istio for Knative.

It needs to install Knative Serving, so it is not correct to explain
it in Istio which is pre-req.
Also, the configuration is instructed in knative installation.

https://knative.dev/docs/install/any-kubernetes-cluster/
> 4. Configure DNS
`Magic DNS (xip.io)`

Fixes https://github.com/knative/docs/issues/2343

### DIFF
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -139,44 +139,6 @@ spec:
 EOF
 ```
 
-### Configuring DNS
-
-Knative dispatches to different services based on their hostname, so it greatly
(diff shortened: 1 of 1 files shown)


nak3:
(Just rebased to solve conflict.)

nak3:
/assign @abrennan89 

Would you mind approve it?

abrennan89:
/lgtm
/approve

Thanks for fixing this, @nak3 !

PR #2574, opened 2020-06-11
Remove AWS from unsupported default-domain in Caveat

nak3:
## Proposed Changes 

Current doc explains that default-domain does not work on AWS as it
does not expose External IP. But it exposes Extenal Host and supported
it by https://github.com/knative/serving/pull/7518.

Hence, this patch removes AWS from Caveat.

/cc @mattmoor @evankanderson

### DIFF
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -293,7 +293,7 @@ We ship a simple Kubernetes Job called "default domain" that will (see caveats) 
 kubectl apply --filename {{< artifact repo="serving" file="serving-default-domain.yaml" >}}
 ```
 
-**Caveat**: This will only work if the cluster LoadBalancer service exposes an IPv4 address, so it will not work with IPv6 clusters, AWS, or local setups like Minikube.  For these, see "Real DNS" or "Temporary DNS".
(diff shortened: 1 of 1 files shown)


docs/install/any-kubernetes-cluster.md
```diff
@@ -293,7 +293,7 @@ We ship a simple Kubernetes Job called "default domain" that will (see caveats)
 kubectl apply --filename {{< artifact repo="serving" file="serving-default-domain.yaml" >}}
 ```
 
-**Caveat**: This will only work if the cluster LoadBalancer service exposes an IPv4 address, so it will not work with IPv6 clusters, AWS, or local setups like Minikube.  For these, see "Real DNS" or "Temporary DNS".
+**Caveat**: This will only work if the cluster LoadBalancer service exposes an IPv4 address or hostname, so it will not work with IPv6 clusters or local setups like Minikube.  For these, see "Real DNS" or "Temporary DNS".
```
abrennan89, line 296:
```suggestion
**Caveat**: This will only work if the cluster LoadBalancer service exposes an IPv4 address or hostname. It does not work for IPv6 clusters or local setups, such as Minikube.
```

abrennan89, line 296: Add a line after this:
For information about IPv6 clusters or local setups, see [then add links to docs about DNS types instead of saying "Real DNS" or "Temporary DNS"].

evankanderson: The "Real DNS" and "Temporary DNS" are other tabs (see lines 300-363).

PR #2439, opened 2020-05-07
Improvements and structure changes for autoscaling docs

abrennan89:
## Proposed Changes

- Split autoscaling into its own sub directory and smaller topics, to make the content easier to consume.
- General content editing.

PR #2603, opened 2020-06-19
Template for verifying AutoTLS was moved

zevisert:
Auto TLS verification used the `autoscale-go` sample, but it was moved in https://github.com/knative/docs/pull/2439, and not updated here.

 - Didn't find any. I came from the documentation's "*View / Edit this page*" link

## Proposed Changes 

- Update the link to the deployment template for verifying auto-tls is correctly configured

### DIFF
--- a/docs/serving/using-auto-tls.md
+++ b/docs/serving/using-auto-tls.md
@@ -308,7 +308,7 @@ be able to handle HTTPS traffic.
 
 1.  Run the following comand to create a Knative Service:
     ```shell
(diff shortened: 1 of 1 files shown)


zevisert:
This PR is against `release-0.15` since that was the default, given the "_View / Edit this page_" link taking me to the markdown doc that's published for v0.15, if I should actually open this against master, let me know!

mpetason:
Usually updates happen with the main branch, I'm not sure if we should target the release or the main branch. After we get an update on that this should get a lgtm or a request to open a new PR. We'll have to wait for someone from OWNERS to weigh in.

PR #2643, opened 2020-07-05
Fix incorrect yaml format in "Kubernetes event using the API Server Source"

nak3:
This patch deletes `---` which creates incorrect yaml rendering.

Please refer to https://knative.dev/docs/eventing/samples/kubernetes-event-source/#deployment-steps
The yaml format in `ServiceAccount` is broken.

image


lionelvillard:
I think we should fix the underlying rendering issue, not the YAML.

nak3:
I see. Then I close this PR.

PR #2644, opened 2020-07-05
Enable grpc server reflection for grpc-ping example

ZhiminXiang:
With [server reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), we can use `grpcurl` to list the gRPC proto server information, and interact with the grpc-ping server more conveniently.


ZhiminXiang:
The integration tests were fixed in a separate PR https://github.com/knative/docs/pull/2716. Can I get another lgtm? :)

PR #2653, opened 2020-07-07
Fix the doc for eventing as yaml files are renamed

houshengbo:
## Proposed Changes 

- This PR changed the doc of eventing, as yaml files have been renamed.
- This PR is based on https://github.com/knative/eventing/pull/3536.

### DIFF
--- a/docs/install/upgrade-installation.md
+++ b/docs/install/upgrade-installation.md
@@ -67,7 +67,7 @@ upgrade, and these are identified in the release notes. For example, upgrading
 from v0.15.0 to v0.16.0 for Eventing you have to run:
 
 ```bash
-kubectl apply --filename {{< artifact repo="eventing" file="pre-install-to-v0.16.0.yaml" >}}
+kubectl apply --filename {{< artifact repo="eventing" file="eventing-pre-install-jobs.yaml" >}}
 ```
 
 ### Upgrade existing resources to the latest stored version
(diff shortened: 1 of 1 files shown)


docs/install/upgrade-installation.md
```diff
@@ -67,7 +67,7 @@ upgrade, and these are identified in the release notes. For example, upgrading
 from v0.15.0 to v0.16.0 for Eventing you have to run:
 
 ```bash
-kubectl apply --filename {{< artifact repo="eventing" file="pre-install-to-v0.16.0.yaml" >}}
```
vaikas, line 70: This one doesn't exist in v0.16.0, so I wonder what the best way to indicate this is? Wonder if it's going to be better to say to install from v0.15.0 to v0.16.0 for eventing you'd do it just like this and then add a section and another example for v0.15.0 to v0.16.1 having this file since these changes will be in that release?

houshengbo: @vaikas My bold thought is whether we can manually change the release from pre-install-to-v0.16.0.yaml to eventing-pre-install-job.
If not, operator will at least wait for another release to enable online download.

houshengbo: Or create eventing-pre-install-job and eventing-post-install-job with the same content as pre-install-to-v0.16.0.yaml and post-install-to-v0.16.0.yaml, for operator. Keep all of them in 0.16. 
Complete the renaming for next release. @vaikas @markusthoemmes

vaikas: I'd be down for having both files there for the release, I just don't know how to release only two additional files into release artifacts. Can you figure out how to do that so they are release artifacts in the .16.0 release, without having to create another 'dot' release and all the fun it entails?
Wonder if it's just a matter of manually making a copy in the gcs and editing the release in the github page?

vaikas: I manually added these files to the .16.0 release, so let's get this in.

PR #2658, opened 2020-07-08
Removed "alias" field from v0.15.0 docs

ellenevans:
Archive 0.15 docs by removing alias field.


ellenevans:
Still not that day.

ellenevans:
Nope

PR #2662, opened 2020-07-10
Clean up eventing docs

abrennan89:
Doesn't fix any one particular issue; general clean up / rewrite of eventing doc sections.

## Proposed Changes

- Clean up of eventing getting started doc

### DIFF
--- a/docs/eventing/getting-started.md
+++ b/docs/eventing/getting-started.md
@@ -87,17 +87,17 @@ specific consumer.
 
    ---
 
-   kind: Service
-   apiVersion: v1
-   metadata:
-     name: hello-display
-   spec:
-     selector:
-       app: hello-display
-     ports:
-     - protocol: TCP
-       port: 80
-       targetPort: 8080
+     kind: Service
+     apiVersion: v1
+     metadata:
+       name: hello-display
+     spec:
+       selector:
+         app: hello-display
+       ports:
+       - protocol: TCP
+         port: 80
+         targetPort: 8080
    END
    ```
 1. To deploy the `goodbye-display` consumer to your cluster, run the following
--- a/docs/eventing/getting-started.md
+++ b/docs/eventing/getting-started.md
@@ -31,7 +31,7 @@ The [broker](./broker/README.md#broker) allows you to route events to different 
     ```
     kubectl create -f - <<EOF
     apiVersion: eventing.knative.dev/v1
-    kind: Broker
+    kind: broker
     metadata:
      name: default
      namespace: event-example
@@ -41,7 +41,7 @@ The [broker](./broker/README.md#broker) allows you to route events to different 
 1. Verify that the broker is working correctly, by entering the following command:
 
     ```
-    kubectl --namespace event-example get Broker default
+    kubectl -n event-example get broker default
     ```
 
     This shows information about your broker. If the broker is working correctly, it shows a `READY` status of `True`:
@@ -63,7 +63,7 @@ demonstrate how you can configure your event producers to target a specific cons
    command:
 
      ```
-     kubectl --namespace event-example apply --filename - << END
+     kubectl -n event-example apply -f << EOF
      apiVersion: apps/v1
      kind: Deployment
      metadata:
@@ -94,14 +94,14 @@ demonstrate how you can configure your event producers to target a specific cons
        - protocol: TCP
          port: 80
          targetPort: 8080
-     END
+     EOF
      ```
 
 1. To deploy the `goodbye-display` consumer to your cluster, run the following
    command:
 
      ```
-     kubectl --namespace event-example apply --filename - << END
+     kubectl -n event-example apply -f - << EOF
      apiVersion: apps/v1
      kind: Deployment
      metadata:
@@ -133,12 +133,12 @@ demonstrate how you can configure your event producers to target a specific cons
        - protocol: TCP
          port: 80
          targetPort: 8080
-     END
+     EOF
      ```
 
 1. Verify that the event consumers are working by entering the following command:
      ```
-     kubectl --namespace event-example get deployments hello-display goodbye-display
+     kubectl -n event-example get deployments hello-display goodbye-display
      ```
    This lists the `hello-display` and `goodbye-display` consumers that you
    deployed:
@@ -158,7 +158,7 @@ Each trigger can specify a filter that enables selection of relevant events base
 
 1. Create a trigger by entering the following command:
    ```
-   kubectl --namespace event-example apply --filename - << END
+   kubectl -n event-example apply -f - << EOF
    apiVersion: eventing.knative.dev/v1
    kind: Trigger
    metadata:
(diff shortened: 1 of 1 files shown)


abrennan89:
Thanks @mpetason 
I have a few formatting issues still to fix before I merge, but thanks for reviewing so fast :slightly_smiling_face: 
https://docs-2662.default.docs-on-the-rocks.io/docs/eventing/getting-started/ (preview)

docs/eventing/getting-started.md
```diff
@@ -155,27 +93,25 @@ specific consumer.
...
-     metadata:
-       name: hello-display
-     spec:
-       selector:
-         app: hello-display
-       ports:
-       - protocol: TCP
-         port: 80
-         targetPort: 8080
+
```
mattmoor, line 96: Remove trailing whitespace:
```suggestion

```

docs/eventing/getting-started.md
```diff
@@ -1,128 +1,66 @@
...
 
-1. If `READY=False`, wait 2 minutes and re-run the command. If you continue to
-   receive the `READY=False`, see the [Debugging Guide](./debugging/README.md)
-   to help troubleshoot the issue.
+  ```
+  NAME      READY   REASON   URL                                                        AGE
+  default   True             http://default-broker.event-example.svc.cluster.local      1m
+  ```
 
-Now that your `Broker` is ready to manage events, you can create and configure
```
vaikas, line 124: I really think this 2 minutes comes from previous incarnations where there were many resources being created, etc. If the Broker doesn't come ready rerunning the command doesn't help anything. I would maybe just say that if Broker doesn't come ready (in the order that all of our resources come, serving included), we should jump to debugging guide. I don't know how we word those in all our resources, point is just that nobody should wait for 2 minutes, and re-running won't fix anything :) I'll stop yammering :)

abrennan89: Makes sense to me, I've removed the mention of 2 mins now.

vaikas: I think you removed it from different place? Still shows it here? :)

docs/eventing/getting-started.md
```diff
@@ -213,37 +138,28 @@ specific consumer.
...
    This lists the `hello-display` and `goodbye-display` consumers that you
    deployed:
-
-   ```sh
+   ```
    NAME              READY   UP-TO-DATE   AVAILABLE   AGE
    hello-display     1/1     1            1           26s
    goodbye-display   1/1     1            1           16s
    ```
+   The number of replicas in the **READY** column should match the number of replicas in the **AVAILABLE** column, which might take a few minutes.
```
vaikas, line 152: Yikes, again here, nothing should be taking a few minutes :) Maybe like few seconds.

vaikas:
Thanks for working on cleaning these up!

docs/eventing/getting-started.md
```diff
@@ -1,140 +1,70 @@
...
-   This shows the `Broker` that you created:
+  ```
+  kubectl create -f - <<EOF
+  apiVersion: eventing.knative.dev/v1
+  kind: Broker
+  metadata:
+   name: default
+   namespace: event-example
+  EOF
+  ```
```
lionelvillard: The indentation is off here.

docs/eventing/getting-started.md
abrennan89: Thanks :)

docs/eventing/getting-started.md
abrennan89: Yeah I'm having to do a bit of messing around with it to stop it breaking how the site renders for some reason. I'll get you to take a look over the samples when I feel like I'm finished if that's cool so you can check it's technically correct as well as rendering OK?

lionelvillard: sounds good!

docs/eventing/getting-started.md
```diff
@@ -1,163 +1,88 @@
...
 
-   This shows the `Broker` that you created:
+1. Verify that the broker is working correctly, by entering the following command:
 
-   ```sh
-   NAME      READY   REASON   URL                                                        AGE
-   default   True             http://default-broker.event-example.svc.cluster.local      1m
-   ```
+    ```
+    kubectl --namespace event-example get Broker default
```
vaikas, line 44: nit: Broker -> brokers
only to match the other examples (deployments for example below).

docs/eventing/getting-started.md
```diff
@@ -1,163 +1,88 @@
...
-           - name: event-display
-             # Source code: https://github.com/knative/eventing-contrib/tree/master/cmd/event_display
-             image: gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/event_display
-
-   ---
-
-   # Service pointing at the previous Deployment. This will be the target for event
-   # consumption.
+     ```
+     kubectl --namespace event-example apply --filename - << END
```
vaikas, line 66: L32 (broker creation uses (and I prefer)):
kubectl --namespace event-example apply -f - << EOF

So, using the EOF as the limiter (for consistency)
using -f instead of --filename (shorter to type, and for consistency).

docs/eventing/getting-started.md
vaikas, line 44: nit number 2.  For folks that are new, might as well give them the shorthand for the namespace:
-n instead of --namespace

Here and throughout.
if we everywhere else always use --namespace, fine :)

docs/eventing/getting-started.md
lionelvillard: would it makes sense to document writing style conventions somewhere?

docs/eventing/getting-started.md
```diff
@@ -285,53 +197,35 @@ events based on the Cloud Event context attributes.
...
+    If this value looks incorrect, see the [Debugging Guide](./debugging/README.md) to troubleshoot the issue.
 
-In this section you will create an event producer that you can use to interact
-with the Knative Eventing subcomponents you created earlier. Most events are
-created systematically, but this guide uses `curl` to manually send individual
-events and demonstrate how these events are received by the correct event
-consumer. Because you can only access the `Broker` from within your Eventing
-cluster, you must create a `Pod` within that cluster to act as your event
-producer.
+## Creating a pod as an event producer
```
lionelvillard, line 221: I'm hoping we can simplify this section. It's too low level for most users and involves a lot of steps just to send an event. I don't know if today we have a reasonable alternative but I'm hoping at some point `kn` will (or is?) support sending events.

Note that I'm not suggesting to totally get rid on this section, instead I'm suggesting to provide a simpler alternative. We could use two tabs, one "with curl", and the other one "with kn".

vaikas:
Cleaning the creation of the pods creation, etc. is perhaps something that we can do as a follow up? @lionelvillard

docs/eventing/getting-started.md
abrennan89: Yep, we're planning to do this in the docs WG just haven't had bandwidth thus far.

abrennan89:
@vaikas I would appreciate that :slightly_smiling_face: 
I was just trying to do a quick cleanup of this doc, I don't doubt that it will need a lot more work, since things like the default broker may also be outdated already, but tidying it up is a start :slightly_smiling_face: we can always do more follow up PRs but I'd like to get this one merged asap.

PR #2677, opened 2020-07-16
Change API Server Source title in the sources table

devguyio:
## Proposed Changes 

- Add a more descriptive name of the API Server source entry in the sources table
- Point to the example of the API Server Source instead of the k8s type source code which was confusing for a user trying to use the API Server Source


lionelvillard:
Thanks! However this page is generated and we are also revamping it completely. More info soon (hopefully)

PR #2682, opened 2020-07-17
Document how to set default for containerConcurrency

julz:
It is possible to default the hard limit (containerConcurrency), just not via config-autoscaler.yaml.

/assign @markusthoemmes

### DIFF
--- a/docs/serving/autoscaling/concurrency.md
+++ b/docs/serving/autoscaling/concurrency.md
@@ -76,16 +76,18 @@ spec:
 
 The hard limit is always specified [per revision](./autoscaling-concepts.md) via the `containerConcurrency` field on the revision spec. This particular per-revision setting is not an annotation; it is present on the revision's spec itself as `containerConcurrency`.
 
-There is no global setting for the hard limit in the autoscaling config map, but a default value can be set for the Revision's `containerConcurrency` field in `config-defaults.yaml`.
+There is no global setting for the hard limit in the autoscaling config map (because `containerConcurrency` has implications outside of autoscaling, such as on buffering and queuing requests), but a default value can be set for the Revision's `containerConcurrency` field in `config-defaults.yaml`.
 
 * The default value is `0`, meaning that there is no limit on the number of requests that are allowed to flow into the revision.
 * A value greater than `0` specifies the exact number of requests that are allowed to flow to the replica at any one time.
 
-* **Global key:** No global key (but a default can be set for the per-revision spec key in `config-defaults.yaml`).
+* **Global key:** container-concurrency (in `config-defaults.yaml`)
 * **Per-revision spec key:** `containerConcurrency`
 * **Possible values:** integer
 * **Default:** `0`, meaning no limit
 
+{{< tabs name="container-concurrency" default="Per Revision" >}}
+{{% tab name="Per Revision" %}}
 **Example:**
 ```yaml
 apiVersion: serving.knative.dev/v1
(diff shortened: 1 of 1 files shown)


docs/serving/autoscaling/concurrency.md
```diff
@@ -74,13 +74,14 @@ spec:
...
-This particular per-revision setting is not an annotation; it is present on the revision's spec itself as `containerConcurrency`.
+The hard limit is always specified [per revision](./autoscaling-concepts.md) via the `containerConcurrency` field on the revision spec. This particular per-revision setting is not an annotation; it is present on the revision's spec itself as `containerConcurrency`.
+
+There is no global setting for the hard limit in the autoscaling config map, but a default value can be set for the Revision's `containerConcurrency` field in `config-defaults.yaml`.
 
 * The default value is `0`, meaning that there is no limit on the number of requests that are allowed to flow into the revision.
 * A value greater than `0` specifies the exact number of requests that are allowed to flow to the replica at any one time.
 
-* **Global key:** No global key.
+* **Global key:** No global key (but a default can be set for the per-revision spec key in `config-defaults.yaml`).
```
markusthoemmes, line 84: Hmm, let's document the global key in the other configmap then and add an example. We can just add a remark that it's in a different CM.

You might want to add that it's not in the AS configMap because it touches more aspects (namely queuing and buffering which are somewhat concerns of networking) which is also why this isn't just yet another autoscaling annotation but an integral part of the spec IIRC

docs/serving/autoscaling/concurrency.md
```diff
@@ -97,6 +100,19 @@ spec:
     spec:
       containerConcurrency: 50
 ```
+{{< /tab >}}
+{{% tab name="Global (Defaults ConfigMap)" %}}
```
markusthoemmes, line 104: Operator example too? :pray:

docs/serving/autoscaling/concurrency.md
julz: okie, gave that a go

docs/serving/autoscaling/concurrency.md
julz: heh, never actually used the operator ☺️, but took an educated guess

docs/serving/autoscaling/concurrency.md
```diff
@@ -74,17 +74,20 @@ spec:
 
 ### Hard limit
 
-The hard limit has no global setting and can only be specified [per revision](./autoscaling-concepts.md).
-This particular per-revision setting is not an annotation; it is present on the revision's spec itself as `containerConcurrency`.
+The hard limit is always specified [per revision](./autoscaling-concepts.md) via the `containerConcurrency` field on the revision spec. This particular per-revision setting is not an annotation; it is present on the revision's spec itself as `containerConcurrency`.
```
abrennan89, line 77:
```suggestion
The hard limit is specified [per Revision](./autoscaling-concepts.md) using the `containerConcurrency` field on the Revision spec. This setting is not an annotation.
```

docs/serving/autoscaling/concurrency.md
```diff
@@ -74,17 +74,20 @@ spec:
...
+
+There is no global setting for the hard limit in the autoscaling config map (because `containerConcurrency` has implications outside of autoscaling, such as on buffering and queuing of requests), but a default value can be set for the Revision's `containerConcurrency` field in `config-defaults.yaml`.
```
abrennan89, line 79:
```suggestion
There is no global setting for the hard limit in the autoscaling ConfigMap, because `containerConcurrency` has implications outside of autoscaling, such as on buffering and queuing of requests. However, a default value can be set for the Revision's `containerConcurrency` field in `config-defaults.yaml`.
```

julz:
Thanks for the suggestions @abrennan89! Applied them both.

julz:
Uh-oh, looks like I confused the google cla gods by using the GitHub interface :(. Let me try a rebase.

abrennan89:
/lgtm
/approve

Thanks @julz :slightly_smiling_face:

PR #2698, opened 2020-07-24
Update the cm name, ChannelBasedBroker -> MTChannelBasedBroker

vaikas:

## Proposed Changes 

- Change the CM name from imc-channel (does not exist) to config-br-default-channel
- Change ChannelBasedBroker -> MTChannelBasedBroker
-


vaikas:
Markdown link check fails with this existing failure:
```
docs/install/any-kubernetes-cluster.md\u0026#xA;	ERROR	http://xip.io\u0026#xA;		Lookup xip.io on 10.7.240.10:53: no such host\u0026#xA;
```

Issue #2702, opened 2020-07-28
Creating a private cluster-local service(docs/serving/cluster-local-route.md)


ajtomato:
## Expected Behavior

To label a Kubernetes service:

kubectl label __svc__ ${SERVICE_NAME} serving.knative.dev/visibility=cluster-local

## Actual Behavior

To label a Kubernetes service:

kubectl label __route__ ${SERVICE_NAME} serving.knative.dev/visibility=cluster-local

## Steps to Reproduce the Problem

None

## Additional Info

None

abrennan89:
@ajtomato can you please provide a link to where you're seeing this in the docs?
Thanks!

ajtomato:
https://knative.dev/docs/serving/cluster-local-route/

nak3:
Thank you @ajtomato 
Fixing it https://github.com/knative/docs/pull/2710.

/kind networking

PR #2710, opened 2020-07-29
Label to k8s service instead of route

nak3:
Fixes https://github.com/knative/docs/issues/2702

## Proposed Changes 

- This patch fixes wrong resource name to label. It should be `service` instead of `route` under the `To label a Kubernetes service:`.

/cc @abrennan89 @ajtomato


abrennan89:
/lgtm
/approve

Thanks @nak3 !

PR #2718, opened 2020-07-30
Switch from dep to go mod

chizhg:
## Proposed Changes 
- `docs` repo is the only one that is still on `dep` (other Knative repos have all been switched to go mod), this PR switches this repo to be also on go mod.


mpetason:
There are a lot of updates in this, @evankanderson do you want to take a look? I understand what the update is doing but I'm not sure if it will break anything.

hack/tools.go
```diff
@@ -11,3 +14,10 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
+*/
+
+package tools
+
+import (
+    _ "knative.dev/test-infra/scripts"
```
mattmoor, line 22: Format Go code:
```suggestion
	_ "knative.dev/test-infra/scripts"
```

PR #2720, opened 2020-07-31
Add docs for initial scale

taragu:
Fixes https://github.com/knative/serving/issues/8826

## Proposed Changes 
- Add docs for initial scale

/cc @julz @vagababov @markusthoemmes

### DIFF
--- a/docs/serving/autoscaling/scale-bounds.md
+++ b/docs/serving/autoscaling/scale-bounds.md
@@ -9,8 +9,6 @@ To apply upper and lower bounds to autoscaling behavior, you can specify scale b
 
 You can also specify the initial target scale of a revision after creation, with either its cluster-wide flag or annotation.
 
-If both `minScale` annotation and `initial-scale` ConfigMap value or `autoscaling.knative.dev/initialScale` annotation are specified, and `minScale` is greater than the calculated initial scale, the initial scale will be ignored.
-
 ## Lower bound
 
 This value controls the minimum number of replicas that each revision should have.
@@ -77,7 +75,7 @@ spec:
 
 ## Initial scale
 
-This value controls the initial target scale of a revision after creation.
+This value controls the initial target scale of a revision after creation. If lower bound is set and is greater than the initial scale, the initial scale will be ignored.
 
 * **Global key:** `initial-scale` in combination with `allow-zero-initial-scale`
 * **Per-revision annotation key:** `autoscaling.knative.dev/initialScale`
--- a/docs/serving/autoscaling/scale-bounds.md
+++ b/docs/serving/autoscaling/scale-bounds.md
@@ -75,9 +75,9 @@ spec:
 
 ## Initial scale
 
-This value controls the initial target scale of a revision after creation. This is the target scale a revision must reach, immediately after first creation, before being marked Ready. Once the revision has reached this scale once, this value is ignored (e.g. the revision will scale down after the initial target scale is reached if the traffic received only needs a smaller scale).
+This value controls the initial target scale of a revision after creation. This is the target scale a revision must reach, immediately after creation, before being marked Ready. Once the revision has reached this scale once, this value is ignored (e.g. the revision will scale down after the initial target scale is reached if the traffic received only needs a smaller scale).
 
-If lower bound is set and is greater than the initial scale, the initial scale will be ignored.
+During creation the revision, the larger of initial scale and lower bound is picked as the target scale.
 
 * **Global key:** `initial-scale` in combination with `allow-zero-initial-scale`
 * **Per-revision annotation key:** `autoscaling.knative.dev/initialScale`
--- a/docs/serving/autoscaling/scale-bounds.md
+++ b/docs/serving/autoscaling/scale-bounds.md
@@ -77,7 +77,7 @@ spec:
 
 This value controls the initial target scale of a revision after creation. This is the target scale a revision must reach, immediately after creation, before being marked Ready. Once the revision has reached this scale once, this value is ignored (e.g. the revision will scale down after the initial target scale is reached if the traffic received only needs a smaller scale).
 
-During creation the revision, the larger of initial scale and lower bound is picked as the target scale.
+During the creation of the revision, the larger of initial scale and lower bound is picked as the target scale.
 
 * **Global key:** `initial-scale` in combination with `allow-zero-initial-scale`
 * **Per-revision annotation key:** `autoscaling.knative.dev/initialScale`
--- a/docs/serving/autoscaling/scale-bounds.md
+++ b/docs/serving/autoscaling/scale-bounds.md
@@ -7,7 +7,7 @@ type: "docs"
 
 To apply upper and lower bounds to autoscaling behavior, you can specify scale bounds in both directions.
 
-You can also specify the initial target scale of a revision after creation, with either its cluster-wide flag or annotation.
+You can also specify the initial scale a revision is scaled to immediately after creation, either as a default for all revisions, or for a specific revision using an annotation.
 
 ## Lower bound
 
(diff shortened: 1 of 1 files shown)


docs/serving/autoscaling/scale-bounds.md
```diff
@@ -7,6 +7,10 @@ type: "docs"
 
 To apply upper and lower bounds to autoscaling behavior, you can specify scale bounds in both directions.
 
+You can also specify the initial target scale of a revision after creation, with either its cluster-wide flag or annotation.
+
+If both `minScale` annotation and initial scale (global or per-revision) are specified, and `minScale` is greater than the calculated initial scale, the initial scale will be ignored.
```
mpetason, line 12:
```suggestion
If both `minScale` and initial scale annotations (global or per-revision) are specified, and `minScale` is greater than the calculated initial scale, the initial scale will be ignored.
```

taragu: I just updated this to make it more explicit (instead of saying "global or per-revision" here)

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -9,7 +9,7 @@ To apply upper and lower bounds to autoscaling behavior, you can specify scale b
 
 You can also specify the initial target scale of a revision after creation, with either its cluster-wide flag or annotation.
 
-If both `minScale` annotation and initial scale (global or per-revision) are specified, and `minScale` is greater than the calculated initial scale, the initial scale will be ignored.
+If both `minScale` annotation and `initial-scale` ConfigMap value or `autoscaling.knative.dev/initialScale` annotation are specified, and `minScale` is greater than the calculated initial scale, the initial scale will be ignored.
```
mpetason, line 12:
```suggestion
If the `minScale` annotation and the `autoscaling.knative.dev/initialScale` annotation are both specified, and `minScale` is greater than the calculated initial scale, the initial scale will be ignored.
```

taragu: I just moved this sentence down because it makes more sense there. I also decided not to mention annotation, because it's not entirely accurate, since it's the combination of annotation value and cluster-wide value that determines the calculated initial scale.

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -71,4 +73,60 @@ spec:
 {{< /tab >}}
 {{< /tabs >}}
 
+## Initial scale
+
+This value controls the initial target scale of a revision after creation. If lower bound is set and is greater than the initial scale, the initial scale will be ignored.
```
julz, line 78: I think this could use an extra sentence to describe what initial scale actually does. Maybe something like:

```suggestion
This value controls the initial target scale of a revision after creation. This is the target scale a revision must reach, immediately after first creation, before being marked Ready. Once the revision has reached this scale once, this value is ignored.

If a lower bound is set, the revision must achieve this scale before being marked Ready, so if it is greater than the initial scale, the initial scale will effectively be ignored.
```

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -7,6 +7,8 @@ type: "docs"
 
 To apply upper and lower bounds to autoscaling behavior, you can specify scale bounds in both directions.
 
+You can also specify the initial target scale of a revision after creation, with either its cluster-wide flag or annotation.
```
julz, line 10:
```suggestion
You can also specify the initial scale a revision is scaled to immediately after creation, either as a default for all revisions, or for a specific revision using an annotation.
```

docs/serving/autoscaling/scale-bounds.md
taragu: Yes this would be very helpful. Should we also mention the revision will be scale down immediately after the initial target scale is reached if the received traffic only needs a smaller scale?

julz: Depends if we think "Once the revision has reached this scale once, this value is ignored" is enough for readers to guess this.. I think it might be, but also I could see any extra sentence being helpful to spell it out.

taragu:
@vagababov @markusthoemmes would you please take a look?

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -71,4 +73,62 @@ spec:
 {{< /tab >}}
 {{< /tabs >}}
 
+## Initial scale
+
+This value controls the initial target scale of a revision after creation. This is the target scale a revision must reach, immediately after first creation, before being marked Ready. Once the revision has reached this scale once, this value is ignored (e.g. the revision will scale down after the initial target scale is reached if the traffic received only needs a smaller scale).
```
vagababov, line 78:
```suggestion
This value controls the initial target scale of a revision after creation. This is the target scale a revision must reach, immediately after creation, before being marked Ready. Once the revision has reached this scale once, this value is ignored (e.g. the revision will scale down after the initial target scale is reached if the traffic received only needs a smaller scale).
```
I think creation is _first_ by definition :)

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -71,4 +73,62 @@ spec:
...
+
+If lower bound is set and is greater than the initial scale, the initial scale will be ignored.
```
vagababov, line 80: Perhaps we can rephrase like:
```
During creation the revision the larger of initial scale and lower bound is picked.
```

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -75,9 +75,9 @@ spec:
 
 ## Initial scale
 
-This value controls the initial target scale of a revision after creation. This is the target scale a revision must reach, immediately after first creation, before being marked Ready. Once the revision has reached this scale once, this value is ignored (e.g. the revision will scale down after the initial target scale is reached if the traffic received only needs a smaller scale).
+This value controls the initial target scale of a revision after creation. This is the target scale a revision must reach, immediately after creation, before being marked Ready. Once the revision has reached this scale once, this value is ignored (e.g. the revision will scale down after the initial target scale is reached if the traffic received only needs a smaller scale).
 
-If lower bound is set and is greater than the initial scale, the initial scale will be ignored.
+During creation the revision, the larger of initial scale and lower bound is picked as the target scale.
```
vagababov, line 80:
```suggestion
During the creation of the revision, the larger of initial scale and lower bound is picked as the target scale.
```
My text might have not been very precise :D

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -7,6 +7,8 @@ type: "docs"
 
 To apply upper and lower bounds to autoscaling behavior, you can specify scale bounds in both directions.
```
abrennan89, line 8:
```suggestion
You can configure upper and lower bounds to control autoscaling behavior.
```

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -7,6 +7,8 @@ type: "docs"
...
 
+You can also specify the initial scale a revision is scaled to immediately after creation, either as a default for all revisions, or for a specific revision using an annotation.
```
abrennan89, line 10:
```suggestion
You can also specify the initial scale that a Revision is scaled to immediately after creation.
This can be a default configuration for all Revisions, or for a specific Revision using an annotation.
```

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -7,6 +7,8 @@ type: "docs"
...
+
 ## Lower bound
 
 This value controls the minimum number of replicas that each revision should have.
```
abrennan89, line 14:
```suggestion
This value controls the minimum number of replicas that each Revision should have.
```

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -71,4 +73,62 @@ spec:
 {{< /tab >}}
 {{< /tabs >}}
 
+## Initial scale
+
+This value controls the initial target scale of a revision after creation. This is the target scale a revision must reach, immediately after creation, before being marked Ready. Once the revision has reached this scale once, this value is ignored (e.g. the revision will scale down after the initial target scale is reached if the traffic received only needs a smaller scale).
```
abrennan89, line 78:
```suggestion
This value controls the initial target scale a Revision must reach immediately after it is created before it is marked as `Ready`.
After the Revision has reached this scale one time, this value is ignored. This means that the Revision will scale down after the initial target scale is reached if the actual traffic received only needs a smaller scale.
```

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -71,4 +73,62 @@ spec:
...
+
+During the creation of the revision, the larger of initial scale and lower bound is picked as the target scale.
```
abrennan89, line 80:
```suggestion
When the Revision is created, the larger of initial scale and lower bound is automatically chosen as the target scale.
```

PR #2722, opened 2020-08-03
Added Docs related to excluding namespaces from the Knative Webhook

mpetason:
Documents how to exclude a specific namespace from the Knative Webhook

https://github.com/knative/pkg/commit/086ff43956418f32abedde99deb4fab952de69f5

## Proposed Changes 

- adds webhook-customizations page
- documents the label to use to exclude Knative namespaces from the webhook

### DIFF
--- a/docs/serving/webhook-customizations.md
+++ b/docs/serving/webhook-customizations.md
@@ -1,12 +1,12 @@
 ---
-title: "Webhook Customizations"
+title: "Exclude namespaces from the Knative webhook"
 weight: 70
 type: "docs"
 ---
 
-## Exclude Namespaces from Webhook
+The Knative webhook examines resources that are created, read, updated, or deleted. This includes system namespaces, which can cause issues during an upgrade if the webhook becomes non-responsive. Cluster administrators may want to disable the Knative webhook on system namespaces to prevent issues during upgrades.
 
-A namespace can use the label `webhooks.knative.dev/exclude` to bypass the Knative Webhook.
+You can configure the label `webhooks.knative.dev/exclude` to allow namespaces to bypass the Knative webhook.
 
 ``` yaml
 apiVersion: v1
@@ -16,5 +16,3 @@ metadata:
   labels: 
     webhooks.knative.dev/exclude: "true"
 ```
-
-By default the webhook will examine resources that are created/read/updated/deleted. This includes system namespaces which can cause issues during an upgrade if the webhook is not responding.  Cluster administrators may want to disable the Knative Webhook on system namespaces to prevent issues during upgrades.
--- a/docs/serving/webhook-customizations.md
+++ b/docs/serving/webhook-customizations.md
@@ -6,7 +6,7 @@ type: "docs"
 
 ## Exclude Namespaces from Webhook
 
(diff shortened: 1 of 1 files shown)


mpetason:
I'm looking for feedback on where this documentation should be added. I added a webhook-customizations page although I don't know if we'll end up with additional content. This may belong as a small section on a different page.

docs/serving/webhook-customizations.md
```diff
@@ -0,0 +1,20 @@
+---
+title: "Webhook Customizations"
+weight: 70
+type: "docs"
+---
+
```
abrennan89, line 6: Add some context information about why a user would want to /  need to do this?

docs/serving/webhook-customizations.md
```diff
@@ -0,0 +1,20 @@
...
+## Exclude Namespaces from Webhook
```
abrennan89, line 7:
```suggestion
## Exclude namespaces from the Knative webhook
```

docs/serving/webhook-customizations.md
```diff
@@ -0,0 +1,20 @@
...
+
+A namespace can use the label `webhooks.knative.dev/exclude` to bypass the Knative Webhook.
```
abrennan89, line 9:
```suggestion
You can configure the label `webhooks.knative.dev/exclude` to allow namespaces to bypass the Knative webhook.
```

docs/serving/webhook-customizations.md
```diff
@@ -0,0 +1,20 @@
...
+``` yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: knative-dev
+  labels: 
+    webhooks.knative.dev/exclude: "true"
+```
+
+By default the webhook will examine resources that are created/read/updated/deleted. This includes system namespaces which can cause issues during an upgrade if the webhook is not responding.  Cluster administrators may want to disable the Knative Webhook on system namespaces to prevent issues during upgrades.
```
abrennan89, line 20:
```suggestion
The Knative webhook examines resources that are created, read, updated, or deleted. This includes system namespaces, which can cause issues during an upgrade if the webhook becomes non-responsive. Cluster administrators may want to disable the Knative webhook on system namespaces to prevent issues during upgrades.
```

abrennan89: This should be the intro :slightly_smiling_face: move to the top of the doc

docs/serving/webhook-customizations.md
abrennan89: Actually I'd rename the doc 
"Webhook Customizations" -> "Exclude namespaces from the Knative webhook"
and then remove this ## heading entirely, since you're not really customizing, just disabling.

mpetason: Good idea. I was trying to set this up as a doc that may include other customizations. Trying to figure out if this is better as a single doc or added to something else. I'll update it like it is always going to be a solo doc and use the title you suggested.

abrennan89: Yeah totally know what you mean :slightly_smiling_face: I used to do this too, but I think it's actually better just to keep it specific and then if it needs to move into a bigger "webhooks" section or something later, it can. Thanks so much for making the changes!

abrennan89:
/lgtm
/approve

Looks good!

PR #2723, opened 2020-08-03
Changed filename from serving.yaml to serving-core.yaml in upgrade docs

### LINKED ISSUE #2669: Upgrading your installation(docs/install/upgrade-installation.md)
Since current version([v0.16.0](https://github.com/knative/serving/releases/)) does not release serving.yaml anymore, the [Performing the upgrade](https://knative.dev/docs/install/upgrade-installation/#performing-the-upgrade) part needs improving.

mpetason:
Fixes #2669 

## Proposed Changes 

- It looks like we moved from serving.yaml to serving-core.yaml so I modified the upgrade file path

### DIFF
--- a/docs/install/upgrade-installation.md
+++ b/docs/install/upgrade-installation.md
@@ -94,7 +94,7 @@ Knative Serving and Eventing components and the monitoring plug-in, the
 following command upgrades the installation to v0.16.0:
 
 ```bash
-kubectl apply --filename https://github.com/knative/serving/releases/download/v0.16.0/serving.yaml \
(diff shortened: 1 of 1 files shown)


nak3:
Although this is correct fix, the upgrade doc still misses "how to upgrade the plugins". As far as I know, the moving serving.yaml to serving-core.yaml dropped some plugin resources like net-istio, etc.

dprotaso:
/hold

good point @nak3 - we'll probably want a blurb about that

dprotaso:
actually - we could do that separately and unblock this

/unhold

nak3:
> actually - we could do that separately and unblock this

Yes, I think it is a little bit complicated task :cry:  We should do it in separated PR.

PR #2727, opened 2020-08-05
Update knative-api-specification-1.0.md

igorbelianski:
This document states : 
      “name	string       (Optional)	Must be one of "http1" or "h2c" (if supported). Defaults to "http1".

I was also following this document  Kantive spec for RevisionTemplateSpec here:
https://knative.dev/docs/reference/serving-api/#serving.knative.dev/v1.RevisionTemplateSpec
That documentation defers to k8s for the definition of  PodSpec Kubernetes core/v1.PodSpec. 
     documented here https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podspec-v1-core
  which refers to containers Container array 
     documented here  https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#container-v1-core
  which refers to ports ContainerPort array
      documented here https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#containerport-v1-core 
   which states:
name. string	If specified, this must be an IANA_SVC_NAME and unique within the pod. Each named port in a pod must have a unique name. Name for the port that can be referred to by services.

I think it is important to explicitly communicate that these 2 definition relate to each other (In my specific case I was working on https://github.com/knative/serving/issues/4283 and it is unclear where new values like "auto" can be added and whether keeping the IANA registered is important. )

I think it is important to express that  this name is a subset of values supported by k8s (must be an IANA_SVC_NAME) also currently the only 2 that may are supported by  Knative are  "http1" and "h2c"(optional)

## Proposed Changes 

- s /
      Must be one of "http1" or "h2c" (if supported). Defaults to "http1".
     /
      If specified, this must be an IANA_SVC_NAME. The only allowed values are "http1" or "h2c" (if supported). Defaults to "http1".
-
-

### DIFF
--- a/docs/serving/spec/knative-api-specification-1.0.md
+++ b/docs/serving/spec/knative-api-specification-1.0.md
@@ -2155,7 +2155,7 @@ Max: 1
 <br>
 (Optional)
    </td>
-   <td>Must be one of "http1" or "h2c" (if supported). Defaults to "http1".
+   <td>If specified, this must be an IANA_SVC_NAME. The only allowed values are "http1" or "h2c" (if supported). Defaults to "http1".
    </td>
    <td>REQUIRED
    </td>


mattmoor:
IANA registration is not important, `IANA_SVC_NAME` is defined as:

>  `An alphanumeric (a-z, and 0-9) string, with a maximum length of 15 characters, with the '-' character allowed anywhere except the first or the last character or adjacent to another '-' character, it must contain at least a (a-z) character`

mattmoor:
Given the fixed arity of choices here, I'm kinda 👎 on merging this because I feel like is just creates further confusions around what the IANA bits means (as the K8s docs seem to).

cc @dprotaso @tcnghia for thoughts

dprotaso:
If we only allow two specific values I don't think we should include a clause that adds constraints.

I opt for succinctness

igorbelianski:
Sorry for delay (was on vacation) . Completely agree on the goal of succinctness. I also see the point of new confusion being introduced by this PR. (I have had a few drafts of the sentence , and that's the least confusing version I could come up with )

Let me articulate motivations for this PR little bit differently.  Here is a risk / concern that I am trying to mitigate:

Scenario: 
Knative contributors design new features / functionality and introduce new const values (as an example "auto" proposed here : ([https://github.com/knative/serving/issues/4283])
Knative community  proceeds with the proposal of introducing "auto" which may  be already a registered IANA entry with potentially different meaning  / interpretation ( or it may not be registered yet, but may get a new registration by 3rd party in the future with unrelated definition meaning). I also agree with Matt that IANA registration is not a requirement , but IMO it is clearly a worthwhile "safety guardrail " that we may try to leverage.

@mattmoor @dprotaso   Given this argument, let me know if you

a) still think this concern is not worth mitigating 
b) there is a different / better mitigation ( may be some internal Knative docs that comunity contributors are not likely to miss?)

mattmoor:
> proceeds with the proposal of introducing "auto" which may be already a registered IANA entry with potentially different meaning / interpretation

My point above was that registration is not a requirement, rather that IANA is being used descriptively to characterize the validation performed.  When placed in users hands in most other contexts these are non-semantic free-form strings, so IANA registration is sort of a silly requirement (given no semantics).

Our use of these semantically is an unfortunate wart we picked up from Istio.  I think that if we could have had "auto" all alone we simply wouldn't have attached semantics to port names at all, and this would never have been an issue.

Alas we are here.  For this user-facing documentation, I'd rather leave this as-is (expanded to encompass any new members of the enumeration, as those are available).

Sounds like Dave agrees (but is on PTO).  Sorry for the delayed response, been juggling a few things.

Issue #2734, opened 2020-08-08
docs for helloworld-go is not rendered correctly


xinydev:
## Expected Behavior

Yaml file was not truncated in docs

## Actual Behavior

the sentence in yaml is rendered as head in hugo website

## Steps to Reproduce the Problem 

1.  https://knative.dev/docs/eventing/samples/helloworld/helloworld-go/   Bottom of this page
1. https://github.com/knative/docs/tree/master/docs/eventing/samples/helloworld/helloworld-go

PR #2739, opened 2020-08-11
Fix rendering issue in helloworld-go eventing

nak3:
Fixes https://github.com/knative/docs/issues/2734

## Proposed Changes 

This patch adds extra indents to the code blockl with `---`.

I am not 100% sure if it fixes the issue on website, but helloworld-python has 
the same block and it renders perfectly. So it seems the indent issue.

https://knative.dev/development/eventing/samples/helloworld/helloworld-go/ (wrong rendering)
https://knative.dev/development/eventing/samples/helloworld/helloworld-python/ (correct rendering)

### DIFF
--- a/docs/eventing/samples/helloworld/helloworld-go/README.md
+++ b/docs/eventing/samples/helloworld/helloworld-go/README.md
@@ -137,74 +137,74 @@ cd knative-docs/docs/eventing/samples/helloworld/helloworld-go
    definition into the file. Make sure to replace `{username}` with your Docker
    Hub username.
 
-   ```yaml
-   # Namespace for sample application
-   apiVersion: v1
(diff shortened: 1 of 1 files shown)


mpetason:
It does seem like this is how the other doc resolved the issue, maybe --- on the same indention was breaking it? I'll add a lgtm tag and wait to see if someone else has any opinions.

/lgtm

nak3:
> It does seem like this is how the other doc resolved the issue, maybe --- on the same indention was breaking it?

I just assumed it from https://knative.dev/development/eventing/samples/helloworld/helloworld-python/, but I guess that's true :smile:

lionelvillard:
should rendered fine. 
/approve

PR #2764, opened 2020-08-20
Fix rendering issue in Apache Kafka Broker page

pierDipi:
## Proposed Changes 

- Fix rendering issue


pierDipi:
/cc @abrennan89 

Thie PR needs to be cherry-picked to 0.17 but there is no label for it.
/label needs-cherry-pick

pierDipi:
Ok, I'm not sure why we have only those labels available. :confused:

PR #2770, opened 2020-08-21
Update upgrade-installation.md

yuchunyu97:


yuchunyu97:
@googlebot I signed it!

markusthoemmes:
Why would we keep updating this? We're only giving an example here after all :thinking:

PR #2775, opened 2020-08-22
Fix default installation without sidecar injection

guillaumeblaquiere:
The default installation doesn't work as described in the documentation.

I fixed it like this

 * manifest apply no longer exists in istioctl CLI.
 * Add runAsRoot to allow targetPort creation bellow 1024.

### DIFF
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -86,12 +86,14 @@ spec:
             - port: 15020
               name: status-port
             - port: 80
+              targetPort: 8080
               name: http2
             - port: 443
+              targetPort: 8443
               name: https
 EOF
 
-istioctl manifest install -f istio-minimal-operator.yaml --set values.gateways.istio-ingressgateway.runAsRoot=true
+istioctl manifest install -f istio-minimal-operator.yaml
 ```
 
 #### Installing Istio with sidecar injection
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -91,7 +91,7 @@ spec:
               name: https
 EOF
 
-istioctl manifest apply -f istio-minimal-operator.yaml
+istioctl manifest install -f istio-minimal-operator.yaml --set values.gateways.istio-ingressgateway.runAsRoot=true
 ```
 
 #### Installing Istio with sidecar injection
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -93,7 +93,7 @@ spec:
               name: https
 EOF
 
-istioctl manifest install -f istio-minimal-operator.yaml
+istioctl install -f istio-minimal-operator.yaml
 ```
 
 #### Installing Istio with sidecar injection


docs/install/installing-istio.md
```diff
@@ -91,7 +91,7 @@ spec:
               name: https
 EOF
 
-istioctl manifest apply -f istio-minimal-operator.yaml
+istioctl manifest install -f istio-minimal-operator.yaml --set values.gateways.istio-ingressgateway.runAsRoot=true
```
JRBANCEL, line 94: Why do we need this?
Last time I installed Istio (1.6.x), it opens ports in the 1024+ range and the service maps the real port to it.
For example, port 80 is mapped to 8080.
Is this not the case in 1.5?

guillaumeblaquiere: I tested early in august, with version 1.6.x it worked (I prepared a demo for the GDG Cloud Melbourne). Recently, I tested the 1.7 version (for a demo to the GDG Nantes (a city in France), tonight: 6pm CEST) and it didn't work. I fixed like this. There might be a better/nicer way.

julz: better than runAsRoot is to modify the configuration so that the requested `targetPorts` are high enough not to require root. Modifying the `ports` section of Istio-minimal-operator.yaml to look like the below should work with 1.7:

~~~~
            - name: status-port
              port: 15021
              targetPort: 15021
            - name: http2
              port: 80
              targetPort: 8080
            - name: https
              port: 443
              targetPort: 8443
~~~~

guillaumeblaquiere: @julz, I committed a new version thank to your comment after having testing it!

upodroid:
I spent several hours debugging around till I noticed cluster-local-gateway can't bind to priveleged ports.

Fresh knative install with istio 1.7

Can we get this merged ASAP?

nak3:
@JRBANCEL @tcnghia @ZhiminXiang  I was also asked about the status of Istio 1.7 for Knative. I am wondering that we should merge this for 1.7 and put a note about the issue we have seen in CI? (I am happy to send a follow up PR.) We may also want to update istio-latest test to 1.7, though.

tcnghia:
+1 on that.

I think at this point I'd go with testing 1.7 in `latest` with reduced parallelism in place. `stable` will still be older but at least we have some coverage.

julz:
FWIW we've seen some big improvements in our load testing of knative with istio 1.7 with this fix applied: https://github.com/istio/istio/pull/26918. It's not in 1.7, but it should be in 1.7.1 in a few days. The tl;dr is there was a race when Endpoints get created before Service that can sometimes cause the istio endpoint cache never to get updated, and this situation seems to occur in knative semi-regularly under load. _Maybe_ this may even fix the problems at higher parallelism, if we're lucky, but either way - fyi.

guillaumeblaquiere:
Any updates on this? Any helps/comments that I can bring to you?

nak3:
I was thinking that we should wait for start testing Istio 1.7.1 which will include https://github.com/istio/istio/pull/26985 but maybe we don't need to wait for it?

/lgtm
/assign @tcnghia @ZhiminXiang @JRBANCEL 
/hold

ZhiminXiang:
I think I may miss some context. What Istio versions could this PR apply to?

nak3:
It depends on `istioctl` version which users use. But without this PR, installation of istio 1.7 always fail. So this PR is kind of applying to Istio 1.7.

JRBANCEL:
I don't think we need to start testing 1.7 since with or without this change Istio until 1.6 included works and this unblock 1.7.

I don't see any risk or downside of this change (except that i might create some downtime but even Istio doesn't seem to care about this :disappointed:)

docs/install/installing-istio.md
```diff
@@ -86,12 +86,14 @@ spec:
...
             - port: 80
+              targetPort: 8080
               name: http2
             - port: 443
+              targetPort: 8443
               name: https
 EOF
 
-istioctl manifest apply -f istio-minimal-operator.yaml
+istioctl manifest install -f istio-minimal-operator.yaml
```
howardjohn, line 96: I think it's just `istioctl install`? Unless there is an alias I am not familiar with

nak3: Yeah, it seems `istioctl manifest install` was forgotten to remove. I am trying to change to remove it https://github.com/istio/istio/pull/27277 so we should use `istioctl install` instead.

guillaumeblaquiere: I removed the manifest in this PR

tcnghia:
/unhold

(there is no objection to JR suggestion https://github.com/knative/docs/pull/2775#issuecomment-689997066 )

tcnghia (approved):
/lgtm

PR #2784, opened 2020-08-28
fix the min k8s version

ZhiminXiang:
## Proposed Changes 

- Fix the min k8s version. It should be 1.16


tnir:
When would the production docs be updated?

I still see `1.16` as the mimimum version of Kubernetes at https://knative.dev/docs/install/any-kubernetes-cluster/#before-you-begin

PR #2805, opened 2020-09-11
Try cloudevents-go with kn

mattmoor:
/hold


mattmoor:
With a local render of the docs:

image

cc @rhuss @maximilien @abrennan89

mattmoor:
I'm going to try staging the helloworld-go sample next.  Let's figure out the shape we want on these, then can "go broad" with the other languages in each of the respective categories.  Then we can go deep on the more focused samples.

WDYT?

matzew:
I think this does look great, @mattmoor

rhuss:
@mattmoor that looks nice, I think this is the way to go offer different alternatives for the same thing. Of course the next question will be what the default tab will be ;-) but this can be discussed on the hello world example.

mattmoor:
I went with `kn` here, and (personally) I'd be happy to bias towards that, but if it'll be at all contentious, let's punt it for a follow-up.

slinkydeveloper:
Cool! Can you do the same with other cloudevents-* examples?  
/lgtm

mattmoor:
@slinkydeveloper I wanted to iterate on a single example of this and `helloworld-go` (stacked PR) as a template and then I was going to put together a spreadsheet to try and crowd source this to scale it horizontally.

Some thing will vary (e.g. `ko` here) and I'm not sure I'm the best to add that nuance in every language, but it should be easier to knock out than the original bootstrap.

docs/serving/samples/cloudevents/cloudevents-go/index.md
```diff
@@ -5,4 +5,159 @@ weight: 1
 type: "docs"
 ---
 
-{{% readfile file="README.md" %}}
+A simple web app written in Go that can receive and send Cloud Events that you
```
n3wscott, line 8: CloudEvents

n3wscott:
one little nit, but /lgtm

slinkydeveloper:
@mattmoor sgtm

docs/serving/samples/cloudevents/cloudevents-go/index.md
mattmoor: That's not really this PR, so you should check all the languages

n3wscott: ok lgtm

n3wscott:
/unhold

let it in, looks good

PR #2820, opened 2020-09-16
GitHubSource is now in the knative-sandbox org.

lionelvillard:
## Proposed Changes 

- Change links pointing to the GitHub source
-
-


lionelvillard:
not needed anymore

PR #2828, opened 2020-09-17
Fix the doc layout so it actually renders the whole file

vagababov:
Now it's cut after hard limit

As discovered in https://github.com/knative/serving/issues/9446

/assign @abrennan89 @mattmoor


tcnghia (approved):
/lgtm

PR #2835, opened 2020-09-21
Fix alignment in example output

pierDipi:
See: https://knative.dev/development/eventing/broker/kafka-broker/#installation

## Proposed Changes 

- Align example output


abrennan89 (approved):
Thanks @pierDipi !
/lgtm
/approve

PR #2840, opened 2020-09-22
Wrong sample link is wrong

matzew:
Signed-off-by: Matthias Wessendorf <mwessend@redhat.com>

The link to the `PingSource` sample is wrong.  Yes it's in the same directory, but the `.md` is wrong here (due to how the site works)

/assign @abrennan89


abrennan89:
image

I thought this looked fine, but the link is failing for pingsource specifically so idk if this is right.
@RichieEscarez would you mind taking a look to double check?

matzew:
I could use some link like: 
https://knative.dev/docs/eventing/sources/pingsource/

:man_shrugging:

mpetason:
I tested this locally too - so will be interested in the fix. I tried to find another example that worked - but it seems like everything else was going up a directory and then referencing sample or sources so "../sources/ping-source" - but I'm not great with markdown to know why that would matter. It links correctly in vscode.

mpetason:
Tagging an issue related to this PR 

fixes #2883

RichieEscarez:
The way these files have been restructured is not valid Markdown so therefore, the Markdown link checker will always fail. 
In other words, the markdown says that a pingsource folder does not exist. 

The doc builds were initially set up to ignore _index.md files but now that README.md files are being removed and content is being directly added in the _index.md files, the build needs to accommodate these files (this would only work if all _index.md files are converted - ie. remove READMEs and relocate their content).

Otherwise, content in _index.md files must not include the .md file extensions (which is not a valid Markdown link) and the Markdown link checker will always fail.

RichieEscarez (approved):
This works (see comment for details)

/lgtm

PR #2847, opened 2020-09-23
[Automated] Upgrade to latest dependencies

mattmoor:
Produced via:
```shell
./hack/update-deps.sh --upgrade
./hack/update-codegen.sh
```
/cc abrennan89 dprotaso vaikas
/assign abrennan89 dprotaso vaikas


dprotaso:
/lgtm

@mattmoor there's no `./hack/update-codegen.sh` so I wonder if your script isn't failing when it should

Issue #2845, opened 2020-09-23
Missing Installation guides


rajitha1998:
## Expected Behavior
* The install guidelines and Introduction docs should be loaded correctly
## Actual Behavior
* Some pages related to a first-time contributor is not working.

## Steps to Reproduce the Problem

1. Visit: [Contributing page](https://knative.dev/contributing/)
2. Try navigating to [install guides](https://knative.dev/contributing/docs/install/), [Getting Started with Knative Serving ](https://knative.dev/contributing/docs/serving/getting-started-knative-app) or [Getting Started With Eventing](https://knative.dev/contributing/docs/eventing/getting-started).

## Additional Info
![Screenshot from 2020-09-23 18-49-07](https://user-images.githubusercontent.com/36359382/94019992-03fa9480-fdd0-11ea-9ea8-5992bb502623.png)

mattmoor:
Indeed, those are all broken 😬

mattmoor:
I believe that's from here, and the links work in markdown form, so something we need to fix: https://github.com/knative/docs/blob/master/CONTRIBUTING.md#getting-started

/kind good-first-issue

PR #2870, opened 2020-09-29
Fix liks on CONTRIBUTING.md

shuheiktgw:
Fixes https://github.com/knative/docs/issues/2845

## Proposed Changes 

- As described in the issue, It seems that links on the CONTRIBUTING.md do not work. It seems that if we add `./` it works as it does on docs/README.md.


abrennan89 (approved):
/lgtm
/approve

PR #2871, opened 2020-09-30
Add global configuration for target-burst-capacity and some corrections

markusthoemmes:
## Proposed Changes 

As per title really. This setting has a global setting too and some of the info in this doc was outdated so gave it a scrub while at it.


julz (commented):
/lgtm

julz:
oops, too slow! cc @abrennan89 just in case you want to take a look over too

PR #2872, opened 2020-09-30
Add tested Istio version for clarity

markusthoemmes:
## Proposed Changes 

We've been asked for "which version of Istio is actually supported" multiple times so I feel like we should clarify this in our docs and update accordingly once new versions stabilize for us.

### DIFF
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -18,13 +18,13 @@ instructions.
 You need:
 
 - A Kubernetes cluster created.
-- [`istioctl`](https://istio.io/docs/setup/install/istioctl/) (v1.5.4 or later) installed.
+- [`istioctl`](https://istio.io/docs/setup/install/istioctl/) (v1.4.10 or later) installed.
 
 ## Supported Istio versions
 
-The current known-to-be-stable version of Istio tested in conjunction with Knative is **v1.5.7**.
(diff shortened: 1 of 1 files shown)


nak3:
/assign @ZhiminXiang @JRBANCEL 

I think @ZhiminXiang and @JRBANCEL are familiar with recent tested Istio version.

docs/install/installing-istio.md
```diff
@@ -20,6 +20,12 @@ You need:
 - A Kubernetes cluster created.
 - [`istioctl`](https://istio.io/docs/setup/install/istioctl/) (v1.5.4 or later) installed.
 
+## Supported Istio versions
+
+The current known-to-be-stable version of Istio tested in conjunction with Knative is **v1.5.7**.
+Versions in the 1.5 line above **v1.5.4** should generally be fine too.
+Versions above the 1.5 line are under test but have not stabilized yet.
```
mattmoor, line 27: It's actually 1.4.x, IIUC net-istio bumped stable before fixing the way serving pinned things.

cc @tcnghia @JRBANCEL

ZhiminXiang: yea. currently this is 1.4.10. The plan is to bump to Istio 1.7 soon.

markusthoemmes: Adjusted.

PR #2888, opened 2020-10-05
Update the links for operator docs

houshengbo:
## Proposed Changes 

- The internal links for operator docs should connect to the local MD files instead of public links.


evankanderson (commented):
/lgtm
/approve

PR #2890, opened 2020-10-06
Explicitly point out not installing ns-cert component for http01 challenge

ZhiminXiang:


docs/serving/using-auto-tls.md
```diff
@@ -154,6 +154,10 @@ and which DNS provider validates those requests.
 
 If you choose to use the mode of provisioning certificate per namespace, you need to install `networking-ns-cert` components.
 
+**Note that provisioning certificate per namespace only works with DNS-01
+ challenge. If you choose to use HTTP-01 challenge, you must NOT install this
```
nak3, line 158: Just curious, ns-cert for HTTP-01 is under development? Or it will never support HTTP-01?

nak3:
/lgtm

Just one small question.

docs/serving/using-auto-tls.md
ZhiminXiang: It will never support HTTP01 as the namespace-level cert is wildcard cert which can only works with DNS01.

nak3: I see. Thank you.

docs/serving/using-auto-tls.md
```diff
@@ -154,6 +154,10 @@ and which DNS provider validates those requests.
...
+ component**
```
abrennan89, line 159:
```suggestion
**IMPORTANT:** Provisioning a certificate per namespace only works with DNS-01
 challenge. This component cannot be used with HTTP-01 challenge.
```

ZhiminXiang: Done

abrennan89 (changes requested):
Updated the format so it looks more like other notes :slightly_smiling_face: 
@RichieEscarez can we look into introducing some CSS formatting or something for the site to add notes that stand out visually? Something like https://www.w3schools.com/w3css/w3css_notes.asp

abrennan89:
@ZhiminXiang if you accept the suggestion I think you need to force push to fix the CLA thing, sorry :disappointed:

abrennan89 (approved):
/lgtm
/approve

RichieEscarez:
> Updated the format so it looks more like other notes 🙂
> @RichieEscarez can we look into introducing some CSS formatting or something for the site to add notes that stand out visually? Something like https://www.w3schools.com/w3css/w3css_notes.asp

Opened issue here to track that enhancement (also includes info for others to attempt the change): https://github.com/knative/website/issues/228

docs/serving/using-auto-tls.md
```diff
@@ -154,6 +154,9 @@ and which DNS provider validates those requests.
 
 If you choose to use the mode of provisioning certificate per namespace, you need to install `networking-ns-cert` components.
 
+**IMPORTANT:** Provisioning a certificate per namespace only works with DNS-01
+ challenge. This component cannot be used with HTTP-01 challenge.
+
```
RichieEscarez, line 159: @ZhiminXiang can you please also add that:

- HTTP-01 does not support provisioning a certificate per namespace

Up above within the list "1.  Using HTTP-01 challenge" (line 30) to make this clear at the time when choosing DNS-01 vs HTTP-01

ZhiminXiang: Done

abrennan89:
/lgtm
Thanks @ZhiminXiang !

PR #2897, opened 2020-10-07
Change event type in Sequence with Broker example

### LINKED ISSUE #2851: Page needs improvement
## Expected Behavior
I can finish the example as described.
https://knative.dev/docs/eventing/samples/sequence/sequence-with-broker-trigger/index.html

## Actual Behavior
It will be an endless loop between the broker and the sequence. Because the ksvc yaml doesn't set the TYPE env so it doesn't change the cloud event type. and the trigger to broker will be always triggered and the trigger to the display will be never triggered.

1.
2.
3.

## Additional Info
After set the TYPE env for step3 to "samples.http.mod3", the problem solved.

pierDipi:
If the event's type isn't changed by the last step in the
sequence the event will be resent to the Trigger that targets
the sequence creating an unwanted loop.

Fixes #2851

## Proposed Changes 

- Change event type in Sequence with Broker example

/label cherrypick-0.18
/label cherrypick-0.17
/label cherrypick-0.16

### DIFF
--- a/docs/eventing/samples/sequence/sequence-with-broker-trigger/README.md
+++ b/docs/eventing/samples/sequence/sequence-with-broker-trigger/README.md
@@ -79,7 +79,8 @@ spec:
           env:
             - name: MESSAGE
               value: " - Handled by 2"
-
+            - name: TYPE
+              value: "samples.http.mod3"
 ---
 
 ```


pierDipi:
I cannot add `cherrypick-*` labels :sob:

matzew (approved):
/lgtm
/approve

abrennan89:
/approve
/lgtm
Thanks @pierDipi - added the cherrypick labels but you need to do this manually I think
@mattmoor any hopes of getting something automated for this? :slightly_smiling_face: :crossed_fingers:

mattmoor:
@abrennan89 I think it would be interesting to set up an action for this, but I'm reluctant to try and build it until https://github.com/knative/community/issues/298 makes some progress.

pierDipi:
```
> Could not GET 'https://jcenter.bintray.com/org/slf4j/slf4j-api/1.7.25/slf4j-api-1.7.25.jar'. Received status code 504 from server: Gateway Time-out
```
/retest

PR #2904, opened 2020-10-07
Removed deprecated observability plugin

abrennan89:
Fixes #2804 

## Proposed Changes 
- Removed observability plugin from install docs

PR #2905, opened 2020-10-07
Removed deprecated observability plugin

abrennan89:
Relates to #2904


mattmoor:
Are we sure that there aren't any things referencing this?

abrennan89:
@mattmoor the monitoring bundle has been removed now for the 0.19 release, so I think it's suitable to remove this now.
There are some other references in docs for monitoring, but we will remove them all and Stavros has opened new issues to track creating replacement docs instead of leaving the old ones in there.

PR #2906, opened 2020-10-07
Do not add visibility label to Route

nak3:
Although https://knative.dev/docs/serving/cluster-local-route/
explains to add the visibility label to `Route`, the label on `Route` is
removed because Route takes over Ksvc's label.

Please see:
```
$ kubectl label route hello-example serving.knative.dev/visibility=cluster-local
route.serving.knative.dev/hello-example labeled

$ kubectl get rt --show-labels
NAME            URL                                        READY   REASON   LABELS
hello-example   http://hello-example.default.example.com   True             serving.knative.dev/service=hello-example
```

If users deployed the app by Route + Configuration instead of Ksvc
this makes sense but it is very rare (I think) and we should remove
this instruction.

/cc @ZhiminXiang @tcnghia

### DIFF
--- a/docs/serving/cluster-local-route.md
+++ b/docs/serving/cluster-local-route.md
@@ -38,12 +38,6 @@ To label the KService:
 kubectl label kservice ${KSVC_NAME} serving.knative.dev/visibility=cluster-local
 ```
 
-To label a route:
-
-```shell
-kubectl label route ${ROUTE_NAME} serving.knative.dev/visibility=cluster-local
-```
-
 To label a Kubernetes service:
 
 ```shell
--- a/docs/serving/cluster-local-route.md
+++ b/docs/serving/cluster-local-route.md
@@ -38,7 +38,7 @@ To label the KService:
 kubectl label kservice ${KSVC_NAME} serving.knative.dev/visibility=cluster-local
 ```
 
-To label a route when you used Route directory without KService:
+To label a route when you used Route directly without KService:
 
 ```shell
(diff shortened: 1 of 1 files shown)


tcnghia:
I think we probably want to clarify this use case instead? It is indeed very rare but Route+Configuration usage is still allowed with the API.

cc @dprotaso @mattmoor for their thoughts about this `ksvc`-less usage.

tcnghia:
Ping @mattmoor @dprotaso

mattmoor:
I'd clarify that folks should label the ksvc is using ksvc, but they can also label Route if used directly.

nak3:
Okay, i added the explanation. (I am still not sure about the use case of Route+Configuration :smile:  but we probably need to add it in another section in the doc.)

docs/serving/cluster-local-route.md
```diff
@@ -38,7 +38,7 @@ To label the KService:
 kubectl label kservice ${KSVC_NAME} serving.knative.dev/visibility=cluster-local
 ```
 
-To label a route:
+To label a route when you used Route directory without KService:
```
tcnghia, line 41:
```suggestion
To label a route when you used Route directly without KService:
```

nak3: Thank you! Updated.

dprotaso:
> I am still not sure about the use case of Route+Configuration

It allows for more advanced routing since you can have a route point to different configurations. 

ie. i have a config for every release branch of my application and a route that points to all of them using tags

tcnghia (approved):
/lgtm

PR #2915, opened 2020-10-08
[release-0.17] Change event type in Sequence with Broker example (#2897)

### LINKED ISSUE #2851: Page needs improvement

pierDipi:
If the type of the event isn't changed by the last step in the
sequence the event will be resent to the Trigger that targets
the sequence creating an unwanted loop.

Fixes #2851

Signed-off-by: Pierangelo Di Pilato <pierangelodipilato@gmail.com>

## Proposed Changes 

- Backport change event type in Sequence with Broker example (#2897)

### DIFF
--- a/docs/eventing/samples/sequence/sequence-with-broker-trigger/README.md
+++ b/docs/eventing/samples/sequence/sequence-with-broker-trigger/README.md
@@ -79,7 +79,8 @@ spec:
           env:
             - name: MESSAGE
               value: " - Handled by 2"
-
+            - name: TYPE
+              value: "samples.http.mod3"
 ---
 
 ```


abrennan89 (approved):
/lgtm
/approve

PR #2945, opened 2020-10-19
Bump golang versions to align all samples

markusthoemmes:
## Proposed Changes 

The eventing sample uses go 1.14 already and we're using it everywhere, across the board. This is mostly for consistency.


vagababov:
May be just 15 outright?

markusthoemmes:
@vagababov 0.15 isn't available at least on my distribution yet (Fedora), so I'd be happy if we're holding back just a little bit longer.

PR #2972, opened 2020-10-26
Remove reference to eventing-contrib

lionelvillard:

## Proposed Changes 

- Remove instructions about adding a source to eventing-contrib
- Add instructions on how to move a repo to knative-sandbox
-


lionelvillard:
@lberk Wanna take a quick look at this one?

lberk:
/lgtm
We might still want to mention that the new contributor will want to copy over a `hack/update-codegen.sh` and `hack/update-deps.sh` but we can save that for a follow up

PR #2989, opened 2020-10-31
Add sample logging configuration using fluentbit as collector and forwarder

evankanderson:
Fixes https://github.com/knative/eventing/issues/3805

## Proposed Changes 

- Add installation guide on how to install and configure Fluent Bit to copy logs from each node to a central location.

Does not yet document how to use this with minikube/docker-for-desktop/kind and a `hostPath` that's mounted from the local machine, but that would be a nice next step. [Duffie Cooley has a writeup for exposting host paths in Kind here](https://mauilion.dev/posts/kind-pvc-localdata/).


docs/install/collecting-logs/README.md
```diff
@@ -0,0 +1,87 @@
+This document describes how to set up [Fluent Bit](https://docs.fluentbit.io/),
+a log processor and forwarder, to collect your kubernetes logs in a central
+directory. This is not required for running Knative, but can be helpful with
+[Knative Serving](../serving), which will automatically delete pods (and their
+associated logs) when they are no longer needed. Note that Fluent Bit support
```
mpetason, line 5:
```suggestion
associated logs) when they are no longer needed. Note that Fluent Bit supports
```

docs/install/collecting-logs/README.md
```diff
@@ -0,0 +1,87 @@
...
+exporting to a number of other log providers; if you already have an existing
+log provider (for example, Splunk, Datadog, ElasticSearch, or Stackdriver), then
+you may only need the second part of setting up and configuring log forwarders.
```
mpetason, line 8: Maybe the second part can link to where you are talking about, to skip ahead?

docs/install/collecting-logs/README.md
```diff
@@ -0,0 +1,87 @@
...
+[Forwarder1]logs->[Collector]
+[Forwarder2]logs->[Collector]
+
+// Add notes
+[Collector]->[shared volume]
+[nginx]-[shared volume]
+-->
+
+The [`fluent-bit-collector.yaml`](./fluent-bit-collector.yaml) defines a
+StatefulSet as well as a Kubernetes Service which allows accessing reading the
```
mpetason, line 33:
```suggestion
StatefulSet as well as a Kubernetes Service which allows accessing and reading the
```

Not sure if you just wanted "accessing" or "reading", prob just "accessing"

docs/install/collecting-logs/README.md
evankanderson: Done.

evankanderson:
Updated.

mpetason:
Looks like there is some white space left. After that fix it looks great. Maybe we need someone to give it a technical review too?

evankanderson:
I think we could have someone give it a try-out besides me, but I was able to get this working on both a real K8s cluster and on docker-desktop, so I think there's something there.

I also have a followup describing how to use a shared volume between docker desktop and the collector... I'll need testers with minikube and kind as well.

docs/install/collecting-logs/README.md
evankanderson: Done.

docs/install/collecting-logs/README.md
evankanderson: Done.

evankanderson:
@mpetason any concerns about giving this an `/lgtm`? I've run the instructions on 3 clusters now: kubeadm, docker-desktop, and Azure.

mpetason:
Doesn't look like anyone else had comments. I think it looks good.

/lgtm

PR #2990, opened 2020-11-02
Better doc for channel config via operator

matzew:
Signed-off-by: Matthias Wessendorf <mwessend@redhat.com>

## Proposed Changes 

Adding a little bit more context for the different "default" channel configs.

The `current` text on master is incorrectly saying that the `` CM will configure the default channel. However it is only setting the deefault for what the configured broker uses (if we use a channel based broker).

Adding text that shows howto set the default channel via the `default-ch-webhook` configmap

### DIFF
--- a/docs/install/operator/configuring-eventing-cr.md
+++ b/docs/install/operator/configuring-eventing-cr.md
@@ -46,11 +46,44 @@ if the current Knative Eventing deployment is version 0.14.x, you must upgrade t
 The Knative Eventing operator CR is configured the same way as the Knative Serving operator CR. Because the operator manages
 the Knative Eventing installation, it will overwrite any updates to the `ConfigMaps` which are used to configure Knative
 Eventing. The `KnativeEventing` custom resource allows you to set values for these ConfigMaps via the operator. Knative
-Eventing has multiple ConfigMaps named with the prefix `config-`. The `spec.config` in `KnativeEventing` has one entry
+Eventing has multiple ConfigMaps, most of them are named with the prefix `config-`. The `spec.config` in `KnativeEventing` has one entry
 `<name>` for each ConfigMap named `config-<name>`, with a value which will be used for the ConfigMap's `data`.
 
+### Setting the default channel
+
 For example, if you would like to change your default channel from `InMemoryChannel` into `KafkaChannel`, here is what
-your Eventing CR looks like, to modify the ConfigMap `config-br-default-channel`:
+your Eventing CR looks like, to modify the ConfigMap `default-ch-webhook`:
+
+```
+apiVersion: operator.knative.dev/v1alpha1
+kind: KnativeEventing
+metadata:
+  name: knative-eventing
+  namespace: knative-eventing
+spec:
+  config:
+    default-ch-webhook:
+      default-ch-config: |
+        clusterDefault:
+          apiVersion: messaging.knative.dev/v1beta1
+          kind: KafkaChannel
+          spec:
+            numPartitions: 10
+            replicationFactor: 1
+        namespaceDefaults:
+          my-namespace:
+            apiVersion: messaging.knative.dev/v1beta1
+            kind: KafkaChannel
+            spec:
+              numPartitions: 10
+              replicationFactor: 1
+```
+
+The `clusterDefault` sets the global, cluster based default. Inside the `namespaceDefaults` you can configure the channel defaults on a per namespace base.
+
+### Setting the default channel for the broker
+
+For example, if you are using a Channel-based Broker and you would like to change the brokers default channel from `InMemoryChannel` into `KafkaChannel`, here is what your Eventing CR looks like, to modify the ConfigMap `config-br-default-channel`:
 
 ```
 apiVersion: operator.knative.dev/v1alpha1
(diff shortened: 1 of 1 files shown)


docs/install/operator/configuring-eventing-cr.md
```diff
@@ -46,11 +46,44 @@ if the current Knative Eventing deployment is version 0.14.x, you must upgrade t
...
+        namespaceDefaults:
+          my-namespace:
+            apiVersion: messaging.knative.dev/v1beta1
+            kind: KafkaChannel
+            spec:
+              numPartitions: 10
+              replicationFactor: 1
+```
+
+The `clusterDefault` sets the global, cluster based default. Inside the `namespaceDefaults` you can configure the channel defaults on a per namespace base.
```
vaikas, line 82: would it be useful to have a difference configuration here for the specified namespace? I wonder if it's confusing because they are both the same, so there's really no need to do this? Maybe we should have a mixed configuration, one using kafka and one using something else?

docs/install/operator/configuring-eventing-cr.md
```diff
@@ -46,11 +46,44 @@ if the current Knative Eventing deployment is version 0.14.x, you must upgrade t
...
+
+### Setting the default channel for the broker
+
+For example, if you are using a Channel-based Broker and you would like to change the brokers default channel from `InMemoryChannel` into `KafkaChannel`, here is what your Eventing CR looks like, to modify the ConfigMap `config-br-default-channel`:
```
vaikas, line 86: I'd take out the 'For example, ' in the beginning and just say something like:
If you are using a Channel-based Broker...

docs/install/operator/configuring-eventing-cr.md
matzew: true :-)

docs/install/operator/configuring-eventing-cr.md
matzew: ok. cool

vaikas:
/approve
@abrennan89 would you mind taking a look to ensure the wording makes sense?

docs/install/operator/configuring-eventing-cr.md
```diff
@@ -46,11 +46,44 @@ if the current Knative Eventing deployment is version 0.14.x, you must upgrade t
...
+            apiVersion: messaging.knative.dev/v1beta1
+            kind: KafkaChannel
+            spec:
+              numPartitions: 1
+              replicationFactor: 1
+```
+
+The `clusterDefault` sets the global, cluster based default. Inside the `namespaceDefaults` you can configure the channel defaults on a per namespace base.
+
+### Setting the default channel for the broker
```
antoineco, line 84: Shouldn't we capitalize object names such as Channel and Broker? Not sure how it's done in the rest of the docs, but I see mixed capitalization in this particular file.

matzew: I've changed that !

docs/install/operator/configuring-eventing-cr.md
abrennan89: @matzew do you plan to update one of these maybe to IMC?

matzew: yeah, I can do that

docs/install/operator/configuring-eventing-cr.md
```diff
@@ -46,11 +46,46 @@ if the current Knative Eventing deployment is version 0.14.x, you must upgrade t
...
+            apiVersion: messaging.knative.dev/v1
+            kind: InMemoryChannel
+            spec:
+              delivery:
+                backoffDelay: PT0.5S
+                backoffPolicy: exponential
+                retry: 5
+```
+
+The `clusterDefault` sets the global, cluster based default. Inside the `namespaceDefaults` you can configure the channel defaults on a per namespace base.
```
mpetason, line 84:
```suggestion
The `clusterDefault` sets the global, cluster based default. Inside the `namespaceDefaults` you can configure the channel defaults on a per namespace basis.
```

something like per need basis, or "you can configure the channel defaults per namespace as needed"

docs/install/operator/configuring-eventing-cr.md
```diff
@@ -46,11 +46,46 @@ if the current Knative Eventing deployment is version 0.14.x, you must upgrade t
...
+
+### Setting the default channel for the broker
+
+If you are using a channel-based broker and you would like to change the brokers default channel from `InMemoryChannel` into `KafkaChannel`, here is what your Eventing CR looks like, to modify the ConfigMap `config-br-default-channel`:
```
mpetason, line 88:
```suggestion
If you are using a channel-based broker and you would like to change the brokers default channel, from `InMemoryChannel` to `KafkaChannel`, here is an example of what your Eventing CR would look like when modifying the ConfigMap `config-br-default-channel`:
```

^ maybe that reads better. changing from to, instead of into maybe? Possibly would sound better above too.

abrennan89:
@matzew you'll need to push it again I think to get rid of the weird CLA issue.
@mpetason you should be able to lgtm this when you're ready to merge.

mpetason:
@googlebot I fixed it.

thisisnotapril:
Flipping CLA bit - commit authors are same person; diff emails

PR #3005, opened 2020-11-06
Add documentation on how to use OpenTelemetry +collector

evankanderson:
I'll dig out relevant issue numbers in the morning; this provides instructions for setting up the OpenTelemetry collector and Prometheus (in a fairly small footprint; probably around 250MB of RAM based on 200MB for Prometheus and 50MB for the otc-collector.

With this documentation, I'd like to start the clock ticking on migrating existing Prometheus and Stackdriver users to the OpenTelemetry collector so that we can drastically simplify the code in `pkg/metrics`.

## Proposed Changes 

- Add documentation on exporting statistics using the OpenTelemetry aggregator.

/assign @MontyCarter @mpetason


csantanapr:
I will review and try the instructions @evankanderson

evankanderson:
If you get stuck, feel free to ping me on Slack.

(Oh, I just realized I forgot to count the two operators in the overall RAM overhead... it looks like they add around 120MB, with the otel one being a bit bigger because it has two containers.)

docs/install/collecting-metrics/README.md
```diff
@@ -0,0 +1,101 @@
...
+[queue-proxy1]->[Collector]
+[queue-proxy2]->[Collector]
+[autoscaler]->[Collector]
+[Collector]<-scrape[Prometheus]
+-->
+
+1. The first step is to install the operator:
+
+   ```shell
+   kubectl apply --filename https://github.com/open-telemetry/opentelemetry-operator/releases/latest/download/opentelemetry-operator.yaml
```
csantanapr, line 31: When I run the command I get two warnings, I'm guessing there are optional CRs in the release yaml. harmless.
```
nable to recognize "https://github.com/open-telemetry/opentelemetry-operator/releases/latest/download/opentelemetry-operator.yaml": no matches for kind "Certificate" in version "cert-manager.io/v1alpha2"
unable to recognize "https://github.com/open-telemetry/opentelemetry-operator/releases/latest/download/opentelemetry-operator.yaml": no matches for kind "Issuer" in version "cert-manager.io/v1alpha2"
```
Should we mention something or not? I'm ok either way.

docs/install/collecting-metrics/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+   ```shell
+   kubectl create namespace metrics
+   ```
+
+3. And then create an OpenTelemetryCollector object to store the configuration
+   for the collector:
+
+   ```shell
+   kubectl apply --filename collector.yaml
```
csantanapr, line 44: We usually instruct in pre-reqs to git clone the repo and change directory, if not it will be a file not found

docs/install/collecting-metrics/README.md
csantanapr: We should add `kubectl wait` here if not on the next step to create a collector CR it would fail
```
Error from server (InternalError): error when creating "collector.yaml": Internal error occurred: failed calling webhook "mopentelemetrycollector.kb.io": Post "https://opentelemetry-operator-webhook-service.opentelemetry-operator-system.svc:443/mutate-opentelemetry-io-v1alpha1-opentelemetrycollector?timeout=30s": dial tcp 10.96.101.77:443: connect: connection refused
```
The operator pod was still creating
```
kubectl get pods -n opentelemetry-operator-system
NAME                                                         READY   STATUS              RESTARTS   AGE
opentelemetry-operator-controller-manager-6c7b945dcf-vqzvv   0/2     ContainerCreating   0          7m46s
```

csantanapr: operator is not installing for me for some reason, I'm using kind
```
Unable to attach or mount volumes: unmounted volumes=[cert],
```
I will reach out on Slack @evankanderson

evankanderson: I forgot to copy the instruction to install cert-manager, which I already had on the cluster for Knative.

I'll also see if there are any plans to avoid having a hard dependency on cert-manager.

docs/install/collecting-metrics/README.md
evankanderson: I'm wondering about linking to `raw.githubusercontent.com` instead, so users don't need to do a clone. What do you think?

csantanapr: Yeah I think is better to provide htttp URL to `raw.githubusercontent.com`

docs/install/collecting-metrics/collector.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+apiVersion: opentelemetry.io/v1alpha1
+kind: OpenTelemetryCollector
+metadata:
+  name: otel
+  namespace: metrics
+spec:
+  config: |
+    receivers:
+      jaeger:
```
csantanapr, line 12: Does this post would cover distributing tracing (ie zipkin/jaeger) or that would be a different doc?

docs/install/collecting-metrics/README.md
```diff
@@ -0,0 +1,106 @@
...
+   ```
+
+1. Finally, update the `config-observability` ConfigMap in Knative Serving and
+   Eventing
+   ```shell
+   kubectl patch --namespace knative-serving configmap/config-observability \
```
csantanapr, line 55: From the diagram it looks like only `queue-proxy` and `activator` would be pushing metrics, what about the control-plane controllers? Would they also be pushing metrics or not?

docs/install/collecting-metrics/README.md
```diff
@@ -0,0 +1,106 @@
...
+   ```
+
+By default, the Prometheus instance will only be exposed on a private service
+named `prometheus-operated`; to access the console in your web browser, run:
+
+```shell
+kubectl port-forward --namespace metrics service/prometheus-operated 9090
+```
+
+And then access the console in your browser via http://localhost:9090.
```
csantanapr, line 106: What should we do for dashboard like grafana? should we provide them or not? Or that could be a different doc

csantanapr: I feel like we should document each component the available metric name and a short description. 
Maybe like a table format that includes component, metric, description

csantanapr:
Using the opentelemetry operator didn't work for me. 

I think also using Prometheus Operator it hides the simplicity of configure prometheus to scrape

I think we should provide a simpler setup, maybe single yaml that defines 1 deployment and 1 ConfigMap for each agent/collector and Prometheus 

If this example is only about metrics I think we should remove any references to traces to avoid confusion. 

Do another docs example for traces.

docs/install/collecting-metrics/collector.yaml
evankanderson: I think it would in the future, I wanted to go one step at a time.

docs/install/collecting-metrics/README.md
evankanderson: Yes, they would too, I can add more elements, thanks for the callout!

docs/install/collecting-metrics/README.md
evankanderson: I think this is a subsequent (post 0.19) work item to document the exported metrics (possibly rationalizing them).

We absolutely should do this, and I expect that we'd need to treat it like other API contracts. I wasn't quite ready to do that here, but I can give 1 or 2 sample queries.

evankanderson:
I'll switch this to direct ConfigMap/Deployment/Service setup, rather than using the operators. Thanks for the feedback.

I'll spin a new version of this this morning, hopefully.

docs/install/collecting-metrics/collector.yaml
```diff
@@ -0,0 +1,98 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: otel-collector-config
+  namespace:
```
MontyCarter, line 5: `metrics`?

docs/install/collecting-metrics/collector.yaml
```diff
@@ -0,0 +1,98 @@
...
+    exporters:
+      logging:
+      prometheus:
+        endpoint: "0.0.0.0:8889"
+
+    service:
+      extensions:
+        health_check:
+        pprof:
+        zpages:
```
MontyCarter, line 21: Do these extensions work if you don't declare them first?  Something like this is what I was expecting: https://github.com/open-telemetry/opentelemetry-collector/blob/master/examples/demo/otel-collector-config.yaml#L30

If these aren't functional, should we take them out?

docs/install/collecting-metrics/collector.yaml
```diff
@@ -0,0 +1,98 @@
...
+          requests:  # Note: these are suitable for a small instance, but may need to be increased for a large instance.
+            memory: 100Mi
+            cpu: 50m
+        ports:
+        - name: otel
+          containerPort: 55678
+        - name: prom-export
+          containerPort: 8889
+        - name: zpages  # A /debug page
+          containerPort: 55679
```
MontyCarter, line 60: I would argue that the internal metrics are useful too, particularly during initial setup.

Maybe announce the 8888 port here?

docs/install/collecting-metrics/README.md
MontyCarter: I think it would be really useful to have a doc listing and rationalizing the exported metrics.

I suspect we'll find a lot that don't warrant being exported by default.

As you suggested, I think it will be useful to have metric for this document that can act as a litmus test, so folks who aren't up to speed on metrics know what they're looking for to gauge success.

MontyCarter (approved):

evankanderson:
@csantanapr Please take another look/try these again.

It turns out that queue-proxy doesn't seem to put the most useful labels on the exported metrics, so I added some `label_replace` examples.

docs/install/collecting-metrics/collector.yaml
evankanderson: Yes, fixed. (I was trying to copy things from one machine to another to test the hand-rolled setup extracted from the operator configuration I'm using on an existing cluster.)

docs/install/collecting-metrics/collector.yaml
evankanderson: These have defaults assuming the key is present. No arguments are needed (I added a port declaration for the `zpages` port; I can add the other two port declarations to the Deployment if you think it helps).

docs/install/collecting-metrics/collector.yaml
evankanderson: The metrics describing the operation of the OpenTelemetry collector?

I didn't want to confuse things too much -- this was intended to be an example rather than a complete solution(for example, there's no dashboard, and I'm personally not interested in trying to maintain something like that dashboard).

docs/install/collecting-metrics/README.md
evankanderson: I'm not sure about not exporting them by default -- the cost of an exported metric should not be particularly high, but I think we *do* need to rationalize both the metric names and labels.

docs/install/collecting-metrics/collector.yaml
MontyCarter: If they will work this way, I'd say leave them in.  I've always declared them above, so wasn't sure.

Also, don't think it's necessary to declare all of the ports.  ZPages is the main one, so.

docs/install/collecting-metrics/collector.yaml
MontyCarter: I see.  This works for me, then.  I see the value in keeping it simpler.

docs/install/collecting-metrics/collector.yaml
```diff
@@ -0,0 +1,96 @@
...
+      opencensus:
+        endpoint: "0.0.0.0:55678"
+
+    exporters:
+      logging:
+      prometheus:
+        endpoint: "0.0.0.0:8889"
+
+    service:
+      extensions:
```
csantanapr, line 18: Are these extensions leverage by knative? Do we required them define here?

csantanapr: Net mind my comment these are collector aspects

https://github.com/open-telemetry/opentelemetry-collector/blob/master/extension/README.md

And they document the default ports for health and zpages

evankanderson: They aren't strictly necessary, but I found zpages to be useful.

docs/install/collecting-metrics/README.md
skonto: We should add Eventing objects like Sources which emit metrics.

docs/install/collecting-metrics/README.md
skonto: Next step should be what metric matters to check first in certain scenarios eg. debugging.

docs/install/collecting-metrics/README.md
```diff
@@ -0,0 +1,97 @@
+This document describes how to set up the
+[OpenTelemetry Collector](https://opentelemetry.io/docs/collector/about/) to
```
skonto, line 2: Link seems dead, an option is `https://opentelemetry.io/docs/collector/`

docs/install/collecting-metrics/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+The collector provides a long-lived location where various Knative components
+can push metrics (and eventually traces) to be retained and collected by a
+monitoring service. For this example, we'll configure a single collector
+instance using a ConfigMap and a Deployment. For more complex deployments, some
+of this can be automated using the
+[opentelemetry-operator](https://github.com/open-telemetry/opentelemetry-operator),
+but it's also easy to manage this service directly.
+
+![Diagram of components reporting to collector, which is scraped by Prometheus](./system-diagram.svg)
```
skonto, line 21: I think we need to make clear that this is an example architecture and provide a link to more options [here](https://github.com/open-telemetry/opentelemetry-collector/blob/master/docs/design.md).
Some users might want to use an Collector as an Agent etc, do we support that?

docs/install/collecting-metrics/README.md
```diff
@@ -0,0 +1,97 @@
+This document describes how to set up the
+[OpenTelemetry Collector](https://opentelemetry.io/docs/collector/about/) to
+receive metrics from the Knative infrastructure components and distribute them
+to Prometheus. [OpenTelemetry](https://opentelemetry.io/) is a CNCF project to
```
skonto, line 4: We could also use the definition of what is advertised at the [official site](https://opentelemetry.io/):

> OpenTelemetry (CNF project) is an observability framework for cloud native software. It provides a collection of tools, APIs, and SDKs for instrumenting, generating, collecting, and exporting telemetry data (metrics, logs, and traces) for analysis in order to understand software's performance and behavior.

abrennan89:
@evankanderson any updates on this one? Is it blocked by anything?

docs/install/collecting-metrics/README.md
evankanderson: Thanks, fixed.

docs/install/collecting-metrics/README.md
evankanderson: I think those should work; I wanted to include a minimal example here.

docs/install/collecting-metrics/README.md
evankanderson: Thanks, copied it over.

evankanderson:
I've addressed the most recent @skonto comments. I think this is still correct -- I've used it intermittently and didn't realize I hadn't gotten the PR submitted. 😁

abrennan89:
/approve
/lgtm

Thanks Evan! :slightly_smiling_face:

PR #3010, opened 2020-11-06
Drop the `-alpine` for dotnet, it isn't multi-arch.

mattmoor:
This is serving from a Graviton EKS cluster built with this sample: https://helloworld-csharp.default.mink.dev

/assign @csantanapr @abrennan89


csantanapr:
I think after 0.19 lands we should update the README of the samples to mention `docker buildx` they can use for building arm64

mattmoor:
@csantanapr Good idea, want to open an issue?

csantanapr:
Thank @mattmoor I verified on raspberry pi4 with `docker buildx`
```
curl https://helloworld-csharp.default.pi.csantanapr.com
Hello World!
```

PR #3016, opened 2020-11-09
Transform Scala example to use a Dockerfile

### LINKED ISSUE #2927: helloworld-scala doesn't use a Dockerfile
## Expected Behavior

Uniformity with other samples, so Dockerfile.

## Actual Behavior

The image is assembled with an sbt plugin.

https://github.com/knative/docs/blob/master/docs/serving/samples/hello-world/helloworld-scala/README.md

markusthoemmes:
Fixes #2927

There's a few more things askew with the Scala sample but I'd like to line those up in a separate PR to keep concerns separate.

### DIFF
--- a/docs/serving/samples/hello-world/helloworld-scala/Dockerfile
+++ b/docs/serving/samples/hello-world/helloworld-scala/Dockerfile
@@ -1,0 +1,22 @@
+# Use an SBT image matching the Scala and JDK version.
+FROM hseeberger/scala-sbt:8u265_1.4.2_2.12.12 as builder
+
+# Copy local code to the container image.
+WORKDIR /app
+COPY build.sbt .
+COPY project ./project
+COPY src ./src
+
+# Build a release artifact.
+RUN sbt assembly
+
+# Use the Official OpenJDK image for a lean production stage of our multi-stage build.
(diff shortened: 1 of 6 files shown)


docs/serving/samples/hello-world/helloworld-scala/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Use an SBT image matching the Scala and JDK version.
+FROM hseeberger/scala-sbt:8u265_1.4.2_2.12.12 as builder
```
mattmoor, line 2: Do you know if there is a multi-arch image we can use here?  All of our other helloworld samples work against arm64 now 😅

markusthoemmes: :thinking: does the build-time image matter, seeing as this is on the JVM?

mattmoor: If I run the Dockerfile build on arm64 it does 😉

markusthoemmes: He, true, quick research didn't yield anything though :/

Issue #1964, opened 2019-11-05
Custom domains doc only explains customizing default domain


ahmetb:
https://knative.dev/docs/serving/using-a-custom-domain/

This doc is titled "Setting up a custom domain" but it actually only covers "Configuring the default domain".

It doesn't explain how can I expose my `hello-go.default.example.com` with a domain name I own like `hello.com`.

jonatasbaldin:
Hhey @ahmetb!

@bsideup wrote this article on how to create custom domains per Service: https://bsideup.github.io/posts/knative_custom_domains/

I'm thinking about documenting this process in the docs.

What do you think?

bsideup:
@jonatasbaldin funny thing - I thought about contributing it to the docs but didn't find time yet :D

I was thinking about doing it this week, but if you do it first - feel free to! 👍

jonatasbaldin:
hey, just made the docs! would love a review from you @bsideup

ahmetb:
I don’t think that blog post quite explains my need. I am looking to map example.com to a ksvc. To do that, I guess you need to drop {{.Name}} because I don’t care about ksvc’s name in my domain?

bsideup:
@ahmetb you can use the same approach as described in the blog post, just instead of using `$subdomain.{{.Domain}}` (where `$subdomain` is the label's value) you can replace it with just `$domain`

ahmetb:
I see. I think that’s the useful scenario docs should actually mention. 1:1 domain ksvc mappings; not templates.

PR #3017, opened 2020-11-09
Add DomainMapping docs

julz:
Documents the [DomainMapping feature](https://github.com/knative/serving/issues/9713) which will be added in alpha in v0.19.

This probably needs lots of work @abrennan89 @mpetason, but opening as a first version, anyway.

In particular looking for guidance on how best to call out that this is an Alpha feature etc.

### DIFF
--- a/docs/serving/creating-domain-mappings.md
+++ b/docs/serving/creating-domain-mappings.md
@@ -26,6 +26,11 @@ to service requests.
 You will also need to point the domain name at your Knative cluster using the
 tools provided by your domain registrar.
 
+Domain Mappings map a single, non-wildcard domain to a specific Knative
+Service. For example in the example yaml below, the "mydomain.com" Domain
+Mapping maps only "mydomain.com" and not "www.mydomain.com". You can create
+multiple Domain Mappings to map multiple domains.
+
 ### Procedure
 
 1. Create a new file named `domainmapping.yaml` containing the following information.
--- a/docs/serving/creating-domain-mappings.md
+++ b/docs/serving/creating-domain-mappings.md
@@ -7,7 +7,7 @@ type: "docs"
 
 If the Domain Mapping feature is installed, you can serve
 custom domains backed by Knative Services. For example, if you
-own the "MyDomain.com" domain name, you can point DNS at your
+own the "mydomain.com" domain name, you can point DNS at your
 Knative cluster, and have this domain name be served by a
 Knative Service.
 
--- a/docs/serving/creating-domain-mappings.md
+++ b/docs/serving/creating-domain-mappings.md
@@ -36,7 +36,7 @@ tools provided by your domain registrar.
    name: mydomain.com
    namespace: default
   spec:
-   ref: 
+   ref:
      name: helloworld-go
      kind: Service
      apiVersion: serving.knative.dev/v1
--- a/docs/serving/creating-domain-mappings.md
+++ b/docs/serving/creating-domain-mappings.md
@@ -10,31 +10,34 @@ cluster configuration, e.g. "mysvc.mynamespace.mydomain". You can also map a
 single custom domain name that you own to a specific Knative Service using the
 Domain Mapping feature, if enabled.
 
-For example, if you own the "mydomain.com" domain name, and configure its DNS
+For example, if you own the "example.org" domain name, and configure its DNS
 to reference your Knative cluster, you can use the DomainMapping feature to
 have this domain be served by a Knative Service.
 
 ## Before you begin
 
 1. You need to enable the DomainMapping feature (and a supported Knative
-   Ingress implementation) to use it. See [the Install instructions](../install/).
+   Ingress implementation) to use it. See [the Install instructions](../install/any-kubernetes-cluster/#optional-serving-extensions).
 1. To map a custom domain to a Knative Service, you should first [create a Knative
 Service](https://knative.dev/docs/serving/creating-services/).
+1. You will need a Domain Name to map, and the ability to change its DNS to
+   point to your Knative Cluster. The details of this step are dependant on
+   your domain registrar.
 
 ## Creating a Domain Mapping
 
-To create a mapping from a custom domain to a Knative Service, you need to
-create a YAML file that defines a Domain Mapping.
-This YAML file specifies the domain name to map and the Knative Service to use
-to service requests.
+To create a mapping from a custom domain name that you control to a Knative
+Service, you need to create a YAML file that defines a Domain Mapping. This
+YAML file specifies the domain name to map and the Knative Service to use to
+service requests.
 
 You will also need to point the domain name at your Knative cluster using the
 tools provided by your domain registrar.
 
 Domain Mappings map a single, non-wildcard domain to a specific Knative
-Service. For example in the example yaml below, the "mydomain.com" Domain
-Mapping maps only "mydomain.com" and not "www.mydomain.com". You can create
-multiple Domain Mappings to map multiple domains.
+Service. For example in the example yaml below, the "example.org" Domain
+Mapping maps only "example.org" and not "www.example.org". You can create
+multiple Domain Mappings to map multiple domains and subdomains.
 
 ### Procedure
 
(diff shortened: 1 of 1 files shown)


csantanapr:
> Documents the DomainMapping feature which will be added in alpha in v0.18.

You meant to say v0.19 ?

docs/serving/creating-domain-mappings.md
```diff
@@ -0,0 +1,57 @@
+---
+title: "Creating a Mapping between a Custom Domain Name and a Knative Service (Alpha)"
+linkTitle: "Creating Domain Mappings (Alpha)"
+weight: 64
+type: "docs"
+---
+
+If the Domain Mapping feature is installed, you can serve
+custom domains backed by Knative Services. For example, if you
+own the "MyDomain.com" domain name, you can point DNS at your
```
csantanapr, line 10: Maybe using lower case is better “mydomain.com”

julz:
> You meant to say v0.19 ?

I did. And as one of the release managers for 0.19: oops 😂 .

docs/serving/creating-domain-mappings.md
julz: d'oh, thanks, fixed

docs/serving/creating-domain-mappings.md
```diff
@@ -0,0 +1,57 @@
+---
+title: "Creating a Mapping between a Custom Domain Name and a Knative Service (Alpha)"
+linkTitle: "Creating Domain Mappings (Alpha)"
+weight: 64
+type: "docs"
+---
+
+If the Domain Mapping feature is installed, you can serve
```
csantanapr, line 8: This sentence sounds like the user needs to install an optional components when installing knative..
Is this true or you mean that user use a knative 0.19+ that contains this CRD and that the state of the feature is alpha and we want feedback early

docs/serving/creating-domain-mappings.md
```diff
@@ -0,0 +1,57 @@
...
+tools provided by your domain registrar.
+
+### Procedure
+
+1. Create a new file named `domainmapping.yaml` containing the following information.
+  ```yaml
+  apiVersion: serving.knative.dev/v1alpha1
+  kind: DomainMapping
+  metadata:
+   name: mydomain.com
```
csantanapr, line 36: Is there a spec.domain field? Because I would be very disappointed if the actual domain value is the name of the CR

docs/serving/creating-domain-mappings.md
julz: Yes, the user does need to install an additional component -- it's not installed by default in this release (there's an extra yaml file, much like installing the monitoring bundle or HPA autoscaler). I'll PR an update to install docs separately with that optional step.

docs/serving/creating-domain-mappings.md
```diff
@@ -0,0 +1,57 @@
...
+
+To map a custom domain to a Knative Service, you should first [create a Knative
+Service](https://knative.dev/docs/serving/creating-services/).
+
+## Creating a Domain Mapping
+
+To create a mapping from a custom domain to a Knative Service, you need to
+create a YAML file that defines a Domain Mapping.
+This YAML file specifies the domain name to map and the Knative Service to use
+to service requests.
```
csantanapr, line 24: Put some reference to the example “mydomain.com” explain if using this value would cover all sub domains ? *.mydomain.com or it needs to a specific domain ?

docs/serving/creating-domain-mappings.md
csantanapr: Add a reference and link here you don’t have to put the actual steps here

docs/serving/creating-domain-mappings.md
julz: It's intentionally the name of the DM to prevent collisions. Here's one of the conversations in the design doc where we decided this: https://docs.google.com/document/d/1yXSuKpQqZM5tczBFpPG8TGQoWSkvmoA17nPwDd1UxEI/edit?disco=AAAAG2yt488.

docs/serving/creating-domain-mappings.md
```diff
@@ -0,0 +1,57 @@
...
+### Procedure
+
+1. Create a new file named `domainmapping.yaml` containing the following information.
+  ```yaml
+  apiVersion: serving.knative.dev/v1alpha1
+  kind: DomainMapping
+  metadata:
+   name: mydomain.com
+   namespace: default
+  spec:
```
csantanapr, line 38: spec.domain missing?

julz: no, see above

docs/serving/creating-domain-mappings.md
julz: added an extra para here, ptal

docs/serving/creating-domain-mappings.md
```diff
@@ -0,0 +1,57 @@
+---
+title: "Creating a Mapping between a Custom Domain Name and a Knative Service (Alpha)"
+linkTitle: "Creating Domain Mappings (Alpha)"
+weight: 64
+type: "docs"
+---
```
csantanapr, line 6: The intro needs some use case if not would very confusing for the end user on why she would need this feature. 

You can put something like 
knative by default offeres a comfigmap “config-domain”  in the control plane namespace  (Ie knative-serving”) that allow an admin with write access to the namespace edit the comfigmap  to configure domain names including a selector to be use which services corresponds to a domain name.. 
But in some cases the admin would not want this ConfigMap to be edited and would prefer users to define the domain name mapping to be fine by users in the namespace where the knative service is deploy. 
This feature would allow to define the mapping in their namespace by creating a Kubernetes resource DomainMapping

docs/serving/creating-domain-mappings.md
julz: I will when I've added that step to the docs :)

docs/serving/creating-domain-mappings.md
julz: How about just "Knative Services are automatically given a default domain name based on the cluster configuration. You can also map a single custom domain name that you own to a specific Knative Service using the Domain Mapping feature."

csantanapr: sure

docs/serving/creating-domain-mappings.md
csantanapr: I think we need to add to this doc the limitations, that not all net ingress support this feature, you can list them here, or mention in the install doc and point there from here.

abrennan89:
@julz is this one ready for docs review / merge?

abrennan89:
Question: is this related to #1964? Might it close that issue?

julz:
> Question: is this related to #1964? Might it close that issue?

I think it might @abrennan89. This documents the feature that - iiuc - @ahmetb was expecting the other doc to be about. Hopefully @ahmetb can confirm if so

ahmetb:
Correct. That issue asks for this PR.

docs/serving/creating-domain-mappings.md
```diff
@@ -0,0 +1,67 @@
...
+Domain Mapping feature, if enabled.
+
+For example, if you own the "mydomain.com" domain name, and configure its DNS
+to reference your Knative cluster, you can use the DomainMapping feature to
+have this domain be served by a Knative Service.
+
+## Before you begin
+
+1. You need to enable the DomainMapping feature (and a supported Knative
+   Ingress implementation) to use it. See [the Install instructions](../install/).
```
duglin, line 20: where on that page should people look? I didn't see anything covering this topic. Perhaps we could point to the exact page of interest?

nak3 (commented):
@julz It seems mydomain.com is already used by some organization so should we change the example domain like mydomain.example.org which is listed in [RFC2606 - 3. Reserved Example Second Level Domain Names](https://tools.ietf.org/html/rfc2606#page-2)?

julz:
oops - thanks @nak3, will update!

julz:
Ready for another look @abrennan89 @nak3

PR #3025, opened 2020-11-11
Add description on how to configure each ingress for knative serving

houshengbo:
## Proposed Changes 

- Knative operator is also able to configure multiple ingresses other than istio. The guide is based on the existing guide in knative. The only change is that Knative operator leverage the serving CR instead of directly working on configMaps.

### DIFF
--- a/docs/install/knative-with-operators.md
+++ b/docs/install/knative-with-operators.md
@@ -539,8 +539,8 @@ To access your application using `curl` using this method:
    curl -H "Host: helloworld-go.default.example.com" http://192.168.39.228:32198
    ```
 
-   In the case of the provided `helloworld-go` sample application, the output
-   should, using the default configuration, be:
+   In the case of the provided `helloworld-go` sample application, using the default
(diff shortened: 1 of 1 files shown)


docs/install/knative-with-operators.md
```diff
@@ -207,6 +207,358 @@ NAME              VERSION             READY   REASON
...
+   networking layer defines your External IP and port to be
+   `http://192.168.39.228:32198` and you wish to access the above
+   `helloworld-go` application, use:
+
+   ```bash
+   curl -H "Host: helloworld-go.default.example.com" http://192.168.39.228:32198
+   ```
+
+   In the case of the provided `helloworld-go` sample application, the output
+   should, using the default configuration, be:
```
mpetason, line 543:
```suggestion
   In the case of the provided `helloworld-go` sample application, using the default configuration, the output
   should be:
```

evankanderson (commented):
/approve

I'll let Mike do the `/lgtm`

PR #3028, opened 2020-11-11
Add information on sharing logs to local disk

evankanderson:
Adds directions to make cluster logs visible on the local disk so you don't need to run `kubectl logs ...` when developing locally.

## Proposed Changes 

- Add directions for Kind and Docker Desktop to share the log collection directory with the host filesystem.

Given the variations for different platforms, I didn't include example YAML this time.


docs/install/collecting-logs/README.md
```diff
@@ -62,6 +62,85 @@ tools:
...
+PersistentVolume. Note that the directory path `./logs` is relative to the
+directory that the Kind cluster was created in.
+
+### Docker Desktop
+
+Docker desktop automatically creates some shared mounts between the host and the
+guest operating systems, so you only need to know the path to your home
+directory. Here are some examples for different operating systems:
+
+| Host OS | `hostPath` | | Mac OS | `/Users/${USER}` | | Windows |
```
csantanapr, line 139: Does this table look correct?
In markdown rendered https://github.com/evankanderson/docs/blob/twigs/docs/install/collecting-logs/README.md#docker-desktop
doesn't look correct

docs/install/collecting-logs/README.md
```diff
@@ -62,6 +62,85 @@ tools:
...
+    name: logs
+  spec:
+    accessModes: ["ReadWriteOnce"]
+    volumeName: shared-logs
+```
+
+### Kind
+
+When creating your cluster, you'll need to use a `kind-config.yaml` when
+creating your cluster and specify `extraMounts` for each node, like so:
```
mpetason, line 113:
```suggestion
When creating your cluster, you'll need to use a `kind-config.yaml` and specify `extraMounts` for each node, like so:
```

docs/install/collecting-logs/README.md
evankanderson: Yeah, not sure what happened here, there were supposed to be newlines to make that a table.

docs/install/collecting-logs/README.md
evankanderson: Thanks! I went back and added the `kind-config.yaml` sentence after I'd started and didn't read the whole sentence at once.

docs/install/collecting-logs/README.md
```diff
@@ -62,6 +62,88 @@ tools:
 kubectl exec --namespace logging --stdin --tty --container nginx log-collector-0
 ```
 
+## Local collector
+
+If you are using a local Kubernetes cluster for development (Kind, Docker
+Desktop, or Minikube), you can create a `hostPath` PersistentVolume to store the
+logs on your desktop OS. This will allow you to use all your normal desktop
+tools on the files without needing Kubernetes-specific tools.
```
abrennan89, line 70:
```suggestion
If you are using a local Kubernetes cluster for development, such as _Kind_, _Docker Desktop_, or _Minikube_, you can create a `hostPath` persistent volume to store log files on your desktop operating system. This allows you to use common desktop tools to interact with the log files without requiring Kubernetes-specific tools.
```

docs/install/collecting-logs/README.md
```diff
@@ -62,6 +62,88 @@ tools:
...
+
+The PersistentVolumeClaim will look something like this, but the `hostPath` will
+vary based on your Kubernetes software and host operating system. Some example
+values are documented below.
```
abrennan89, line 74:
```suggestion
The `PersistentVolumeClaim` object looks similar to the following example, except that the `hostPath` will
vary based on your Kubernetes software and host operating system:
```

docs/install/collecting-logs/README.md
```diff
@@ -62,6 +62,88 @@ tools:
...
+    name: logs-log-collector-0
+    namespace: logging
+  capacity:
+    storage: 5Gi
+  hostPath:
+    path: /run/desktop/mnt/host/c/Users/evank/logs/
+```
+
+And then you'll need to update the StatefulSet's `volumeClaimTemplates` to
+reference the `shared-logs` volume, like this fragment of yaml:
```
abrennan89, line 99:
```suggestion
You must update the `volumeClaimTemplates` value of the `StatefulSet` object to
reference the `shared-logs` volume:
```

docs/install/collecting-logs/README.md
```diff
@@ -62,6 +62,88 @@ tools:
...
+```yaml
+volumeClaimTemplates:
+  metadata:
+    name: logs
+  spec:
+    accessModes: ["ReadWriteOnce"]
+    volumeName: shared-logs
+```
+
+### Kind
```
abrennan89, line 110: Should we be adding platform-specific docs? :thinking: Didn't we intentionally remove all docs about things like Minikube previously?

abrennan89 (changes requested):
Added some comments, however I'm not sure we should be adding platform specific docs at all? Wasn't there a specific effort to remove things like this before and just add the Knative Offerings page instead?

cc @pmorie @mattmoor

mattmoor:
We removed all of the vendor specific cluster provisioning and setup pages because they were largely identical, drifting, and pretty low signal/noise.

I'm receptive to documenting how to use open local options, but still think we should aim to minimize them because there are so many options (Kind, minikube, Docker Desktop, microk8s, k3s, k0s, minishift, ....).

My $0.02.

abrennan89:
> I'm receptive to documenting how to use open local options, but still think we should aim to minimize them because there are so many options (Kind, minikube, Docker Desktop, microk8s, k3s, k0s, minishift, ....).

The issue I have is how do we "minimize" them or draw the line with what gets documented and what doesn't? I feel like if we allow mentioning one and specific steps for one like Kind we have to allow them all, and that to me is a slippery slope. I get that it would be useful for folks to have open local options, but I don't think our docs are the place for those platform specific instructions to live. I think eventually it will end up bloating our docs and creating maintenance work that shouldn't be under the scope of the Knative docs WG.

mattmoor:
> The issue I have is how do we "minimize" them or draw the line with what gets documented and what doesn't?

Ack, I don't have a good answer.  I think for upstream docs folks generally shouldn't need an account with X to walk through the instructions, but that's a really low bar.

Another alternative here would be to pick one option (e.g. KinD) and document that option fairly thoroughly, but largely punt on other options.  cc @omerbensaadon @csantanapr.

abrennan89:
>  I think for upstream docs folks generally shouldn't need an account with X to walk through the instructions, but that's a really low bar.

I would agree, but I also think that a prerequisite for using something like Knative should be a basic understanding of a Kubernetes cluster; there needs to be some level of knowledge that can be assumed here, particularly I think with a project this complicated where everything is Kubernetes APIs and resources.
I don't think it's really up to us to walk people through how to do everything on a specific cluster type, that's something they should be requesting in the docs for the platform itself IMO.
I would like to think we can just pick one and document it thoroughly, but I think there will always be disagreement about what is the best platform to do that for, and we will no doubt get requests for others if we open that can of worms, and then we have no real way to justify it to users if we doc one and not the others.

evankanderson:
Updated (moved text from `README.md` to `_index.md`), and added a section for Minikube based on documentation.

evankanderson:
I could switch the Kind/Docker Desktop/Minikube sections to tabs if you think that would look cleaner.

If the functionality were any closer, I'd try to converge on a single set of instructions, but the current state is:

* KinD: config file before cluster start
* Docker Desktop: automatic, sort of
* Minikube: command run _after_ cluster start

/meow

abrennan89 (approved):
/lgtm

I think it's OK as is for now, thanks Evan.

How did I not know about "meow" until now?! 😹

PR #3038, opened 2020-11-16
Replace wrong description of kubectl attach

EmmanuelKasper:
No SSH was ever harmed when starting a shell via kubectl attach to execute shell commands.
So correct that from the documentation.

## Proposed Changes 

-
-
-

### DIFF
--- a/docs/eventing/getting-started.md
+++ b/docs/eventing/getting-started.md
@@ -235,7 +235,7 @@ metadata:
   name: curl
 spec:
   containers:
-    # This could be any image that we can SSH into and has curl.
+    # This could be any image which has curl and a shell we can attach to
   - image: radial/busyboxplus:curl
     imagePullPolicy: IfNotPresent
     name: curl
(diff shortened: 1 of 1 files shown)


EmmanuelKasper:
Hi, I force pushed the PR with  correct email adress, CLA is now signed, should be good to merge.

abrennan89:
/lgtm
/hold

Needs a technical review, so I have tagged some folks.
There are some tests failing on this too, any idea why @mattmoor @RichieEscarez ?

mattmoor:
@abrennan89 this is because the action definitions on the release branch are using a feature of actions that got disabled.  These should be safe to ignore in the meantime, since it only impacts cherrypicks.  We can cherrypick the latest actions into the release branch if this becomes a recurring problem.

abrennan89:
/unhold
@vagababov said this looks OK, waiting for him to approve.

EmmanuelKasper:
Thanks for feedback and approval, humans and robots.

PR #3039, opened 2020-11-16
Adding docs for the sugar controller

n3wscott:
## Proposed Changes 

- Document the sugar controller for Namespaces
- Document the sugar controller for Triggers

### DIFF
--- a/docs/eventing/sugar/README.md
+++ b/docs/eventing/sugar/README.md
@@ -1,0 +1,86 @@
+Knative Eventing Sugar Controller will react to special labels and annotations to produce or control eventing resources in a cluster or namespace.
+
+## Installing
+
+The following command installs the Eventing Sugar Controller:
+
+```bash
+kubectl apply --filename {{< artifact repo="eventing" file="eventing-sugar-controller.yaml" >}}
+```
+
+## Automatic Broker Creation
+
+One way to create a Broker is to manually apply a resource to a cluster using the default settings:
+
+```shell
+kubectl create -f - <<EOF
+apiVersion: eventing.knative.dev/v1
+kind: Broker
+metadata:
+  name: default
+  namespace: default
+EOF
+```
+
+There might be cases where automated Broker creation is desirable, such as on namespace creation, or on Trigger creation. The Sugar controller enables those use-cases:
+
+- When a Namespace is labeled with `eventing.knative.dev/injection=enabled`, the
+  controller will create a default broker in that namespace.
+- When a Trigger is annotated with `eventing.knative.dev/injection=enabled`, the
+  controller will create a Broker named by that Trigger in the Trigger's
+  Namespace.
+
+When a Broker is deleted and the above labels or annotations are in-use, the Sugar Controller will automatically recreate a default Broker.
+
+### Namespace Examples
+
+Creating a "default" Broker when creating a Namespace:
+
+```shell
+kubectl apply -f - <<EOF
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: example
+  labels:
+    eventing.knative.dev/injection: enabled
+EOF
+```
+
+To automatically create a Broker after a namespace exists, label the Namespace:
+
+```shell
+kubectl label namespace default eventing.knative.dev/injection=enabled
+```
+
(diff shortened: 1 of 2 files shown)


docs/eventing/sugar/README.md
```diff
@@ -0,0 +1,92 @@
...
+```shell
+kubectl label namespace default eventing.knative.dev/injection=enabled
+```
+
+If the Broker named "default" already exists in the Namespace, the Sugar
+Controller will do nothing.
+
+### Trigger Examples
+
+To create a Broker if none exists for the provided Broker name in the Namespace
```
vaikas, line 66: This reads funny to me. You might say something that it's possible to create a Trigger even if a broker doesn't yet exist. Might want to make a caveat however that if you typo, you'll get a broker that you might not want. Anyways, I had to read this couple of times to figure out what you meant.

docs/eventing/sugar/README.md
```diff
@@ -0,0 +1,92 @@
...
+kind: Broker
+metadata:
+  name: default
+  namespace: default
+EOF
+```
+
+There might be cases where automated Broker creation is desirable, such as on
+namespace creation, or on Trigger creation. The Sugar controller enables those
+use-cases:
```
vaikas, line 29: I thought there was a way to have it autogen a Broker for any new namespace without adding an annotation? opt-out?

docs/eventing/sugar/README.md
evankanderson, line 66: This seems like it's a useful pattern to allow defining an application without creating conflicts on the Broker resource if you have several applications sharing a Broker. Is it worth calling out that pattern?

docs/eventing/sugar/README.md
evankanderson: I was thinking that the namespace version of sugaring seemed more useful as an opt-out, since you'd need edit on namespaces to be able to apply the label.

n3wscott: I think you are thinking of the SinkBinding @vaikas.

n3wscott: > I was thinking that the namespace version of sugaring seemed more useful as an opt-out, since you'd need edit on namespaces to be able to apply the label.

We can think about that as a followup feature, but the thinking if if you are provided a namespace, and have no control over the labels, it is an option when the namespace is created provided by the admin of that cluster.

docs/eventing/sugar/README.md
n3wscott: Reworded, I think it is simpler now.

n3wscott: I am not sure I understand this comment, what is the suggestion?

docs/eventing/sugar/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+- When a Namespace is labeled with `eventing.knative.dev/injection=enabled`, the
+  controller will create a default broker in that namespace.
```
grantr, line 35: Unless it's previously defined what name a "default broker" has, I think this should be made more explicit: 

```
the controller will create a default broker named `default` in that namespace.
```

Then subsequent references can either refer to the name of the broker object (`default`) or the name of the concept ("default broker").

n3wscott: ok good point.

n3wscott: Updated.

docs/eventing/sugar/README.md
grantr: I think Evan is suggesting that you add a possible use case to help the reader understand, why would I use this?

IIUC, the use case is, I have multiple apps that will share a broker, and I want the broker to be created when any of those apps are created. BUT I don't want those apps to have to worry about keeping the broker definitions in sync in their yaml.

grantr (approved):
/lgtm

/hold
Holding in case Scott wants to add Evan's suggestion, but since there's a followup for this already in #3040, he might want to get this in now and focus there.

n3wscott:
> Holding in case Scott wants to add Evan's suggestion, but since there's a followup for this already in #3040, he might want to get this in now and focus there.

Thanks so much, I will take care of it in the followup!

/unhold

PR #3049, opened 2020-11-19
sample: Convert to BusyBox httpd for the shell sample

rhuss:
nc has issues because of the constant listen and unlisten on port 8080.
Therefor switching over to busybox's [httpd](https://git.busybox.net/busybox/tree/networking/httpd.c)
which is should be much more robust.

The shell example still stays simple and reuses an existing command.

### DIFF
--- a/docs/serving/samples/hello-world/helloworld-shell/Dockerfile
+++ b/docs/serving/samples/hello-world/helloworld-shell/Dockerfile
@@ -1,13 +1,12 @@
-# Use the official Alpine image for a lean production container.
-# https://hub.docker.com/_/alpine
-FROM alpine:3
+# Busybox image that contains the simple 'httpd'
+# https://git.busybox.net/busybox/tree/networking/httpd.c
+FROM busybox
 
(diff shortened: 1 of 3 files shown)


rhuss:
The integration test fails on the Kotlin sample where it can't find a dependency. However this is unrelated to this PR.

markusthoemmes:
Kotlin failures are supposed to be fixed via #3050

mattmoor (commented):
/lgtm
/approve

Let's give it a shot

csantanapr:
@rhuss @mattmoor 
I didn't know `busybox` came with `httpd` that's so cool.

One concern I have is that updates change the semantics of the function, now the script doesn't run every time there is a new invocation/request like before.

I updated the example to run `script.sh` on every invoke here what do you think if I submit a PR or is there is something wrong with my approach?
https://github.com/csantanapr/knative-docs/blob/hello-shell/docs/serving/samples/hello-world/helloworld-shell/index.md#building

PR #3053, opened 2020-11-20
helloworld shell sample: Switch to use CGI bin for the output

rhuss:
Also, httpd is now started by init for proper SIGTERM handling.
If you run the image itself within Docker, note that SIGINT (CTRL-C) is not captured.
Use `docker --init` for being able to stop the containe with CTRL-C
(This is not relevant when used for Knative)

Thanks to @csantanapr for the idea to switch to CGI.


test/sampleapp/config.yaml
```diff
@@ -59,6 +59,6 @@ languages:
   - language: "shell"
     expectedOutput: "Hello Shell Sample v1!"
```
csantanapr, line 60: The function should return just this string or Hello World right?

I don't think the script is matching this contract now. 

Would be good to show in the script how the user can do this 3 things 
1. write to back the response to http client
2. write to stdout (to show in kubectl logs)
3. Write to stderr (to show in kubectl logs)

The dynamic info and env info you would get from kubectl logs and the http client only gets Hello World or Hello $TARGET

rhuss: Ah, right, the test will probably fail. I wonder whether the match could be more flexible, let me check.

I tried to write to stdout / stderr, however this is not easy: If running `httpd` without `-f` it is not logging to stdout. If running with `-f` and redirecting to stdout to `/proc/1/fd/1` this work, but then `init` won't catch the TERM signal.

Let me emphasize that it is very important to have a proper SIGTERM handling, otherwise you don't play well with K8 as you need to wait to the end of the grace period.

I will check the tests, so that the are more flexible on matching the output (via regexp). For the stdout / stderr redirection there is imo not yet a good solution (except when we find a way to log the background httpd, too)

csantanapr: Yes I understand sigterm is very important, understand why, and strongly agree. 

The PR looks good, but I would just have the script  output Hello $TARGET or Hello World and nothing else about env or dates to match the other samples in other languages and not mess with the tests

rhuss: yeah, you are probably right. Maybe I just wanted to add some reasoning to the point of that guy by adding some fancy CGI stuff ;-P

image

csantanapr: Well exposing this cgi is kind of also diminishing knative also “knative is also a cgi script”

I’m starting to like what we had before I think it was @n3wscott  to have go binary that listens on port 8080 and executes a command, this command can be “sh script.sh” 

Similar to what openfaas has in theirs https://github.com/openfaas/templates/blob/master/template/dockerfile/function/Dockerfile

The. We have our sigterm, and user only logs to stderr, and what ever the print is the response body text

rhuss: but was the difference then to the `go-sample` really ? It was 90% identical.

csantanapr: I would just copy the docker file and claim that openfaas users can easily run their containers in knative also 😸

rhuss: BTW, I think supporting CGI as an example is kind of proof, too,  that "Knative is much more" but can easily simulate CGI, also. There are so many arguments and its so evident that Knative is not CGI.

My comment above was a joke btw. And we should let CGI fade out in grace (not sure if many still know it really ....).

I'm happy with this sample when stripping away the dynamic part (although found it quite helpful, too)

csantanapr:
thanks for the changes @rhuss 
/approve
/lgtm

Issue #3048, opened 2020-11-19
`istioctl install` command not found


futureCreator:
Hello, Istio 👋 

## Expected Behavior
https://knative.dev/v0.18-docs/install/installing-istio/#installing-istio-without-sidecar-injection
```
istioctl install -f istio-minimal-operator.yaml
```
Istio installation is expected.
`Istioctl` version: 1.4.10

## Actual Behavior

```
Error: unknown command "install" for "istioctl"
Run 'istioctl --help' for usage.
```

The v1.4.10 is recommended in the [Knative document](https://knative.dev/v0.18-docs/install/installing-istio/#before-you-begin).
> You need:
> - A Kubernetes cluster created.
> - istioctl (v1.4.10 or later) installed.
> 

But `install` command does not exist in [v1.4.10](https://istio.io/v1.4/docs/reference/commands/istioctl/).

## Steps to Reproduce the Problem

1. Follow [this step](https://knative.dev/v0.18-docs/install/installing-istio/#installing-istio-without-sidecar-injection).

## Additional Info
**Additional context**
n/a

**Install information:**
 - Platform (GKE, IKS, AKS, etc.): Plain K8s
 - Knative Version: v0.18

nak3:
Hello @futureCreator
Thank you for your report. Yes, it is a bug and fixing by https://github.com/knative/docs/pull/3056.

And if you don't mind, could you please use Knative v0.19?
0.19 is the latest version and the doc already has the fix - 0.19 doc is here https://knative.dev/v0.19-docs/install/installing-istio/#installing-istio-without-sidecar-injection

futureCreator:
@nak3 
Thank you! 🙏 
I'm going to install the 0.19 version and proceed.

brun0xff:
I'm trying to install Istio with the 1.5.0 version and I had the same error, any help?

PR #3056, opened 2020-11-22
[release-0.18] Update tested Istio version to v1.7

nak3:
The installation manifest uses `istioctl` command but it mentioned the
tested version is Istio v1.4.

This patch fixes the tested version to v1.7 in release 0.18 branch.

Fixes https://github.com/knative/docs/issues/3048

/cc @ZhiminXiang @JRBANCEL @arturenault

### DIFF
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -18,13 +18,13 @@ instructions.
 You need:
 
 - A Kubernetes cluster created.
-- [`istioctl`](https://istio.io/docs/setup/install/istioctl/) (v1.4.10 or later) installed.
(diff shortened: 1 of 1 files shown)


ZhiminXiang:
hmm, I am not sure if we have tested knative 0.18 with Istio 1.7. I think when we released, we only tested it with 1.4.

And I recently had a try, but could not make it work https://github.com/knative/serving/pull/10144

abrennan89:
@ZhiminXiang @nak3 any update on this one?

nak3:
Sorry for the delay. Let's leave as it is.

PR #3057, opened 2020-11-22
Update eventing broker README

xtreme-sameer-vohra:
## Proposed Changes 
Add RabbitMQ as an example of a channel provider

### DIFF
--- a/docs/eventing/broker/README.md
+++ b/docs/eventing/broker/README.md
@@ -1,6 +1,6 @@
 Knative provides a multi-tenant, channel-based broker implementation that uses channels for event routing.
 
-Before you can use the Knative Channel-based Broker, you must install a channel provider, such as InMemoryChannel, Kafka or Nats.
+Before you can use the Knative Channel-based Broker, you must install a channel provider, such as InMemoryChannel, RabbitMQ, Kafka or Nats.
 
 **NOTE:** InMemoryChannel channels are for development use only and must not be used in a production deployment.
 


xtreme-sameer-vohra:
@googlebot I fixed it.

docs/eventing/broker/README.md
```diff
@@ -1,6 +1,6 @@
 Knative provides a multi-tenant, channel-based broker implementation that uses channels for event routing.
 
-Before you can use the Knative Channel-based Broker, you must install a channel provider, such as InMemoryChannel, Kafka or Nats.
+Before you can use the Knative Channel-based Broker, you must install a channel provider, such as InMemoryChannel, RabbitMQ, Kafka or Nats.
```
abrennan89, line 3:
```suggestion
Before you can use the Knative channel-based broker, you must install a channel provider, such as InMemoryChannel, RabbitMQ, Kafka or Nats.
```

abrennan89 (changes requested):
If we're going to mention RabbitMQ, it should probably also be added here: https://knative.dev/docs/eventing/channels/channels-crds/ as an entry in the table.

We should try to keep this table in sync with the channel implementations we mention in this line probably, although I'm not sure we want to mention every single possible implementation here for "such as" - maybe just the core implementations, and add RabbitMQ to the table only?

WDYT @knative/eventing-wg-leads ?

vaikas:
Actually, there's no RabbitMQ channel yet, and as such you can't install a RabbitMQ channel and use it with the channel based broker. It should be a separate broker implementation like this here:
Section 4 in the installing eventing component:
```Install a Broker (eventing) layer:```

just like there's Kafka broker, we should have a RabbitMQ broker as a separate tab.

My most sincere apologies for not catching this sooner, I just added the ok-to-test and was going to go look at it, and totally spaced, I'm sorry :(

abrennan89:
Ahhh I see @vaikas - that makes sense!
Don't worry about it at all, it's no problem :slightly_smiling_face: it didn't merge or anything yet anyway.

@xtreme-sameer-vohra I am going to close this PR though without merging since this isn't accurate, hope that's OK.

xtreme-sameer-vohra:
Thanks for the clarification. 
I hadn't looked further to check if RabbitMQ had a Channel implementation 🤦

PR #3064, opened 2020-11-25
Update Eventing README

xtreme-sameer-vohra:
## Proposed Changes

Switch `service` to `building block` when referring to internal Knative Eventing.

### Why 
Disambiguate `service` to refer to an abstraction which has an http endpoint and primarily referring to a K8s Service or Knative Service.
Here, also calling Knative Eventing build blocks as `services` creates confusion even if the usage is technically correct.


abrennan89 (approved):
/lgtm

PR #3069, opened 2020-11-30
Fix Kafka Binding example

xtreme-sameer-vohra:
## Proposed Changes 
- Drop namespace of the Job resource used in the Kafka example as none of the other resources use a namespace.
- Add job verification step


abrennan89:
/assign @lionelvillard @matzew @lberk 
lgtm, but I'll let someone more technical add the approved label for it, since Mike has done a docs review + lgtm. Thanks @xtreme-sameer-vohra !

PR #3095, opened 2020-12-08
Set the targetUtilizationPercentage to 100

zhongj0802:
Set the targetUtilizationPercentage to 100 in Autoscale Sample App - Go service.yaml

Signed-off-by: zhongjun <zhongjun0802@gmail.com>

- set the revision targetUtilizationPercentage to 100 so as to override the default value 70.  otherwise the autoscaler will created 8 pods ,not 5 pods according the concurrency algorithm in the page https://knative.dev/docs/serving/autoscaling/autoscale-go/
-
-

### DIFF
--- a/docs/serving/autoscaling/autoscale-go/service.yaml
+++ b/docs/serving/autoscaling/autoscale-go/service.yaml
@@ -22,6 +22,7 @@ spec:
       annotations:
         # Target 10 in-flight-requests per pod.
         autoscaling.knative.dev/target: "10"
+        autoscaling.knative.dev/targetUtilizationPercentage: "100"
     spec:
       containers:
       - image: gcr.io/knative-samples/autoscale-go:0.1


markusthoemmes:
I wonder if we should instead introduce the targetUtilization concept and keep this setting at it's default instead :thinking:. Any thoughts @zhongj0802 ?

rhuss:
As I suppose that 100% utilization is really an edge case that you won't find often in the wild, I would stick to the default and explain the behaviour. This is also important that people understand the behaviour of their services (not samples) when they question why they have more pods than `int(concurrent requests / concurrency target) + 1`. So +1 to @markusthoemmes suggestion to introduce the concept of utilization already in the sample. If not, the user might actually question what this annotation is for anyway.

In general, I don't think that samples should tune their behaviour to become edge cases but should describe the default behaviour first and then go on to describe samples with different configurations as advanced examples.

csantanapr:
/unassign @mpetason 

This is now assigned to Scaling WG

vagababov:
I'd rather we changed the docs and described how targetUtilization fits in.

abrennan89:
@markusthoemmes @rhuss @csantanapr @vagababov - should we merge this or...?

markusthoemmes:
Nope, the comments above are valid.

abrennan89:
@zhongj0802 I'm going to close this PR since it shouldn't be merged then; folks are asking that `targetUtilization` be included / explained properly rather than just changing the sample.

@csantanapr when you say this is assigned to the scaling WG, do you mean someone is working on proper `targetUtilization` docs? Otherwise can someone please open a docs issue to track this and include the info that needs to be included, any feature track docs, etc? cc @markusthoemmes @vagababov

csantanapr:
@abrennan89 no one is working on the new doc.

Yes agree someone needs to open a doc issue on what needs to be documented, I don't know what this is.

All we have is @vagababov comment
>I'd rather we changed the docs and described how targetUtilization fits in.

markusthoemmes:
Target utilization is documented here: https://knative.dev/docs/serving/autoscaling/concurrency/#target-utilization

PR #3114, opened 2020-12-18
Add the description on how to to append customized the manifests

houshengbo:
The operator has introduced a new field called `spec.additionalManifests` to specify the customized manifests, based on the default manifests. We need instructions on how to leverage this field.

## Proposed Changes 

- The PR adds the instruction on how to install the customized manifests in the append mode.


tcnghia (approved):
/lgtm

PR #3118, opened 2020-12-23
Add docs on mTLS and Istio authorization policy

arturenault:
Fixes #3115

## Proposed Changes 

- Add docs on mTLS and Istio authorization policy

/assign @nak3 @JRBANCEL

PR #3120, opened 2020-12-28
Fix typo in istio authorization docs

arturenault:
Forgot to add the knative-serving namespace, which is mentioned in the previous paragraph.


arturenault:
Also, just saw the instructions in the PR markdown for merging into master and then cherry-picking into release. This and #3118  only went into release-0.19 because that was the default. Once this is merged I'll cherry-pick both into master.

tcnghia (approved):
/lgtm

PR #3138, opened 2021-01-11
Add .gitattributes for docs

evankanderson:
Cleanup to indicate git attributes so that we can make lint checks depend on Git attributes rather than ad-hoc patterns.

## Proposed Changes 

- Set `git` attributes, mark certain graphics files as `binary`.


matzew (approved):
/lgtm
/approve

PR #3149, opened 2021-01-14
Add vagababov to eng-approvers and sort and trim people not longer in Knative project

vagababov:
Also add `eng-approvers` as approvers to the serving docs.
Since that seems the reason for this group in the first place.

/assign mattmoor @abrennan89

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -35,13 +35,6 @@ aliases:
   - steuhs
   - yt3liu
 
-  serving-eng-reviewers:
-  - dprotaso
-  - markusthoemmes
-  - mattmoor
-  - tcnghia
-  - vagababov
-
   eventing-eng-approvers:
   - grantr
   - lionelvillard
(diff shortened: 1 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -1,14 +1,15 @@
 aliases:
   eng-approvers:
```
mattmoor, line 2: IDK why we have both eng-approvers and serving-eng-reviewers, this seems.... complex.

@abrennan89 WDYT about simplifying things?

vagababov: I can merge both lists.
THough usually we have both everywhere.

abrennan89 (approved):
/lgtm
/hold

I'm fine with this either way @mattmoor :slightly_smiling_face: 

Since this list isn't really maintained by docs and just happens to live here I'd say it's up to the Serving WG how they want to proceed. Gonna go ahead and approve but I'll put the hold on this so you can figure out what you want to do and remove it when you want to merge.

OWNERS_ALIASES
vagababov: Merged the two

mattmoor (commented):
We could go further, but this seems fine for now.
/unhold
/lgtm

PR #3173, opened 2021-01-19
update gen-api-reference-docs script

ahmetb:
Updating the tool version to v0.3.0 tag.
Ran locally against v0.20.0 tag for serving/eventing repos, seems to be
working fine.

cc: @RichieEscarez

### DIFF
--- a/docs/reference/api/eventing/eventing.md
+++ b/docs/reference/api/eventing/eventing.md
@@ -69,7 +69,7 @@ Resource Types:
 <tbody>
 <tr>
 <td>
-<code>apiVersion</code></br>
+<code>apiVersion</code><br/>
 string</td>
 <td>
 <code>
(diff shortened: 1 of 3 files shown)


RichieEscarez:
The linter wont let us merge any trailing white space: https://github.com/knative/docs/pull/3173/checks?check_run_id=1731455825

Otherwise, THANK YOU!!!

/lgtm
/approve

ahmetb:
> The linter wont let us merge any trailing white space: [#3173 (checks)](https://github.com/knative/docs/pull/3173/checks?check_run_id=1731455825)

@RichieEscarez  we have two options to address this:

1. I can fix it one time in the tool (but can't guarantee it'll break again in the future, it almost certainly will, because the tool doesn't try to produce perfect HTML, and trailing spaces have no significance on a "generated" pages that aren't manually maintained).

2. We can find a way to exclude "trailing whitespace" linter from processing such _generated_ pages.

ahmetb:
Also worth noting, these trailing whitespaces have always been there. The linter check is newer than the format of these pages.

RichieEscarez:
Thanks @ahmetb ! For the time being, I simply added a step in our "how to generate the API" topic for removing whitespace: https://github.com/knative/docs/blob/master/docs/reference/api/README.md

Yes, the linter is new and does help keep the docs clean in general but its not perfect.  I'd actually like for the GitHub action to propose/comment in the PR with a fix rather than just an error. I generally don't think we need to take any action here but if you are interested in opening a PR, i'd be happy to approve it for you.

PR #3190, opened 2021-01-25
Improve CronJob yaml in SinkBinding example

KauzClay:
## Proposed Changes 
* use correct apiVersion
* fix schedule so it runs every minute like the comment suggests
* use busybox as default example image, following suit from [Kubernetes CronJob docs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/)

### DIFF
--- a/docs/eventing/sources/sinkbinding.md
+++ b/docs/eventing/sources/sinkbinding.md
@@ -87,13 +87,13 @@ Create a `CronJob` object:
 
 1. Copy the sample YAML into a `cronjob.yaml` file:
     ```yaml
-    apiVersion: batch/v1
+     apiVersion: batch/v1beta1
     kind: CronJob
     metadata:
       name: heartbeat-cron
     spec:
       # Run every minute
-      schedule: "* * * * *"
+      schedule: "*/1 * * * *"
       jobTemplate:
         metadata:
           labels:
@@ -104,9 +104,11 @@ Create a `CronJob` object:
               restartPolicy: Never
               containers:
                 - name: single-heartbeat
-                  image: <FILL IN YOUR IMAGE HERE>
+                  image: busybox # Or your preferred image
                   args:
-                    - --period=1
+                  - /bin/sh
+                  - -c
+                  - date; echo "Heartbeat from $POD_NAME"
                   env:
                     - name: ONE_SHOT
                       value: "true"
--- a/docs/eventing/sources/sinkbinding.md
+++ b/docs/eventing/sources/sinkbinding.md
@@ -87,7 +87,7 @@ Create a `CronJob` object:
 
 1. Copy the sample YAML into a `cronjob.yaml` file:
     ```yaml
-     apiVersion: batch/v1beta1
+    apiVersion: batch/v1beta1
     kind: CronJob
     metadata:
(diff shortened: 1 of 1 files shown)


docs/eventing/sources/sinkbinding.md
```diff
@@ -104,9 +104,11 @@ Create a `CronJob` object:
               restartPolicy: Never
               containers:
                 - name: single-heartbeat
-                  image: <FILL IN YOUR IMAGE HERE>
+                  image: busybox # Or your preferred image
                   args:
-                    - --period=1
+                  - /bin/sh
```
csantanapr, line 109: I don't think this is a good change, as the example wants to give the user a cronjob example with an image that send a HTTP post cloud event to the to be injected K_SINK which in this case is asking the user to replace with heartbeats image.

Maybe what's need it is in image more info about the heartbeat image location to be inserted, or we can put a predefined image from gcr and skip the step on creating the image.

docs/eventing/sources/sinkbinding.md
```diff
@@ -87,13 +87,13 @@ Create a `CronJob` object:
 
 1. Copy the sample YAML into a `cronjob.yaml` file:
     ```yaml
-    apiVersion: batch/v1
+     apiVersion: batch/v1beta1
```
abrennan89, line 90: Is this indent correct? :thinking:

docs/eventing/sources/sinkbinding.md
abrennan89: Agreed, to me `busybox` etc doesn't mean anything / isn't helpful here.
Can we work on getting some standard image to use here @csantanapr or can someone provide more info to clarify this? Could you assign an appropriate SME?

abrennan89:
/hold

Needs clarification from SMEs on technical accuracy as per @csantanapr 's comment.

docs/eventing/sources/sinkbinding.md
lionelvillard: the heartbeat image is published there: gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats.

KauzClay: I see, I didn't realize the image needed for this chronjob example needed to send an HTTP post cloud event. Busybox doesn't make sense for that.

> Maybe what's need it is in image more info about the heartbeat image location to be inserted

This would help, at least making a specific point that not just any chronjob will do. You need one with the proper cloud event behavior.

docs/eventing/sources/sinkbinding.md
KauzClay: no, should be fixed now.

docs/eventing/sources/sinkbinding.md
csantanapr: Like @lionelvillard indicated use the image `gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats`

csantanapr (changes requested):
use image `gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats` remove busydbox

csantanapr (approved):

PR #3191, opened 2021-01-26
[#3085] Enhance PingSource example with kn

### LINKED ISSUE #3085: PingSource example(docs/eventing/samples/ping-source/index.md) enhancement
This page has a nice and simple to follow example. 
It would be awesome to leverage the tabs for choices at multiple sections to showcase [kn](https://knative.dev/docs/install/install-kn/#install-kn-using-an-executable-binary) rather than showing 2 kubectl alternatives. 

Why?
- it starts teaching folks how & when to use kn
- by having the plain yaml beside it shows the declarative way of achieving things ( following the same pattern as kubectl)

**Additional context**
Replace the "By filename" tabs on each section with showing the "Kn" option

xtreme-sameer-vohra:
Fixes #3085

## Proposed Changes
- use kn rather than `kubectl -f some-file.yaml` as thats a wasted
  opportunity and doesn't add any value.
- show examples of kn side by side with the declarative form

### DIFF
--- a/docs/eventing/samples/ping-source/index.md
+++ b/docs/eventing/samples/ping-source/index.md
@@ -20,8 +20,8 @@ a Knative Service.
 To verify that `PingSource` is working, create a simple Knative
 Service that dumps incoming messages to its log.
 
-{{< tabs name="create-service" default="By YAML" >}}
-{{% tab name="By YAML" %}}
+{{< tabs name="create-service" default="YAML" >}}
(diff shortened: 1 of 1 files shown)


omerbensaadon (approved):
I think this is an easy and manageable way to expose the kn cli to users without overwhelming them. 

It also means that that folks don't need to dig around yaml and can see what changes they are affecting with each command.

/lgtm

PR #3204, opened 2021-02-01
[#3196] Follow up on Evan's comments

### LINKED ISSUE #3196: Move Concept information into introduction
There is currently only one overview page in the Concepts doc section: https://knative.dev/docs/concepts/overview/
This is confusing because it's really just a duplicate / information that should belong in the landing page. It doesn't really have any in-depth information explaining Knative concepts.
I'm proposing to remove this section until we can create an actual concepts section with information about different terms, since for now it's just adding additional topics to the nav for no real payoff.
I will open a PR to move the information from this landing page to the overview if there are no objections.

**Additional context**
https://knative.dev/docs/concepts/overview/

abrennan89:
Fixes #3196 

## Proposed Changes
- Follow up @evankanderson 's comments from #3197

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -3,31 +3,31 @@ title: "Welcome to Knative"
 linkTitle: "Documentation"
 weight: 10
 type: "docs"
+showlandingtoc: "false"
 aliases:
    - /docs/concepts/overview.md
 ---
 
-The Knative project provides a set of [Kubernetes controllers](https://kubernetes.io/docs/concepts/architecture/controller/) that help to simplify developer and administrator experiences on [Kubernetes](https://kubernetes.io) clusters, by introducing an event-driven, serverless architecture.
+The Knative project provides a set of [Kubernetes](https://kubernetes.io) components that introduce event-driven, serverless capabilities for Kubernetes clusters.
 
-There are two core Knative components, that can be installed and used together or independently to provide different functions:
+Knative APIs build on existing Kubernetes APIs, so that Knative resources are compatible with other Kubernetes-native resources, and can be managed by cluster administrators using existing Kubernetes tools.
 
-* [Knative Serving](https://knative.dev/docs/serving/): Run stateless services more easily on Kubernetes,
-   by making autoscaling, networking and rollouts easier.
-
-* [Knative Eventing](https://knative.dev/docs/eventing/): Create subscriptions to event sources declaratively,
-   and route events to Kubernetes endpoints.
-
-These components are delivered as Kubernetes custom resource definitions (CRDs), which can be configured by a cluster administrator to provide default settings for developer-created applications and event workflow components.
-
-**Note**: Earlier versions of Knative included a Build component.  That component has been broken out into its own
-project, [Tekton Pipelines](https://tekton.dev/).
-
-Knative registers its own API types to the Kubernetes API, enabling developers and cluster administrators to use familiar languages and frameworks to solve common use cases, such as:
+Common languages and frameworks, which include Kubernetes-friendly tooling, work smoothly with Knative to reduce the time spent solving common deployment issues, such as:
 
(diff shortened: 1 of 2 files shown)


docs/_index.md
```diff
@@ -3,32 +3,32 @@ title: "Welcome to Knative"
 linkTitle: "Documentation"
 weight: 10
 type: "docs"
+showlandingtoc: "false"
 aliases:
    - /docs/concepts/overview.md
 ---
 
-The Knative project provides a set of [Kubernetes controllers](https://kubernetes.io/docs/concepts/architecture/controller/) that help to simplify developer and administrator experiences on [Kubernetes](https://kubernetes.io) clusters, by introducing an event-driven, serverless architecture.
+The Knative project provides a set of [Kubernetes](https://kubernetes.io) components that introduce event-driven, serverless capabilities for Kubernetes clusters.
```
evankanderson, line 11: "event driven and serverless" -- you can choose either, but don't need to take both.

docs/_index.md
```diff
@@ -3,32 +3,32 @@ title: "Welcome to Knative"
...
-* [Knative Eventing](https://knative.dev/docs/eventing/): Create subscriptions to event sources declaratively,
-   and route events to Kubernetes endpoints.
-
-These components are delivered as Kubernetes custom resource definitions (CRDs), which can be configured by a cluster administrator to provide default settings for developer-created applications and event workflow components.
-
-**Note**: Earlier versions of Knative included a Build component.  That component has been broken out into its own
-project, [Tekton Pipelines](https://tekton.dev/).
-
-Knative registers its own API types to the Kubernetes API, enabling developers and cluster administrators to use familiar languages and frameworks to solve common use cases, such as:
+Common languages and frameworks, which include Kubernetes-friendly tooling, work smoothly with Knative to reduce the time spent solving common deployment issues, such as:
```
evankanderson, line 15: I think putting a `, which` here confuses things -- I think we mean "languages and frameworks which include K8s-friendly tooling" in contrast to languages and tools which don't have K8s-friendly tooling. (For example, Bash doesn't have a whole lot of tools to really help with K8s.)

This probably means removing both commas, since I think the "include K8s-friendly tooling" was meant as an aside, but in this case I think it's a qualifier instead.

evankanderson (commented):
Two points (amusingly, both involve commas).

I think we do need to fix the second one; the first one is a nit.

docs/_index.md
abrennan89: Good point

docs/_index.md
abrennan89: lol actually similar to the other comment on the other PR, but in this case it would then be "that", not "which". That is effectively the difference between the two; "that" is for a specific, restrictive clause, "which" is for non-restrictive clauses.
Good point though again, thanks for the attention to detail :heart_eyes: 

https://developers.google.com/style/pronouns?hl=en#relative-pronouns

abrennan89:
@evankanderson should be ready for a merge now hopefully.
Do you know if there's any good way to do automatic cherrypicking to the last couple of versions?
If not it can just be for the new ones, no biggy, just curious. Thanks!

omerbensaadon (commented):
lgtm!

csantanapr:
@abrennan89 once the PR is merged we can add a comment to cherry pick to specific branch/release
https://github.com/knative/docs/blob/master/CONTRIBUTING.md#choosing-the-correct-branch

abrennan89:
Couple of the links seem broken so I'll update those before removing the hold

abrennan89:
/lgtm
Re-adding the lgtm since @csantanapr had already approved this and I just needed to fix the links

abrennan89:
FFS :joy: can someone who ISN'T me lgtm this please

abrennan89:
Just gonna CP it manually...

PR #3206, opened 2021-02-02
update docs for topic Installing Istio for Knative

james426759:
Signed-off-by: james759426 <james759426@gmail.com>

Fixes #

## Proposed Changes 

- update docs for topic Installing Istio for Knative v.0.20

   old: Replace the `local-gateway.knative-serving.cluster-local-gateway` field with the custom service. As an example, if you name both

   new: Replace the `local-gateway.knative-serving.knative-local-gateway` field with the custom service. As an example, if you name both

### DIFF
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -141,7 +141,7 @@ need to update gateway configmap `config-istio` under the `knative-serving` name
 kubectl edit configmap config-istio -n knative-serving
 ```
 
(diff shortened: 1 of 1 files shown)


nak3:
/assign @arturenault @JRBANCEL @ZhiminXiang 
/ok-to-test

We need to update all `cluster-local-gateway` in the docs https://knative.dev/docs/install/installing-istio/ should be updated to `knative-local-gateway`?

csantanapr (approved):
Thanks for the update the details are in the config map for net-istio https://github.com/knative-sandbox/net-istio/blob/master/config/config.yaml

PR #3215, opened 2021-02-05
[#2781] Removing unnecessary version labels

### LINKED ISSUE #2781: Feature state labels are unclear in certain contexts (we should change '@' to 'since')
## Expected Behavior

1. On pages such as https://knative.dev/docs/install/any-kubernetes-cluster/#installing-the-serving-component, the Serving link should not be suffixed by "stable @ v0.9" in the opening paragraph, because to at least one reader (hi! 👋) this looks like we're suggesting the current stable version is 0.9.
1. It's fine to have "stable @ v0.9" when this is preceded by "Feature State:" since this makes it clear what the '@' means, but:
1. Even better (imo) would be just to use "stable since v0.9" throughout, since this is clear in all contexts

## Actual Behavior

1. The opposite of the above.

1. Visit https://knative.dev/docs/install/any-kubernetes-cluster/#installing-the-serving-component
1. Read the opening paragraph
1. Observe "Serving `stable @ v0.9` provides an abstraction for stateless request-based scale-to-zero services." rather than something like "Serving `stable since v0.9` provides an abstraction for stateless request-based scale-to-zero services."
1. Repeat throughout docs.

/assign @abrennan89

abrennan89:
Fixes #2781 

Removing all the labels and some other minor general cleanup.

IMHO we should remove whatever we can from the install docs ahead of reorganizing them to take out all the optional stuff like networking layers (just keep the default, everything else goes to its own section). It'll make cleaning this up easier later.

### DIFF
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -27,11 +27,9 @@ Knative has two components, which can be installed and used independently or
 together. To help you pick and choose the pieces that are right for you, here is
 a brief description of each:
 
-- [**Serving**](#installing-the-serving-component)
-  {{< feature-state version="v0.9" state="stable" short=true >}} provides an
+- [**Serving**](#installing-the-serving-component) provides an
   abstraction for stateless request-based scale-to-zero services.
-- [**Eventing**](#installing-the-eventing-component)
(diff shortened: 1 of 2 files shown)


csantanapr (approved):
Love to see this confusing versions removed

PR #3217, opened 2021-02-06
fix kn example for pingsource

### LINKED ISSUE #3216: pingsource example not working with kn cli
## Expected Behavior

Be able to do create a pingsource using the know cli

## Actual Behavior
There are two problems.
1. Missing `/` on the command, it doesn't work
2. The sink URL is wrong missing namespace

1. Create event-display pod and service
1. Create kn with command
```
kn source ping create test-ping-source \
  --namespace pingsource-example
  --schedule "*/1 * * * *" \
  --data '{"message": "Hello world!"}' \
  --sink http://event-display.svc.cluster.local
```

It errors out
```
Error: required flag(s) "sink" not set
Run 'kn --help' for usage
zsh: command not found: --schedule
```

After fixing, then event display doesn't get hit with HTTP request by dispatcher
Debug the dispatcher for pingsource
Using stern
```
stern -n knative-eventing -l eventing.knative.dev/source=ping-source-controller
```

```
{"level":"error","ts":"2021-02-06T15:55:03.248Z","caller":"mtping/runner.go:133","msg":"failed to send cloudevent result: {result 26 0  Post \"http://event-display.svc.cluster.local\": dial tcp: lookup event-display.svc.cluster.local on 10.96.0.10:53: no such host}
```

Doc needs updating

Capture all steps in this issue as it serves as troubleshooting/debug info

Most people would not know to look into a different namespace like `knative-eventing` and then look for the pingsource pod

Also users might not have access to the `knative-eventing` namespace to debug

Maybe the pingsource controller should update the pingsource resource status with an error reason condition?
```
Status:
  Ce Attributes:
    Source:  /apis/v1/namespaces/default/pingsources/test-ping-source
    Type:    dev.knative.sources.ping
  Conditions:
    Last Transition Time:  2021-02-06T16:03:47Z
    Status:                True
    Type:                  Deployed
    Last Transition Time:  2021-02-06T16:03:47Z
    Status:                True
    Type:                  Ready
    Last Transition Time:  2021-02-06T16:03:47Z
    Status:                True
    Type:                  SinkProvided
  Observed Generation:     1
  Sink Uri:                http://event-display.svc.cluster.local
```

There is no indication what's wrong in top level printer

```
❯ k get pingsource
NAME               SINK                                     SCHEDULE      AGE     READY   REASON
test-ping-source   http://event-display.svc.cluster.local   */1 * * * *   2m50s   True
```
What you think @n3wscott @vaikas on how to best indicate an error as one of condition maybe

csantanapr:
Closes #3216

### DIFF
--- a/docs/eventing/sources/pingsource.md
+++ b/docs/eventing/sources/pingsource.md
@@ -98,13 +98,13 @@ EOF
 {{< /tab >}}
 
 {{% tab name="kn" %}}
-
+Notice that the namespace is specified in two places in the command in `--namespace` and the `--sink` hostname
 ```shell
 kn source ping create test-ping-source \
-  --namespace pingsource-example
+  --namespace pingsource-example \
   --schedule "*/1 * * * *" \
   --data '{"message": "Hello world!"}' \
-  --sink http://event-display.svc.cluster.local
+  --sink http://event-display.pingsource-example.svc.cluster.local
 ```
 
 {{< /tab >}}


csantanapr:
This is a good issue + pr to tag with something related to `ux/getting-started`

csantanapr:
@omerbensaadon 

>This is a good issue + pr to tag with something related to ux/getting-started

docs/eventing/sources/pingsource.md
```diff
@@ -98,13 +98,13 @@ EOF
...
-
+Notice that the namespace is specified in two places in the command in `--namespace` and the `--sink` hostname
 ```shell
 kn source ping create test-ping-source \
-  --namespace pingsource-example
+  --namespace pingsource-example \
   --schedule "*/1 * * * *" \
   --data '{"message": "Hello world!"}' \
-  --sink http://event-display.svc.cluster.local
+  --sink http://event-display.pingsource-example.svc.cluster.local
```
rhuss, line 107: Any particual reason why the URL is used for the sing and e.g. not `--sink event-display` (if targeting a Knative service within the same namespace) ?

rhuss: Also, you can skip the first namespace argument if you just want to create the source in the _current_ namespace (which also save the step of creating a particular namespace). imo its ok to assume that the user know that commands work on a _current namespace_ if no namespace is provided (an idiom known from `kubectl`)

rhuss: The URL format should be only used if there is no better way to address the sink (e.g. across namespaces). That said, it would be of course nice to support a namespace in the cli sink-syntax, too.

csantanapr: I tried to figured out how to specify the service from CLI help it showed <kind>:name 

The service is not ksvc just a normal k8s service. 

I think tried service:event-display and svc:event-display and I remember getting an error.

csantanapr: About specifying the namespace in the kubectl is following the example on creating the namespace at the start then using the namespace in every kubectl command. 

We don't want to introduce complexity to the new user on using the syntax of switching namespace context (ie kubectl config set-context --current --namsoace ns)

I don't design this example but all.examples in eventing use the pattern so we should be consistent

csantanapr:
Hi @rhuss 

Can you take a look at this PR again.
>

The reason we can't use `--sink event-display` is because event-display in this case is NOT a ksvc/Kantive is just a simple normal k8s service.

I try the following which I think the `kn` CLI should support any resources that is addresable using duck typing or at minimum hard code support for `--sink svc:event-display` or `--sink service:event-display` for normal `Kind: Service`

Here what I tried
```
❯ kn source ping create test-ping-source \
  --namespace pingsource-example \
  --schedule "*/1 * * * *" \
  --data '{"message": "Hello world!"}' \
  --sink event-display
Error: services.serving.knative.dev "event-display" not found
Run 'kn --help' for usage
❯ kn source ping create test-ping-source \
  --namespace pingsource-example \
  --schedule "*/1 * * * *" \
  --data '{"message": "Hello world!"}' \
  --sink service:event-display
Error: unsupported sink prefix: 'service', please use prefix 'ksvc' for knative service
Run 'kn --help' for usage
```

Let's see if we can move forward and address what this PR is trying to do solve which is a broken sample in the docs

vaikas:
To support any arbitrary duck type (say, sequence, broker, etc.), you'd have to know the GVK, so starting with service:NAME or svc: NAME makes sense to me. We could later add more, or just have support for gvk.

csantanapr:
@vaikas thanks for the review, I opened an issue to capture the feature request for the CLI

Could we merge it now?

csantanapr:
/assign @navidshaikh @sixolet 

Assigned this PR to the Client WG according to this https://github.com/knative/community/blob/master/working-groups/WORKING-GROUPS.md#client

Hope WG leads can find a resolution to this PR soon

csantanapr:
/unassign @vaikas 

Unassigning Ville since this PR is assigned to Client WG leads to find a final resolution to this PR

docs/eventing/sources/pingsource.md
rhuss: Sorry, overlooked that it is a standard service.  Yes, then URL is the best option if you don't want to include a mapping for a k8s in the kn configuration. We have not added regular K8s services to the standard mapping because it's outside of Knative. However if people have a strong feeling to include it nevertheless (and e.g. connect it to the prefix  `svc:`), then we could consider this. It's just that it would lead to an error message on non-k8s implementations of Knative (e.g. Cloud Run)

rhuss (commented):
Looks good to me

/lgtm
/approve

PR #3235, opened 2021-02-10
Update supported Istio versions.

arturenault:
This paragraph isn't present in master, so committing directly to release-0.20.


csantanapr:
@arturenault This PR is now against release-20, please create a PR against master, and then we can cherry pick back to other release branches.

Issue #3202, opened 2021-02-01
Document SASL and SSL support for Kafka Broker and KafkaSink


pierDipi:
## What is changing? (Please include as many details as possible.)

Kafka `Broker` and `KafkaSink` support SASL and SSL configurations, we should document how to use them.

## How will this impact our users?

Users can configure SASL and SSL for client authorization and authentication against their Apache Kafka clusters.

## In what release will this take happen (to the best of your knowledge)?

0.21

## Context
Link to associated PRs or issues from other repos here.

1. PR: https://github.com/knative-sandbox/eventing-kafka-broker/pull/534
1. Issue: https://github.com/knative-sandbox/eventing-kafka-broker/issues/368

PR #3239, opened 2021-02-12
Document SASL and SSL support for Kafka Broker

pierDipi:
Part of #3202

## Proposed Changes 

- Document SASL and SSL support for Kafka Broker

/cc @abrennan89 @matzew @slinkydeveloper

### DIFF
--- a/docs/eventing/broker/kafka-broker.md
+++ b/docs/eventing/broker/kafka-broker.md
@@ -144,6 +144,88 @@ data:
         namespace: knative-eventing
 ```
 
+## Security
+
+Apache Kafka supports different security features, Knative supports the followings:
+
+- [Authentication using `SASL` without encryption](#authentication-using-sasl)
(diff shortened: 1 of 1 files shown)


csantanapr:
Thank you @pierDipi for the contribution, I saw a lot of users asking how to configure kafka auth for both Source and Broker having this helps a lot 🙏

csantanapr:
@pierDipi Support for this goes back to 0.17 (oldest documented version, 0.18 soon) right, so after merging this we would need to cherry-pick back to the other branches

pierDipi:
@csantanapr this feature will be released in v0.21 and it's not available in any of the previous versions. 
However, the Broker implementation can be used when a previous version of Knative Eventing is installed.
For example, I can use Knative Eventing v0.18 and the Kafka Broker v0.21 without issues.
Perhaps should we clarify this aspect in our docs?

PR #3251, opened 2021-02-15
Change the title for upgrading with the operator

houshengbo:
## Proposed Changes 

- Use a different title for the upgrading instruction with Knative operator.


tcnghia (approved):
/lgtm

Issue #3222, opened 2021-02-08
Setting up a custom domain(docs/serving/using-a-custom-domain.md)


yangyunfeng007:
There is a dead link.
https://github.com/knative/serving/blob/master/config/config-domain.yaml
![image](https://user-images.githubusercontent.com/17523539/107201759-62997580-6a34-11eb-8482-40893608118b.png)

nak3:
Thank you @yangyunfeng007 for the report. https://github.com/knative/docs/pull/3261 is fixing the issue.

PR #3261, opened 2021-02-18
Update path for config-domain configmap

nak3:
Fixes https://github.com/knative/docs/issues/3222

`config-domain.yaml` has a symlink before https://github.com/knative/serving/blob/release-0.17/config/config-domain.yaml, but the symlink was removed.

So this patch fixes the yaml path to the original file https://github.com/knative/serving/blob/master/config/core/configmaps/domain.yaml.


abrennan89 (approved):
/lgtm

PR #3262, opened 2021-02-18
updated progressDeadline value to seconds format[600s]

vinayGudi:

## Proposed Changes 

-
-
-

### DIFF
--- a/docs/serving/services/deployment.md
+++ b/docs/serving/services/deployment.md
@@ -42,5 +42,5 @@ metadata:
   namespace: knative-serving
 data:
 ...
-  progressDeadline: "10m"
+  progressDeadline: "600s"
 ...


docs/serving/services/deployment.md
```diff
@@ -42,5 +42,5 @@ metadata:
   namespace: knative-serving
 data:
 ...
-  progressDeadline: "10m"
+  progressDeadline: "600s"
```
csantanapr, line 45: Why are you changing to seconds? 

The description above mentions 10 minutes so "10m" seems more natural. 

is the description mentioned 600 seconds then using "600s" then is more natural. 

using such a hi number seconds in an example comes with cognitive load. 

if the example needs to in seconds then a value of 120s or less is easier to reason about.

vagababov: 👍 
I see no resaon for this change.

vinayGudi: but when i checked the actual deployed configmap yaml it is in minutes.
Screenshot from 2021-02-23 11-29-59

csantanapr: Ah I see perhaps would be good idea to update the configMap embedded docs to indicate that a and m are supported and change 600s to 10m

Issue #3271, opened 2021-02-24
Need to replace cluster-local-gateway with knative-local-gateway


nak3:
## Actual Behavior

Since https://github.com/knative-sandbox/net-istio/pull/411, cluster-local-gateway is removed but there are many `cluster-local-gateway` in the docs.

## Additional Info
**Additional context**

For example,
- https://knative.dev/docs/install/installing-istio/ 
- https://knative.dev/docs/serving/setting-up-custom-ingress-gateway/

PR #3289, opened 2021-02-28
Remove note about cluster-local-gateway and istio-knative-extras.yaml

nak3:
`istio-knative-extras.yaml` does not exist since long time ago.
And `cluster-local-gateway` was removed and `knative-local-gateway` is always deployed when `net-istio` is used.

This patch removes the note.

Fixes https://github.com/knative/docs/issues/3271


abrennan89 (approved):
/lgtm

abrennan89:
@nak3 does this need to be cherrypicked to some earlier releases also?

nak3:
Yes, it was removed around v0.20 https://github.com/knative-sandbox/net-istio/commit/058432d749435ef1fc61aa2b1fd048d0c75460ee. Let's cheery-pick for 0.20 and 0.21.

PR #3297, opened 2021-03-01
Cleanup install docs

abrennan89:
- removed README
- cleaned up list of deployment options and headings / structure
- Removed fluff, improved readability / findability
- Fixed TOC / nav headings

### DIFF
--- a/docs/install/_index.md
+++ b/docs/install/_index.md
@@ -1,8 +1,21 @@
 ---
 title: "Installing Knative"
-linkTitle: ""
 weight: 05
 type: "docs"
+showlandingtoc: "false"
 ---
 
-{{% readfile file="README.md" %}}
+You can install the Serving component, Eventing component, or both on your cluster by using one of the following deployment options:
+
+- Using a [YAML-based installation](./any-kubernetes-cluster).
+- Using the [Knative Operator](./knative-with-operators).
+- Following the documentation for vendor managed [Knative offerings](../knative-offerings).
+
+You can also [upgrade an existing Knative installation](./upgrade-installation).
+
+**NOTE:** Currently Knative instructions are only provided for Mac and Linux environments.
+<!-- TODO: Link to provisioning guide for advanced installation -->
+
+## Next steps
+
+- Install the [Knative CLI](./install-kn) to use `kn` commands.
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -1,6 +1,6 @@
 ---
-title: "Installing Knative"
-weight: 06
+title: "YAML-based installation"
+weight: 01
 type: "docs"
 aliases:
   - /docs/install/knative-with-any-k8s
@@ -17,36 +17,17 @@ aliases:
   - /docs/install/knative-with-minikube
   - /docs/install/knative-with-minishift
   - /docs/install/knative-with-pks
+showlandingtoc: "false"
 ---
 
-This guide walks you through the installation of the latest version of Knative.
-Note if you are upgrading an existing installation, follow the
-[instructions here](./upgrade-installation.md).
+You can install Knative by applying YAML files using the `kubectl` CLI.
 
-Knative has two components, which can be installed and used independently or
-together. To help you pick and choose the pieces that are right for you, here is
-a brief description of each:
+## Prerequisites
 
-- [**Serving**](#installing-the-serving-component) provides an
-  abstraction for stateless request-based scale-to-zero services.
-- [**Eventing**](#installing-the-eventing-component) provides
-  abstractions to enable binding event sources (e.g. Github Webhooks, Kafka) and
-  consumers (e.g. Kubernetes or Knative Services).
-
-<!-- TODO: Mention CLI here too and point to CLI docs?-->
-
-## Before you begin
-
-This guide assumes that you want to install an upstream Knative release on a
-Kubernetes cluster. A growing number of vendors have managed Knative offerings;
-see the [Knative Offerings](../knative-offerings.md) page for a full list.
-
-Knative {{< version >}} requires a Kubernetes cluster v1.17 or newer, as well as
-a compatible `kubectl`. This guide assumes that you've already created a
-Kubernetes cluster, and that you are using bash in a Mac or Linux environment;
-some commands will need to be adjusted for use in a Windows environment.
-
-<!-- TODO: Link to provisioning guide for advanced installation -->
+- You have a cluster that uses Kubernetes v1.17 or newer.
+- You have installed the [`kubectl` CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
+- If you have only one node in your cluster, you will need at least 6 CPUs, 6 GB of memory, and 30 GB of disk storage.
+- If you have multiple nodes in your cluster, for each node you will need at least 2 CPUs, 4 GB of memory, and 20 GB of disk storage.
 
 ## Installing the Serving component
 
@@ -493,13 +474,6 @@ kubectl apply --filename {{< artifact repo="serving" file="serving-domainmapping
 
 {{< /tab >}} {{< /tabs >}}
 
-### Getting started with Serving
-
-Deploy your first app with the
-[getting started with Knative app deployment](../serving/getting-started-knative-app.md)
-guide. You can also find a number of samples for Knative Serving
-[here](../serving/samples/).
-
 ## Installing the Eventing component
 
 1. Install the
@@ -850,9 +824,3 @@ To learn more about the VMware sources and bindings, try
 {{< /tab >}}
 
 {{< /tabs >}}
-
-### Getting started with Eventing
-
-You can find a number of samples for Knative Eventing
-[here](../eventing/samples/README.md). A quick-start guide is available
-[here](../eventing/getting-started.md).
--- a/docs/install/knative-with-operators.md
+++ b/docs/install/knative-with-operators.md
@@ -1,84 +1,74 @@
 ---
-title: "Installing Knative components using Operator"
-weight: 10
+title: "Knative Operator installation"
+weight: 02
 type: "docs"
+showlandingtoc: "false"
 ---
 
-Knative provides an [operator](https://github.com/knative/operator) as a tool to install, configure and manage Knative. The Knative operator leverages custom objects
-in the cluster to define and manage the installed Knative software. This guide explains how to install and uninstall
-Knative using Knative operator.
+Knative provides a [Kubernetes Operator](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/) to install, configure and manage Knative.
 
-## Before you begin
-
-Knative installation using the Operator requires the following:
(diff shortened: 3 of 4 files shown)


docs/install/any-kubernetes-cluster.md
```diff
@@ -17,36 +17,17 @@ aliases:
...
   - /docs/install/knative-with-pks
+showlandingtoc: "false"
 ---
 
-This guide walks you through the installation of the latest version of Knative.
-Note if you are upgrading an existing installation, follow the
-[instructions here](./upgrade-installation.md).
+You can install Knative by applying YAML files using the `kubectl` CLI.
 
-Knative has two components, which can be installed and used independently or
```
julz, line 26: Checking this is intentional: I think this drops the fact that Serving and Eventing can be installed independently, which means if you just wanted to install (e.g.) serving, the instructions are now effectively twice as long. I _think_ quite a few people only want one or the other. Do we intend to drop this fact here?

docs/install/any-kubernetes-cluster.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: "Installing Knative"
-weight: 06
+title: "YAML-based installation"
```
evankanderson, line 2: +1, much better title!

docs/install/any-kubernetes-cluster.md
```diff
@@ -17,36 +17,17 @@ aliases:
...
-
-Knative {{< version >}} requires a Kubernetes cluster v1.17 or newer, as well as
-a compatible `kubectl`. This guide assumes that you've already created a
-Kubernetes cluster, and that you are using bash in a Mac or Linux environment;
-some commands will need to be adjusted for use in a Windows environment.
-
-<!-- TODO: Link to provisioning guide for advanced installation -->
+- You have a cluster that uses Kubernetes v1.17 or newer.
+- You have installed the [`kubectl` CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
+- If you have only one node in your cluster, you will need at least 6 CPUs, 6 GB of memory, and 30 GB of disk storage.
```
evankanderson, line 29: Do we really need this much CPU? I suspect 2-4 CPUs are sufficient.

(Fine to leave as a `<!-- TODO: verify -->` comment...)

docs/install/knative-with-operators.md
```diff
@@ -1,84 +1,74 @@
...
 
-```
-git clone https://github.com/knative/operator.git
-```
+    ```
+    git clone https://github.com/knative/operator.git
+    ```
 
-1. Install the operator in the root directory of the source using the following command:
+1. Install the Operator in the root directory of the source:
```
julz, line 42:
```suggestion
1. Install the Operator from the root directory of the source:
```

docs/install/knative-with-operators.md
```diff
@@ -1,84 +1,74 @@
...
 
-## Install Knative with the Knative Operator
+## Prerequisites
 
-You can find the release information of Knative Operator on the [Releases page](https://github.com/knative/operator/releases).
+- You have a cluster that uses Kubernetes v1.17 or newer.
+- You have installed the `kubectl` CLI.
+- If you have only one node in your cluster, you will need at least 6 CPUs, 6 GB of memory, and 30 GB of disk storage.
+- If you have multiple nodes in your cluster, for each node you will need at least 2 CPUs, 4 GB of memory, and 20 GB of disk storage.
+- Your Kubernetes cluster must have access to the internet, since the Knative Operator must download images.
```
evankanderson, line 19: I think the YAML version requires the same thing -- I don't think the Operator is downloading images directly anymore, but Kubernetes will still need to fetch images.

@houshengbo to confirm

docs/install/knative-with-operators.md
```diff
@@ -1,84 +1,74 @@
...
-
-Knative Operator is still in Alpha phase. It has not been tested in a production environment, and should be used
+**NOTE:** The Knative Operator is still in Alpha phase. It has not been tested in a production environment, and should be used
 for development or test purposes only.
 
-## Install Knative with the Knative Operator
+## Prerequisites
 
-You can find the release information of Knative Operator on the [Releases page](https://github.com/knative/operator/releases).
+- You have a cluster that uses Kubernetes v1.17 or newer.
```
julz, line 15: This updated v1.16 to v1.17, but the minimum version is actually now v1.18 I think

docs/install/knative-with-operators.md
```diff
@@ -1,84 +1,74 @@
...
+## Installing the latest release
 
-__From releases__:
+You can find the release information about the Knative Operator on the [Releases page](https://github.com/knative/operator/releases).
 
-Install the latest Knative operator with the following command:
+To install the latest stable Operator release, enter:
 
 ```
 kubectl apply -f {{< artifact org="knative" repo="operator" file="operator.yaml" >}}
```
evankanderson, line 29: Historically, we've been using `--filename` (long-format) flags rather than `-f` (short-format) flags, but I don't have an opinion here.

docs/install/knative-with-operators.md
```diff
@@ -1,84 +1,74 @@
...
 ```
 
-__From source code__:
+## Installing from source
 
-You can also install Knative Operator from source using `ko`.
+You can install the Knative Operator from the source code using the [ko](https://github.com/google/ko) build tool.
```
evankanderson, line 34: Might be worth explaining why you would want to do this, or moving this to https://github.com/knative/operator/tree/master/DEVELOPMENT.md

@houshengbo , WDYT?

docs/install/knative-with-operators.md
evankanderson, line 42: We lost the `ko` requirement and link for build-from-source, but I think we should probably also think about the cases that would lead to building from source in a non-contributor role.

evankanderson (commented):
Three questions for @houshengbo, mostly about the install mode of the operator and checked-in vs internet-fetched YAML manifests.

Other than that, two small nits, but I think we can address the operator fixes in a subsequent PR.

/lgtm

docs/install/_index.md
```diff
@@ -1,8 +1,21 @@
...
-{{% readfile file="README.md" %}}
+You can install the Serving component, Eventing component, or both on your cluster by using one of the following deployment options:
+
+- Using a [YAML-based installation](./any-kubernetes-cluster).
+- Using the [Knative Operator](./knative-with-operators).
+- Following the documentation for vendor managed [Knative offerings](../knative-offerings).
+
+You can also [upgrade an existing Knative installation](./upgrade-installation).
+
+**NOTE:** Currently Knative instructions are only provided for Mac and Linux environments.
```
julz, line 16: I read this as being a bit stronger than it needs to be (or was before). The instructions assume Mac or Linux, but should work with pretty minor modifications on windows. WDYT about something like:

```suggestion
**NOTE:** Currently the Knative install instructions assume you are running Mac or Linux with a bash shell.
```

julz (commented):
lgtm, couple minor comments/questions

docs/install/knative-with-operators.md
csantanapr: It should be phrased in different way the cluster doesn’t need access to the internet but it needs access to the container registry that the referenced images are stored. 
this is because admin has the choice to do airgap installations by copying g the images to internal registry and modify the default image references

docs/install/knative-with-operators.md
csantanapr: I’m a fan of “-f” everywhere

csantanapr: It’s shorter and avoids the cut off of long strings in some web views

docs/install/knative-with-operators.md
csantanapr: This should not be in user land docs it should be in github dev/contrib md

docs/install/knative-with-operators.md
csantanapr: Leave building from source be an advanced topic in sync with github repo so the instructions should be in github serving both users that want to compile and user that want to contribute code and need to compile

docs/install/any-kubernetes-cluster.md
evankanderson: If we're going to have it, we should move it up to `install/_index.md`. I think it's an important call-out, so I'll move it up there.

docs/install/knative-with-operators.md
houshengbo: @evankanderson Right, kube still needs to fetch images, like gcr.io/knative-releases/knative.dev/operator/cmd/operator:v0.21.0.
The operator itself does not.

docs/install/knative-with-operators.md
houshengbo: @evankanderson I can move this section into the dev part of operator.

abrennan89:
OOF, this was merged before I even got to read the feedback :/ I'll add another PR I guess to follow up

docs/install/knative-with-operators.md
abrennan89: +1 for `-f` IMHO

docs/install/any-kubernetes-cluster.md
abrennan89: Agreed @evankanderson - it's actually called out on the main landing page because I agree it's important (https://knative.dev/docs/) but I guess my thinking here was just trying to remove fluff where possible; we're planning to split this into two sections for installing serving and eventing separately soon anyway. I'll add it to the index though in another PR since this one already got merged :confused:

abrennan89: It's in the section above now https://github.com/knative/docs/pull/3297/files/17839389f802f56108e13a19938374327b61891d#diff-8ff3dcba7bbec8163f1d3589ca5a14ed90107bc5b31715ff4f79649f478970e2R8

It's also called out on the main landing page

docs/install/any-kubernetes-cluster.md
abrennan89: Adding a comment

docs/install/knative-with-operators.md
abrennan89: > this is because admin has the choice to do airgap installations by copying g the images to internal registry and modify the default image references

@csantanapr is the air gapped installation documented somewhere? Otherwise I will leave this as "access to the internet" because mentioning this opens a can of worms that I don't have time to dig into documenting right now. Can you please open an issue in the docs repo for air gapped installs if it's not documented or tracked yet?

@houshengbo thanks for confirming, I'm following up the comments in another PR and will update this to say kube instead of the Operator.

docs/install/knative-with-operators.md
abrennan89: Removing this bit in follow up PR

docs/install/knative-with-operators.md
abrennan89: I'm removing the install from source bit anyway as per the comments above

abrennan89: Removing in follow up PR

PR #3302, opened 2021-03-02
Clean up and add to CLI docs

abrennan89:
## Proposed Changes

- Moved kn and kubectl information into one central CLI tools section and provided descriptions on use for each.
- Updated external links so that they will open in a new tab rather than taking users out of knative.dev
- Removed `ko` - cc @houshengbo (please open a different PR to copy this removed content to the developer docs if you want to preserve it, it really doesn't belong in knative.dev)


docs/client/_index.md
```diff
@@ -0,0 +1,70 @@
...
+- <a href="https://kubernetes.io/docs/tasks/tools/install-kubectl/" target="_blank">Install the `kubectl` CLI</a>
+
+## kn CLI
+
+The `kn` CLI cannot be used to install Knative components, but provides a quick and easy interface for creating Knative resources such as services and event sources, without the need to create or modify YAML files directly. The `kn` CLI also simplifies completion of otherwise complex procedures such as autoscaling and traffic splitting.
+
+### Next steps
+
+- [Install the `kn` CLI](./install-kn/)
+- [Connect the `kn` CLI to your cluster](./install-kn#connecting-the-kn-cli-to-your-cluster)
```
rhuss, line 27: `kn` and `kubectl` share the same configuration files to connect to the cluster, so either add this to `kubectl`, too, or leave it out here.

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+aliases:
+  - /docs/install/install-kn
+  - /docs/client/connecting-kn-to-your-cluster
+---
+
+This guide provides details about how you can install the Knative `kn` client (CLI) using various methods.
+
+## Install kn using an executable binary
+
+You can install `kn` by downloading the executable binary for your system and placing it in the system path.
```
rhuss, line 14: Another indication to remove the CLI part above, to be consistent.

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+weight: 10
+type: "docs"
+aliases:
+  - /docs/install/install-kn
+  - /docs/client/connecting-kn-to-your-cluster
+---
+
+This guide provides details about how you can install the Knative `kn` client (CLI) using various methods.
+
+## Install kn using an executable binary
```
rhuss, line 12: kn is always an executable binary. I think the point here is "by downloading from GitHub", so I would add this as title.

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+
+You can install `kn` by downloading the executable binary for your system and placing it in the system path.
+
+Links to the latest stable executable binary releases are available on the <a href="https://github.com/knative/client/releases" target="_blank">`kn` release page</a>.
+
+
+### Install kn using nightly executable binary
```
rhuss, line 19: kn is not only executable at night ;-) It's nightly _built_ so, "by downloading a nightly built executable" would be more appropriate.

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+
+Nightly executable binaries are available for users who want to install the latest pre-release build of `kn`.
```
rhuss, line 21: same here

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+
+**WARNING:** Nightly executable binaries include features which may not be included in the latest Knative release and are not considered to be stable.
```
rhuss, line 23: dito

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+
+Links to the latest nightly executable binaries are available here:
```
rhuss, line 25: dito

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+
+- <a href="https://storage.googleapis.com/knative-nightly/client/latest/kn-darwin-amd64" target="_blank">macOS</a>
+- <a href="https://storage.googleapis.com/knative-nightly/client/latest/kn-linux-amd64" target="_blank">Linux</a>
+- <a href="https://storage.googleapis.com/knative-nightly/client/latest/kn-windows-amd64.exe" target="_blank">Windows</a>
+
+## Install kn using Go
```
rhuss, line 31: "Build kn yourself" ???

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+
+1. Move `kn` into your system path, and verify that `kn` commands are working properly. For example:
+
+      ```
+      kn version
+      ```
+
+## Install kn using brew
+
+For macOs, you can install `kn` by using <a href="https://github.com/knative/homebrew-client" target="_blank">brew</a>.
```
rhuss, line 54:
```suggestion
For macOs and Linux, you can install `kn` by using <a href="https://github.com/knative/homebrew-client" target="_blank">brew</a>.
```

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Installing kn using container images
```
rhuss, line 56:
```suggestion
## Run kn using container images
```

docs/client/install-kn.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+## Using kn with Tekton
+
+See the <a href="http://hub.tekton.dev/tekton/task/kn" target="_blank">Tekton documentation</a>.
+
+## Connecting the kn CLI to your cluster
+
+The `kn` Client uses your `kubectl` client configuration, the kubeconfig file, to connect to your cluster. This file is usually automatically created when you create a Kubernetes cluster. `kn` looks for your kubeconfig file in the default location of `$HOME/.kube/config`.
+
```
rhuss, line 78: We could also add here, that you can specify the configuration also via `--config path/to/config` or by setting the env variable `KUBECONFIG` that points to the config file.

rhuss (commented):
Thanks for cleaning up! Please find my comments inline, happy to help with rephrasing or pimping up the `kn` section in the client overview document.

docs/client/_index.md
```diff
@@ -0,0 +1,70 @@
...
+
+You can use the `kubectl` CLI to apply the YAML files required to install Knative components, and also to create Knative resources, such as services and event sources using YAML.
+
+### Next steps
+
+- <a href="https://kubernetes.io/docs/tasks/tools/install-kubectl/" target="_blank">Install the `kubectl` CLI</a>
+
+## kn CLI
+
+The `kn` CLI cannot be used to install Knative components, but provides a quick and easy interface for creating Knative resources such as services and event sources, without the need to create or modify YAML files directly. The `kn` CLI also simplifies completion of otherwise complex procedures such as autoscaling and traffic splitting.
```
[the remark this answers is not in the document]
abrennan89: I'll move this to the end maybe, but I think it's important to note somewhere that it's basically not a replacement for `kubectl` in that way?

docs/client/_index.md
```diff
@@ -0,0 +1,70 @@
...
+- [Connect the `kn` CLI to your cluster](./install-kn#connecting-the-kn-cli-to-your-cluster)
+
+## ko CLI
+
+The <a href="https://github.com/google/ko" target="_blank">`ko` CLI</a> is designed to simplify the development of Go apps on Kubernetes by
+abstracting away the container image being used, and instead referring to Go
+packages by their <a href="https://golang.org/doc/code.html#ImportPaths" target="_blank">import paths</a>, for example `github.com/kaniko/serving/cmd/controller`.
+
+`ko` is used in the development and release of Knative components, but is not
+intended to be a requirement for end users. Users are only required to `kubectl apply` the released configuration files generated by `ko`.
```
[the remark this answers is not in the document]
abrennan89: Yeah, I didn't really get it before but I've been doing some work on the install section too and we're removing `ko` to install things from source there too, so I get it now. I think this belongs in dev docs, not in knative.dev. I will probably just remove all mentions of `ko` from knative.dev.

docs/client/_index.md
abrennan89: Good to know, I'll maybe move that whole section to here in the `_index` file then

docs/client/install-kn.md
abrennan89: Saying build it yourself doesn't give a clear idea of what's actually required to do that, I think "using Go" is better tbh

docs/client/install-kn.md
abrennan89: How is "running" different to installing it? If there's a difference here we should explain it, otherwise it seems odd to have commands to "run" something in the install section

docs/client/_index.md
```diff
@@ -0,0 +1,70 @@
...
+`KO_DOCKER_REPO=gcr.io/my-docker-repo`.
+
+`ko apply` then passes the generated YAML configuration to `kubectl apply`.
+
+`ko` also supports the following tasks:
+
+- `ko publish` to simply push images and not produce configs.
+- `ko resolve` to push images and output the generated configs, but not
+  `kubectl apply` them.
+- `ko delete` to simply passthrough to `kubectl delete` for convenience.
```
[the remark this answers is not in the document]
abrennan89: as above, I'm gonna remove `ko` completely

docs/client/install-kn.md
abrennan89: Not really, as a user I wouldn't care whether this was downloaded from github or somewhere else, the important part to me is that it's a binary so I know I can just download it and move it to the path and that's it. We don't actually even mention github in this section so to me it doesn't make sense as a title.

docs/client/install-kn.md
abrennan89: If we're going to state this we should show a full example of how both of these are used / done properly, e.g. example commands setting the variable or using the `--config` flag - can you provide this please? Otherwise I think just mentioning it and not explaining how to do it would be confusing for users.

abrennan89 (commented):
Added some responses to your comments @rhuss

docs/client/install-kn.md
rhuss: "Installing" means putting the binary into a permanent place on your system. "Running" it doesn't do it but just executes it. Next time to run it this way you again have potentially download the image again.

The general concept of containers is that you can run code without installing it locally. Therefore you don't use "installing a container image" but "running a container" when describing this action.

Maybe just switch the section title from "installing" to "using" or "starting" ? I think the user is mostly concerned about how she can *run* `kn`, regardless of whether it is installed locally or started from within a container image (or maybe even run with `go run` which would be even another option).

docs/client/install-kn.md
rhuss: I can add an example. Should I do it here or do you prefer a separate PR ?

docs/client/install-kn.md
abrennan89: That was kind of what I thought but thank you for clarifying it :) Yeah I think it makes sense maybe to change it to "using kn" rather than installing in this context, will give this an update.

docs/client/install-kn.md
abrennan89: Totally up to you :) if you want to send me details I can add it in this PR or you can add one on top after this gets merged.

docs/client/install-kn.md
abrennan89: Maybe actually something like "setting up" or "getting started with kn"

abrennan89: > Next time to run it this way you again have potentially download the image again.

If we call this out I think it makes the difference clearer too

rhuss:
/lgtm

I added https://github.com/knative/client/issues/1251 for adding an example for how to connect.

thanks !

rhuss (approved):

PR #3305, opened 2021-03-03
Follow up install doc improvements

abrennan89:
Following up some comments from #3297 

- Removed some fluff (links should go in context docs not procedures, distracting)
- Updated kube min version
- `--filename` -> `-f` for consistency
- Removed install from source, belongs in dev docs

### DIFF
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -24,16 +24,18 @@ You can install Knative by applying YAML files using the `kubectl` CLI.
 
 ## Prerequisites
 
-- You have a cluster that uses Kubernetes v1.17 or newer.
+- You have a cluster that uses Kubernetes v1.18 or newer.
 - You have installed the [`kubectl` CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
 - If you have only one node in your cluster, you will need at least 6 CPUs, 6 GB of memory, and 30 GB of disk storage.
 - If you have multiple nodes in your cluster, for each node you will need at least 2 CPUs, 4 GB of memory, and 20 GB of disk storage.
+<!--TODO: Verify these requirements-->
+- Your Kubernetes cluster must have access to the internet, since Kubernetes needs to be able to fetch images.
 
 ## Installing the Serving component
 
 1. Install the
-   [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
-   (aka CRDs):
+   [custom resource definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
+   (CRDs):
 
    ```bash
    kubectl apply --filename {{< artifact repo="serving" file="serving-crds.yaml" >}}
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -33,18 +33,16 @@ You can install Knative by applying YAML files using the `kubectl` CLI.
 
 ## Installing the Serving component
 
-1. Install the
-   [custom resource definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
-   (CRDs):
+1. Install the required custom resources:
 
    ```bash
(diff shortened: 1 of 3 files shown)


docs/install/knative-with-operators.md
```diff
@@ -12,39 +12,24 @@ for development or test purposes only.
...
 - If you have multiple nodes in your cluster, for each node you will need at least 2 CPUs, 4 GB of memory, and 20 GB of disk storage.
-- Your Kubernetes cluster must have access to the internet, since the Knative Operator must download images.
+- Your Kubernetes cluster must have access to the internet, since Kubernetes needs to be able to fetch images, such as `gcr.io/knative-releases/knative.dev/operator/cmd/operator:<version>`.
+<!--TODO: Verify these requirements-->
 - You have installed [Istio](./installing-istio.md).
 
 ## Installing the latest release
 
-You can find the release information about the Knative Operator on the [Releases page](https://github.com/knative/operator/releases).
+You can information about the different released versions of the Knative Operator on the [Releases page](https://github.com/knative/operator/releases).
```
julz, line 25: missing a verb after 'can' I think

docs/install/any-kubernetes-cluster.md
```diff
@@ -24,25 +24,25 @@ You can install Knative by applying YAML files using the `kubectl` CLI.
...
 - If you have multiple nodes in your cluster, for each node you will need at least 2 CPUs, 4 GB of memory, and 20 GB of disk storage.
+<!--TODO: Verify these requirements-->
+- Your Kubernetes cluster must have access to the internet, since Kubernetes needs to be able to fetch images.
 
 ## Installing the Serving component
 
-1. Install the
-   [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
-   (aka CRDs):
+1. Install the required custom resources:
```
julz, line 36: I think you still need the word 'definitions' here to distinguish from an instance of the CRD, but I may well be being too pedantic and noone cares 😅.

```suggestion
1. Install the required custom resource definitions:
```

julz (commented):
lgtm with a couple tiny nits

docs/install/knative-with-operators.md
abrennan89: Oops! Thank you :slightly_smiling_face:

docs/install/any-kubernetes-cluster.md
abrennan89: I wondered this, however, in the kube docs, they seem to use CustomResourceDefinitions more specifically when talking about the API, and custom resources more as the thing that's actually installed?
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/

I've been trying to nail down for a while when we should refer to what thing. If you think this is important to say though I'm OK with that, I think in general though we maybe need to find a way to make it clearer across docs the difference and when we mean CRs vs CRDs?

abrennan89: I guess maybe it should be CustomResourceDefinition unless we mean an instance of a custom resource like it is in the kube docs, I just hate things being formatted this way in docs :sob: does it make sense to change it to CustomResourceDefinition though?

abrennan89: Imma say "CustomResourceDefinition object" everywhere I think for this...

julz: hah, no, I'm convinced by the argument that the Kube docs say the CR is the thing installed by the CRD and that this is, in fact, a total bike shed. Let's go with your simpler version :D.

abrennan89: So...which one? You know more than me so if you make a call I'm happy to go with it :) sorry just not sure which is the "simpler version you mean" :joy:

julz: I personally think, reflecting on this and the fact it seems like a hellmouth to the bike-shed dimension, we should just go with the one with less words. So I vote "custom resources". If anyone _really_ cares that this is technically wrong (Im not even sure it is now, honestly 😅) we can always change it. I regret bringing it up 😂.

julz (approved):
/lgtm

Issue #3133, opened 2021-01-07
Missing Subscription documentation Knative Eventing(docs/eventing/_index.md)


omerbensaadon:
As I was fixing #2942 (see PR #3132) I realized the "Subscriptions" link also links to code and not documentation. 

I looked for docs for Subscriptions but it doesn't seem to exists. 

## Expected Behavior
When I click on "Subscriptions" in https://knative.dev/docs/eventing/#event-channels-and-subscriptions I should be taken to documentation that explains Subscriptions

## Actual Behavior
This documentation does not exist so it links me to lines of code. 

**Describe the change you'd like to see**
We find a place to put Subscription documentation and link to it in https://knative.dev/docs/eventing/#event-channels-and-subscriptions

Issue #2942, opened 2020-10-19
Channel Link in Knative Eventing(docs/eventing/_index.md)


yelloman:
## Expected Behavior
In my option it is better to bring the reader do documentation than source code if documentation is availibly. 

The Link Channels should be show to (channels/README.md)

## Actual Behavior

In section Event channels and subscriptions the link to Channel bring me to https://github.com/knative/eventing/blob/master/pkg/apis/messaging/v1/channel_types.go#L57. But I think there is a good documentation for Channels unter (channels/README.md).

## Steps to Reproduce the Problem

1. Open the URL: https://knative.dev/docs/eventing/#event-channels-and-subscriptions
1. Use the Link to Channels in the first sentence.

abrennan89:
@omerbensaadon when I look at the latest development version of the docs: https://knative.dev/development/eventing/#event-channels-and-subscriptions
It doesn't appear as if #3132 actually fixed this issue, since it still links to the code? Do you still want to work on this one?

abrennan89:
@omerbensaadon oh I see the issue, the fix was made against the 0.19 branch. Changes like this should be made against master if they need to be carried through to newer versions of the docs rather than just the one specific version. Can you cherry pick that PR to the master branch please?
/assign @omerbensaadon

omerbensaadon:
Sorry @abrennan89 I am just getting used to following up on Github notifications, they were going to an email I never checked for awhile there... 

Pardon the ball drop and thanks for picking this back up :)

PR #3306, opened 2021-03-03
Add Subscription docs with kn commands

abrennan89:
This PR adds a section for subscriptions under the channel docs, with kubectl / kn commands for create, list, describe and delete.

**NOTE:** Need to follow up with `kn subscription update` in another PR because there seem to be a lot of options here and I'm not sure how to doc them / what they all do.


docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+
+    ```yaml
+    apiVersion: messaging.knative.dev/v1beta1
+    kind: Subscription
+    metadata:
+      name: <subscription_name> # Name of the subscription.
+      namespace: default
+    spec:
+      channel:
+        apiVersion: messaging.knative.dev/v1beta1
```
pierDipi, line 47:
```suggestion
        apiVersion: messaging.knative.dev/v1
```

pierDipi (commented):
A small nit.

abrennan89:
Made some additions, including some cleanup and fixes for:
- #2942 
- #3133

abrennan89:
cc @yelloman @omerbensaadon 
This should fix the issues ya'll raised

docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+Create a subscription between a channel and a sink:
+
+```
+kn subscription create <subscription_name> \
+  --channel <Group:Version:Kind>:<channel_name> \
+  --sink <sink_prefix>:<sink_name> \
+  --sink-reply <sink_prefix>:<sink_name> \
+  --sink-dead-letter <sink_prefix>:<sink_name>
+```
+
```
rhuss, line 24: I think we should either explain all options (like `--sink-reply` and `--sink-dead-letter`) and also the naming convention here, i.e. `<sink_prefix>`, or reduce the example to the minimum required.

Let me try to give some raw input here, that we could use here to explain:

* `--channel` is the source for the cloud-events that should be processed. You have to provide the channel name here. If you don't use a default channel (backed by the `Channel` resource), you have to prefix the channel name with the Group/Version/Kind coordinates of the typed channel, e.g. "
" for a Kafka backed channel.

docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+Example command:
+
+```
+kn subscription create mysubscription --channel mychannel --sink ksvc:myservice
+```
+
+This example command creates a channel named `mysubscription`, that routes events from a channel named `mychannel` to a service named `myservice`.
```
rhuss, line 31:
```suggestion
This example command creates a channel named `mysubscription`, that routes events from a channel named `mychannel` to a Knative service named `myservice`. The `ksvc:` prefix is optional here and could be omitted.
```

docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+        deadLetterSink:
+          ref:
+            apiVersion: serving.knative.dev/v1
+            kind: Service
+            name: <service_name>
+            # Configuration settings for event delivery.
+            # This tells the subscription what happens to events that cannot be delivered to the subscriber.
+            # When this is configured, events that failed to be consumed are sent to the deadLetterSink.
+            # The event is dropped, no re-delivery of the event is attempted, and an error is logged in the system.
+            # The deadLetterSink value must be a Destination.
```
rhuss, line 60: Should we add the `reply:` field here, too, so that it shows all possible configurations ? // @knative/eventing-writers (and maybe there are even more options for a subscription that should be added here ?)

docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+{{< /tab >}} {{< /tabs >}}
+
+## List subscriptions
+
+You can list all existing subscriptions by using the `kn` CLI tool.
+
+- List all subscriptions:
+
+    ```
+    kn subscription list
```
rhuss, line 83: Should we add here a sample output so that the users can see what they expect ?

docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+    ```
+- List subscriptions in YAML format:
+
+    ```
+    kn subscription list -o yaml
```
rhuss, line 88: same here (but maybe in a truncated format ...)

docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+    ```
+
+## Describe a subscription
+
+You can print details about a subscription by using the `kn` CLI tool:
+
+```
+kn subscription describe <subscription_name>
+```
+<!--TODO: Add an example command and output-->
```
rhuss, line 98: I can post an example with output later (need to setup a cluster first ;-P

docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+<!--TODO: Add details for kn subscription update - existing generated docs weren't clear enough, need better explained examples-->
```
rhuss, line 99: Update is quite simple, it takes the same options as `create`. I will add some examples later, when I'm doing the second round (and post the example for the commands above)

docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Deleting subscriptions
```
rhuss, line 101:
```suggestion
## Delete subscriptions
```

to be consistent with the other headings ?

docs/eventing/channels/subscriptions.md
```diff
@@ -0,0 +1,120 @@
...
+```
+kn subscription delete <subscription_name>
+```
+
+{{< /tab >}}
+
+{{% tab name="kubectl" %}}
+
+```
+kubectl delete <custom_resource_filename>
```
rhuss, line 117: That's not correct.

It should be either `kubectl delete -f <custom_resource_file>` with `custom_resource_file` must contain the subscription definition (but actually can contain any resource which then will be deleted, also) or better:

```
kubectl delete subscription <subscription_name>
```

which is more specific for subscriptions.

rhuss (commented):
Thanks ! Here's my first round of comments, going to add the promised examples later in a second round.

docs/eventing/channels/subscriptions.md
abrennan89: Yeah we can, could you provide this please with the other examples? Any ETA on these? Should I merge this PR without them and you can iterate in a follow up PR to add examples?

docs/eventing/channels/subscriptions.md
abrennan89: Added an update.

docs/eventing/channels/subscriptions.md
abrennan89: > We could also split off the description of the sink format into a separate document as it can be reused when describing the --sink option for source and triggers.

Let's open a new issue for this / follow up in a separate PR. I'm wary of having too much scope creep in this one.

rhuss (approved):
thanks, let's get that merged and fine tune it in a second round PR.

/lgtm
/approve

PR #3314, opened 2021-03-05
Update Rust version and installation commands

glasnt:
Fixes helloworld-rust example to be buildable. 

## Proposed Changes

Updates installation command based on reported error when building

```
warning: To build the current package use `cargo build`, to install the current package run `cargo install --path .`
```

Updates Rust version from 1.27 to current latest 1.50, which makes the sample build successfully without any other package version changes. 

Successfully builds with these changes.


dprotaso:
/lgtm
/approve

thanks for the fix!

PR #3316, opened 2021-03-05
Fix broken numbering in procedure

snneji:
Add correct spacing in markdown to fix broken numbering in the "Install kn using Go" section of the page.


snneji:
@googlebot I signed it!

PR #3317, opened 2021-03-05
Put command under steps

snneji:
Add spacing to align command with steps and fix broken numbering.


abrennan89 (approved):
Think we removed / are planning to remove `ko` in the new version since it's just for development really, but thanks for fixing this in the meantime :)

/lgtm

PR #3318, opened 2021-03-07
Removed the Go code link

### LINKED ISSUE #3309: Remove links to Go code
We have some pages that link to Go code (especially Go types).
Most of the time while reading the documentation is not expected to see that a link brings you to a low-level representation of the resource you're learning.
In addition to that, users might not be familiar with Go and they should not think that Go is required for learning or using Knative.

## Expected Behavior

No link to Go code.

## Additional information

We should find all links to Go code, and then for each of them, we can replace with an example using YAML or the `kn` CLI

_If you want to tackle a specific page or link, please comment on this issue, so that there is no duplicate effort from multiple people_

/good-first-issue

Yashks1994:
Fixes #3309 - Remove links to Go code

## Proposed Changes 

- Removed the Go code path as mentioned in the fixes.

### DIFF
--- a/docs/eventing/event-registry.md
+++ b/docs/eventing/event-registry.md
@@ -7,7 +7,7 @@ type: "docs"
 ## Overview
 
 The event registry maintains a catalog of event types that can be consumed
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
/lgtm

PR #3321, opened 2021-03-07
Removed the Go path link from the ./eventing/event-delivery.md

### LINKED ISSUE #3309: Remove links to Go code

Yashks1994:
Fixes #3309 - Remove links to Go code

## Proposed Changes 

- Removed the Go path link for Delivery Spec
- Removed the Go path link for Destination

### DIFF
--- a/docs/eventing/event-delivery.md
+++ b/docs/eventing/event-delivery.md
@@ -72,7 +72,33 @@ after the specified number of retries.
 
 ## Common Delivery Parameters
 
-The `delivery` value must be a Delivery Spec.
+The `delivery` value must be a Delivery Spec, which is a partial schema that is embedded in resources like `Broker`, `Trigger` and `Subscription`.
+
+```yaml
+# DeadLetterSink is the sink receiving event that could not be sent to
+# a destination.
+deadLetterSink:
+  ref:
+    apiVersion: v1
+    kind: Service
+    name: my-service
+  uri: /my-path
+
+# Retry is the minimum number of retries the sender should attempt when
+# sending an event before moving it to the dead letter sink.
+retry: 5
+
+// BackoffPolicy is the retry backoff policy (linear, exponential).
+backoffPolicy: exponential
+
+# BackoffDelay is the delay before retrying.
+# More information on Duration format:
+#  - https://www.iso.org/iso-8601-date-and-time-format.html
+#  - https://en.wikipedia.org/wiki/ISO_8601
+#
+# For linear policy, backoff delay is backoffDelay*<numberOfRetries>.
+# For exponential policy, backoff delay is backoffDelay*2^<numberOfRetries>.
+backoffDelay: PT2S```
 
 ### deadLetterSink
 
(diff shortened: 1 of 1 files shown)


docs/eventing/event-delivery.md
```diff
@@ -72,14 +72,14 @@ after the specified number of retries.
 
 ## Common Delivery Parameters
 
-The `delivery` value must be a [Delivery Spec](https://pkg.go.dev/knative.dev/eventing/pkg/apis/duck/v1?tab=doc#DeliverySpec)
+The `delivery` value must be a Delivery Spec.
```
pierDipi, line 75:
```suggestion
The `delivery` value must be a Delivery Spec, which is a partial schema that is embedded in resources like `Broker`, `Trigger` and `Subscription`.

```yaml
# DeadLetterSink is the sink receiving event that could not be sent to
# a destination.
deadLetterSink:
  ref:
    apiVersion: v1
    kind: Service
    name: my-service
  uri: /my-path

# Retry is the minimum number of retries the sender should attempt when
# sending an event before moving it to the dead letter sink.
retry: 5

// BackoffPolicy is the retry backoff policy (linear, exponential).
backoffPolicy: exponential

# BackoffDelay is the delay before retrying.
# More information on Duration format:
#  - https://www.iso.org/iso-8601-date-and-time-format.html
#  - https://en.wikipedia.org/wiki/ISO_8601
#
# For linear policy, backoff delay is backoffDelay*<numberOfRetries>.
# For exponential policy, backoff delay is backoffDelay*2^<numberOfRetries>.
backoffDelay: PT2S```
```

docs/eventing/event-delivery.md
```diff
@@ -72,14 +72,14 @@ after the specified number of retries.
...
 
 ### deadLetterSink
 
 When present, events that failed to be consumed are sent to the `deadLetterSink`.
 In case of failure, the event is dropped and an error is logged into the system.
 
-The `deadLetterSink` value must be a [Destination](https://pkg.go.dev/knative.dev/pkg/apis/duck/v1#Destination).
+The `deadLetterSink` value must be a Destination.
```
pierDipi, line 82: Can you update the example below (line 87) by replacing `<Destination>` with the below YAML?
```YAML
# DeadLetterSink is the sink receiving event that could not be sent to
# a destination.
deadLetterSink:
  ref:
    apiVersion: v1
    kind: Service
    name: my-service
  uri: /my-path
```

docs/eventing/event-delivery.md
Yashks1994: Done

docs/eventing/event-delivery.md
Yashks1994: Removed the existing yaml with the new one.

pierDipi (approved):
/lgtm

PR #3332, opened 2021-03-09
Drop master references to knative repos

markusthoemmes:
A simple search & replace to fix all of the master references to be main.


markusthoemmes:
Ugh, worked from the wrong HEAD...

Issue #3337, opened 2021-03-11
Debugging issues with your application(docs/serving/debugging-application-issues.md)


MandarJKulkarni:
## Expected Behavior
Below 2 links on the [Debugging issues with your application](https://knative.dev/docs/serving/debugging-application-issues/) documentation page are not working.

Check application logs
Check Istio routing

The links are in the [Check revision Status](https://knative.dev/docs/serving/debugging-application-issues/#check-revision-status) section below Check Pod Status sub-section

## Actual Behavior
The links should work.

## Steps to Reproduce the Problem

1. Refer the [Check revision Status](https://knative.dev/docs/serving/debugging-application-issues/#check-revision-status) section in the knative documentation.

PR #3343, opened 2021-03-15
Fix obsolete links in debugging guide

nak3:
This patch fixes obsolete links in debugging guide.

- `Check application logs` is removed by 15153b3e469931b7dcf79aa56fcc0fcf23a42f13 so remove link as well.
- `Check Istio routing` is updated to `Check Ingress/Istio routing` by bbda7cf64b2859ef6fd68d90e82ef05fa458fdcd so rename the link as well.

Fix https://github.com/knative/docs/issues/3337

### DIFF
--- a/docs/serving/debugging-application-issues.md
+++ b/docs/serving/debugging-application-issues.md
@@ -133,8 +133,7 @@ conditions:
 If you see this condition, check the following to continue debugging:
 
 - [Check Pod status](#check-pod-status)
-- [Check application logs](#check-application-logs)
-- [Check Istio routing](#check-clusteringressistio-routing)
+- [Check Istio routing](#check-ingressistio-routing)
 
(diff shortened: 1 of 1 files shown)


docs/serving/debugging-application-issues.md
```diff
@@ -133,8 +133,7 @@ conditions:
 If you see this condition, check the following to continue debugging:
 
 - [Check Pod status](#check-pod-status)
-- [Check application logs](#check-application-logs)
-- [Check Istio routing](#check-clusteringressistio-routing)
+- [Check Istio routing](#check-ingressistio-routing)
```
abrennan89, line 136: Does this link definitely work?
Is it linking from "### Check Ingress/Istio routing", and if so can we change that title to be one or the other instead of with the `/`?

abrennan89 (approved):
/lgtm

PR #3347, opened 2021-03-15
Fixed comments and moved verification step

### LINKED ISSUE #3336: Exposed comments and odd formatting in YAML-based installation
## Actual Behavior

In the topic [YAML-based installation](https://knative.dev/development/install/any-kubernetes-cluster/) in the `main` branch:
1. There are visible comments in a few places on the page. These should be hidden.
2. The final step of the installation procedure looks odd (step beginning with "Monitor the Knative components..."). This applies to both Serving and Eventing sections. This step used to be step 5 in the serving/eventing procedures before the procedure was broken up into sub-sections. This step should be fixed.

Screenshot of the issues mentioned above:

Screen Shot 2021-03-10 at 18 23 20

abrennan89:
Fixes #3336 

## Proposed Changes

- Fixed comments so they're not exposed
- Instead of having monitoring step squashed in under the tabs, I have moved this out under a verification heading, since that's really what it is.
- Removed a few lines of fluff, having verification steps now should make it self explanatory that this is the basic install and everything else is optional, also there's an overhaul of the install docs happening anyway.

### DIFF
--- a/docs/install/any-kubernetes-cluster.md
+++ b/docs/install/any-kubernetes-cluster.md
@@ -53,9 +53,8 @@ To install the serving component:
 
 Follow the procedure for the networking layer of your choice:
 
-      <!-- TODO: Link to document/diagram describing what is a networking layer.  -->
-
-      <!-- This indentation is important for things to render properly. -->
+<!-- TODO: Link to document/diagram describing what is a networking layer.  -->
+<!-- This indentation is important for things to render properly. -->
 
    {{< tabs name="serving_networking" default="Istio" >}}
    {{% tab name="Ambassador" %}}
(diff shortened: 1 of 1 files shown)


docs/install/any-kubernetes-cluster.md
```diff
@@ -53,9 +53,8 @@ To install the serving component:
 
 Follow the procedure for the networking layer of your choice:
 
-      <!-- TODO: Link to document/diagram describing what is a networking layer.  -->
-
-      <!-- This indentation is important for things to render properly. -->
+<!-- TODO: Link to document/diagram describing what is a networking layer.  -->
+<!-- This indentation is important for things to render properly. -->
```
csantanapr, line 57: oh this is the trick to hide them

docs/install/any-kubernetes-cluster.md
```diff
@@ -270,12 +268,21 @@ The following commands install Kourier and enable its Knative integration.
 
 {{< /tab >}} {{< /tabs >}}
 
+
+### Verify the installation
```
csantanapr, line 272: Much better wording "Verify" than "Monitoring"

PR #3357, opened 2021-03-17
Remote serving API spec and point to new location in specs repo

tayarani:
Fixes https://github.com/knative/specs/issues/4

## Proposed Changes 

- Part of https://github.com/knative/specs/issues/4
- Serving API spec moved to specs repo as part of https://github.com/knative/specs/pull/7
-


evankanderson:
This is part of moving all the Knative specifications to a repo associated with the conformance tests so that we can do conformance certification.

I don't think we need this level of detail on the website, so moving it seems fine.

/lgtm
/approve

PR #3379, opened 2021-03-22
Removed README and moved content over to _index.md to resolve issue

### LINKED ISSUE #3236: Broken link to eventing sugar controller URL
## Expected Behavior
The `Installing` section on https://knative.dev/docs/eventing/sugar/ should give a link to an installable YAML URL.

## Actual Behavior
`{{< artifact ...` is shown instead.

mpetason:
with artifact links being built properly

Fixes #3236 

## Proposed Changes 

- removed readme file since it seems like website build is causing issues with resolving an artifact
- moved content into _index.md to match other documentation


docs/eventing/sugar/_index.md
```diff
@@ -6,4 +6,101 @@ type: "docs"
...
+  broker: sugar
+  subscriber:
+    ref:
+      apiVersion: v1
+      kind: Service
+      name: event-display
+EOF
+```
+
+> _Note_: If the named Broker already exists, the Sugar controller will do
```
vaikas, line 99: This seems redundant with L105-106? It's also in L73-74 which makes sense since it's in the Broker section, but here it seems that we don't need it twice?

vaikas (commented):
small nit.

docs/eventing/sugar/_index.md
evankanderson: I agree that this is probably redundant, but it seems to be a direct copy-paste of the existing content (91-92 in README.md), so opening another PR to improve these docs seems equally quick, and it would let this fix go in.

evankanderson (commented):
/lgtm
/approve

PR #3385, opened 2021-03-24
Use a unique name for the scale-to-zero-retention tabs

markusthoemmes:
## Proposed Changes 

The tabs of the retention period are currently broken because they actually point at the tabs from the grace period. This fixes that.


abrennan89 (approved):
/lgtm

PR #3386, opened 2021-03-24
Update links and descriptions of TriggerMesh sources

antoineco:
## Proposed Changes

- Update list of AWS event sources maintained by TriggerMesh
- Fix dead links for the Slack and Zendesk event sources (also maintained by TriggerMesh)

cc @omerbensaadon


snneji (commented):
Looks like an improvement to me. I suggested that you change "a" to "an" in a few locations.

Also, it's good practice to spell out abbreviations such as AWS SNS on first use. For example, the [Amazon website](https://aws.amazon.com/sns/?whats-new-cards.sort-by=item.additionalFields.postDateTime&whats-new-cards.sort-order=desc) uses "Amazon Simple Notification Service (Amazon SNS)" on first use. Not sure if you want to use that here, or if there is a good reason to make an exception to the rule for spelling out abbreviations.

csantanapr:
Thanks, @snneji for the review ❤️ 

/hold

antoineco:
@snneji About the acronyms, that's a good suggestion. The reason why I didn't spell them out is because I have personally never seen anyone referring to Amazon SNS as "Amazon Simple Notification Service". I would even bet a lot of people wouldn't immediately connect the dots if we phrased it that way.

Second reason, it's a table, so very long names like that would their 2 lines, but I'm really just bikeshedding on that one.

The final decision is yours :)

omerbensaadon:
thank you @antoineco 🥇

csantanapr:
I have to agree with @antoineco this AWS services are better known by their acronyms like S3 and SQS and  itself

csantanapr:
@snneji 

You would need to comment on this PR with “@googlebot I consent” to make the CLA bot happy

snneji:
Sure, since you want to save space in the table making an exception for these abbreviations is OK.

antoineco:
The length wasn't my main argument here.

PR #3390, opened 2021-03-24
Fix broken link and tweak wording

snneji:
- Fixed a broken link
- Tweaked the serving and eventing extensions headings so that they are more consistent with the other headings on the page
- Changed the wording in optional extension opening sentences


csantanapr:
thanks @snneji 

/lgtm

PR #3409, opened 2021-04-02
add namespace

wyyxd2017:
[add namespcae](https://github.com/wyyxd2017/docs/commit/852ba29592814de4ecdf436b0d0b7a69270877ec)

### DIFF
--- a/docs/eventing/getting-started.md
+++ b/docs/eventing/getting-started.md
@@ -68,6 +68,7 @@ demonstrate how you can configure your event producers to target a specific cons
      kind: Deployment
      metadata:
        name: hello-display
+       namespace: event-example
      spec:
        replicas: 1
        selector:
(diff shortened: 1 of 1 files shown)


docs/eventing/getting-started.md
```diff
@@ -68,6 +68,7 @@ demonstrate how you can configure your event producers to target a specific cons
      kind: Deployment
      metadata:
        name: hello-display
+       namespace: event-example
```
csantanapr, line 71: The namespace is being handled in the command line `kubectl -n event-example` do not add in the yaml. 
Doing in the command line is more flexible as allows the resource to be deployed to another namespace.
/hold

csantanapr (changes requested):
/hold

docs/eventing/getting-started.md
hatowang: ok，thanks

Issue #3431, opened 2021-04-07
Document Duck Types at a high level


n3wscott:
We need to explain the concept in the docs at a high level to make eventing docs make sense. 

/assign

PR #3434, opened 2021-04-07
Adding duck type concept page

n3wscott:
First pass on what a duck type is.

### DIFF
--- a/docs/concepts/README.md
+++ b/docs/concepts/README.md
@@ -93,7 +93,7 @@ knowlage of `Human` or `Dog`.
 
 ## Knative Duck Types
 
-Knative defines Several duck type contracts that are in use across the project:
+Knative defines several duck type contracts that are in use across the project:
 
 - [Addressable](#addressable)
 - [Binding](#binding)
--- a/docs/concepts/README.md
+++ b/docs/concepts/README.md
@@ -5,7 +5,7 @@ Knative uses duck typing to keep various components loosely coupled, but
 reasoning about a resource's control plane shape and behaviors based on a common
 definition of that contact. If a resource has the same fields in the same schema
 locations as the contract specifies, and the control/data plane behaviors as the
-contract specifies, Knative can use that resource as if it is the generic duck
+common definition specifies, Knative can use that resource as if it is the generic duck
 type with little specific knowledge about the resource type. Some resources may
 choose to opt-in or be opt'ed-in to multiple duck types.
 
--- a/docs/concepts/README.md
+++ b/docs/concepts/README.md
@@ -11,9 +11,9 @@ choose to opt-in or be opt'ed-in to multiple duck types.
 
 <!-- TODO: point to Discovery ClusterDuckType documentation. -->
 
-The most basic usage of duck typing in Knative is our usage of object refs in
-resource specs to "point" to another resource. The contract of the object
-holding the ref will prescribe the expected duck type of the "pointee".
+A fundamental use of duck typing in Knative is the use of object references in
+resource _specs_ to point to another resource. The definition of the object
+containing the reference prescribes the expected duck type of the resource being referenced.
 
 In the following example, a Knative `Example` resource named `pointer` references a
 `Dog` resource named `pointee` in its spec:
--- a/docs/concepts/README.md
+++ b/docs/concepts/README.md
@@ -1,0 +1,200 @@
+# Duck Typing
+
+Knative uses duck typing to keep various components loosely coupled, but [what is duck typing](https://en.wikipedia.org/wiki/Duck_typing)? It is reasoning about a resource's control plane shape and behaviors based on a common definition of that contact. If a resource has the same fields in the same schema locations as the contract specifies, and the control/data plane behaviors as the contract specifies, Knative can use that resource as if it is the generic duck type with little specific knowledge about the resource type. Some resources may choose to opt-in or be opt'ed-in to multiple duck types.
+
+<!-- TODO: point to Discovery ClusterDuckType documentation. -->
+
+The most basic usage of duck typing in Knative is our usage of object refs in resource specs to "point" to another resource. The contract of the object holding the ref will prescribe the expected duck type of the "pointee". 
+
+For example, take this example Knative resource `pointer` that is pointing to a `Dog` named `pointee`:
+
+```yaml
+apiVersion: sample.knative.dev/v1
+kind: Example
+metadata:
+  name: pointer
+spec:
+  size:
+    apiVersion: extension.example.com/v1
+    kind: Dog
+    name: pointee
+````
+
+Assume the expected shape of a "Sizable" duck type is that in the status, we expect the following schema shape:
+
+```yaml
+<standard metadata>
+<spec ignored for Sizable>
+status: 
+  height: <in inches>
+  weight: <in pounds>
+```
+
+Now the instance of `pointee` could look like this:
+
+```yaml
+apiVersion: extension.example.com/v1
+kind: Dog
+metadata:
+  name: pointee
+spec:
+  owner: Smith Family
+  etc: more here
+status: 
+  lastFeeding: 2 hours ago
+  hungry: true
+  age: 2
+  height: 27
+  weight: 70
+```
+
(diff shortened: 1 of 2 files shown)


docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
+# Duck Typing
+
+Knative uses duck typing to keep various components loosely coupled, but
+[what is duck typing](https://en.wikipedia.org/wiki/Duck_typing)? It is
+reasoning about a resource's control plane shape and behaviors based on a common
+definition of that contact. If a resource has the same fields in the same schema
+locations as the contract specifies, and the control/data plane behaviors as the
```
mpetason, line 7:
```suggestion
locations as the contract specifies, and the control/data plane behaves as the
```
Maybe behaves? I could be wrong.

docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
...
+  etc: more here
+status:
+  lastFeeding: 2 hours ago
+  hungry: true
+  age: 2
+  height: 27
+  weight: 70
+```
+
+When the `Example` resource needs to do it's work, it only acts on the
```
mpetason, line 62:
```suggestion
When the `Example` resource needs to do its work, it only acts on the
```

docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
...
+  height: 62
+  weight: 120
+```
+
+The `Example` resource was able to do the logic it is set to do without explicit
+knowlage of `Human` or `Dog`.
+
+## Knative Duck Types
+
+Knative defines Several duck type contracts that are in use across the project:
```
mpetason, line 99:
```suggestion
Knative defines several duck type contracts that are in use across the project:
```

mpetason (commented):
small updates. looks good!

docs/concepts/_index.md
```diff
@@ -0,0 +1,8 @@
+---
+title: "Knative concepts"
+linkTitle: "concepts"
+weight: 100
+type: "docs"
+---
+
+{{% readfile file="README.md" %}}
```
abrennan89, line 8: We actually don't need to do this anymore - you can add the content straight into the _index file and delete the README.

FWIW, if the only thing in this file is duck typing, I'd just make the file be called ducktyping.md instead of index, since I think this is a big enough doc to have it's own section like Concepts (folder) > Duck typing

docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
+# Duck Typing
+
+Knative uses duck typing to keep various components loosely coupled, but
+[what is duck typing](https://en.wikipedia.org/wiki/Duck_typing)? It is
+reasoning about a resource's control plane shape and behaviors based on a common
+definition of that contact. If a resource has the same fields in the same schema
```
abrennan89, line 6:
```suggestion
Knative enables [loose coupling](https://en.wikipedia.org/wiki/Loose_coupling) of its components by using [duck typing](https://en.wikipedia.org/wiki/Duck_typing). Duck typing means that the compatibility of a resource for use in a Knative system is determined by certain properties being present that can be used to identify the resource control plane shape and behaviors. These properties are based on a set of common definitions for different types of resources, called duck types. If a resource has the same fields in the same schema
```

abrennan89: You don't have to use this but I suggested some stuff that made this make more sense to me 🙂

docs/concepts/README.md
abrennan89, line 7:
```suggestion
locations as the common definition specifies, and the same control or data plane behaviors as the
```

docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
...
+locations as the contract specifies, and the control/data plane behaviors as the
+contract specifies, Knative can use that resource as if it is the generic duck
```
abrennan89, line 8:
```suggestion
common definition specifies, Knative can use that resource as if it is the generic duck
```

docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
...
+type with little specific knowledge about the resource type. Some resources may
+choose to opt-in or be opt'ed-in to multiple duck types.
```
abrennan89, line 10:
```suggestion
type, without specific knowledge about the resource type. Some resources may
choose to opt-in to multiple duck types.
```

abrennan89: I'd explain what it means to "opt in" better - does this just mean that they can use properties that would make them acceptable as more than one duck type?

docs/concepts/README.md
abrennan89: I'd suggest changing this just because the term contract seems very abstract when we're already explaining some pretty abstract stuff, but if contract here means something specific feel free to ignore me

docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+<!-- TODO: point to Discovery ClusterDuckType documentation. -->
```
abrennan89, line 12: I'd make Discovery a different topic to Duck types under the Concepts folder, like I mentioned in the other comment, re not having this all in the index file. I think it'll be easier to read if it's a page per topic.

docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+The most basic usage of duck typing in Knative is our usage of object refs in
+resource specs to "point" to another resource. The contract of the object
+holding the ref will prescribe the expected duck type of the "pointee".
```
abrennan89, line 16:
```suggestion
A fundamental use of duck typing in Knative is the use of object references in
resource _specs_ to point to another resource. The definition of the object
containing the reference prescribes the expected duck type of the resource being referenced.
```

docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+For example, take this example Knative resource `pointer` that is pointing to a
+`Dog` named `pointee`:
```
abrennan89, line 19:
```suggestion
In the following example, a Knative `Example` resource named `pointer` references a
`Dog` resource named `pointee` in its spec:
```

docs/concepts/README.md
```diff
@@ -0,0 +1,217 @@
...
+Assume the expected shape of a "Sizable" duck type is that in the status, we
+expect the following schema shape:
+
+```yaml
+<standard metadata>
+<spec ignored for Sizable>
+status:
+  height: <in inches>
+  weight: <in pounds>
+```
```
abrennan89, line 42: Since we don't talk about schemas anywhere else really until now, it might be good to explain what this actually is in terms of Knative resources? If the YAML example is basically an `Example` resource the same as the one above, this isn't obvious from the explanation, because it's not obvious that a schema and a resource are the same thing. I'd stick with just saying "resource" to avoid confusion if that's the case. I also think we should never abbreviate examples with things like `<standard metadata` because we're assuming a reader is more familiar with this than they might actually be, and again it's not clear if its the same thing as the resource metadata above.

abrennan89: Reading through this bit and the Human example bit I get what you're saying, but I think the way it's written is sort of confusing. I don't have time really to rewrite/edit it all for now, so I'm happy enough to merge as is when you're ready, but let's maybe put it on our to do list to revisit how we're explaining this later to see if it can be clearer 🙂

abrennan89 (changes requested):
Added some comments, I think this could be edited for readability later but the information so far is excellent, thanks so much @n3wscott 🎉

docs/concepts/README.md
n3wscott: Sure! I was just rolling with one concept as a start

docs/concepts/README.md
n3wscott: it means I can make a third party resource into a duck type (if it conforms to the shape) without the third party realizing this. the double hop integration

docs/concepts/README.md
n3wscott: the whole point I am drafting it and showing it :D make it make sense to others...

docs/concepts/_index.md
n3wscott: ACK, as a followup (maybe on your side?) :D

n3wscott:
Thanks for the review!! looks like CLA bot is ok with the edits 😌

docs/concepts/_index.md
abrennan89: lol I suppoooooose so.... 😜

omerbensaadon:
Closes https://github.com/knative/docs/issues/3431 ?

abrennan89:
/approve
/lgtm

Let's get this merged and iterate improvements later @n3wscott ?

PR #3438, opened 2021-04-08
[KafkaSink] Fix auth secret reference

pierDipi:
The YAML key for the secret reference is wrong.

## Proposed Changes 

- Fix auth secret reference


evankanderson (commented):
/lgtm
/approve

PR #3441, opened 2021-04-08
Add context to network layer install section

### LINKED ISSUE #3377: Add context as to why users need to add a Networking layer, add a recommendation for prototyping (development/install/any-kubernetes-cluster.md)
## Expected Behavior
When a new users comes to the "Installing a networking layer" section of the YAML-based install page, they should understand why Knative Serving needs a Networking layer, what it enables them to do and should be instructed to choose the Networking layer they are most familiar with/is already installed in their cluster. 

In the event users do not have a Networking layer they are familiar with (or the one they are familiar with is not supported in the documentation), we should make a recommendation as to which Networking layer to use **strictly for prototyping purposes.**

Based on a fairly lax Slack survey, I recommend we tell users Kourier would probably be easiest, while noting that production use-cases may work better with Istio. 

## Actual Behavior
When a new users comes to the "Installing a networking layer" section of the YAML-based install page, they are given no context as to why they are being asked to install Networking. Moreover, the Networking multi-select defaults to "Istio" which is a heavy-handed networking layer to use to "Get Started" 

This sends users on a goose-chase through the Istio docs, where they leave the site at an average rate of 45% (15% higher than site average). In both rounds of interviews we did, we had a grouping of User Feedback which was titled "Istio! Arg!" 

1. Go to https://knative.dev/development/install/any-kubernetes-cluster/#installing-a-networking-layer
1. Commence goose-chase

## Additional Info
**Additional context**
Child issue of knative/ux#26 part of the ongoing efforts in knative/ux#1

snneji:
Closes #3377

### DIFF
--- a/docs/install/install-serving-with-yaml.md
+++ b/docs/install/install-serving-with-yaml.md
@@ -34,7 +34,8 @@ To install the serving component:
 
 ## Install a networking layer
 
-The networking layer in Knative Serving routes incoming and outgoing traffic for your Knative installation.
+The networking layer in Knative Serving is responsible for incoming requests and the associated
+responses for your Knative installation.
 Knative enables you to choose a networking layer and then translates Knative’s resources
 into the specific configuration for your chosen networking solution.
 
--- a/docs/install/install-serving-with-yaml.md
+++ b/docs/install/install-serving-with-yaml.md
@@ -35,13 +35,13 @@ To install the serving component:
 ## Install a networking layer
 
 The networking layer in Knative Serving routes incoming and outgoing traffic for your Knative installation.
-Knative enables you to choose a networking layer and then converts Knative’s resources
+Knative enables you to choose a networking layer and then translates Knative’s resources
 into the specific configuration for your chosen networking solution.
 
 You can choose a networking layer based on what you are familiar with, what is
 already installed on your cluster, or what is suitable for your use case.
 For prototyping purposes, if you have no preference for a networking layer, choose Kourier.
-For production use cases, Istio might be a better choice because it has more features.
+For production use cases, Istio is a common choice.
 
 The tabs below expand to show instructions for installing a networking layer.
 Follow the procedure for the networking layer of your choice:
--- a/docs/install/install-serving-with-yaml.md
+++ b/docs/install/install-serving-with-yaml.md
@@ -34,6 +34,17 @@ To install the serving component:
 
 ## Install a networking layer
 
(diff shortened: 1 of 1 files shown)


docs/install/install-serving-with-yaml.md
```diff
@@ -34,6 +34,15 @@ To install the serving component:
 
 ## Install a networking layer
 
+The networking layer in Knative Serving routes incoming and outgoing traffic for your Knative installation.
```
evankanderson, line 37: I'd sort of prefer to refer to this as "HTTP router" or "HTTP ingress" (it's more precise, and avoids confusion with e.g. Kubernetes CNI layer).

The networking layer in question here is responsible for incoming requests and the associated responses. It doesn't do anything for outbound traffic from the user containers.

docs/install/install-serving-with-yaml.md
```diff
@@ -34,6 +34,15 @@ To install the serving component:
...
+Knative enables you to choose a networking layer and then converts Knative’s resources
+into the specific configuration for your chosen networking solution.
```
evankanderson, line 39: This is accurate, but I might say "translates" rather than "converts"; I'm not sure I have a particular rationale for that word choice, except that the original Knative resources still exist.

docs/install/install-serving-with-yaml.md
```diff
@@ -34,6 +34,15 @@ To install the serving component:
...
+
+You can choose a networking layer based on what you are familiar with, what is
+already installed on your cluster, or what is suitable for your use case.
+For prototyping purposes, if you have no preference for a networking layer, choose Kourier.
+For production use cases, Istio might be a better choice because it has more features.
```
evankanderson, line 44: How about "Istio is a common choice for production use."

I think highlighting Kourier for prototyping use is super-useful. I'm not quite as sure about making a recommendation for production use, but I don't fele strongly to suggest a different approach here.

evankanderson (commented):
A few terminology commens, but I'm okay with the current language. If you want this to go in as-is, reply with `/hold cancel`. If you decide to make further changes, that will remove the `/lgtm` label and you'll need a re-review (and also to cancel the hold).

/approve
/hold

docs/install/install-serving-with-yaml.md
snneji: Thanks @evankanderson for your suggestions. Would it make sense for me to change the wording to something like:

> The networking layer in Knative Serving is responsible for incoming requests and the associated responses for your Knative installation.

PR #3443, opened 2021-04-08
Make OWNERS_ALIASES valid yaml

evankanderson:
It turns out when I write a large YAML document by hand, I don't produce valid YAML.

## Proposed Changes 

- Make OWNERS_ALIASES valid YAML.


chaodaiG:
I believe prow uses OWNERS file in main branch, so someone will need to manually apply approve label

PR #3460, opened 2021-04-14
Updates to services docs

### LINKED ISSUE #978: Can't find docs on how to create cluster internal-only Knative service

abrennan89:
Fixes #978 

- Cleanup Knative services docs section


mpetason (commented):
Small updates. It looks good. The only thing I really saw was related to something that was written a while ago, that could be a little bit more clear - the image section. Other than that Service is lowercase in some areas and uppercase in others - for Kubernetes Service and Knative Service.

docs/serving/services/creating-services.md
```diff
@@ -1,76 +1,71 @@
...
+       containers:
+        - image: gcr.io/knative-samples/helloworld-go
+          env:
+            - name: TARGET
+              value: "Go Sample v1"
+    ```
+    * `apiVersion`: The current Knative version.
+    * `name`(metadata): The name of the application.
+    * `namespace`: The namespace that the application will use.
+    * `image`: The URL to the image of the application.
```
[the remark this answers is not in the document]
abrennan89: "The URL for the image of the application" I'd say is correct maybe?

docs/serving/services/_index.md
```diff
@@ -1,6 +1,7 @@
 ---
-title: "Knative Services"
-linkTitle: "Services"
-weight: 10
+title: "Knative services"
+weight: 02
 type: "docs"
 ---
+
+Knative services are used to deploy an application. Each Knative service is defined by a route and a configuration, which have the same name as the service, and are contained in a YAML file. Each time the configuration is updated, a new revision is created. Revisions are each backed by a deployment with two [Kubernetes Services](../knative-kubernetes-services).
```
evankanderson, line 7: Reviewdog points out trailing whitespace on this line.

evankanderson, line 7: I'm confused by "are contained in a YAML file" -- the configuration and route are created by the service controller, and there isn't a YAML file used to define them -- their configuration is derived from the configuration of the service.

evankanderson, line 7: The "Revisions are backed by a deployment with two K8s services" seems like an implementation detail.

Maybe "Revisions are immutable snapshots of a particular configuration, and use underlying Kubernetes resources to scale the number of Pods based on traffic." ?

docs/serving/services/creating-services.md
```diff
@@ -1,76 +1,71 @@
 ---
 title: "Creating Knative services"
-weight: 10
+linkTitle: "Creating a service"
+weight: 02
 type: "docs"
 ---
 
-Knative Services are used to deploy an application. Each Knative Service is defined by a Route and a Configuration, which have the same name as the Service, contained in a YAML file. Every time the Configuration is updated, a new Revision is created.
+To create an application, you need to create a YAML file that defines a Knative service. This YAML file specifies metadata about the application, points to the hosted image of the app and allows the service to be configured.
```
evankanderson, line 8: I'm not sure I like "to create an application", because there are lots of other ways to create an application. Maybe "To create an application with Knative, ..." ?

docs/serving/services/creating-services.md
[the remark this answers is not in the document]
evankanderson: I'd say the "URL of the docker image" or "The URL of the docker container".

I tend to take a more expansive view of an application, which might include configuration and backing databases (not covered here for this simple application).

docs/serving/services/creating-services.md
```diff
@@ -1,76 +1,71 @@
...
-   > Note: If you’re deploying an image of your own app, update the name of the app and the URL of the image accordingly.
+     template:
+      spec:
+       containers:
+        - image: gcr.io/knative-samples/helloworld-go
+          env:
+            - name: TARGET
+              value: "Go Sample v1"
+    ```
+    * `apiVersion`: The current Knative version.
```
evankanderson, line 41: Not sure why we call out `apiVersion` but not `kind`. Both of these should be stable in all these examples at this point.

docs/serving/services/using-a-custom-domain-per-service.md
```diff
@@ -0,0 +1,61 @@
...
+1. In a service definition, add the `custom-hostname` annotation:
+
+    ```yaml
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+       metadata:
+       name: hello-world
+       annotations:
+       # the Service FQDN will become hello-world.{default-domain}
+    custom-hostname: hello-world
```
evankanderson, line 47: This needs to be indented 6 more spaces (2 more levels):

```suggestion
          # the Service FQDN will become hello-world.{default-domain}
          custom-hostname: hello-world
```

evankanderson (commented):
A few suggestions, but the move and rewrite overall looks good.

FWIW, it would be super-easy to review a file move (separate from content changes), but I realize that may be harder to handle on your end.

docs/serving/services/creating-services.md
pmbanugo: I would lean more towards `deploy an application`

docs/serving/services/creating-services.md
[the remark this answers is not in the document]
pmbanugo: Maybe also `The URL of the application's image` 😄

docs/serving/creating-domain-mappings.md
```diff
@@ -5,23 +5,21 @@ weight: 64
 type: "docs"
```
pmbanugo, line 5: I just noticed this exist because I've searched the docs for such info and only found https://knative.dev/docs/serving/using-a-custom-domain-per-service/. Although, I think that the title fits both pages. Maybe they can be combined into one.

pmbanugo, line 5: I just noticed this exist because I've searched the docs for such info and only found https://knative.dev/docs/serving/using-a-custom-domain-per-service/. Although, I think that the title fits both pages. Maybe they can be combined into one.

docs/serving/services/creating-services.md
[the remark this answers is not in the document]
abrennan89: > Maybe also The URL of the application's image

There are a lot of reasons I prefer not to use "application's", but in short it's because it personifies an inanimate object and would be confusing if we ever went to have the docs translated. I believe it's not often used in tech writing for those and other reasons.

CC @omerbensaadon probably a good one for the style guide?

"The URL of the container" I guess to keep it simple @evankanderson ? Since it wouldn't necesssarily be docker if they used a different image?

docs/serving/services/creating-services.md
abrennan89: idk either, this was already there before I did the edits so I'd prefer not to block the PR on it since I think this whole bit could do with rewriting / reformatting to fit properly what we're doing now with "callouts" - can you maybe add a follow up issue for this?

docs/serving/services/_index.md
abrennan89: Yeah, reading this back it's confusing, it was supposed to mean basically I think that routes and configs are their own resources / YAML files, but I'll reword it.

docs/serving/creating-domain-mappings.md
abrennan89: I think maybe we should raise an additional issue for this since I don't want to expand the scope here too much; the initial reason for this PR was to make the private services docs more visible, so maybe could you raise one "to clean up custom domain docs"?

pmbanugo: Make sense. I'll create an issue for that.

docs/serving/services/creating-services.md
```diff
@@ -1,76 +1,71 @@
 ---
 title: "Creating Knative services"
-weight: 10
+linkTitle: "Creating a service"
+weight: 02
 type: "docs"
 ---
 
-Knative Services are used to deploy an application. Each Knative Service is defined by a Route and a Configuration, which have the same name as the Service, contained in a YAML file. Every time the Configuration is updated, a new Revision is created.
+To create an application using Knative, you must create a YAML file that defines a Knative service. This YAML file specifies metadata about the application, points to the hosted image of the app and allows the service to be configured.
```
evankanderson, line 8: I'm going to `/lgtm` and `/approve`, but I'm not sure about the "you must create a YAML file", since we also support using the `kn` CLI client and may have other mechanisms (web form?) in the future.

docs/serving/services/creating-services.md
evankanderson: If you don't want to rewrite this, I'm happy to have this be the place where we draw the line on updates.

evankanderson (commented):
/lgtm
/approve

Issue #1923, opened 2019-10-23
Restructure "Eventing component" section


RichieEscarez:
Align content with the developer flow and arrange content into logical sections.

[View proposed nav structure](https://docs.google.com/spreadsheets/d/1L3EAeTX7RRdZhcQZeNmxwN4DCncuImCXA5A1_KQvwSk/edit?ts=5d826691#gid=413198505&range=121:121)

- For all moved or renamed files, add a redirect to that file's frontmatter (#1925)
- [x] Create a Eventing only install section
- [ ] Create new “Event consumers” section
- [ ] Create new “Event producers” section
- [x] Create new “Debugging” section
- [x] Organize “Debugging (and Observabiltiy)” content into new section

abrennan89:
I think the eventing only install section is covered now in the docs here https://knative.dev/docs/install/any-kubernetes-cluster/ ? @carieshmarie wdyt? Can we check off that item in the list?

RichieEscarez:
Reopening for now but we need revisit this

abrennan89:
I think creating issues like this that are one big long list actually end up getting ignored a lot of the time because it's too much scope / too much like an epic rather than a single issue to track a docs change. Maybe try splitting the list up?

abrennan89:
Note to self: Conceptual info re producers / consumers in this PR can be added: https://github.com/knative/docs/pull/3476

abrennan89:
Note to self: Related to https://github.com/knative/ux/issues/39

abrennan89:
For observability and debugging, these require engineering input/work that are already tracked as other issues:
- https://github.com/knative/docs/issues/3044 debugging
- https://github.com/knative/docs/issues/2931 observability
Checking these off the list for this issue

abrennan89:
Once you split out the tasks from this issue (I removed some of the original list items and made them their own issues), what's left is a pretty generic request for information. I think just adding info about "event producers and consumers" out of context doesn't really add any value, so I'd like for us to close this one for now and revisit this when we have bandwidth to flesh out some more conceptual topics about Knative as a whole. Probably makes more sense to do this after we've split docs into different "guides" for different personas as well, so that we know what type of info is required.

PR #3476, opened 2021-04-19
Add to pingsource.md

RichardJJG:
Added material from this Google Doc:

https://docs.google.com/document/d/1V_6Cjqbt0khUhewXnT_3_Hc2L0RDFThy4lHo8z-sEQ8/edit#heading=h.wowd2jl072p0


docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
 ---
-title: "PingSource"
+title: "PingSource example"
 linkTitle: "PingSource"
-weight: 31
+weight: 10
 type: "docs"
+aliases:
+  - ../cronjob-source
```
abrennan89, line 7: I think this needs to be an absolute link, e.g. /docs/whatever/path

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 ---
 
-![version](https://img.shields.io/badge/API_Version-v1beta2-red?style=flat-square)
```
abrennan89, line 8: Not sure why the badge was removed?
@omerbensaadon is there a decision about whether we're keeping these or not? Eng added them afaik.

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+This topic explains event-driven architecture in terms of producers and consumers and, to that end,
+shows how to configure PingSource as an event source targeting a Knative Service.
```
abrennan89, line 11:
```suggestion
This topic explains event-driven architecture in terms of producers and consumers, and shows how to configure PingSource as an event source targeting a Knative service.
```

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
 ---
-title: "PingSource"
+title: "PingSource example"
```
abrennan89, line 2: I'm not sure about the heading, is this just an example topic? Where would context info about the source itself be if it's not here?

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 
-A PingSource produces events with a fixed payload on a specified cron schedule.
+## Before you begin
 
-## Installation
+1. Set up [Knative Serving](../../../serving).
+1. Set up [Knative Eventing](../../../eventing).
 
-The PingSource source type is enabled by default when you install Knative Eventing.
+## Overview
```
abrennan89, line 18: I really don't like overview as a heading - can we just move all the context info to above the "before you begin" instead? That way we're also not mixing steps / instructions and contextual info.

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 
-## Example
+In event-driven architecture the most straightforward use case is a producer creating an event and a
+consumer acting on that event.
```
abrennan89, line 21: I don't think this needs to be repeated since we have "This topic explains event-driven architecture..." above. I'd choose one and remove the other. I think the previous statement is better but its up to you.

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+Producers create events. Using GitHub as an example, the producer is the GitHub webhook that sends
+information when a pull request is created.
+Essentially, you ask the producer for events that they are capable of sending and you tell the
+producer where to send those events.
 
-Create a new namespace called `pingsource-example` by entering the following
-command:
+Producers send the event data per specifications that make it easy for consumers to handle it.
+Knative uses the CloudEvents specifications by default.
+For more information, see the [Cloud Events](https://cloudevents.io/) website.
```
abrennan89, line 32: I don't really think this belongs here honestly, I think all of this info about event-driven etc should be in a main, generic section about event sources, not preceding a procedure on creating an example PingSource?

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 
-```shell
-kubectl create namespace pingsource-example
-```
+### Consumers
```
abrennan89, line 34: same as above, seems like generic info for an "Event sources" landing page

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
-### Creating the Event Display Service
+In the GitHub example above, a consumer might update a ticket in the ticketing system with the data
+from the pull request.
 
-In this step, you create one event consumer, `event-display` to verify that
-`PingSource` is properly working.
 
-To deploy the `event-display` consumer to your cluster, run the following
-command:
+## Create a consumer (Knative service)
```
abrennan89, line 42: Let's not use parentheses in headings, especially since there are many types of consumers not just a ksvc.
It's explained that it's a service in the next paragraph so there's no need to have it here IMO.

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+
+To create a consumer, and verify that `PingSource` is working, create a simple Knative service that
+receives CloudEvents and writes them to stdout.
```
abrennan89, line 45: Formatting - either use `PingSource` or PingSource, cc @omerbensaadon for style guide. Again, if we're going with the k8s, I'd leave out the backticks.

abrennan89, line 45:
```suggestion
To verify that the PingSource is working, create a simple Knative service that receives CloudEvents and writes them to `stdout`.
```

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+
+{{< tabs name="create-service" default="YAML" >}}
+{{% tab name="YAML" %}}
+Create the service from stdin by running:
```
abrennan89, line 49:
```suggestion
Create the service from `stdin` by running:
```
What is `stdin`? I don't think this is explained.

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
-    matchLabels: &labels
-      app: event-display
   template:
-    metadata:
-      labels: *labels
     spec:
       containers:
-        - name: event-display
-          image: gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/event_display
+        - image: gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/event_display
```
abrennan89, line 61: Why remove all of this instead of explaining it?
Also, is it not `event-display`? I could be wrong but would check this with eng.

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+EOF
+```
+{{< /tab >}}
 
----
+{{% tab name="kn" %}}
+Create the service using the kn command-line interface by running:
```
abrennan89, line 67:
```suggestion
Create a service using the `kn` CLI:
```
CLI is a common enough term to abbreviate IMO, we use CLI most other places I believe.

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
-spec:
-  selector:
-    app: event-display
-  ports:
-  - protocol: TCP
-    port: 80
-    targetPort: 8080
-EOF
+```shell
+kn service create event-display --image gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/event_display
```
abrennan89, line 70: again, I'd check this image isn't `event-display` rather than `event_display`

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 ```
+{{< /tab >}}
+{{< /tabs >}}
+
+## Create a producer (PingSource)
```
abrennan89, line 75: Again, remove parentheses, I don't think there's a need to even mention concepts like producer vs consumer in this doc honestly w/ the exception of if you're explaining them as concepts. I'd just say "Create a PingSource"

@omerbensaadon can we also decide whether to use gerunds in headings please in the style guide. It might be nice if we could have something like the IBM style guide IMO as a reference but idk how licensing and funding for that would work upstream.

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 
-### Creating the PingSource
+Create a producer (PingSource) that sends events to your consumer every two minutes.
+The `sink` element in the metadata describes where to send events.
+In this case, events are sent to a service with the name `event-display`.
+This is a tight connection between the producer and consumer.
```
abrennan89, line 80: Again, remove parentheses. Feels over-explained.
I don't think these all need to be separate lines either, it could just be a paragraph? The flow just feels off here IMO.

I think in general this includes too much context between steps of a procedure - IMO all the conceptual info should go in a generic event sources concept doc and the steps should be in a simple procedural topic.
Also, rather than explaining things like `sink` here, can we use the callout workaround style that was established; https://docs.google.com/document/d/1tigolRzKhronvgoAJfszE-XHWWd0LQ8o_j0D73O9KvM/edit#heading=h.ka7avn97fckt

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
       kind: Service
       name: event-display
 EOF
 ```
-
 {{< /tab >}}
 
 {{% tab name="kn" %}}
-Notice that the namespace is specified in two places in the command in `--namespace` and the `--sink` hostname
+Create the event source from the `ping-source.yaml` file by running:
```
abrennan89, line 109:
```suggestion
Create a PingSource by using the `kn` CLI:
```

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
-  --sink http://event-display.pingsource-example.svc.cluster.local
+  --schedule "*/2 * * * *" --data \
+  '{ "message": "Hello world!" }' \
+  --sink ksvc:event-display
 ```
-
 {{< /tab >}}
 {{< /tabs >}}
 
 ## (Optional) Create a PingSource with binary data
```
abrennan89, line 120: I'd make this its own section instead of being optional in an example. By the time a reader gets to here they've created a PingSource with the other methods. Also, please no more parentheses in titles 😂

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 
-Sometimes you may want to send binary data, which cannot be directly serialized in yaml, to downstream. This can be achieved by using `dataBase64` as the payload. As the name suggests, `dataBase64` should carry data that is base64 encoded.
+You might sometimes want to send binary data, which cannot be directly serialized in YAML, to
+downstream. You can do so using `dataBase64` as the payload.
+As the name suggests, `dataBase64` should carry data that is base64-encoded.
```
abrennan89, line 124:
```suggestion
You can send binary data, that cannot be directly serialized by using YAML, to downstream, by using `dataBase64` as the payload. `dataBase64` carries data that is base64-encoded.
```
What is "downstream" here?

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 
-Please note that `data` and `dataBase64` cannot co-exist.
+**Note:** `data` and `dataBase64` cannot co-exist.
```
abrennan89, line 126:
```suggestion
**NOTE:** You can use either `data` or `dataBase64`, but not both.
```
Explain why?

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+
+Create the event source with binary data from stdin by running:
```
abrennan89, line 128: I'd just use PingSource throughout > event source

I think things like `stdin` that refer to something specific should always be in back ticks, and I really think we should explain this cos I don't know what it is. At least maybe add a link explaining it.

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 
-### Verify
+## Verify the message was sent
+
+Verify that the message was sent to the Knative Eventing system by looking at message dumper logs.
+
+{{< tabs name="view-event" default="kubectl" >}}
+{{% tab name="kubectl" %}}
+
+View the logs of the event-display service by running:
```
abrennan89, line 155: Be consistent with names - `event-display`

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+  time: 2019-12-04T14:24:00.000702251Z
   datacontenttype: application/json
 Data,
   {
     "message": "Hello world!"
   }
 ```
 
-If you created a PingSource with binary data, you should also see the following:
+If you created a PingSource with binary data, you also see the following:
```
abrennan89, line 194:
```suggestion
If you created a PingSource with binary data, the output will be similar to the following:
```

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
   datacontenttype: text/plain
 Data,
-  data
+  ZGF0YQ==
 ```
 
-### Cleanup
+## Clean up
+
+Delete the PingSource instance:
```
abrennan89, line 212: I'd remove this because it's duplicated in the tabs

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
 
-Delete the `pingsource-example` namespace and all of its resources from your
-cluster by entering the following command:
+{{< tabs name="delete-source" default="kubectl" >}}
+{{% tab name="kubectl" %}}
+
+Delete the PingSource instance by running:
```
abrennan89, line 217:
```suggestion
Delete the PingSource instance:
```

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+kubectl delete pingsources.sources.knative.dev test-ping-source-binary
 ```
 
-## Reference Documentation
+{{< /tab >}}
 
-See the [PingSource specification](../../reference/api/eventing/#sources.knative.dev/v1beta2.PingSource).
+{{% tab name="kn" %}}
+
+Delete the PingSource instance by running:
```
abrennan89, line 228:
```suggestion
Delete the PingSource instance:
```

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
-## Contact
 
-For any inquiries about this source, please reach out on to the
-[Knative users group](https://groups.google.com/forum/#!forum/knative-users).
+Delete the service instance:
+
+{{< tabs name="delete-service" default="kubectl" >}}
+{{% tab name="kubectl" %}}
+
+Delete the Service instance by running:
```
abrennan89, line 244:
```suggestion
Delete the Service instance:
```

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+```
+
+{{< /tab >}}
+{{< /tabs >}}
 
-## Contact
 
-For any inquiries about this source, please reach out on to the
-[Knative users group](https://groups.google.com/forum/#!forum/knative-users).
+Delete the service instance:
```
abrennan89, line 239: Again, delete, repeated in tabs

abrennan89: Otherwise I'd make it a step, e.g.

1. Delete the service instance:
    `<tab>`

but not include the same line in the tabs

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,261 @@
...
+
+```shell
+kubectl delete service.serving.knative.dev event-display
+```
+
+{{< /tab >}}
+
+{{% tab name="kn" %}}
+
+Delete the service instance by running:
```
abrennan89, line 254:
```suggestion
Delete the service instance:
```

abrennan89 (changes requested):
Left some comments

docs/eventing/sources/pingsource.md
RichardJJG: @omerbensaadon wants it here, AFAIK. We could (should?) re-title this as more of a Getting Started document than a reference topic for PingSource buried way down the subnav.

docs/eventing/sources/pingsource.md
RichardJJG: Same as above, @omerbensaadon asked for this doc to be better furnished with detail as part of an effort to making getting started with Eventing easier.

docs/eventing/sources/pingsource.md
RichardJJG: I'm assuming it's standard input -- https://en.wikipedia.org/wiki/Standard_streams#Standard_input_(stdin)
Is that the case, @lionelvillard?
If stdin, I don't believe it requires backticks. I've just added suggested rules for this to our style guide.

docs/eventing/sources/pingsource.md
RichardJJG: Google Style says to spell it out: https://developers.google.com/style/word-list#letter-c
We can over-rule, of course, but if so that needs to go in our Word & Phrase list first.

docs/eventing/sources/pingsource.md
RichardJJG: The idea is to add just enough conceptual information to put the reader at better ease with the procedure. I can drop the parentheses, though. I like them as a hand-holding way of walking through what's actually going on, but I can see how the reader might think PingSource is the de facto producer.

docs/eventing/sources/pingsource.md
RichardJJG: @lionelvillard, why can't you use both?

docs/eventing/sources/pingsource.md
RichardJJG: I've just added code formatting rules to the Style Guide, which people are free to support/dispute.

docs/eventing/sources/pingsource.md
omerbensaadon: Don't look at me!

omerbensaadon: We may want to update them for v1 though, yeah?

docs/eventing/sources/pingsource.md
RichardJJG: I think "by running" matters because it better orients readers, especially blind ones. I've written "Documenting Commands" rules in the Style Guide given Google's Style Guide is a little light on them.

docs/eventing/sources/pingsource.md
RichardJJG: I think "by running" matters because it better orients readers, especially blind ones. I've written "Documenting Commands" rules in the Style Guide given Google's Style Guide is a little light on them.

docs/eventing/sources/pingsource.md
RichardJJG: I think "by running" matters because it better orients readers, especially blind ones. I've written "Documenting Commands" rules in the Style Guide given Google's Style Guide is a little light on them.

docs/eventing/sources/pingsource.md
RichardJJG: I think "by running" matters because it better orients readers, especially blind ones. I've written "Documenting Commands" rules in the Style Guide given Google's Style Guide is a little light on them.

docs/eventing/sources/pingsource.md
lionelvillard: pingsource is v1beta2. No v1 yet.

docs/eventing/sources/pingsource.md
lionelvillard: Personally I would remove `from stdin` (all over). Brings no value, only confusion.

docs/eventing/sources/pingsource.md
abrennan89: @RichardJJG I'd leave these in since eng added them please.

docs/eventing/sources/pingsource.md
abrennan89: @lionelvillard agree

docs/eventing/sources/pingsource.md
abrennan89: Added to the list, let's overrule it please, I don't think this is an uncommon enough phrase for us to need to spell it out every time.

docs/eventing/sources/pingsource.md
RichardJJG: @abrennan89, it's good content regardless. Where would you rather it be placed?

docs/eventing/sources/pingsource.md
RichardJJG: @abrennan89, it's good content regardless. Where would you rather it be placed?

abrennan89:
Re the comments above, can we go with "by entering the command:" rather than "by running"? Otherwise a good point about accessibility @RichardJJG

docs/eventing/sources/pingsource.md
RichardJJG: Will do (I deleted this by accident anyway, I think)

docs/eventing/sources/pingsource.md
RichardJJG: Yep, I'll remove it.

docs/eventing/sources/pingsource.md
RichardJJG: Given no alias was present before (I think it was added through some merging accident) I'll simply delete it.

docs/eventing/sources/pingsource.md
RichardJJG: Yep

docs/eventing/sources/pingsource.md
RichardJJG: Okay, will do.

docs/eventing/sources/pingsource.md
RichardJJG: Okay

docs/eventing/sources/pingsource.md
RichardJJG: I'll drop the backticks.

docs/eventing/sources/pingsource.md
RichardJJG: I received the much abridged example from the Google doc the engineer wrote; I've not deviated from what the engineer wants.

RichardJJG: Also, good catch -- @vaikas, this should be `event-display` rather than `event_display`, right?

docs/eventing/sources/pingsource.md
RichardJJG: `event-display` rather than `event_display`, right, @vaikas?

docs/eventing/sources/pingsource.md
RichardJJG: Okay, I'll cut back a bit.

docs/eventing/sources/pingsource.md
RichardJJG: Yep, makes sense, thanks.

docs/eventing/sources/pingsource.md
RichardJJG: Good catch, thanks.

docs/eventing/sources/pingsource.md
RichardJJG: I'll delete the duplicate within the tabs.

docs/eventing/sources/pingsource.md
RichardJJG: Formatting for numbered steps across two tabbed sections seems impossible (or very near it -- I've tried many times). I'll simply leave the full instruction outside the tabs as above.

RichardJJG:
> Re the comments above, can we go with "by entering the command:" rather than "by running"? Otherwise a good point about accessibility @RichardJJG

Why do you prefer the wordier "by entering the command:"? 
I've written a "Documenting Commands" section in the style guide, by the way, that makes the case for "running". We can change it to "by entering the command" if that's the better option.

abrennan89:
@RichardJJG I think they're both equally valid and it's mostly personal preference (FWIW I'm not sure there's a better case for using "run" based on what's in the style guide atm?), however I'd prefer:
"To [do X], enter the command:"
over “X the Y by running:” 

We should probably base a decision on what would be easier for non-native speakers / translation down the road, which I believe would be "enter", but we probably need some research to make an informed decision about that. @omerbensaadon can you help with this?

docs/eventing/sources/pingsource.md
abrennan89: https://knative.dev/development/concepts/ maybe? Or the main eventing landing page, but both of those require cleanup first. FWIW there's a task for producers and consumers as part of https://github.com/knative/docs/issues/1923, so I'd do it as part of that rather than adding to pingsource.

docs/eventing/sources/pingsource.md
abrennan89: see above

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
+
+{{< tabs name="create-service" default="YAML" >}}
+{{% tab name="YAML" %}}
+Create the service from standard input by running:
 
 ```shell
-kubectl -n pingsource-example apply -f - << EOF
-apiVersion: apps/v1
-kind: Deployment
+cat <<EOF | kubectl create -f -
```
abrennan89, line 49: not sure what the reasoning is for using `cat` here as opposed to
```
kubectl apply -f - <<EOF
```
Slack comment from @evankanderson :
"I think we should do the following:
- Use kubectl apply for files/objects that the user creates — it works for both “create” and “update”, and the source of truth is their local files.
- Use kubectl edit for files which are shipped as part of the Knative software, like the serving/eventing ConfigMaps."

abrennan89: Added something to the style guide for this

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
+as the destination.
 
 {{< tabs name="create-source" default="YAML" >}}
 {{% tab name="YAML" %}}
+Create the event source from standard in by running:
 
 ```shell
-kubectl create -n pingsource-example -f - <<EOF
-apiVersion: sources.knative.dev/v1beta2
+cat <<EOF | kubectl create -f -
```
abrennan89, line 86: same as above

docs/eventing/sources/pingsource.md
abrennan89: ping @lionelvillard

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
 
-Sometimes you may want to send binary data, which cannot be directly serialized in yaml, to downstream. This can be achieved by using `dataBase64` as the payload. As the name suggests, `dataBase64` should carry data that is base64 encoded.
+**Note:** `data` and `dataBase64` cannot co-exist.
 
-Please note that `data` and `dataBase64` cannot co-exist.
+Create the event source with binary data from standard in by running:
 
 ```shell
-kubectl create -n pingsource-example -f - <<EOF
+cat <<EOF | kubectl create -f -
```
abrennan89, line 124: same as above

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
-## (Optional) Create a PingSource with binary data
+{{% tab name="binary" %}}
+You can send binary data, that cannot be directly serialized by using YAML, to downstream, by using
+`dataBase64` as the payload. `dataBase64` carries data that is base64-encoded.
 
-Sometimes you may want to send binary data, which cannot be directly serialized in yaml, to downstream. This can be achieved by using `dataBase64` as the payload. As the name suggests, `dataBase64` should carry data that is base64 encoded.
+**Note:** `data` and `dataBase64` cannot co-exist.
 
-Please note that `data` and `dataBase64` cannot co-exist.
+Create the event source with binary data from standard in by running:
```
abrennan89, line 121: I'd leave out "from standard" entirely here.
As an example, I'd prefer
"To create an event source by using binary data, enter the command:"

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
       name: event-display
 EOF
 ```
+{{< /tab >}}
+{{< /tabs >}}
+
+
+## Verify the message was sent
+
+Verify that the message was sent to the Knative Eventing system by looking at message dumper logs.
```
abrennan89, line 146: Use "Verify that the message was sent" both places?

docs/eventing/sources/pingsource.md
abrennan89: don't forget the backticks, not sure why these were removed

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
+{{< /tab >}}
+{{% tab name="kail" %}}
+
+Use `kail` to tail the logs of the subscriber by running:
+
+```shell
+kail -l serving.knative.dev/service=event-display -c user-container --since=10m
+```
+
+For more information, see the [`kail`](https://github.com/boz/kail) repository in GitHub.
```
abrennan89, line 166: I'd call this out earlier as an optional prereq, e.g. "if you want to view logs, you need to install kail" and add the link there, so it's not being introduced mid procedure

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
+```
+
+## Clean up
+
+Delete the PingSource instance:
+
+{{< tabs name="delete-source" default="kubectl" >}}
+{{% tab name="kubectl" %}}
+
+Run:
```
abrennan89, line 213: "To delete the PingSource, enter the command:"

I'd also prefer we have one command per code block, it's not clear to me why both are here, does a user need both, or is the second only if they used the binary data?

docs/eventing/sources/pingsource.md
lionelvillard: Both `data` and `dataBase64` specify the CloudEvent payload, that's why those fields are mutually-exclusive. 

Does that answer your question?

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
+```shell
+kubectl delete pingsources.sources.knative.dev test-ping-source
+kubectl delete pingsources.sources.knative.dev test-ping-source-binary
+```
+
+{{< /tab >}}
+
+{{% tab name="kn" %}}
+
+Run:
```
abrennan89, line 224: same as above

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
+
+
+Delete the service instance:
 
-Delete the `pingsource-example` namespace and all of its resources from your
-cluster by entering the following command:
+{{< tabs name="delete-service" default="kubectl" >}}
+{{% tab name="kubectl" %}}
+
+Run:
```
abrennan89, line 240: same as above, explain what the step is maybe

docs/eventing/sources/pingsource.md
```diff
@@ -1,196 +1,257 @@
...
+kubectl delete service.serving.knative.dev event-display
 ```
 
-## Reference Documentation
+{{< /tab >}}
+
+{{% tab name="kn" %}}
 
-See the [PingSource specification](../../reference/api/eventing/#sources.knative.dev/v1beta2.PingSource).
+Run:
```
abrennan89, line 250: same as above

docs/install/any-kubernetes-cluster.md
```diff
@@ -0,0 +1,836 @@
...
+  - /docs/install/knative-with-minishift
+  - /docs/install/knative-with-pks
+showlandingtoc: "false"
+---
+
+You can install Knative by applying YAML files using the `kubectl` CLI.
+
+## Prerequisites
+
+Before installation, you must meet the following prerequisites:
```
abrennan89:
```suggestion
Before you install Knative, ensure that:
```
Just a less wordy suggestion

docs/install/any-kubernetes-cluster.md
```diff
@@ -0,0 +1,836 @@
...
+- You have a cluster that uses Kubernetes v1.18 or newer.
+- You have installed the [`kubectl` CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
+- If you have only one node in your cluster, you will need at least 6 CPUs, 6 GB of memory, and 30 GB of disk storage.
+- If you have multiple nodes in your cluster, for each node you will need at least 2 CPUs, 4 GB of memory, and 20 GB of disk storage.
+<!--TODO: Verify these requirements-->
+- Your Kubernetes cluster must have access to the internet, since Kubernetes needs to be able to fetch images.
+
+## Installing the Serving component
+
+To install the serving component:
```
abrennan89: I thought per the word list we're using Knative Serving throughout now?

docs/install/any-kubernetes-cluster.md
```diff
@@ -0,0 +1,836 @@
...
+   kubectl patch clusterrolebinding ambassador -p '{"subjects":[{"kind": "ServiceAccount", "name": "ambassador", "namespace": "ambassador"}]}'
+   ```
+
+1. Enable Knative support in Ambassador:
+
+   ```bash
+   kubectl set env --namespace ambassador  deployments/ambassador AMBASSADOR_KNATIVE_SUPPORT=true
+   ```
+
+1. To configure Knative Serving to use Ambassador by default:
```
abrennan89:
```suggestion
1. To configure Knative Serving to use Ambassador by default, enter the command:
```

abrennan89: All the procedure steps in this part don't seem to have "run" or "enter" the command. In future I'd maybe keep this as a separate PR to the pingsource stuff, i.e. keep topics separated so it doesn't blur the PR scope

docs/install/any-kubernetes-cluster.md
```diff
@@ -0,0 +1,836 @@
...
+kubectl apply -f {{< artifact repo="serving" file="serving-domainmapping-crds.yaml" >}}
+kubectl wait --for=condition=Established --all crd
+kubectl apply -f {{< artifact repo="serving" file="serving-domainmapping.yaml" >}}
+```
+
+{{< /tab >}} {{< /tabs >}}
+
+## Installing the Eventing component
+
+To install the Eventing component:
```
abrennan89: same as above, Knative Eventing throughout?

abrennan89 (changes requested):
Added comments.

Q: Was the install stuff added to this PR by mistake?

docs/eventing/sources/pingsource.md
evankanderson: Maybe the following text, to clarify why:

```markdown
**NOTE:** Because both `data` and `dataBase64` specify the event payload, only one field may be set.
```

docs/eventing/sources/pingsource.md
evankanderson: Alternatively, I might call this "stdin", "standard input" (weird, but what STDIN is abbreviated from) or "terminal input".

docs/eventing/sources/pingsource.md
evankanderson: Since we don't have a place for this text right now, would it unblock this PR to add a `<!-- TODO(<issue link>): move this to conceptual documentation when that exists -->` here?

docs/eventing/sources/pingsource.md
evankanderson: I think there are two points:

1. This is an "unnecessary cat" shell construction; I'd agree with the format that Ashleigh suggests
2. Prefer `kubectl apply` over `kubectl create` unless you are using `metadata.generateName`, which is probably not a good choice in a tutorial, because you can't predict the actual object name.

docs/eventing/sources/pingsource.md
evankanderson, line 146: I was going to point out that Knative Services may scale down to zero and have their logs recycled, but it occurs to me that the PingSource might be firing just frequently enough to avoid this.

I suspect this will work whether or not the service is getting scaled down to zero because there will be at least one pod running with at least one request in the logs (because the pingsource keeps sending events), so this is mostly a note for other examples.

docs/eventing/sources/pingsource.md
evankanderson: +1

You could also reference using log aggregation via https://knative.dev/docs/install/collecting-logs/ as another way to view the logs if desired, but I'm not super-happy with that yet so I don't want to add it to this PR.

evankanderson (commented):
/meow

For an extra "unnecessary cat".

abrennan89:
Agreed with Richard to close this one in favor of https://github.com/knative/docs/pull/3509
TODO: I'll add another PR to move the conceptual info for producers etc

PR #3479, opened 2021-04-19
Fix API doc generator to properly link to duck, networking

julz:
## Proposed Changes 

The api reference doc generator's config.json was a bit out of date. It was referencing the duck package as `github.com/knative/pkg` rather than as `knative.dev/pkg` where it now lives (see updated `hack/reference-docs-gen-config.json` in this PR). It also didn't know about the `knative.dev/networking` package since (I suspect) that was created after the config was last updated. There were also a couple other packages that the tool was printing warnings about that I added.

The result is these packages now properly link to the relevant external docs in the output, which appears to have been the original intention here (also ./hack/gen-api-reference-docs.sh no longer complains about "not found external link source for type" \o/).

/assign @evankanderson @abrennan89

### DIFF
--- a/docs/reference/api/eventing/eventing.md
+++ b/docs/reference/api/eventing/eventing.md
@@ -229,7 +229,9 @@ global delivery spec.</p>
 <td>
 <code>Status</code><br/>
 <em>
+<a href="https://pkg.go.dev/knative.dev/pkg/apis/duck/v1#Status">
 knative.dev/pkg/apis/duck/v1.Status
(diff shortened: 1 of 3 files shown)


evankanderson:
I think the knative.dev/networking packages are all _internal_, no?

julz:
ha I was _just_ typing what you said in the comment above. Yeah it looks like it leaks out due to documenting internal autoscaler types 🤷 .

PR #3495, opened 2021-04-22
Add kn config file docs and section for plugins

### LINKED ISSUE #2803: Documentation for kn configuration

An on documentation section should include the description of the`kn` configuration syntax and usage examples.

In addition the documentation should include the default location of that configuration file and how this can be overwritten with on the command line.

Possible options (needs to be verified):

* sink mappings
* plugin directory locations

abrennan89:
Fixes #2803


docs/client/configure-kn.md
```diff
@@ -0,0 +1,32 @@
+---
+title: "Customizing kn"
+weight: 03
+type: "docs"
+---
+
+You can customize your `kn` CLI setup by creating a `config.yaml` configuration file. The CLI will look for this file under the home directory of the user at `$HOME/.config/kn/config.yaml`, however you can create this file anywhere and use the `--config` flag to specify its path.
```
navidshaikh, line 7:
```suggestion
You can customize your `kn` CLI setup by creating a `config.yaml` configuration file. The configuration can be either provided via a `--config` flag or is picked up from a default location. The default configuration location conforms to the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and is different for Unix systems and Windows systems. If `XDG_CONFIG_HOME` env var is set, the default config location `kn` looks for is `$XDG_CONFIG_HOME/kn`, otherwise file is looked up under the home directory of the user at `$HOME/.config/kn/config.yaml`. For Windows systems, the default `kn` configuration location is `$APPDATA\kn`.
```

docs/client/configure-kn.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+Where
+
+- `lookupPluginsInPath` specifies whether `kn` should look for [plugins](./kn-plugins) in the specified `PATH` environment variable. This is a boolean configuration option. The default value is `false`.
+
+- `pluginsdir` specifies the directory where `kn` will look for plugins. The defaults path is `~/.config/kn/plugins`. This can be any directory that is visible to the user.
+
+- `sink` defines the prefix that is used to refer to Kubernetes Addressable resources. To configure a sink prefix, you must define following fields, as shown in the example:
+    - `prefix`: The prefix you want to use to describe your sink. Service, `svc`, and `broker` are predefined prefixes in `kn`.
```
rhuss, line 28: Could be that the original documentation out of sync, but the currently predefined prefixes are:

* `ksvc` for a Knative service
* `broker` for a broker
* `channel` for a channel.

docs/client/configure-kn.md
```diff
@@ -0,0 +1,32 @@
...
+## Example configuration file
+
+```yaml
+lookupPluginsInPath: true
+pluginsdir: ~/.config/kn/plugins
+sink:
+- prefix: myprefix
+  group: eventing.knative.dev
+  version: v1alpha1
+  resource: brokers
```
rhuss, line 18:
```suggestion
plugins:
  pathLookup: true
  directory: ~/.config/kn/plugins
eventing:
  sink:
  - prefix: svc
    group: core
    version: v1
    resource: services
```

Apologies, that the original documentation is not up to date (going to send a PR to fix https://github.com/knative/client/tree/main/docs#kn-config).

(also I updated the sample mapping to a more realistic example, as this could be used directly to map Kubernetes Services to a `svc:` prefixes (while Knative Services are mapped to `ksvc:` already by default)

This means, we have introduced a grouping in the configuration on the first level (`plugins` and `eventing`), this might deserve separate sub-sections maybe in the documentation ?

rhuss (changes requested):
Thanks for kicking this off!

Unfortunately, I just found out that the documentation in https://github.com/knative/client/tree/main/docs#kn-config is outdated. I added relevant comments in this PR but will also update the documentation over there in the client repo.

rhuss:
We could also add the documentation for the channel mapping. When I update https://github.com/knative/client/tree/main/docs#kn-config I will add that part there, too.

abrennan89:
Thanks for the comments folks, I'll update the PR.

@rhuss re
> We could also add the documentation for the channel mapping. 

Let's add that in a follow up PR so that it doesn't block this, since it's not in the current docs.

docs/client/configure-kn.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+Where
+
+- `lookupPluginsInPath` specifies whether `kn` should look for [plugins](./kn-plugins) in the specified `PATH` environment variable. This is a boolean configuration option. The default value is `false`.
```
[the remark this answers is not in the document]
abrennan89: removed "specified", thanks for clarifying

docs/client/configure-kn.md
abrennan89: Made some updates. I think giving them separate sections for now might be overkill tbh until we get to adding more complex docs about configuration for `kn`. IMO for now it's enough to explain them in one example that can be iterated later.

abrennan89:
Made some updates. I do want to limit the scope of this PR to what currently exists so that I'm not blocked on completing it while I am waiting for eng input, so for anything after this let's iterate the docs in a new PR. I believe that what we have for now is enough to close the original issue.

abrennan89:
One last thing actually I'd like to include @rhuss @navidshaikh if someone can provide an example:
"In addition the documentation should include the default location of that configuration file and how this can be overwritten with on the command line."

I asked in Slack previously I think, but can someone give me an example command for how to update this using CLI and what a sample output/confirmation msg looks like, or how one can verify that it was successfully overwritten?

rhuss:
> "In addition the documentation should include the default location of that configuration file and how this can be overwritten with on the command line."

This just means that you can specify a `--config path/to/config.yaml` to every command for using that specified configuration file instead of the default configuration file.

docs/client/configure-kn.md
```diff
@@ -0,0 +1,36 @@
...
+- For Windows systems, the default `kn` configuration location is `%APPDATA%\kn`.
+
+## Example configuration file
+
+```yaml
+plugins:
+  pathLookup: true
+  directory: ~/.config/kn/plugins
+eventing:
+  sink:
```
rhuss, line 20:
```suggestion
  sink-mappings:
```

docs/client/configure-kn.md
```diff
@@ -0,0 +1,36 @@
...
+
+- If the `XDG_CONFIG_HOME` environment variable is set, the default configuration location that `kn` looks for is `$XDG_CONFIG_HOME/kn`.
+- If the `XDG_CONFIG_HOME` environment variable is not set, `kn` looks for the configuration in the home directory of the user at `$HOME/.config/kn/config.yaml`.
+- For Windows systems, the default `kn` configuration location is `%APPDATA%\kn`.
+
+## Example configuration file
+
+```yaml
+plugins:
+  pathLookup: true
```
rhuss, line 17:
```suggestion
  path-lookup: true
```

rhuss (changes requested):
Update to the config keys (which I got wrong). Verified that these are the correct keys now.

rhuss:
I'm good with this PR as soon as the update to the examples have been updated.

docs/client/configure-kn.md
```diff
@@ -0,0 +1,36 @@
...
+  sink:
+  - prefix: svc
+    group: core
+    version: v1
+    resource: services
+```
+
+Where
+
+- `pathLookup` specifies whether `kn` should look for [plugins](./kn-plugins) in the `PATH` environment variable. This is a boolean configuration option. The default value is `false`.
```
rhuss, line 29:
```suggestion
- `path-lookup` specifies whether `kn` should look for [plugins](./kn-plugins) in the `PATH` environment variable. This is a boolean configuration option. The default value is `false`.
```

docs/client/configure-kn.md
```diff
@@ -0,0 +1,36 @@
...
+    group: core
+    version: v1
+    resource: services
+```
+
+Where
+
+- `path-lookup` specifies whether `kn` should look for [plugins](./kn-plugins) in the `PATH` environment variable. This is a boolean configuration option. The default value is `false`.
+- `directory` specifies the directory where `kn` will look for plugins. The default path depends on the operating system, as described above. This can be any directory that is visible to the user.
+- `sink` defines the Kubernetes Addressable resource that is used when you use the `--sink` flag with a `kn` CLI command.
```
abrennan89, line 31:
```suggestion
- `sink-mappings` defines the Kubernetes Addressable resource that is used when you use the `--sink` flag with a `kn` CLI command.
```

rhuss (approved):

rhuss:
Thanks ! looks good for now, I'm just about to update the client's README to add `channel-type-mappings` to the documentation, and will ping you again for an update of the docs.

thanks, lgtm
/lgtm
/approve

PR #3506, opened 2021-04-28
config-tracing configmap fix

Shashankft9:
Noticed that the key `enable` is not even in the structure of tracing config: 
```
type Config struct {
    Backend              BackendType
    ZipkinEndpoint       string
    StackdriverProjectID string
    Debug                bool
    SampleRate           float64
}
```
also noticed here: https://github.com/knative/eventing/issues/1889#issuecomment-753998185
also its not mentioned in the `configuration options` in the doc, so I figured it might be something left over from previous versions. Let me know if it really is required? 

## Proposed Changes 

- just replaced `enable` key with `backend`
-
-

### DIFF
--- a/docs/eventing/accessing-traces.md
+++ b/docs/eventing/accessing-traces.md
@@ -41,7 +41,7 @@ metadata:
   name: config-tracing
   namespace: knative-eventing
 data:
-  enable: "true"
+  backend: "zipkin"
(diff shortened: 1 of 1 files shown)


matzew (approved):
/lgtm
/approve

abrennan89 (approved):
/lgtm

PR #3545, opened 2021-05-09
Remove responsive-revision-gc config

nak3:
Since https://github.com/knative/serving/pull/10084 deleted old GC and
started using responsive GC by default, `responsive-revision-gc` in
`config-feature` is not used at all.

This patch removes the instruction for the configuration.

https://github.com/knative/serving/pull/11309 removes the configuration in serving repo.


abrennan89 (approved):
/lgtm

Issue #3541, opened 2021-05-07
Ensure the Knative Community tab is hosted in the old Hugo format 


omerbensaadon:
## Description: 
Since we want to limit the scope of our work in this first release of the MkDocs site, we have opted to keep our Community tab in the existing Hugo template and wire it together with the MkDocs prototype. 

See: https://knative.dev/blog/

---
## Story
```
As Danny the Developer
I want to check out the Knative Community Tab  
So that I can find information on the Knative Community
```

## Acceptance Criteria
`Community Tab Present` 
```
GIVEN I am on the new MkDocs site
WHEN I look at the main site navigation tabs  
THEN I see a tab named "Community" 
```

`MkDocs -> Hugo`
```
GIVEN I am on the new MkDocs site 
WHEN I click on "Community" in the main site navigation tabs 
THEN I am brought to the Knative Community site (currently knative.dev/community) hosted in Hugo
```

`Hugo -> MkDocs`
```
GIVEN I am on the Knative Community site hosted in Hugo 
WHEN I click on "Documentation" in the header
THEN I am brought back to the homepage in the MkDocs site
```

## Known Issues and Considerations: 
- This issue does not cover broken links and will be accepted nearly broken. We'll do clean-up in a different issue.

omerbensaadon:
Closed by #3558

Issue #3539, opened 2021-05-07
Ensure the Knative Blog is hosted in the old Hugo format


omerbensaadon:
## Description: 
Since we want to limit the scope of our work in this first release of the MkDocs site, we have opted to keep our Blog in the existing Hugo template and wire it together with the MkDocs prototype. 

See: https://knative.dev/blog/

---
## Story
```
As Danny the Developer
I want to check out the Knative blog 
So that I can see release and feature announcements, articles and social events. 
```

## Acceptance Criteria
`Blog Tab Present` 
```
GIVEN I am on the new MkDocs site
WHEN I look at the main site navigation tabs 
THEN I see a tab named "Blog" 
```

`MkDocs -> Hugo`
```
GIVEN I am on the new MkDocs site 
WHEN I click on "Blog" in the main site navigation tabs 
THEN I am brought to the Knative Blog (currently knative.dev/blog) hosted in Hugo
```

`Hugo -> MkDocs`
```
GIVEN I am on the Knative Blog hosted in Hugo 
WHEN I click on "Documentation" in the header
THEN I am brought back to the homepage in the MkDocs site
```

## Known Issues and Considerations: 
- Currently, there is no support for the community tab, you can ignore this for now. 
- Currently, there is no support for posting new blogs, you can ignore this for now.

omerbensaadon:
Closed by #3558

PR #3558, opened 2021-05-12
Build blog and community with hugo

julz:
Fixes https://github.com/knative/docs/issues/3539.
Fixes https://github.com/knative/docs/issues/3541.

mkhugo (1)

Builds the blog and community tabs with hugo.

Also added a little CSS to visually distinguish these links.

This works by cloning the website and community repos locally, copying blog and community in to the right places (same trick website uses in https://github.com/knative/website/blob/main/scripts/processsourcefiles.sh#L107), and then building those pieces using hugo in to the same site dir as mkdocs builds the docs.

Running:
- Same as today: `mkdocs serve` previews just the main site (with live reloading previews)
- New: `./hack/build-with-blog.sh` builds to `/site/` and is what we should run in the netlify build/github action.
- New: `./hack/build-with-blog.sh serve` builds to `/site/` and **starts a local preview server with blog/community included** (but no live reloads, sorry!). 
  - Please run `npm install -g http-server` first.

Follow on work: allow community/website to be checked out as sibling directories/cached to avoid rerunning the git clone / allow local testing of changes to these repositories.

~~Known issue: the library Im using to do the local preview (`npm install -g serve`) doesn't cope with URLs that have dots in them, like `http://localhost:5000/blog/2021/01/15/version-v0.20-release/`, this is just a local preview issue which can be solved by using some other tool to preview the built `site/` directory, but I think we can solve that separately.~~ 
.. Fixed! Using `http-server` instead.

### DIFF
--- a/hack/build-with-blog.sh
+++ b/hack/build-with-blog.sh
@@ -2,7 +2,7 @@
 
 # Builds blog and community into the site by cloning the website repo, copying blog/community dirs in, running hugo.
 # - Results are written to site/ as normal.
-# - Run as "./hack/build-with-blog.sh serve" to run a local preview server on site/ afterwards (requires `npm install -g serve`).
+# - Run as "./hack/build-with-blog.sh serve" to run a local preview server on site/ afterwards (requires `npm install -g http-server`).
 
 # Quit on error
 set -e
@@ -38,7 +38,6 @@ hugo
 popd
 
 # Hugo builds to public/, just copy over to site/ to match up with mkdocs
-# This means you can run `mkdocs build` and then `build-blog.sh` and get a combined site/.
 mv temp/website/public/blog site/
 mv temp/website/public/community site/
 mv temp/website/public/css site/
@@ -73,6 +72,6 @@ rm -rf temp
 
 if [ "$1" = "serve" ]; then
   pushd site
-  serve
+  http-server
   popd
 fi
--- a/hack/build-with-blog.sh
+++ b/hack/build-with-blog.sh
@@ -72,6 +72,6 @@ rm -rf temp
 
 if [ "$1" = "serve" ]; then
   pushd site
-  http-server
+  npx http-server
   popd
 fi
--- a/hack/build-with-blog.sh
+++ b/hack/build-with-blog.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+# Builds blog and community into the site by cloning the website repo, copying blog/community dirs in, running hugo.
+# - Results are written to site/ as normal.
+# - Run as "./hack/build-with-blog.sh serve" to run a local preview server on site/ afterwards (requires `npm install -g serve`).
+
+# Quit on error
+set -e
+# Echo each line
+set -x
+
+# First, build the main site with mkdocs
+rm -rf site/
+mkdocs build -d site/docs
+
+# Re-Clone
+# TODO(jz) Cache this and just do a pull/update for local dev flow.
+# Maybe also support local checkout in siblings?
+rm -rf temp
+mkdir temp
+git clone --recurse-submodules https://github.com/knative/website temp/website
+git clone https://github.com/knative/community temp/community
+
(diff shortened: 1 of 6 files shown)


omerbensaadon:
Love the CSS!! Also tips you off that you are about to experience a different format to the one you are currently on. 

Heads up, got this error: 

```
Error: Error building site: POSTCSS: failed to transform "scss/main.css" (text/css). Check your PostCSS installation; install with "npm install postcss-cli". See https://gohugo.io/hugo-pipes/postcss/
```

Going to install postcss and try again

julz:
oh yeah, you do indeed need postcss for the hugo build, and there's another library too I think it needs I forgot to document, `npm install autoprefixer` (maybe you already had it?). Hopefully we'll throw all this in a docker container so folks dont have to care, but in the meantime these should go in contributor guide.

omerbensaadon:
It looks like this build off of the existing website repo, so none of the processes for posting to the blog or editing community will change? 

Also, still getting this error after running `npm install autoprefixer` and `npm install postcss-cli`

julz:
> It looks like this build off of the existing website repo, so none of the processes for posting to the blog or editing community will change?

correct, the build for those parts is essentially much unmodified (just being driven from docs rather than other way round), so all the author workflows for docs/community should stay the same, I think

omerbensaadon:
`npm install -g postcss postcss-cli autoprefixer` fixed my problem!

omerbensaadon:
There will need to be a clean up story written to ensure that folks cannot escape into the old site, but for now, this is absolutely sufficiently de-risked. Good work @julz!

csantanapr:
Using
```
npm install -g postcss postcss-cli autoprefixer
```
Doesn't work for me
```
Total in 1150 ms
Error: Error building site: POSTCSS: failed to transform "scss/main.css" (text/css): Error: Loading PostCSS Plugin failed: Cannot find module 'autoprefixer'
```

csantanapr:
They need to be local for me to work
```
npm install postcss postcss-cli autoprefixer
```
Then it works.
```json
❯ cat package.json
{
  "dependencies": {
    "autoprefixer": "^10.2.5",
    "postcss": "^8.2.15",
    "postcss-cli": "^8.3.1"
  }
}
```

csantanapr (changes requested):
Add package.json and package-lock.json

```
{
  "dependencies": {
    "autoprefixer": "^10.2.5",
    "postcss": "^8.2.15",
    "postcss-cli": "^8.3.1"
  }
}
```
this way user can just `npm install` at the root before running the script

csantanapr:
add `temp` to `.gitignore`

hack/build-with-blog.sh
```diff
@@ -0,0 +1,77 @@
...
+</body>
+</html>
+EOF
+
+# Clean up
+rm -rf temp
+
+if [ "$1" = "serve" ]; then
+  pushd site
+  http-server
```
csantanapr, line 75: You think is better to use `python3 -m http.server 8000` ?

csantanapr:
I'm done with the review, is works great. 
That was very clever stuff you did here @julz 🔥 

I would say lets add the package.json and package-lock.json, and the .gitignore for `temp` and merge it

hack/build-with-blog.sh
julz: Dunno, I just picked the first thing I found that worked 😅 we need npm for autoprefixer and postcss anyway tho so doesn't seem any big reason to switch 🤷‍♂️. Might be able to make it a one liner with "npx run http-server" tho 🤔

julz:
ok, copied the package.json over from website and deleted the libraries we don't use. Also added temp and node_modules to gitignore and changed the `http-server` command to go via `npx` so everything should work after only `npm i`. 

@csantanapr ptal!

csantanapr:
works like a charm
/unhold

csantanapr (approved):

PR #3563, opened 2021-05-12
remove unnecessary requrirements to do a release

csantanapr:
- No need to have netlify admin access to cut a release
- No need to have Hugo installed to cut a release


evankanderson (commented):
/lgtm
/approve

PR #3551, opened 2021-05-11
Move mkdocs work over from julz/mkdocs branch

julz:
Fixes #3547.

Moves over the mkdocs work from julz/mkdocs, and re-converts serving/eventing docs to mkdocs.

This is almost entirely automated. I copied over mkdocs.yml, extra.css,
overrides/* (the home page), macros.py, index.md, and the getting started
directory and then ran the following scripts:

- [mkhugo](http://github.com/julz/mkhugo) to convert shortcodes to nice mkdocs markdown.
- Rename _index.md to README.md where it did not contain a readfile shortcode.
- [h1f1](http://github.com/julz/h1f1) to add top-level headers where missing
  (looks nicer on github, and makes mkdocs material search results look nicer -- this is the
  majority of the diff 🤷).
- [mknav](http://github.com/julz/mknav) to generate the nav section of the yaml
  for serving and eventing based on the `weight`s in the existing metadata (and
    then a bit of manual touch-up after to make the structure a bit nicer).

Having done this - since the scripts had done all the work of converting them - I dropped the excludes
from mkdocs.yml that had been hiding most of the samples \o/.

To test, check out the branch and run `mkdocs serve`. You should see no errors
(other than some broken in-page links we need to fix up - these look like they're mostly missing aliases),
and you should be able to correctly click through all the pages.

All the above scripts are idempontent, so it should be fine to rerun them to
fix up any outstanding PRs that we want to cherry-pick after this.

/assign @omerbensaadon @csantanapr

PR #3566, opened 2021-05-12
Regenerate the client nav

julz:
Updates the client section nav (via [mknav](http://github.com/julz/mknav)). (Forgot to do this when I did #3551).

/assign @omerbensaadon


csantanapr:
@julz there are conflicts

PR #3572, opened 2021-05-13
Update Github action to build with hugo and mkdocs

csantanapr:
- Update gh action to build with mkdocs and hugo
- Handle better redirects to be compatible with personal forks with different names than `docs`
- Handle the requirement of having postcss in the PATH


julz (approved):
🔥 

/lgtm

PR #3574, opened 2021-05-13
Replace xip.io with sslip.io

benjamintanweihao:
## Proposed Changes

`xip.io` doesn't seem to be working anymore. This change replaces it with `sslip.io`.
-
-
-


nak3:
Thank you @benjamintanweihao 
Serving started using `sslip.io` rather than `nip.io` so it would be better to use `sslip.io` consistently.

/cc @julz @evankanderson

benjamintanweihao:
> Thank you @benjamintanweihao
> Serving started using `sslip.io` rather than `nip.io` so it would be better to use `sslip.io` consistently.
> 
> /cc @julz @evankanderson

Ok!  I shall change it to sslio.io then

evankanderson (commented):
/lgtm
/approve

PR #3576, opened 2021-05-14
EventType is not yet part of the v1 eventing API

lenalebt:
## Proposed Changes 

I tried adding an EventType as proposed in the example - but `EventType` still is in `v1beta1`, although the example states it would be in `v1` already.

### DIFF
--- a/docs/eventing/event-registry.md
+++ b/docs/eventing/event-registry.md
@@ -60,7 +60,7 @@ kubectl get eventtype dev.knative.source.github.push-34cnb -o yaml
 Omitting irrelevant fields:
 
 ```yaml
-apiVersion: eventing.knative.dev/v1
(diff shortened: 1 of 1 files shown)


lenalebt:
@googlebot I signed it!

omerbensaadon:
Great catch, looks like it's showing that way in the spec.

docs/eventing/event-registry.md
```diff
@@ -60,7 +60,7 @@ kubectl get eventtype dev.knative.source.github.push-34cnb -o yaml
 Omitting irrelevant fields:
 
 ```yaml
-apiVersion: eventing.knative.dev/v1
+apiVersion: eventing.knative.dev/v1beta1
```
csantanapr, line 63: Thank you @lenalebt for the update, this is correct

PR #3589, opened 2021-05-17
add correct bin PATH for postcss

csantanapr:
fix build script.

Need to add gh-action later to verify build script

/assign @julz


julz (approved):
/lgtm

PR #3593, opened 2021-05-18
make netlify beta site return to home page

csantanapr:
make netlify hostname default to make it usable on netlify for now
This to fix when clicking on latest version
image

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,5 +1,5 @@
 site_name: Knative
-site_url: https://knative.dev/docs
+site_url: https://beta-knative.netlify.app/docs
 site_description: Knative Documentation
 extra_css:
   - stylesheets/extra.css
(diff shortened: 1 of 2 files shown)


csantanapr:
@julz this change didn't affect `{{ base_url }}`
Where is this coming from?

julz:
hm.. not sure.. maybe we can try swapping the whole href to be {{ site_url }}. Guess it'll all fix itself when we go live anyway since knative.dev/docs will be the right link then! 🤷 .

csantanapr:
Let's make hard code to netlify domain for now, then will adjust as we move the domain name

julz:
so put site_url back to knative.dev/docs, and hardcode the "latest version" url to the netlify url in main.html for now? yeah, works for me

julz (approved):
/lgtm

csantanapr:
The change needs to be done on @julz repo branches

PR #3594, opened 2021-05-18
Remove pycache

nak3:
As per title, this path removes python cache which should not be necessary.

/cc @evankanderson @abrennan89


nak3:
It seems that the cache was included in https://github.com/knative/docs/commit/135e998d49cf4895678a7ea530ff44730ab29443.

julz (approved):
/lgtm

I guess this snuck in because it's gitignored in the mkdocs branch but not in main.

Issue #3560, opened 2021-05-12
incorrect issuer name in the guide for Enabling automatic TLS certificate provisioning


draksler:
## Expected Behavior
Configuring knative as described in [this](https://knative.dev/docs/serving/using-auto-tls/) guide, using the HTTP-01 challenge, should lead to automatic certificate provisioning.

## Actual Behavior
It doesn't.

## Steps to Reproduce the Problem
Follow the [guide](https://knative.dev/docs/serving/using-auto-tls/) - HTTP-01 challenge mode.

## Additional Info
Looking through the logs the following error can be found
`2021-05-12 15:57:19.513 CEST
Event(v1.ObjectReference{Kind:"Certificate", Namespace:"default", Name:"route-bbc3a90b-4aa2-4816-b510-e34a263a63c7", UID:"d537db59-1ca8-4936-9c2a-d78517a0ff6e", APIVersion:"networking.internal.knative.dev/v1alpha1", ResourceVersion:"1283952", FieldPath:""}): type: 'Warning' reason: 'InternalError' clusterissuer.cert-manager.io "letsencrypt-issuer" not found`

the issuer is incorrectly named. Renaming it from
`metadata:
  name: letsencrypt-http01-issuer`
to
`metadata:
  name: letsencrypt-issuer`

fixes the issue and automatic certificate provisioning starts to work.

**Install information:**
 - Platform (GKE, IKS, AKS, etc.): GKE
 - Knative Version: "v0.22.0"

draksler:
or potentially if it is prefered to keep the specific name, the correct one could be referenced in the configmap

nak3:
> or potentially if it is prefered to keep the specific name, the correct one could be referenced in the configmap

Yes, agree. The doc created `letsencrypt-http01-issuer` ClusterIssure in https://knative.dev/docs/serving/using-auto-tls/#clusterissuer-for-http-01-challenge, so we probably want to use `letsencrypt-http01-issuer`.

I think we can just remove the following snippet example from the doc.

```
...
data:
...
  issuerRef: |
    kind: ClusterIssuer
    name: letsencrypt-issuer
```

PR #3595, opened 2021-05-18
Use correct ClusterIssue name

nak3:
The doc uses ClusterIssuer `letsencrypt-http01-issuer` but the configuration uses
`letsencrypt-issuer` so it is confusable. Actual report is https://github.com/knative/docs/issues/3560.

So this patch changes the ClusterIssuer name to
`letsencrypt-http01-issuer` which is used in this doc section.

Fix https://github.com/knative/docs/issues/3560


abrennan89 (approved):
/lgtm

PR #3599, opened 2021-05-18
WIP: test skipping with go.mod

julz:
Go build ./... skips these packages with a go.mod, but the custom code to find packages with tags (https://github.com/knative/hack/blob/93ad9129c268d068d4be6b9f2238626ea4aff171/presubmit-tests.sh#L151-L170) does not, leading to spurious errors like in https://github.com/knative/docs/pull/3532.

This is just to test if skipping fixes stuff, if it does Ill close this and PR in hack.


julz:
hm, works, but doesn't cope with packages in subdirectories of the one with the go.mod file

Issue #3617, opened 2021-05-19
Ensure "Pre-Release" on Knative.dev points to `mkdocs` branch


omerbensaadon:

csantanapr:
I think what this means is make the pre-release drop down in the version in the current hugo knative.dev site redirect folks to the dev-knative.netlify.app site.

@julz can you help with this ^^

julz:
I guess we could hard-code the URL in the href `if eq .version $preReleaseLabel` [here](https://github.com/knative/website/blob/39b28522b71121fade15595fd8a379eb3153c8f2/layouts/partials/navbar-version-selector.html#L44) to the new URL? But swapping back won't work unless we _also_ modify the mkdocs version of the dropdown which may be confusing 🤔

julz:
actually I wonder if changing the url [in params.toml](https://github.com/knative/website/blob/main/config/production/params.toml#L86) would work (but we still have the problem of going the other way, from the new site back to the old)

csantanapr:
I think we do it with netlify redirects both ways

csantanapr:
I propose we try with redirects between the 2 sites with netlify redirects in netlify.toml on both repos
| knative.dev  | dev-knative.netlify.app |
| ------------- | ------------- |
| /development -->  | /development  |
| /v0.23-docs/ | <-- /v0.23-docs/  |
| /v0.22-docs/ | <-- /v0.22-docs/  |
| /v0.21-docs/ | <-- /v0.21-docs/  |
| /v0.20-docs/ | <-- /v0.20-docs/  |

csantanapr:
@julz what you think?
https://github.com/knative/website/pull/307

omerbensaadon:
done!

PR #3628, opened 2021-05-20
handle old versions with redirects

csantanapr:
This will allow the drop down for versions to redirect user to old site

Related to #3617 

/assign @julz


omerbensaadon:
@csantanapr I think one of the minimum requirements here is to get "Pre-Release" on the old site to redirect to the new site

csantanapr:
The pre-release redirect PR is here https://github.com/knative/website/pull/307

PR #3632, opened 2021-05-20
Fix a mismatch in 'a' vs. plural

vaikas:

## Proposed Changes 

- drop the a since later on it uses plural for services.
-
-


omerbensaadon:
/lgtm 
/approve 

Thanks Ville!

PR #3635, opened 2021-05-20
Fix syntax error that causes some CSS to not apply

markusthoemmes:
Title. This caused the announce bar to appear.


julz (approved):
heh, that explains it! thanks :) 

/lgtm
/approve

omerbensaadon:
Relevant to #3640

PR #3636, opened 2021-05-20
Be a bit more subtle with the gradient's colors

markusthoemmes:
Mostly playing around a bit. This is totally a design bikeshed.


omerbensaadon:
Relevant to #3640

PR #3637, opened 2021-05-20
Even out the margins and paddings on the homepage

markusthoemmes:
The margins and paddings are a bit all over the place on the homepage. This aims at evening that out, while also reducing the amount of different font-styles used for a more "calm" picture.


omerbensaadon:
/lgtm 
/approve 

Thanks @markusthoemmes !

omerbensaadon:
Relevant to #3640

PR #3638, opened 2021-05-20
Fix rocket image on small screens

markusthoemmes:
Sadly, I broke this, so here I am fixing it. Also slightly adjusting the positioning on small screens so it doesn't hug the screen corner as much.

### DIFF
--- a/overrides/assets/stylesheets/home.css
+++ b/overrides/assets/stylesheets/home.css
@@ -103,7 +103,7 @@ body {
   }
 
   .rocket {
-    left: -190px;
+    left: -170px;
   }
 
   .intro-container {
--- a/overrides/home.html
+++ b/overrides/home.html
@@ -63,7 +63,7 @@
   </section>
 
   <section class="intro-container">
-    <div class="md-grid md-typeset">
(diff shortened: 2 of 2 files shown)


omerbensaadon:
I really want this rocket to be on the left side of the text... :D

omerbensaadon:
Relevant to #3640

shinigambit:
For me the text seems like the bigger issue, I would expect the rocket to become part of the text's background and the text to occupy the whole width or maybe even show at the top of the text instead.

I was thinking something like this but centering the rocket idk:
Screen Shot 2021-05-22 at 12 08 10
Sorry the image looks so huge

markusthoemmes:
IMO, the rocket shouldn't be on the left side as a default as it breaks the left eye-line, where all text uniformly starts there. That'd look weird I think.

shinigambit:
@markusthoemmes yeah, my default responsive view was Galaxy S9 and the text looks stretched I can see how it looks better in most devices.

For reference this is how it looks in a Galaxy S9 / Galaxy S5:

PR #3646, opened 2021-05-20
Developer guide mkdocs

abrennan89:


gabo1208 (approved):
Just have that suggestion but appart from that everything looks good :)

PR #3657, opened 2021-05-24
fixes #3656

### LINKED ISSUE #3656: YAML in traffic split returns an error
## Expected Behavior
When I apply the YAML from https://dev-knative.netlify.app/development/getting-started/first-traffic-split/#creating-a-new-revision I expect it to work

## Actual Behavior
It returns an error

1. Make the hello.yaml file 
1. Try to apply it 
1. Read it and weep

## Additional Info
**Additional context**

**Install information:**
 - Platform (GKE, IKS, AKS, etc.): Kind
 - Knative Version: Whatever KonK uses

omerbensaadon:
Fixed the first YAML

### DIFF
--- a/docs/getting-started/first-traffic-split.md
+++ b/docs/getting-started/first-traffic-split.md
@@ -113,11 +113,10 @@ By default, when Knative creates a brand new `Revision` it directs 100% of traff
     apiVersion: serving.knative.dev/v1
     kind: Route
     metadata:
-      name: route-hello
+      name: hello
     spec:
-    # Unfortunately, @latest is a handy pointer which is only supported in the kn CLI
       traffic:
-        - revisionName: hello-knative
+        - revisionName: latest
           percent: 50
         - revisionName: hello-world
           percent: 50
--- a/docs/getting-started/first-traffic-split.md
+++ b/docs/getting-started/first-traffic-split.md
@@ -23,8 +23,8 @@ Instead of `TARGET`="World," let's update the environment variable `TARGET` on o
     ```
 
 === "YAML"
-
     ``` bash
+    ---
     apiVersion: serving.knative.dev/v1
     kind: Service
     metadata:
@@ -32,7 +32,7 @@ Instead of `TARGET`="World," let's update the environment variable `TARGET` on o
     spec:
       template:
         metadata:
-          name: knative
+          name: hello-knative
         spec:
           containers:
             - image: gcr.io/knative-samples/helloworld-go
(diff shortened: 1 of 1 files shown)


docs/getting-started/first-traffic-split.md
```diff
@@ -23,16 +23,16 @@ Instead of `TARGET`="World," let's update the environment variable `TARGET` on o
...
+    ---
     apiVersion: serving.knative.dev/v1
     kind: Service
     metadata:
       name: hello
     spec:
       template:
         metadata:
-          name: knative
+          name: hello-knative
```
csantanapr, line 35: The YAML is not matching the `kn` CLI 

If you want the revision to be `world` and `knative` and the service name to be named `hello` then you have to make the CLI refer to it as `world` and `knative` and not `hello-world` or `hello-knative`

docs/getting-started/first-traffic-split.md
```diff
@@ -142,7 +143,7 @@ Verify traffic split configure correctly by listing the revisions again.
 ```{ .bash .no-copy }
 NAME            SERVICE   TRAFFIC   TAGS   GENERATION   AGE   CONDITIONS   READY   REASON
 hello-knative   hello     50%              2            10m   3 OK / 4     True
-hello-world     hello     50%              1            36m   3 OK / 4     True 
+hello-world     hello     50%              1            36m   3 OK / 4     True
```
csantanapr, line 146: If you name your revisions `world` and `knative` then remove `hello-` from the column `NAME` the column `SERVICE` containers `hello`

Also you need to make the CLI commands match to be consistence for example:
Instead of using revision `hello-world`
```
kn service update hello  --traffic hello-world=50  --traffic @latest=50
```
Use revision name `world`
```
kn service update hello  --traffic world=50   --traffic @latest=50
```

csantanapr (changes requested):
The YAML and the `kn` CLI should be consistent use of revision.

In some places revision is used as `hello-world` and in some other places `world`

docs/getting-started/first-traffic-split.md
csantanapr: yeah this is correct I double check

docs/getting-started/first-traffic-split.md
csantanapr: I double check if revision is `knative` the name will always have the service name prefix

csantanapr (approved):

docs/getting-started/first-traffic-split.md
omerbensaadon: I think this is already fixed?

Issue #3644, opened 2021-05-20
Fix all broken links in `/eventing`


omerbensaadon:
## Description: 
We want to make sure the eventing content which lives in the Beta site (https://github.com/knative/docs/tree/mkdocs) does not contain broken links

---

## Story
```
As an Eli the End User 
I want to go through the pages under the "Eventing" section of the site without finding broken links
So that I can feel like the Knative website is crafted with care 
And I assign more credibility to Knative
```

## Acceptance Criteria
`Legibility, Credibility Check`
```
GIVEN I find my way to any of the pages in the sub-directory https://dev-knative.netlify.app/development/eventing/
WHEN I click every link 
THEN I do not find any broken links 
```

## Known Issues and Considerations: 
- When you build the site using `mkdocs serve` mkdocs will spit out all internal links it finds which are broken. If you do `Ctrl+F` eventing/ you should be able to work through all of the broken internal references
- If you do `npm install broken-link-checker -g` and then do `blc https://dev-knative.netlify.app/development/eventing/ -ro --exclude-internal` you should get a list of all broken external links

## Who should probably pick up this issue? 
- [ ] Existing Contributors 
- [x] New Contributors

omerbensaadon:
done! Thanks @gabo1208

PR #3663, opened 2021-05-25
Create user-story

omerbensaadon:

## Proposed Changes 

-
-
-

### DIFF
--- a/.github/ISSUE_TEMPLATE/user-story
+++ b/.github/ISSUE_TEMPLATE/user-story
@@ -1,0 +1,33 @@
+## Description: 
+
+---
+
+## Story
+```
+As {PERSONA} 
+I want to {USER NEED}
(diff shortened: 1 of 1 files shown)


abrennan89:
Is docs really the place for this vs. eng repos? I don't see how we would implement user stories for example if someone is asking for a feature that doesn't exist. It seems like a more rational workflow that things like this would come to us from eng?

omerbensaadon:
@abrennan89 this is me institutionalizing something I am already doing 

ex: #3644 

Do you feel I should just copy + paste this template as needed? I know issue templates are often viewed as something for new contributors but it makes my job a little easier. 

If that's coming at the cost of contributor friction I can hold off

omerbensaadon:
/hold 

need to write a description for this issue template

PR #3666, opened 2021-05-26
Adding docs for pod capabilities feature flat (see serving PR #11410)

psschwei:
Documentation in support of [Serving PR #11410](https://github.com/knative/serving/pull/11410)

## Proposed Changes 

- Adds documentation for a feature gate that allows users to add Linux capabilities to a container

### DIFF
--- a/docs/serving/feature-flags.md
+++ b/docs/serving/feature-flags.md
@@ -275,6 +275,34 @@ spec:
         ...
 ```
 
+## Kubernetes Security Context Capabilities
+
+* **Type**: extension
+* **ConfigMap key**: `kubernetes.containerspec-addcapabilities`
+
+This flag controls whether users can add capabilities on the `securityContext` of the container.
+
+When set to `enabled` or `allowed` it allows [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) to be added to the container.
+
+```yaml
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: helloworld-go
+spec:
+ template:
+  spec:
+   containers:
+     - image: gcr.io/knative-samples/helloworld-go
+       env:
+         - name: TARGET
+           value: "Go Sample v1"
+       securityContext:
+         capabilities:
+           add:
+             - NET_BIND_SERVICE
+```
+
 ## Tag Header Based Routing
 * **Type**: extension
 * **ConfigMap key:** `tag-header-based-routing`


omerbensaadon:
@psschwei thanks for the PR! Should we wait for https://github.com/knative/serving/pull/11410 to merge before merging this?

docs/serving/feature-flags.md
```diff
@@ -275,6 +275,30 @@ spec:
...
+spec:
+ template:
+  spec:
+   containers:
+      securityContext:
+        capabilities:
+          add:
+            - NET_BIND_SERVICE
+            ...
+```
```
abrennan89, line 300: Can we make this a full sample just to avoid confusion? It seems truncated?

docs/serving/feature-flags.md
```diff
@@ -275,6 +275,30 @@ spec:
         ...
 ```
 
+## Kubernetes Security Context Capabilities
+
+* **Type**: extension
+* **ConfigMap key**: `kubernetes.podspec-addcapabilities`
+
+This flag controls whether users can add capabilities on the Pod's securityContext.
```
abrennan89, line 283:
```suggestion
This flag controls whether users can add capabilities on the `securityContext` of a pod.
```

docs/serving/feature-flags.md
```diff
@@ -275,6 +275,30 @@ spec:
...
+
+When set to "enabled" or "allowed" it allows [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) to be added to the container.
```
abrennan89, line 285:
```suggestion
When set to `enabled` or `allowed` it allows [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) to be added to the container.
```

Is enabled / allowed a literal setting? Where does someone switch this on/off?

abrennan89 (changes requested):
Some clarification required IMHO, it doesn't look like there's any example shown of how this is actually enabled in the configmap?

docs/serving/feature-flags.md
psschwei: A full sample is okay with me. Though it looks like most of the other yaml examples in this doc are also snippets rather than full samples -- would it be better to make a separate PR to update all of them?

docs/serving/feature-flags.md
psschwei: This is set in the [`config-features` configmap](https://github.com/psschwei/serving/blob/add-add-capability/config/core/configmaps/features.yaml#L112) in the `knative-serving` namespace. Definitions for the enabled / allowed values are provided in the [`docs/feature-flags.md`](https://github.com/knative/docs/blob/mkdocs/docs/serving/feature-flags.md#control) page.

On `enabled` vs. "enabled" -- happy to make that change. There's a number of other places in the doc where the values are in quotes, would it be better to update them all at once in a separate PR (maybe together with the yaml examples mentioned in a previous comment)?

abrennan89:
@psschwei for other instances of "enabled" and code snippets that need to be updated, can you please open an issue that points to any other instances you're aware of that need to be updated?

I think it's fine to update the ones affected in this PR for now and open an issue for someone to update the others later. Might as well make small improvements in this PR for now.

docs/serving/feature-flags.md
abrennan89: > This is set in the config-features configmap in the knative-serving namespace. Definitions for the enabled / allowed values are provided in the docs/feature-flags.md page.

I'd at least call that out and link to that page maybe, as someone reading this I wouldn't know how to enable it in the configmap without the pointer.

psschwei: Sorry, I got the link wrong... those definitions are actually on this same page (in the first section).

abrennan89 (approved):
/lgtm

PR #3667, opened 2021-05-26
WIP: Make intro section smarter when breaking up

markusthoemmes:
As we've discussed a bit how this **should** break, I took another whack (see discussion in #3638)


omerbensaadon:
love it! 

/lgtm

PR #3670, opened 2021-05-26
Bump cert manager API verson

irbekrm:
`cert-manager` is deprecating all alpha and beta API versions in the next release (see [cert-manager#4021](https://github.com/jetstack/cert-manager/pull/4021)).

This PR changes `cert-manager` examples in docs to use `v1` API instead of `v1alpha2`.

It also bumps the minimum required version of `cert-manager` to be `v1.0.0`- this is the first stable version and also the first with `v1` API.

Please let me know if this is the wrong branch to PR against- I saw that most PRs were made against this branch rather than `main` so did the same.

Potentially related https://github.com/knative-sandbox/net-certmanager/issues/219

### DIFF
--- a/docs/admin/install/install-extensions.md
+++ b/docs/admin/install/install-extensions.md
@@ -47,7 +47,7 @@ The tabs below expand to show instructions for installing each Serving extension
     via cert-manager.
 
     1. First, install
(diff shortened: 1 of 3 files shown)


omerbensaadon:
@irbekrm yup you're totally right to open against this branch! 

Thanks for the PR :)

PR #3676, opened 2021-05-27
Added the experimental features page

slinkydeveloper:
Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

## Proposed Changes 

- This adds a page to document eventing experimental features. For more details: https://docs.google.com/document/d/1AoH0FyLeuHIg5snrlCKzELQv-NEA6bbjTU2Qv3zlW5k/edit?usp=sharing and https://github.com/knative/eventing/blob/main/docs/experimental-features.md

### DIFF
--- a/docs/eventing/experimental-features.md
+++ b/docs/eventing/experimental-features.md
@@ -1,9 +1,3 @@
----
-title: "Experimental features"
-weight: 99
-type: "docs"
----
-
 # Experimental features
 
 In Knative we want to keep the innovation alive, experimenting and delivering new features without affecting the stability of the project.
--- a/docs/eventing/experimental-features.md
+++ b/docs/eventing/experimental-features.md
@@ -8,7 +8,7 @@ type: "docs"
 
 In Knative Eventing we want to keep the innovation alive, experimenting and delivering new features without affecting the stability of the project.
 
-In order to do achieve that goal, we have a process to include new features. 
+In order to achieve that goal, we have a process to include new features.
 This allows users like you to try out new features and provide feedback back to the project.
 
 This document explains how to enable experimental features and which ones are available today.
@@ -21,7 +21,7 @@ You must have a Knative cluster running with the Eventing component installed. [
 
 ## Experimental features configuration
 
-When installing Eventing, the `config-experimental-features` ConfigMap is added to your cluster in the `knative-eventing` namespace. 
+When installing Eventing, the `config-experimental-features` ConfigMap is added to your cluster in the `knative-eventing` namespace.
 In order to enable a feature, you just need to add it to the config map and set its value to `"true"`.
 For example, to enable `new-cool-feature`:
 
--- a/docs/eventing/experimental-features.md
+++ b/docs/eventing/experimental-features.md
@@ -6,9 +6,9 @@ type: "docs"
 
 # Experimental features
 
-In Knative Eventing we want to keep the innovation alive, experimenting and delivering new features without affecting the stability of the project.
+In Knative we want to keep the innovation alive, experimenting and delivering new features without affecting the stability of the project.
 
-In order to achieve that goal, we have a process to include new features.
+In order to achieve that goal in Knative Eventing, we have a process to include new features.
 This allows users like you to try out new features and provide feedback back to the project.
 
 This document explains how to enable experimental features and which ones are available today.
--- a/docs/eventing/experimental-features.md
+++ b/docs/eventing/experimental-features.md
@@ -1,0 +1,60 @@
+---
+title: "Experimental features"
+weight: 99
+type: "docs"
+---
+
+# Experimental features
+
+In Knative Eventing we want to keep the innovation alive, experimenting and delivering new features without affecting the stability of the project.
+
+In order to do achieve that goal, we have a process to include new features. 
+This allows users like you to try out new features and provide feedback back to the project.
+
+This document explains how to enable experimental features and which ones are available today.
+
+For more details about the process, the feature phases, quality requirements and guarantees, check out the [Experimental features process documentation](https://github.com/knative/eventing/blob/main/docs/experimental-features.md).
+
+## Before you begin
+
+You must have a Knative cluster running with the Eventing component installed. [Learn more](../install/)
+
+## Experimental features configuration
+
+When installing Eventing, the `config-experimental-features` ConfigMap is added to your cluster in the `knative-eventing` namespace. 
+In order to enable a feature, you just need to add it to the config map and set its value to `"true"`.
+For example, to enable `new-cool-feature`:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-experimental-features
+  namespace: knative-eventing
+  labels:
+    eventing.knative.dev/release: devel
+    knative.dev/config-propagation: original
+    knative.dev/config-category: eventing
+data:
+  new-cool-feature: "true"
+```
+
+In order to disable it, you can either remove the flag or set it to `"false"`:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-experimental-features
+  namespace: knative-eventing
+  labels:
+    eventing.knative.dev/release: devel
+    knative.dev/config-propagation: original
+    knative.dev/config-category: eventing
+data:
+  new-cool-feature: "false"
+```
+
+## Features list
+
+<!-- TODO there are no experimental features at the moment -->
--- a/docs/eventing/experimental-features.md
+++ b/docs/eventing/experimental-features.md
@@ -15,38 +15,38 @@ You must have a Knative cluster running with the Eventing component installed. [
 
 ## Experimental features configuration
 
-When installing Eventing, the `config-experimental-features` ConfigMap is added to your cluster in the `knative-eventing` namespace.
-In order to enable a feature, you just need to add it to the config map and set its value to `"true"`.
+When installing Eventing, the `config-features` ConfigMap is added to your cluster in the `knative-eventing` namespace.
+In order to enable a feature, you just need to add it to the config map and set its value to `"enabled"`.
 For example, to enable `new-cool-feature`:
 
 ```yaml
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: config-experimental-features
+  name: config-features
   namespace: knative-eventing
   labels:
     eventing.knative.dev/release: devel
     knative.dev/config-propagation: original
     knative.dev/config-category: eventing
 data:
-  new-cool-feature: "true"
+  new-cool-feature: "enabled"
 ```
 
-In order to disable it, you can either remove the flag or set it to `"false"`:
+In order to disable it, you can either remove the flag or set it to `"disabled"`:
 
 ```yaml
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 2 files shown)


docs/eventing/experimental-features.md
```diff
@@ -0,0 +1,60 @@
...
+kind: ConfigMap
+metadata:
+  name: config-experimental-features
+  namespace: knative-eventing
+  labels:
+    eventing.knative.dev/release: devel
+    knative.dev/config-propagation: original
+    knative.dev/config-category: eventing
+data:
+  new-cool-feature: "false"
```
julz, line 55: I guess this is really a comment on the eventing implementation of this rather than the doc, but Im only noticing this here now :).. Would it make sense (I think the proposal doc said we were going to) to align this with the serving features configmap where the settings are "enabled", "disabled", "allowed"?

docs/eventing/experimental-features.md
```diff
@@ -0,0 +1,60 @@
...
+
+When installing Eventing, the `config-experimental-features` ConfigMap is added to your cluster in the `knative-eventing` namespace.
+In order to enable a feature, you just need to add it to the config map and set its value to `"true"`.
+For example, to enable `new-cool-feature`:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-experimental-features
```
julz, line 32: ditto that this is really a comment on the implementation not the doc, but making it here because this is where Im seeing it -- could this be `config-features` to align with the same thing in knative-serving namespace? (I'd say we should rename the serving one, but that would pose a backwards-compat challenge now).

docs/eventing/experimental-features.md
```diff
@@ -0,0 +1,60 @@
...
+weight: 99
+type: "docs"
+---
+
+# Experimental features
+
+In Knative Eventing we want to keep the innovation alive, experimenting and delivering new features without affecting the stability of the project.
+
+In order to achieve that goal, we have a process to include new features.
+This allows users like you to try out new features and provide feedback back to the project.
```
julz, line 12: Super nit, and this may just be how I read this, but I think specifying that "Knative Eventing" wants to keep the innovation alive reads a bit like we're saying that Serving doesnt 😅 which is obviously not the case! WDYT about something like: 
```suggestion
In Knative we want to keep the innovation alive, experimenting and delivering new features without affecting the stability of the project.

In order to achieve that goal in Knative Eventing, we have a process to include new features.
This allows users like you to try out new features and provide feedback back to the project.
```

docs/eventing/experimental-features.md
slinkydeveloper: We talked about that, but we decided to avoid having the opt-in (see the mail thread) features in that same CM (like you have in serving), hence this CM is only about "experimental feature on/off".

In any case, feel free to reply to that mail thread if you have some concrete proposal around it :). Can we ignore this discussion here for the sake of this PR?

julz: FWIW I think some of the serving features also only have "enabled" & "disabled", so I don't _think_ there's a reason we can't align, even if you don't have a need for an "allowed" state yet. Anyway let me go look for the email and reply there, I didn't realise there was a mail thread going on as well as the proposal doc (I've been filtering out stuff tagged [Eventing])

docs/eventing/experimental-features.md
csantanapr: UX feedback 
I think the word “experimental” should not be in any configuration to indicate the level of stability or to indicate that the configuration/feature might get removed

a better approach perhaps would be using docs to indicate this
For example a document comment in the yaml
A message on the command line like when using deprecated features in kubectl, they do this in Kubernetes know and even with colors text yellow to indicate warning
In the action website docs in both the section explaining how to use the feature and in the Go generate re fence html 

the benefit would be that as the user upgrades from old version to new version, and it happens that the feature is stable and not longer experimental 
There is not much work for the administrators to go and update the config in many places or coordinate with others to change the config in many places. Instead he can announce/doc that a new version of Knative is used and that feature X is not longer experimental 

for Knative to move it out from experimental we don’t modify tests or code. 

If the feature is removed from experimental then is a mute point the experience remains the same for that case as oppose feature remaining they have better experience

docs/eventing/experimental-features.md
```diff
@@ -0,0 +1,60 @@
+---
+title: "Experimental features"
+weight: 99
+type: "docs"
+---
+
```
abrennan89, line 6: This doesn't need to be added anymore for mkdocs afaik :)

docs/eventing/experimental-features.md
slinkydeveloper: So this CM is an entrypoint for people that wants to try out experimental features, and when a feature is stable, the flag won't be here anymore. The name experimental is just to keep it consistent with the _Experimental features process_.

The CM gets commented too with details about each specific flag, and then this page will host the documentation for each flag.

> For example a document comment in the yaml
A message on the command line like when using deprecated features in kubectl, they do this in Kubernetes know and even with colors text yellow to indicate warning
In the actual website docs in both the section explaining how to use the feature and in the Go generate rerefence html

For experimental features adding new fields to CRDs, we don't add them to the schema: https://github.com/knative/eventing/blob/main/docs/experimental-features.md#strategies-to-add-new-apis-for-experimental-features

dprotaso: > So this CM is an entrypoint for people that wants to try out experimental features, and when a feature is stable, the flag won't be here anymore.

Sure but like @julz and @csantanapr have pointed out someone coming from serving to enable some functionality will now have a _different_ approach will easily be confused and tripped up.

docs/eventing/experimental-features.md
dprotaso: The original doc it had the following clause - I'm surprised after the discussion in the mailing list the doc wasn't updated and comments resolved.

> We store in a config map a map of enabled/disabled features. Pretty much the same as Kubernetes Feature Gates or https://github.com/knative/serving/blob/master/pkg/apis/config/features.go

The tri-state is more expressive and allows for _users_ to individually opt into experimental behaviour - in contrast to the operator enabling it for everyone.

This allows for a gradual rollout of features.

slinkydeveloper: > The tri-state is more expressive and allows for users to individually opt into experimental behaviour - in contrast to the operator enabling it for everyone.

This was discussed as well both in the doc and in the mailing list: depending on the feature, we're going to decide if we want it opt-in and opt-out by default, but because eventing has plenty of CMs, most of them already namespace scoped, it doesn't make sense to have this ConfigMap too for enabling component features.

csantanapr:
From UX I agree with a tri-state is not that a bad option, but having consistency in the project as “this is how native handles experimental features” vs having fragmented experiences like Eventing works like x and Serving works like y is not a good UX

slinkydeveloper:
I've updated the page per https://github.com/knative/eventing/pull/5459, but I didn't added the `allowed` flag documentation, since we don't have any particular feature today that use it.

I suggest we document it once we have a feature that is going to use the `allowed` state, so we can properly describe why we need it and how to use it.

docs/eventing/experimental-features.md
```diff
@@ -0,0 +1,54 @@
...
+For more details about the process, the feature phases, quality requirements and guarantees, check out the [Experimental features process documentation](https://github.com/knative/eventing/blob/main/docs/experimental-features.md).
+
+## Before you begin
+
+You must have a Knative cluster running with the Eventing component installed. [Learn more](../install/)
+
+## Experimental features configuration
+
+When installing Eventing, the `config-features` ConfigMap is added to your cluster in the `knative-eventing` namespace.
+In order to enable a feature, you just need to add it to the config map and set its value to `"enabled"`.
```
abrennan89, line 19: use backticks for literal values, e.g. `enabled`

abrennan89, line 19: Note: @eric-sap related to the other PR you're doing, it looks like it's both `` and "" 🙂 so `"enabled"`

docs/eventing/experimental-features.md
```diff
@@ -0,0 +1,54 @@
...
+metadata:
+  name: config-features
+  namespace: knative-eventing
+  labels:
+    eventing.knative.dev/release: devel
+    knative.dev/config-propagation: original
+    knative.dev/config-category: eventing
+data:
+  new-cool-feature: "enabled"
+```
```
abrennan89, line 34: Can you add the kubectl command part as per the [style guide](https://docs.google.com/document/d/1tigolRzKhronvgoAJfszE-XHWWd0LQ8o_j0D73O9KvM/edit#heading=h.qqz6in2pxqzk) for any YAML files

docs/eventing/experimental-features.md
```diff
@@ -0,0 +1,54 @@
...
+    eventing.knative.dev/release: devel
+    knative.dev/config-propagation: original
+    knative.dev/config-category: eventing
+data:
+  new-cool-feature: "disabled"
+```
+
+## Features list
+
+<!-- TODO there are no experimental features at the moment -->
```
abrennan89, line 54: Should we add this at all if there are no features to doc yet? Seems strange to add a blank doc / placeholder

mkdocs.yml
```diff
@@ -187,6 +187,7 @@ nav:
       - Debugging: eventing/debugging/index.md
       - Accessing CloudEvent traces: eventing/accessing-traces.md
       - Metrics API: eventing/metrics.md
+      - Experimental Features: eventing/experimental-features.md
```
abrennan89, line 190: Should probably go in the admin guide if they all require configmap editing?

abrennan89 (changes requested):
Couple of comments

docs/eventing/experimental-features.md
eric-sap: In this particular case (judging by the new-cool-feature: example text below this sentence) it looks like the quotes are literally supposed to be part of the configmap (as opposed to, say, the metadata.name or namespace, which do not have quotes).  I'm unsure as to whether the yaml parser cares one way or the other.  I only mention it here because in the config-kafka configmap, there are no fields that require literal double-quotes around them (nor are there examples of such in the document).

csantanapr:
/lgtm
thanks for addressing my comments @slinkydeveloper

csantanapr (approved):

docs/eventing/experimental-features.md
slinkydeveloper: Will add in the next PR :)

mkdocs.yml
slinkydeveloper: :shrug: I see it's the same for serving...

docs/eventing/experimental-features.md
slinkydeveloper: I don't know if there's a kubectl counterpart for modifying the config maps...

docs/eventing/experimental-features.md
slinkydeveloper: That's my bad, i was using true and false as values before, which needs escapes. https://github.com/knative/docs/pull/3714

csantanapr:
don't know why this merged when the PR had a "Request changes" from @abrennan89

slinkydeveloper:
@csantanapr :shrug:, I'm addressing the comments in this other PR: https://github.com/knative/docs/pull/3714

PR #3683, opened 2021-05-27
Fix redirect to cope with query strings

julz:
Hopefully this will fix the netlify preview infinite redirects 🤞.


julz:
looks like that worked

omerbensaadon:
noice!!! thanks for this

PR #3697, opened 2021-06-01
hotfix for broken sinkbinding redirect 

omerbensaadon:


mkdocs.yml
```diff
@@ -278,8 +278,6 @@ plugins:
       strict: false
   - redirects:
       redirect_maps:
-        'eventing/samples/sinkbinding/': 'eventing/sources/sinkbinding/README.md'
```
snneji, line 281: Does this line need to be deleted as well? It is redirecting `eventing/samples/` not `eventing/sources/`?

omerbensaadon: oops you are totally right, good catch

Issue #3710, opened 2021-06-02
Organize additional content into admin vs developer guides


abrennan89:
**Describe the change you'd like to see**

There's a lot of stuff still in the Serving and Eventing sections that can be better organized by persona, so I'm going to try to move it incrementally, do some editing, fix some formatting, etc.

abrennan89:
I have some time on the calendar on the 2nd to talk to @devguyio re event sources, and Markus should be back next week to talk about the load balancing / autoscaling docs thing. Will update then, likely I'll finish this after my vacation in Aug.

abrennan89:
I feel like we've come to a point where we are realizing the issues / limitations of this type of structure.
It's apparent by looking at what's left in the serving and eventing doc sections, where there are things that don't necessarily fit in one persona or the other, e.g. autoscaling has its own section outside of developer or admin topics, because there are things in there that touch both: https://knative.dev/development/serving/autoscaling/

We've also seen a few folks raising issues with not intuitively knowing where to look for installation docs now that we have split them up this way.

I think our time will be better spent if we close this issue and get the card sorts / new structure done ASAP rather than investing much more time trying to figure out where things go in a structure that maybe doesn't work. At least we tried it 🤷🏻‍♀️ Feel free to reopen if anyone disagrees.

PR #3712, opened 2021-06-02
#3710: Move some serving topics to dev guide

abrennan89:
Part of the solution for #3710 (does not complete this issue, there will be multiple PRs)

## Proposed Changes 

- add dev guide troubleshooting for serving
- move custom domains to dev guide -> serving -> services
- content editing / formatting


mkdocs.yml
```diff
@@ -47,23 +48,26 @@ nav:
...
             - Creating a Service: developer/serving/services/creating-services.md
-            - Configure resource requests and limits: developer/serving/services/configure-requests-limits-services.md
             - Configuring private Services: developer/serving/services/private-services.md
-    - Serving Component:
+            - Configuring custom domains: developer/serving/services/custom-domains.md
+            - Configure resource requests and limits: developer/serving/services/configure-requests-limits-services.md
+          - Troubleshooting:
+            - Debugging application issues: developer/serving/troubleshooting/debugging-application-issues.md
+    # Serving
+    - Knative Serving:
```
csantanapr, line 64: We could drop `Knative` from `Knative Serving` if we need space to fit more top level items.
Same for the guides it can be "Administration" "Development"

abrennan89: Let's not do that please, I think we need "guide" for context and as part of our attempt to make docs more consistent we should be using Knative Serving and Knative Eventing throughout - the comments here are just for my own reference to help me find things in the long-winded yml file

PR #3633, opened 2021-05-20
Remove violet highlights from website

markusthoemmes:
The violet hover states and highlights hurt a bit in the otherwise purely blue them, imo.

PR #3717, opened 2021-06-03
Bringing back the purple button 

omerbensaadon:
People have been totally missing this button in user interviews and I want to see if it's because of the color 

Reverts #3633


omerbensaadon:
They used to go right to this button, so I figure it's either the color or the Copy change that's the problem

julz (approved):
Fwiw I cannot explain why (& it obviously doesn't match the rest of the theme) but for whatever reason I do personally like the purple button also 🤷‍♂️

Anyway let's ask some users :)

/lgtm

PR #3718, opened 2021-06-03
(Fixes #3719) Make installing kn into a re-usable snippet 

### LINKED ISSUE #3719: Create re-usable snippet for installing `kn`
We want to poke at out our emerging re-usable content strategy by getting this low-hanging fruit delivered

omerbensaadon:


docs/snippets/README.md
```diff
@@ -0,0 +1,8 @@
+# Snippets
```
julz, line 1: might be worth putting the `snippets` dir in `base_paths` so the path doesn't need to be specified (see https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#specifying-snippet-locations). Also if the snippets dir lives under docs/ we should probably add it to [exclude](https://github.com/knative/docs/blob/mkdocs/mkdocs.yml#L276) so it doesn't get built.

docs/snippets/install-kn.md
```diff
@@ -0,0 +1,50 @@
...
+
+        - <a href="https://gcr.io/knative-releases/knative.dev/client/cmd/kn" target="_blank">Latest release</a>
+
+        You can run `kn` from a container image. For example:
+
+        ```
+        docker run --rm -v "$HOME/.kube/config:/root/.kube/config" gcr.io/knative-releases/knative.dev/client/cmd/kn:latest service list
+        ```
+
+        !!! note
```
abrennan89, line 49: One super minor nit but I'd take this out of the note and put it before the numbered steps in this bit, just cos it's important and also a note inside a tab like this looks kinda odd. Your call though.

abrennan89 (commented):
One minor suggestion but this is 🔥 overall, thanks @omerbensaadon !

docs/snippets/install-kn.md
abrennan89: This isn't a blocker so I'll let @julz lgtm cos it looks like he had suggestions also

docs/snippets/README.md
omerbensaadon: I will do this now thank you!!

omerbensaadon:
found the problem, testing on GH pages and pushing fix

omerbensaadon:
ok this works!

julz (approved):
/lgtm

PR #3727, opened 2021-06-04
Drop install instructions for Gloo until we have better CI coverage

julz:
As discussed in the recent TOC meeting, we want to be sure the Ingresses
we list in the install instructions are ones we're confident will work
well with Knative, to avoid folks having a bad first experience.

Currently we don't have good (any) CI coverage for Gloo, so removing it
from the list for the moment until we do.

/assign @dprotaso @nak3 @ZhiminXiang

### DIFF
--- a/docs/admin/install/knative-with-operators.md
+++ b/docs/admin/install/knative-with-operators.md
@@ -286,53 +286,6 @@ kubectl logs -f deploy/knative-operator
 
           Save this for configuring DNS below.
 
-    === "Gloo"
-
-        _For a detailed guide on Gloo integration, see
-        [Installing Gloo for Knative](https://docs.solo.io/gloo/latest/installation/knative/)
-        in the Gloo documentation._
-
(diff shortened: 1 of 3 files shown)


omerbensaadon:
I would relish the opportunity to drop these

ZhiminXiang:
From the TOC meeting yesterday, sounds like Gloo does not maintain their kingress implementation. @dprotaso do you know about it?

dprotaso:
@ZhiminXiang I reached out and they said they have their own CI testing Knative upstream. I double checked it and it hasn't been updated in two years.

julz:
Rebased and updated to just remove Gloo (since Kong is covered in #3745) so we can revert independently when/if we have better coverage.

ZhiminXiang:
/lgtm

> @ZhiminXiang I reached out and gloo folks said they have their own CI testing Knative upstream. I double checked it and it hasn't been updated in two years.

@dprotaso Thanks for the confirmation!

PR #3730, opened 2021-06-04
Fix active page highlight

julz:
This should fix the CSS so the current page is highlighted in blue in the sidebar again.


abrennan89:
Beautiful, thanks @julz 
/lgtm
/approve

PR #3732, opened 2021-06-04
#3447: Update installation docs and links

abrennan89:
Partial fix for #3447 - still have to do Operator docs.
Fixing broken links, split serving and eventing into own install sections.
Made prereqs into a snippet.


csantanapr:
/lgtm
I verified this in preview

PR #3742, opened 2021-06-07
Drop unneccesary important specifiers

julz:
The `!important` specifier should [generally be avoided because it makes css hard to maintain](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception). Luckily, it doesn't look like we actually need it in any of these cases :).


abrennan89 (approved):
/lgtm

omerbensaadon:
CC: @soham283

PR #3743, opened 2021-06-07
Issue 3642 finishing up eventing sanity check 

omerbensaadon:
Fixes issue #3642 and #3644

Also, replaces all shell with bash


vaikas:
I'd personally change the PR description to something more descriptive as this just seemingly does s/shell/bash/ :) But maybe that's sanity :)
/lgtm

PR #3757, opened 2021-06-09
Revert changes to community

julz:
This commit reverts changes made to community which assumed it was being built by mkdocs (since it's still built with hugo) and caused formatting to break in the community tab.

/assign @omerbensaadon I did a straight `git checkout HEAD~10 community/` so please check I didn't revert something we actually wanted here, but it looks right to me

### DIFF
--- a/community/_index.md
+++ b/community/_index.md
@@ -1,0 +1,12 @@
+---
+title: "Join the Knative community"
+linkTitle: "Community"
+type: "docs"
+showlandingtoc: "false"
+---
+
+Knative is an open source project that anyone in the community can use, improve, and enjoy. We'd love ❣ you to join us!
+
+Here are a few ways to find out what's happening and get involved.
+
+{{< showpartial "community_links.html" >}}
--- a/community/annual_reports/_index.md
+++ b/community/annual_reports/_index.md
@@ -1,0 +1,14 @@
+---
+title: "Knative Annual Reports"
+linkTitle: "Annual reports"
+weight: 40
+type: "docs"
+---
+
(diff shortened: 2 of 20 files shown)


omerbensaadon:
Can you do a find and replace for shell -> bash?

julz:
> Can you do a find and replace for shell -> bash?

Do we care about applying these rules to community? Couldn't we/shouldn't we leave that folder till after GA when we fix up community to use mkdocs?

Either way, if we do want to do that again I'd suggest we re-apply it in a different PR so each PR does one thing as much as possible: having PRs making lots of unrelated changes made this a bit harder to track down via git bisect than it needed to be

omerbensaadon:
Sorry, I wasn't clear, it was already fixed but this PR changes it back to shell, so it's a regressive change

omerbensaadon:
@julz I will leave it up to you just in case you did understand my first message as I meant it!

julz:
I personally think we should probably just leave community/ alone for now, but if it was intentional to fix shell->bash there as well as elsewhere I'm happy to do that conversion again in a follow-on PR once this lands. (It took longer than it should have to track this bug down, and it was harder to revert cleanly and safely, because the commit that introduced the problem made lots of changes at once, so I'd like to avoid repeating that in the fix iyswim).

/unhold

julz:
as soon as this merges Ill do a follow-on to s/shell/bash/ in community, and we can double check that all looks right with hugo (I assume it does, but since the broader change this came in with actually broke community and we didn't realise, we should check!)

PR #3766, opened 2021-06-09
Adding quickstart instructions for minikube

psschwei:
Currently, the "Getting Started" section has instructions for getting set up using Kind. However, some developers prefer to use Minikube instead or have issues getting Kind working on their systems. To support those users, this PR adds instructions for getting started with Knative using Minikube.

The [script](https://github.com/psschwei/knative-on-minikube/blob/main/knative-on-minikube.sh) used for setting up Minikube is currently housed in my personal repo, but it should probably be moved into the docs repo somewhere.

## Proposed Changes 

- Adds quickstart instructions to the docs for setting up a local cluster using minikube

### DIFF
--- a/docs/getting-started/README.md
+++ b/docs/getting-started/README.md
@@ -5,6 +5,8 @@ Before you can get started with a Knative Quickstart deployment you must set up 
 
 You can use [`kind`](https://kind.sigs.k8s.io/docs/user/quick-start){target=_blank} (Kubernetes in Docker) to run a local Kubernetes cluster with Docker container nodes.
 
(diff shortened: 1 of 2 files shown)


csantanapr:
/hold
We don't want to document instructions a quick-start for every local k8s flavor (ie minikube, docker-desktop, kind, k3d, rancher-desktop) including the different matrix of architecture and operating system.

There is a discussion on creating a `kn quickstart` plugin to target one simple local sandbox, we need to flesh out what implementation we want and the UX for an mvp

psschwei:
If we do do minikube, it should be handled by the [quickstart plugin](https://github.com/knative-sandbox/kn-plugin-quickstart).

Issue #3696, opened 2021-06-01
Update style for H4 headings


snneji:
# Issue
H4 headings look like bold text. It would be nice for these headings to have a distinctive style. 

# Example

**Screenshot of H4 formatting:**
(https://knative.dev/docs/developer/eventing/sources/sinkbinding/reference/#cloudevent-overrides)

<img width="739" alt="Screen Shot 2021-05-28 at 18 18 33" src="https://user-images.githubusercontent.com/33425661/120304186-77419800-c2c7-11eb-8309-6ed9e9dc80dc.png">

# Resources
The CSS file for customizing the layout can be found here: https://github.com/knative/docs/blob/mkdocs/docs/stylesheets/extra.css

soham283:
I would like to work on this issue, @snneji can you please assign this issue ?

snneji:
Hi @soham283, I assigned you to this issue. I believe you can also assign yourself to issues using the `/assign` command in a comment. The Knative prow robot then assigns you to the issue.

soham283:
please review the pr #3767  @snneji and @omerbensaadon

snneji:
Done in #3767

PR #3767, opened 2021-06-09
h4 tag change in font weight for fixing issue (#3696)

soham283:
## Proposed Changes 

- changed font weight for h4 tag similar value with other heading tag like h3.
- For fixing issue #3696
- 
knative_h4


omerbensaadon:
CC @snneji @csantanapr

docs/stylesheets/extra.css
```diff
@@ -68,3 +68,9 @@
 .md-typeset .admonition {
   font-size: 100%;
 }
+
+/* Giving h4 tag distinctive style because it is non-differntial with bold text */
+.md-typeset h4 {
+  font-weight: 400;
```
julz, line 74: `font-weight: 400` is the same as normal text (see https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#values), so Im not sure this is really an improvement :(.

soham283: The #3696 issue it was stated that bold text and h4 tags context must have distinct style so that it can be recognized easily
that's why I came up with this

julz: yes, but this makes it so that the h4 tags are impossible to distinguish from regular text, which is even worse than being hard to distinguish from bold text?

soham283: maybe increasing font-size may work initially i thinked about that but thought if styling can make difference

soham283:
Now updated the commit by changing font size, and removed font weight which were making is difficult to differentiate with the normal text.
@snneji @julz @omerbensaadon 
have a look on the new commit made

snneji:
Hi @soham283, thanks for working on this!

I think 1.0rem H4 font size looks a bit too large and heavy compared to H3 headings. I did a few tests in my browser. My personal preference is ~0.88rem. I was split between having the standard or a lighter font weight. My preferences are as follows:

- For the standard font weight (700), I prefer the H4 font size to be between 0.86 and 0.88 rem. Example page:
  Comparative heading size 2 - weight 700
  Headings within text - H4 weight 700

- For a lighter font weight (400), I prefer the H4 font size to be a bit larger — between 0.88 and 0.9 rem. Example page:
  Comparative heading size 3 - weight 400
  Headings within text - H4 weight 400

I suppose this is a matter of personal taste so feel free to disagree @soham283, @julz, and @omerbensaadon!

csantanapr:
@snneji can you take another look when you have a chance?

julz (approved):
/lgtm
/approve

Issue #3703, opened 2021-06-01
Fix formatting issues in Feature Flags doc


psschwei:
Based on discussion in #3666, there's a couple of formatting updates we can make to [docs/serving/feature-flags.md](https://github.com/knative/docs/blob/mkdocs/docs/serving/feature-flags.md):

1. Full examples for yaml samples rather than snippets (for example, [here](https://github.com/knative/docs/blame/mkdocs/docs/serving/feature-flags.md#L104-L121) and [here](https://github.com/knative/docs/blame/mkdocs/docs/serving/feature-flags.md#L129-L141))
2. Using code blocks instead of quotes for key values (i.e. `enabled` instead of "enabled", for example [here](https://github.com/knative/docs/blame/mkdocs/docs/serving/feature-flags.md#L208-L209) and [here](https://github.com/knative/docs/blame/mkdocs/docs/serving/feature-flags.md#L248))

/kind mkdocs
/kind style
/kind good-first-issue

abrennan89:
I think wrt the snippets of code, `...` has been added now in each of them to show that it's just an excerpt, so that's acceptable IMO?

The code text > quotes does still need to be updated though.

abrennan89:
Relates to #3710 since this should be in the admin guide (configmap docs).

PR #3774, opened 2021-06-10
Edit ContainerSource docs

snneji:
Closes https://github.com/knative/eventing/issues/4912.

Docs should now provide:

- Getting Started demo using the component.
- Sample minimum of component
- Explanation of resource.
- All spec fields
- With kubectl cli
- With kn cli

This should probably have a thorough technical review because I guessed at the reference page using info from [containersource.yaml](https://github.com/knative/eventing/blob/1bbc0130b53a21ad592fe745712c42efbcbc7c41/config/core/resources/containersource.yaml).


mkdocs.yml
```diff
@@ -147,7 +147,9 @@ nav:
         - API server source:
           - Overview: eventing/sources/apiserversource/README.md
           - Getting started: eventing/sources/apiserversource/getting-started/README.md
-        - ContainerSource: eventing/sources/containersource.md
+        - ContainerSource:
+          - Creating a ContainerSource object: eventing/sources/containersource/README.md
+          - ContainerSource Reference: eventing/sources/containersource/reference.md
```
abrennan89, line 152: While we're moving this I'd go ahead and move it to the developer guide instead?

docs/eventing/sources/containersource/reference.md
```diff
@@ -0,0 +1,131 @@
...
+### Template parameter
+
+This is a `template` in the shape of `Deployment.spec.template` to use for the ContainerSource.
+For more information, see the [Kubernetes Documentation](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/).
+
+<!-- not sure what the required and optional fields are for this. -->
+
+#### Example: template parameter
+
+  ```yaml
```
abrennan89, line 73: I think for any YAML examples we should show a whole YAML shape rather than a snippet to show where abouts the field goes in the CR structure?

docs/eventing/sources/containersource/reference.md
```diff
@@ -0,0 +1,131 @@
...
+
+#### Example: sink parameter
+
+Given the following YAML, if `ref` resolves into
+`"http://mysink.default.svc.cluster.local"`, then `uri` is added to this
+resulting in `"http://mysink.default.svc.cluster.local/extra/path"`.
+
+<!-- TODO we should have a page to point to describing the ref+uri destinations and the rules we use to resolve those and reuse the page. -->
+
+```yaml
```
abrennan89, line 49: same as below re YAMLs, I think it would be better if we show example bits in context of a whole resource

docs/eventing/sources/containersource/README.md
```diff
@@ -0,0 +1,224 @@
...
+    === "kn"
+
+        ```bash
+        kn service create event-display --port 8080 --image gcr.io/knative-releases/knative.dev/eventing/cmd/event_display
+        ```
+
+    === "YAML"
+
+        ```yaml
+        kubectl -n containersource-example apply -f - << EOF
```
abrennan89, line 68: same as below, for all of `<< EOF` I think it should be `<<EOF` - if not, our style guide needs an update https://dev-knative.netlify.app/development/help/contributor/style-guide/documenting-code/#documenting-yaml

idk if this actually matters but just thought I'd point it out in case, not sure if you're testing these commands

docs/eventing/sources/containersource/reference.md
snneji: My aim with these snippets was to show an example using the fields for the specific parameter (eg the `sink` parameter). The full resource is shown on the procedure page, which links out to this page for more info about the options to use.

Would you be able to explain further how I can improve this or maybe give an example of what you mean? I am open to ideas :)

mkdocs.yml
snneji: I created an Eventing section in the dev guide and moved the ContainerSource topics over. Let me know if that lines up with what you were thinking.

docs/eventing/sources/containersource/reference.md
abrennan89: I'd like for our snippets to be sort of like the example below, to show _where_ `sink` sits within the full YAML if that makes sense. Wrt to linking pages, I think we shouldn't assume that users will always see something that's on a different page, e.g. maybe they already know how to create a ContainerSource but they're just coming directly to this page as a quick reference to see where/how they'd add a sink in an existing YAML, and we don't really show them that if it's just a snippet without context.

I'd like to see this added to the style guide as well since there's at least one other open issue related to this https://github.com/knative/docs/issues/3703

```apiVersion: sources.knative.dev/v1
kind: ContainerSource
metadata:
  name: test-heartbeats
spec:
  ...
  sink:
    ref:
      kind: Service
      name: event-display

snneji: I see what you mean. We do snippets like that in our docs but I was unsure whether to use that style in Knative. I'll update the snippets. :)

abrennan89: Opened an issue to add this to the style guide also https://github.com/knative/docs/issues/3794

abrennan89 (approved):
/lgtm

PR #3798, opened 2021-06-14
3710: Move event delivery docs to developer guide

abrennan89:
Another partial fix for #3710 

## Proposed Changes

- Move all non-configmap event delivery docs into one topic
- Move this topic to the developer guide


csantanapr:
/lgtm
thanks for putting this under one page

PR #3806, opened 2021-06-15
Generate Golang vanity URL redirects

### LINKED ISSUE #3785: 🛑 Add ability to import Go packages site-wide
## Description: 
In order to be able to import go packages as e.g. knative.dev/serving/foo/bar instead of github.com/knative/serving/foo/bar you have to prove you own knative.dev by serving a meta tag in the page when knative.dev/serving/foo/bar is requested with a particular header set

So if we forget, all the imports in the code wont work :smile:

---

## Story
```
As Julz
I want to make Go imports work site-wide
So that the site will work 
```

## Acceptance Criteria
If @csantanapr thinks this is /lgtm, we should be OK 

## Who should probably pick up this issue? 
- [x] Existing Contributors 
- [ ] New Contributors

julz:
Moves tools/redir-gen from website to docs so we can do all the generation here. Modifies the build script to add the generated redirects to the output.

Fixes #3785.

To test: 
1. `curl https://<netlify preview url>/serving?go-get=1`
1. Validate output exactly matches `curl https://knative.dev/serving?go-get=1`
1. If you like, try a few other repos

To generate redirects again (e.g. if a new repo is added to one of the knative orgs):
1. Go to tools/redir-gen, do `go install .`
2. Run `redir-gen` in root folder of docs.


tools/redir-gen/main.go
```diff
@@ -0,0 +1,184 @@
...
+	return nil
+}
+
+var fileTemplate = template.Must(template.New("gohtml").Parse(goHTML))
+var redirTemplate = template.Must(template.New("redir").Parse(redirText))
+
+// createGoGetFile creates a static HTML file providing a knative.dev mapping
+// for the specified org and repo.
+func createGoGetFile(ri repoInfo) error {
+	filename := fmt.Sprintf("golang/%s.html", ri.Repo)
```
csantanapr, line 160: nit
this fails if `golang` directory doesn't exist. I worked around create the directory first.
```
2021/06/15 16:09:23 Unable to create go mod file for knative-sandbox/actions-downstream-test: open golang/actions-downstream-test.html: no such file or directory
```

csantanapr, line 160:
```suggestion
	dirPath := "golang"
	if _, err := os.Stat(dirPath); os.IsNotExist(err) {
		if err = os.MkdirAll(dirPath, 0777); err != nil {
			return fmt.Errorf("error creating directory: %w", err)
		}
	}
	filename := fmt.Sprintf("%s/%s.html", dirPath, ri.Repo)
```

csantanapr (approved):
/approve
/lgtm

tools/redir-gen/main.go
csantanapr: you can address this in another PR if you want to address it

julz: How can it not exist? The directory is added in this commit

csantanapr: humm interesting, let me double check

csantanapr: I ran it from root of the repo and got some problem, then I ran it from the tools directory

csantanapr: Ah I got this when I tried to run the script
```
❯ go run tools/redir-gen/main.go
tools/redir-gen/main.go:27:2: cannot find package "." in:
	/Users/csantana23/dev/knative/docs/vendor/github.com/google/go-github/v32/github
```

csantanapr: How is the script suppose to be run?

csantanapr: Then I did
```
❯ cd tools/redir-gen/
❯ go run main.go
```

julz: This is a straight copy/paste of the existing redir-gen in website repo; while we _could_ tidy it up here/improve it, I'd rather do that separately. As I say above, the directory must exist, or this is being run from the wrong location and it's better to error in that case.

(FWIW if we did want to create the directory there's no need to do the os.Stat since os.MkdirAll does not error if the directory already exists. It would also probably be better to use `filepath.Join` to create the filename)

julz: Instructions in the PR description :)

> To generate redirects again (e.g. if a new repo is added to one of the knative orgs):
> 
> Go to tools/redir-gen, do go install .
> Run redir-gen in root folder of docs.

csantanapr: I felt that whatever I pointed out how to create the directory was going to be wrong.
I looked up an example from the knative source code first :-)
https://github.com/knative/serving/blob/561d9d0c6a04256fcd56fd304107fbe3b2e77dff/vendor/knative.dev/pkg/test/helpers/dir.go#L37-L38

julz: Heh, I might PR a tidy up of that tomorrow then. Stat + mkdir makes sense, or just MkdirAll on its own, but stat + mkdirall is redundant - mkdirall already does a stat. I got worried I was missing something for a sec but StackOverflow agrees 😂 https://stackoverflow.com/questions/35661967/checking-with-os-stat-vs-just-os-mkdirall.

PR #3807, opened 2021-06-15
Relocating ContainerSource docs to dev guide

snneji:
In PR #3774 I moved the ContainerSource docs to the developer guide in the navigation, but forgot to relocate the files in the directory structure. This PR fixes that.

### DIFF
--- a/docs/developer/eventing/sources/containersource/README.md
+++ b/docs/developer/eventing/sources/containersource/README.md
@@ -1,0 +1,224 @@
+# Creating a ContainerSource object
+
+![API version v1](https://img.shields.io/badge/API_Version-v1-red?style=flat-square)
+
+This topic describes how to configure ContainerSource as an event source for
+functions.
+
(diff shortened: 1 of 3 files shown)


snneji:
@abrennan89 as you suggested I moved the ContainerSource docs to the developer section of the directory.

abrennan89:
Thanks @snneji !

I just wanted to double check, is there any further update needed to mkdocs.yaml for this PR?

The links seem to work fine, e.g. https://deploy-preview-3807--dev-knative.netlify.app/development/eventing/sources/containersource/ but I think it should be (development)/developer/eventing/sources/containersource, so it seems like the links are still pointing to just eventing/sources if that makes sense, rather than developer/eventing/sources?

snneji:
@abrennan89 thanks for spotting that issue with the URL. I believe I've fixed it.

abrennan89 (approved):
/lgtm

PR #3814, opened 2021-06-16
Cleaning up sink docs

abrennan89:
No related issue in /docs - this PR cleans up some of the sink docs so that they're grouped in one place in the developer guide, since sinks are not exclusive to SinkBinding objects


navidshaikh (approved):
/lgtm

PR #3819, opened 2021-06-17
Document missing activator-capacity

markusthoemmes:
As per title, this setting was missing in our documentation and completes the autoscaling docs again.

/assign @julz

### DIFF
--- a/docs/serving/load-balancing/README.md
+++ b/docs/serving/load-balancing/README.md
@@ -15,7 +15,7 @@ You can turn on Knative load balancing, by placing the _Activator service_ in th
 Activator pods are scaled horizontally, so there may be multiple Activators in a deployment. In general, the system will perform best if the number of revision pods is larger than the number of Activator pods, and those numbers divide equally.
 <!--TODO(#2472): Add better documentation about what the activator is; explain the components of load balancing; maybe add a diagram-->
 
-Knative assigns a subset of Activators for each revision, depending on the revision size. More revision pods will mean a greater number of Activators for that revision.
+Knative assigns a subset of Activators for each revision, depending on the revision size and the configured capacity per Activator. More revision pods will mean a greater number of Activators for that revision.
 
 The Activator load balancing algorithm works as follows:
 
@@ -24,12 +24,13 @@ The Activator load balancing algorithm works as follows:
 
 For more information, see the documentation on [concurrency](../../serving/autoscaling/concurrency).
 
-## Configuring target burst capacity
+## Configuring target burst capacity behavior
 
 Target burst capacity is mainly responsible for determining whether the Activator is in the request path outside of scale-from-zero scenarios.
 
-Target burst capacity can be configured using a combination of the following parameters:
+Target burst capacity and it's effect of the system can be configured using a combination of the following parameters:
 
 - Setting the targeted concurrency limits for the revision. See [concurrency](../../serving/autoscaling/concurrency).
 - Setting the target utilization parameters. See [target utilization](../../serving/autoscaling/concurrency#target-utilization).
 - Setting the target burst capacity. You can configure target burst capacity using the `autoscaling.knative.dev/targetBurstCapacity` annotation key in the `config-autoscaler` ConfigMap. See [Setting the target burst capacity](./target-burst-capacity#setting-the-target-burst-capacity).
+- Setting the Activator capacity via the `config-autoscaler` ConfigMap. See [Setting the Activator capacity](./target-burst-capacity#setting-the-activator-capacity)
--- a/docs/serving/load-balancing/README.md
+++ b/docs/serving/load-balancing/README.md
@@ -24,13 +24,13 @@ The Activator load balancing algorithm works as follows:
 
 For more information, see the documentation on [concurrency](../../serving/autoscaling/concurrency).
 
-## Configuring target burst capacity behavior
+## Configuring target burst capacity
 
 Target burst capacity is mainly responsible for determining whether the Activator is in the request path outside of scale-from-zero scenarios.
 
-Target burst capacity and its effect of the system can be configured using a combination of the following parameters:
+Target burst capacity can be configured using a combination of the following parameters:
 
 - Setting the targeted concurrency limits for the revision. See [concurrency](../../serving/autoscaling/concurrency).
 - Setting the target utilization parameters. See [target utilization](../../serving/autoscaling/concurrency#target-utilization).
 - Setting the target burst capacity. You can configure target burst capacity using the `autoscaling.knative.dev/targetBurstCapacity` annotation key in the `config-autoscaler` ConfigMap. See [Setting the target burst capacity](./target-burst-capacity#setting-the-target-burst-capacity).
-- Setting the Activator capacity via the `config-autoscaler` ConfigMap. See [Setting the Activator capacity](./activator-capacity#setting-the-activator-capacity)
+- Setting the Activator capacity by using the `config-autoscaler` ConfigMap. See [Setting the Activator capacity](./activator-capacity#setting-the-activator-capacity)
--- a/docs/serving/load-balancing/README.md
+++ b/docs/serving/load-balancing/README.md
@@ -15,7 +15,7 @@ You can turn on Knative load balancing, by placing the _Activator service_ in th
 Activator pods are scaled horizontally, so there may be multiple Activators in a deployment. In general, the system will perform best if the number of revision pods is larger than the number of Activator pods, and those numbers divide equally.
 <!--TODO(#2472): Add better documentation about what the activator is; explain the components of load balancing; maybe add a diagram-->
 
-Knative assigns a subset of Activators for each revision, depending on the revision size and the configured capacity per Activator. More revision pods will mean a greater number of Activators for that revision.
+Knative assigns a subset of Activators for each revision, depending on the revision size. More revision pods will mean a greater number of Activators for that revision.
 
 The Activator load balancing algorithm works as follows:
 
@@ -28,7 +28,7 @@ For more information, see the documentation on [concurrency](../../serving/autos
 
 Target burst capacity is mainly responsible for determining whether the Activator is in the request path outside of scale-from-zero scenarios.
 
-Target burst capacity and it's effect of the system can be configured using a combination of the following parameters:
+Target burst capacity and its effect of the system can be configured using a combination of the following parameters:
 
 - Setting the targeted concurrency limits for the revision. See [concurrency](../../serving/autoscaling/concurrency).
 - Setting the target utilization parameters. See [target utilization](../../serving/autoscaling/concurrency#target-utilization).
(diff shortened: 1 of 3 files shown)


julz:
Lgtm from engineering perspective, do we want to update the page title now it includes activator-capacity as well as tbc?

markusthoemmes:
Hmm, tbh I'd leave it as-is @julz as activator-capacity only ever becomes relevant in the context of target burst capacity :thinking:

julz:
Hmm.. my mental model may be wrong but I don't quite understand that. As an operator mightn't I want to tweak how many requests each activator handles - for example because I changed the resources assigned to the activators - even if TBC is set to the default value (ie 200)? While it's true that if you set TBC to 0 the activator isn't in the path in the first place, I'm not sure that means an administrator would be able to guess that a document called "Configuring Target Burst Capacity" would be where the doc on this flag would live?

markusthoemmes:
Good point, I totally neglected that TBC isn't 0 by default anymore. 

As such, should I create a new page next to tbc to document this?

julz:
Yeah I was thinking a title that captures both like "Configuring Activator Burst Handling" or something might work, but an extra page seems fine too (and I admit I'm struggling to come up with a great title so.. maybe best 😃)

docs/serving/load-balancing/README.md
```diff
@@ -15,7 +15,7 @@ You can turn on Knative load balancing, by placing the _Activator service_ in th
 Activator pods are scaled horizontally, so there may be multiple Activators in a deployment. In general, the system will perform best if the number of revision pods is larger than the number of Activator pods, and those numbers divide equally.
 <!--TODO(#2472): Add better documentation about what the activator is; explain the components of load balancing; maybe add a diagram-->
 
-Knative assigns a subset of Activators for each revision, depending on the revision size. More revision pods will mean a greater number of Activators for that revision.
+Knative assigns a subset of Activators for each revision, depending on the revision size and the configured capacity per Activator. More revision pods will mean a greater number of Activators for that revision.
```
julz, line 18: being a pedant, is this also dependent on tbc?

docs/serving/load-balancing/README.md
```diff
@@ -24,12 +24,13 @@ The Activator load balancing algorithm works as follows:
 
 For more information, see the documentation on [concurrency](../../serving/autoscaling/concurrency).
 
-## Configuring target burst capacity
+## Configuring target burst capacity behavior
 
 Target burst capacity is mainly responsible for determining whether the Activator is in the request path outside of scale-from-zero scenarios.
 
-Target burst capacity can be configured using a combination of the following parameters:
+Target burst capacity and it's effect of the system can be configured using a combination of the following parameters:
```
julz, line 31:
```suggestion
Target burst capacity and its effect of the system can be configured using a combination of the following parameters:
```

docs/serving/load-balancing/README.md
markusthoemmes: Oh I guess I should strip both out again as those concepts have not yet been introduced.

docs/serving/load-balancing/activator-capacity.md
```diff
@@ -0,0 +1,41 @@
+---
+title: "Configuring Activator capacity"
+linkTitle: "Configuring Activator capacity"
+weight: 50
+type: "docs"
+---
+
+# Configuring Activator capacity
+
+If there's more than one Activator in the system, Knative will put as many of them on the request path as it thinks is necessary (if target burst capacity is not set to 0). If available, it'll pick at least two for high availability reasons. The actual number of Activators is calculated via a given _Activator capacity_ like such: `(replicas * target + targetBurstCapacity)/activatorCapacity`. That means, there are enough Activators in the routing path to handle the theoretic capacity of the existing application, including any additional target burst capacity.
```
julz, line 10: How about something (for the first sentence) like: "..Knative will put as many of them on to the request path as it thinks is necessary to handle the current request load plus the Target Burst Capacity. If Target Burst Capacity is 0 Knative only puts the activator in to the path when the revision is scaled to zero"

julz: Also because I happen to know the style guide will not let us use parentheticals :D

docs/serving/load-balancing/activator-capacity.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Configuring Activator capacity"
+linkTitle: "Configuring Activator capacity"
+weight: 50
+type: "docs"
+---
+
+# Configuring Activator capacity
+
+If there's more than one Activator in the system, Knative will put as many of them on the request path as it thinks is necessary to handle the current request load plus the target burst capacity. If the target burst capacity is 0, Knative only puts the Activator into the into the request path if the Revision is scaled to zero.
```
julz:
```suggestion
If there's more than one Activator in the system, Knative will put as many of them on the request path as it thinks is necessary to handle the current request load plus the target burst capacity. If the target burst capacity is 0, Knative only puts the Activator into the request path if the Revision is scaled to zero. 
```

docs/serving/load-balancing/activator-capacity.md
```diff
@@ -0,0 +1,43 @@
...
+linkTitle: "Configuring Activator capacity"
+weight: 50
+type: "docs"
+---
+
+# Configuring Activator capacity
+
+If there's more than one Activator in the system, Knative will put as many of them on the request path as it thinks is necessary to handle the current request load plus the target burst capacity. If the target burst capacity is 0, Knative only puts the Activator into the into the request path if the Revision is scaled to zero. 
+
+If available, Knative will pick at least two Activators for high availability reasons. The actual number of Activators is calculated via a given _Activator capacity_ like such: `(replicas * target + targetBurstCapacity)/activatorCapacity`. That means, there are enough Activators in the routing path to handle the theoretic capacity of the existing application, including any additional target burst capacity.
```
julz, line 12:
```suggestion
If available, Knative will pick at least two Activators for high availability reasons. The actual number of Activators is calculated via a given _Activator capacity_ using the formula: `(replicas * target + targetBurstCapacity)/activatorCapacity`. That means, there are enough Activators in the routing path to handle the theoretic capacity of the existing application, including any additional target burst capacity.
```

docs/serving/load-balancing/activator-capacity.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Configuring Activator capacity"
+linkTitle: "Configuring Activator capacity"
+weight: 50
+type: "docs"
+---
+
+# Configuring Activator capacity
+
+If there's more than one Activator in the system, Knative will put as many of them on the request path as it thinks is necessary to handle the current request load plus the target burst capacity. If the target burst capacity is 0, Knative only puts the Activator into the into the request path if the Revision is scaled to zero.
```
julz, line 10:
```suggestion
If there's more than one Activator in the system, Knative will put as many of them on the request path as it thinks is necessary to handle the current request load plus the target burst capacity. If the target burst capacity is 0, Knative only puts the Activator into the request path if the Revision is scaled to zero.
```

julz (commented):
only nits, lgtm from an engineering POV

docs/serving/load-balancing/activator-capacity.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Configuring Activator capacity"
+linkTitle: "Configuring Activator capacity"
+weight: 50
+type: "docs"
+---
+
```
abrennan89, line 7: This frontmatter doesn't need to be added anymore since we moved to mkdocs

docs/serving/load-balancing/activator-capacity.md
```diff
@@ -0,0 +1,43 @@
...
+# Configuring Activator capacity
+
+If there's more than one Activator in the system, Knative will put as many of them on the request path as it thinks is necessary to handle the current request load plus the target burst capacity. If the target burst capacity is 0, Knative only puts the Activator into the request path if the Revision is scaled to zero.
```
abrennan89, line 10:
```suggestion
If there is more than one Activator in the system, Knative puts as many Activators on the request path as required to handle the current request load plus the target burst capacity. If the target burst capacity is 0, Knative only puts the Activator into the request path if the Revision is scaled to zero.
```

docs/serving/load-balancing/activator-capacity.md
```diff
@@ -0,0 +1,43 @@
...
+
+If available, Knative will pick at least two Activators for high availability reasons. The actual number of Activators is calculated via a given _Activator capacity_ using the formula: `(replicas * target + targetBurstCapacity)/activatorCapacity`. That means, there are enough Activators in the routing path to handle the theoretical capacity of the existing application, including any additional target burst capacity.
```
abrennan89, line 12:
```suggestion
Knative uses at least two Activators to enable high availability if possible. The actual number of Activators is calculated as the _Activator capacity_, by using the formula `(replicas * target + targetBurstCapacity)/activatorCapacity`. This means that there are enough Activators in the routing path to handle the theoretical capacity of the existing application, including any additional target burst capacity.
```

docs/serving/load-balancing/README.md
```diff
@@ -24,12 +24,13 @@ The Activator load balancing algorithm works as follows:
...
 
 Target burst capacity is mainly responsible for determining whether the Activator is in the request path outside of scale-from-zero scenarios.
 
-Target burst capacity can be configured using a combination of the following parameters:
+Target burst capacity and its effect of the system can be configured using a combination of the following parameters:
 
 - Setting the targeted concurrency limits for the revision. See [concurrency](../../serving/autoscaling/concurrency).
 - Setting the target utilization parameters. See [target utilization](../../serving/autoscaling/concurrency#target-utilization).
 - Setting the target burst capacity. You can configure target burst capacity using the `autoscaling.knative.dev/targetBurstCapacity` annotation key in the `config-autoscaler` ConfigMap. See [Setting the target burst capacity](./target-burst-capacity#setting-the-target-burst-capacity).
+- Setting the Activator capacity via the `config-autoscaler` ConfigMap. See [Setting the Activator capacity](./activator-capacity#setting-the-activator-capacity)
```
abrennan89, line 36:
```suggestion
- Setting the Activator capacity by using the `config-autoscaler` ConfigMap. See [Setting the Activator capacity](./activator-capacity#setting-the-activator-capacity)
```
@markusthoemmes not for future, please use "by using" etc instead of "via" 😉

docs/serving/load-balancing/activator-capacity.md
markusthoemmes: This is not quite accurate. The actual number is not the Activator capacity, but the Activator capacity is a factor in the formula.

docs/serving/load-balancing/README.md
```diff
@@ -24,12 +24,13 @@ The Activator load balancing algorithm works as follows:
 
 For more information, see the documentation on [concurrency](../../serving/autoscaling/concurrency).
 
-## Configuring target burst capacity
+## Configuring target burst capacity behavior
 
 Target burst capacity is mainly responsible for determining whether the Activator is in the request path outside of scale-from-zero scenarios.
 
-Target burst capacity can be configured using a combination of the following parameters:
+Target burst capacity and its effect of the system can be configured using a combination of the following parameters:
```
abrennan89, line 31: I don't believe that it's necessary to change this line. I don't think "and its effect of the system" adds anything here, it just makes me wonder "what effects is this referring to?" and we don't really explain how it works here. I'd put it back how it was unless there's some reason it has to change?

abrennan89 (approved):
Some comments and suggestions, looks good so far though

abrennan89:
I actually approved this by mistake 😂 so I put a hold on it til we're ready to merge

docs/serving/load-balancing/activator-capacity.md
abrennan89: Updated it slightly

markusthoemmes: I've already changed the wording a bit, please check if that's fine too.

abrennan89 (approved):
/lgtm
/unhold

PR #3823, opened 2021-06-18
Clarify Magic DNS Caveat

julz:
Magic DNS does work locally on minikube, so long as `minikube tunnel` is
running. This caused confusion for a user in slack.

### DIFF
--- a/docs/admin/install/knative-with-operators.md
+++ b/docs/admin/install/knative-with-operators.md
@@ -328,9 +328,14 @@ kubectl logs -f deploy/knative-operator
     kubectl apply --filename {{artifact(repo="serving",file="serving-default-domain.yaml")}}
     ```
 
-    **Caveat**: This will only work if the cluster LoadBalancer service exposes an
-    IPv4 address or hostname, so it will not work with IPv6 clusters or local setups
-    like Minikube. For these, see "Real DNS" or "Temporary DNS".
+    !!! info "Caveat"
+        This will only work if the cluster `LoadBalancer` service exposes an
+        IPv4 address or hostname, so it will not work with IPv6 clusters or local setups
+        like Minikube unless [`minikube tunnel`](https://minikube.sigs.k8s.io/docs/commands/tunnel/)
+        is running.
+
+        For these, see the "Real DNS" or "Temporary DNS" tabs.
+
 
 === "Real DNS"
     To configure DNS for Knative, take the External IP
--- a/docs/admin/install/serving/install-serving-with-yaml.md
+++ b/docs/admin/install/serving/install-serving-with-yaml.md
@@ -194,10 +194,13 @@ Follow the procedure for the DNS of your choice:
     kubectl apply -f {{ artifact(repo="serving",file="serving-default-domain.yaml")}}
     ```
 
-    !!! info "CAVEAT"
-        This will only work if the cluster LoadBalancer service exposes an
+    !!! info "Caveat"
+        This will only work if the cluster `LoadBalancer` service exposes an
         IPv4 address or hostname, so it will not work with IPv6 clusters or local setups
-        like Minikube. For these, see "Real DNS" or "Temporary DNS".
+        like Minikube unless [`minikube tunnel`](https://minikube.sigs.k8s.io/docs/commands/tunnel/)
+        is running.
+
+        For these, see the "Real DNS" or "Temporary DNS" tabs.
 
 === "Real DNS"
 


docs/admin/install/knative-with-operators.md
```diff
@@ -328,9 +328,14 @@ kubectl logs -f deploy/knative-operator
     kubectl apply --filename {{artifact(repo="serving",file="serving-default-domain.yaml")}}
     ```
 
-    **Caveat**: This will only work if the cluster LoadBalancer service exposes an
-    IPv4 address or hostname, so it will not work with IPv6 clusters or local setups
-    like Minikube. For these, see "Real DNS" or "Temporary DNS".
+    !!! info "Caveat"
```
abrennan89, line 331: I'd make this a "warning" or something instead since it can potentially cause issues for users, and put it before "We ship..."

abrennan89: Would you mind also updating that "we ship" paragraph to something like:

"Knative provides a Kubernetes Job called `default-domain` that configures Knative Serving to use..."

No need for (see caveats) IMO and it's in our style guide now to avoid parentheses afaik.

julz: this PR is just correcting the existing caveat, so maybe we could do that in a separate PR?

docs/admin/install/knative-with-operators.md
```diff
@@ -328,9 +328,14 @@ kubectl logs -f deploy/knative-operator
...
+        This will only work if the cluster `LoadBalancer` service exposes an
```
abrennan89, line 332: Is `LoadBalancer` the literal name of the service? Otherwise I'd write "load balancer". Also, service -> Service as per https://dev-knative.netlify.app/development/help/contributor/style-guide/word-and-phrase-list/ and Minikube -> minikube

docs/admin/install/knative-with-operators.md
```diff
@@ -328,9 +328,14 @@ kubectl logs -f deploy/knative-operator
...
+        IPv4 address or hostname, so it will not work with IPv6 clusters or local setups
+        like Minikube unless [`minikube tunnel`](https://minikube.sigs.k8s.io/docs/commands/tunnel/)
+        is running.
+
+        For these, see the "Real DNS" or "Temporary DNS" tabs.
```
abrennan89, line 337: What are "these"? Can you phrase this better?

docs/admin/install/serving/install-serving-with-yaml.md
```diff
@@ -194,10 +194,13 @@ Follow the procedure for the DNS of your choice:
     kubectl apply -f {{ artifact(repo="serving",file="serving-default-domain.yaml")}}
     ```
 
-    !!! info "CAVEAT"
-        This will only work if the cluster LoadBalancer service exposes an
+    !!! info "Caveat"
```
abrennan89, line 197: again, can we make a rule maybe to use warnings instead of "caveats", this seems like overcomplicating it

same comments as above per work list etc for following paragraph

docs/admin/install/knative-with-operators.md
julz: I didn't change this, it's the existing text.

docs/admin/install/serving/install-serving-with-yaml.md
abrennan89: Actually, could this be a snippet instead? That way the info doesn't need to be updated in two places now or in the future?

julz: We can, but this is the existing text, and Id like to avoid blowing up a small PR to correct some misleading info in to a generic improvement of the document (though Im happy to PR improvements in another issue)

abrennan89 (changes requested):
some comments

docs/admin/install/serving/install-serving-with-yaml.md
abrennan89: It's not a huge change though, it seems like it would be more work to open an issue and raise another PR for this to just change one line...

docs/admin/install/knative-with-operators.md
julz: yes, it's literally a LoadBalancer service type (https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer), but again I didn't change this

docs/admin/install/knative-with-operators.md
abrennan89: OK, opening an issue

abrennan89:
Created an issue to follow up on the style and formatting https://github.com/knative/docs/issues/3824
FWIW I think engineering should also be doing their part to clean up docs and not just throwing it over to writers

abrennan89 (approved):
/lgtm

docs/admin/install/serving/install-serving-with-yaml.md
julz: I mean we've literally here started talking about creating a Snippet. Which is fine, but we can iterate on that in a different PR and get it right without tying it in with this correction?

julz:
> FWIW I think engineering should also be doing their part to clean up docs and not just throwing it over to writers

I dont think anyone disagrees with that sentence. But this was a _half line change_ and there are 4 requested changes to text it didn't introduce, as well as asking for it to move the para to use snippets. That feels like it warrants a separate PR to me.

PR #3825, opened 2021-06-18
#3349: Updated traffic management docs, blue/green deployment

### LINKED ISSUE #3349: Update blue/green deployment docs
This issue is to update the docs about using the blue/green traffic routing pattern.
Existing docs are here: https://knative.dev/development/serving/samples/blue-green-deployment/

Tasks required:

- [x] Move these docs from the obscurity of serving/samples to their own section in the docs under dev guide.
- [ ] Add alternative kn CLI commands as well as cleaning up kubectl/YAML docs.

May require some additional info from eng re commands to add.

abrennan89:
Fixes  #3349

## Proposed Changes
- Moved docs from samples to developer guide
- Clean up / editing content
- Removed superfluous sample files, included examples directly in docs instead to reduce confusion


docs/developer/serving/traffic-management.md
```diff
@@ -0,0 +1,200 @@
+# Traffic management
+
+You can manage traffic routing to different Revisions of a Knative Service by modifying the `traffic` spec of the Service resource.
+
+When you create a Knative Service, it does not have any default `traffic` spec settings. By setting the `traffic` spec, you can split traffic over any number of fixed Revisions, or send traffic to the latest Revision by setting `latestRevision: true` in the spec for a Service.
+
+## Traffic routing examples
```
abrennan89, line 7: Add kn example:

> 
> Here's a CLI command to split traffic:
> 
> kn service update example --traffic green=80 --traffic blue=20
> 
> To assign 80% of traffic to green and 20% to blue revisions of the "example" Service. Note that the actual Revision names will probably be more like "example-green" and "example-blue" -- you'll probably want to use the user-selected revision name option in the YAML/the --revision-name flag to kn service create and update commands.

docs/developer/serving/traffic-management.md
```diff
@@ -0,0 +1,220 @@
...
+    latestRevision: true
+    percent: 0
+```
+
+## Routing and managing traffic by using the Knative CLI
+
+You can use the following `kn` CLI command to split traffic between revisions:
+
+```bash
+kn service update <service-name> --traffic <revision-name>=<percent>
```
rhuss, line 75: We should add an example here also how to set a tag for a revision name, and the usage of `@latest` to configure the entry for the `latestRevision: true` entry.

For example

```
# Add a tag green and blue  (blue is pointing to the `latest: true` entry)
kn service update example --tag green=revision-0001 --tag blue=@latest

# Set a traffic split with 80% to latest and 20% to "v1"
kn service update example --traffic @latest=80 --traffic v1=20
```

docs/developer/serving/traffic-management.md
```diff
@@ -0,0 +1,220 @@
...
+
+You can safely reroute traffic from a live version of an application to a new version by using a [blue/green deployment strategy](https://en.wikipedia.org/wiki/Blue-green_deployment).
+
+### Procedure
+
+1. Create and deploy an app as a Knative Service.
+1. Find the name of the first Revision that was created when you deployed the Service, by running the command:
+
+    ```bash
+    kubectl get configurations <service-name> -o=jsonpath='{.status.latestCreatedRevisionName}'
```
rhuss, line 100: @dsimansk @navidshaikh wdyt, should we add support for `kn` for getting the "lastCreatedRevisionName" so that you don't have to memorize this cryptic kubectl call ?

docs/developer/serving/traffic-management.md
```diff
@@ -0,0 +1,220 @@
...
+
+```bash
+kn service update example --traffic green=80 --traffic blue=20
+```
+
+## Routing and managing traffic with blue/green deployment
+
+You can safely reroute traffic from a live version of an application to a new version by using a [blue/green deployment strategy](https://en.wikipedia.org/wiki/Blue-green_deployment).
+
+### Procedure
```
rhuss, line 94: @abrennan89 I'm fine with the documentation, that is copied over, but for the long term we should definitely describe it in `kn` terms. The example does not have a smooth UX (cryptic kubectl calls, editing yaml files) and kn should definitely cover this use case (and if not, we need to fix kn)

rhuss: I created https://github.com/knative/docs/issues/3951 to track this additional update.

rhuss (commented):
Thanks, looks good to me ! I would add two additional `kn` samples for how to set a tag and how to use @latest. But please note, that we are currently still rethinking the UI surface of the command (like adding additional @previous or @current to specify the `latestCreatedRevision` that is often needed to cover various use cases), so maybe we should rework the `kn` section when we have nailed down the new options, to avoid extra work ?

docs/developer/serving/traffic-management.md
abrennan89: +1, for now I'm just moving it for now and removing unnecessary files etc, but it would be nice to clean this up properly when the UX is figured out.

docs/developer/serving/traffic-management.md
abrennan89: Out of scope for this PR, please continue that discussion on your new issue 😉

docs/developer/serving/traffic-management.md
rhuss: Yes, I understand this and it looks much better now !

Issue #3837, opened 2021-06-23
Fix CSS for tables


abrennan89:
## Expected Behavior

Tables should look nice and be easy to read.

## Actual Behavior

Tables have a lot of ugly bits, like not being consistent widths, lines breaking in the middle of words, etc.

## Steps to Reproduce the Problem

See Slack thread https://knative.slack.com/archives/C9CV04DNJ/p1624464518093800

abrennan89:
Markus fixed the line break issue: https://github.com/knative/docs/pull/3838

PR #3838, opened 2021-06-23
Fix wrapping behavior of code in tables

markusthoemmes:
Ref #3837 

This is supposed to fix code sections breaking in the middle of the word rather than enlarging the table.


abrennan89 (approved):
/lgtm

PR #3842, opened 2021-06-24
Add some marketing pages

evankanderson:
## Proposed Changes

- Clean up the front page a little bit.
- Add new "what's the value" (marketing) pages for Administrators and Developers.
- Remove old `_index.html` that isn't used anymore.


evankanderson:
/hold

This is not what I intended.

evankanderson (commented):
This needs more cleanup for the new content; I'll get to it tonight/tomorrow

evankanderson (commented):
/hold cancel

A couple items where I'm looking for feedback:

1. I'm redirecting the "getting started" call-to-actions to additional marketing pages (and then on into the existing content). Since there is the top-nav, I don't feel too bad about this, but wanted to double-check.

2. I didn't find any icons for the new marketing pages like the nifty rocketship we have on the first page. I'd like to do that at some point, but didn't want to block the content on the icon.

3. Should I add the "getting started" call to action to the top banner on the "for developers" and "for administrators" pages?

omerbensaadon:
/hold 

Thank you for your work on this, I think you're right in your sentiments here, which is that we need more "MKTing" content, feedback below. 

It looks like you've made a change to the homepage which makes the buttons bigger (I like this!) can we spin this out into a **separate PR**?

I'm not sure these pages ought to be more-of-the-same (in terms of format) marketing content that's on the homepage, we have gotten feedback that the homepage is "too much text" and I think linking people to pages with more text will lead to greater drop-off rates. 

Perhaps **submitting a PR to the homepage** to include more of this sort of content would be good (i.e. changing the highlights section?). Since we are working with external contractors to redo the homepage in a couple of weeks, I'll be sure to ask then what they think as well.

I think there may be room for this sort of stuff in the Dev + Admin guides, what do you think about **submitting a PR** to the READMEs of the Dev + Admin Guides with some of this content?

julz:
My 2 cents on this: I _love_ the new content and want to get it on the site, I'm less sure getting it on the site is worth an extra click between home page and actual content. I'd like to see what it would look like if we incorporated the new text in to the existing home page (especially with a designer's help); obviously this would lose the complete stratification of admin vs dev in to different landing pages in favour of putting both sections on to the home page and making people scroll a bit, but I think that trade-off might be ok for a more streamlined experience, personally.

evankanderson:
I think this may end up being background material for future pages of this type that @omerbensaadon is developing.

omerbensaadon:
@evankanderson the team at Lionsway is helping us with this sort of content, I'll link them to this PR and ask if there is anything they would find useful, but it's safe to say we can close this PR. 

Please reopen if you feel differently. 

@abrennan89 thanks for the nudge! Sorry for the delay

omerbensaadon:
PS: if you want to spin out updates to homepage to a separate PR, would be happy to review those

PR #3845, opened 2021-06-24
docs updates

dprotaso:
- include a section about disabling autotls for a namespace
- handle rename of namespace certificate controller
- handle rename of other net-* controllers


ZhiminXiang (commented):
/lgtm

PR #3851, opened 2021-06-25
Adjust breakpoint for new max-width

markusthoemmes:
As per title, this fixes the breakpoints (and lacking paddings) on the main page. Also moves the rocket image breaking to even smaller screens as there's no point breaking that this early.

/assign @julz


julz (approved):
yay, thanks @markusthoemmes

/lgtm

PR #3854, opened 2021-06-25
install isstio.yaml in two passes

dprotaso:
istio.yaml has CRDs and instances of those resources - thus for a successful install we should apply everything in two passes


julz (approved):
/lgtm

Issue #3715, opened 2021-06-03
shortcode {{< branch >}} not working in mkdocs development version


csantanapr:
## Expected Behavior

To be able to run the samples in development version

## Actual Behavior

Can't run samples in development version

## Steps to Reproduce the Problem

1. Go to any sample serving or eventing page that state to clone the docs repo like the Hello Worlds https://dev-knative.netlify.app/development/serving/samples/hello-world/helloworld-go/
1. Copy the git clone line 
```
git clone -b "v0.23.0" https://github.com/knative/docs knative-docs
cd knative-docs/docs/serving/samples/hello-world/helloworld-go
```
1. Run the code and you get the following error
```
Cloning into 'knative-docs'...
fatal: Remote branch v0.23.0 not found in upstream origin
cd: no such file or directory: knative-docs/docs/serving/samples/hello-world/helloworld-go
```
## Additional Info
There is no branch with the name of `v0.23.0` only `release-v0.23` in docs repo.
In development we want to use the branch `mkdocs` for now and later time switch to `main`

The mkdocs build process should take the short code `{{< branch >}}` and replace it with `mkdocs`
```

   git clone -b "{{< branch >}}" https://github.com/knative/docs knative-docs
   cd knative-docs/docs/serving/samples/hello-world/helloworld-go

```

/kind bug
/kind mkdocs
/assign @julz

PR #3860, opened 2021-06-28
Use correct branch for samples

julz:
Sets the branch for code samples (i.e. the `{{ branch }}` variable) to either `mkdocs` (when building the dev docs), or to the version of docs being built (e.g. release-0.23 etc). This makes sure we create the correct links for the `git clones` in the samples.

Fixes https://github.com/knative/docs/issues/3715.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -22,7 +22,7 @@ set -x
 VERSIONS=("0.23" "0.22" "0.21")
 # 3) For now, set branches too. (This will go away when all branches are release-NN).
 DOCS_BRANCHES=("mkrelease-0.23" "mkrelease-0.22" "mkrelease-0.22")
-RELEASE_BRANCHES=("v0.23.0" "v0.22.0" "v0.21.0")
+RELEASE_BRANCHES=("release-0.23" "release-0.22" "release-0.21")
 REPOS=("julz" "julz" "julz")
 # 4) PR the result to main.
 # 5) Party.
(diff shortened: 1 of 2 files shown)


omerbensaadon:
Weirdly I am still getting the same error here? 

https://deploy-preview-3860--dev-knative.netlify.app/docs/eventing/samples/helloworld/helloworld-go/

Can confirm that pre-release has turned into "mkdocs" and we should probably create an issue for "before go live" to switch it to main...

julz:
that's because I didn't bother rebuilding the v0.23 docs, this will work fine when we build 0.24

omerbensaadon:
How do I test this then? Just ensure than `main` is now `mkdocs`?

julz:
if you want to do a quick test, a nice'n'easy way would be to change `DOCS_BRANCHES` at the top of build.sh to `"branchbranch" "mkrelease-0.22" "mkrelease-0.22")` and then run `./hack/build.sh serve` (that will make the "0.23" docs build from this PRs branch instead of the mkrelease-0.23 release I made a while ago).

omerbensaadon:
I am getting the error: 
```
sed: 1: "./content/en/blog/relea ...": invalid command code .
```

Not sure if this is just my machine though. Let me see if there's anyone else who would be up to review this...

julz:
Think this is a requirement of the Hugo stuff, see https://knative.dev/help/contributor/publishing/#sed (or you can probably set the `SKIP_BLOG` env to ` to skip that stuff)

csantanapr:
nice fix @julz

PR #3873, opened 2021-06-29
Fix broken links: create-default-channel in page of channel-types-defaults.md return 404

chengjingtao:
link of "Create an InMemoryChannel" in https://knative.dev/docs/eventing/channels/channel-types-defaults/#next-steps return 404

Signed-off-by: jtcheng <jtcheng@alauda.io>


omerbensaadon:
Thank you for the PR!

PR #3886, opened 2021-06-30
Use correct version in artefact URLs

### LINKED ISSUE #3885: [mkdoc] Yaml URL misses prefix "v" and micro version in v0.23 page
## Expected Behavior

The URL for yaml in installation page is not correct.

For example, https://dev-knative.netlify.app/docs/install/install-serving-with-yaml/#install-the-serving-component

### CURRENT
```
kubectl apply -f https://github.com/knative/serving/releases/download/0.23/serving-crds.yaml
```

### TOBE
```
kubectl apply -f https://github.com/knative/serving/releases/download/v0.23.0/serving-crds.yaml
```

## Actual Behavior

It seems the URL was defined by `artifact`:
https://github.com/knative/docs/blob/0435903a210b752879b12b3da5af2537895e6429/hack/macros.py#L20

and `{version}` causes the problem. So, most probably 0.24 page will experience the same issue.

1. Go to https://dev-knative.netlify.app/docs/install/install-serving-with-yaml/#install-the-serving-component
1. Try `kubectl apply -f https://github.com/knative/serving/releases/download/0.23/serving-crds.yaml` in the page.
1. You will get 404 error.

julz:
Fixes #3885.

Totally goofed on which variable to use for the artefact URLs -- was using the version of _docs_ instead of the serving/eventing release version. Fixed, and added some comments. Thanks @nak3 for spotting this!

/assign @nak3 @omerbensaadon

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -19,10 +19,10 @@ set -x
 # 1) Make a release-NN branch as normal.
 # 2) Update VERSIONS below (on main) to include the new version.
 #    Order matters :-), Most recent first.
-VERSIONS=("0.23" "0.22" "0.21")
-# 3) For now, set branches too. (This will go away when all branches are release-NN).
+VERSIONS=("0.23" "0.22" "0.21")                  # Docs version, results in the url e.g. knative.dev/docs-0.23/..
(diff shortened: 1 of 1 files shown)


nak3:
LGTM! Thank you for the quick fix.

Issue #3864, opened 2021-06-28
Unpicking existing version re-directs


omerbensaadon:
**Describe the change you'd like to see**
Currently, our site links old versions to knative.dev since our beta site is hosted on Netlify, we need to make sure we unpick these redirects before we "go live" because knative.dev will be where the new mkdocs site lives once it's off of the beta domain. 

**Additional context**

Issue #3801, opened 2021-06-15
🛑 Render Old Versions of Knative.dev as static HTML 


omerbensaadon:
## Description: 
Since we don't want to have to move everything over to `mkdocs`, we're forced to live with a "Frankenstein's Monster" for old versions of the website (newest version = `mkdocs`, all previous versions = `Hugo` ) 

---

## Story
```
As Pyotr Alekseevich the Production User 
I want to see old versions of Knative documentation on the website 
So that I am assured the older version of Knative which I have in production has documentation
```

## Acceptance Criteria
- [x] `Visit old versions`
```
GIVEN I have loaded the current version on the Knative site (rendered in mkdocs) 
WHEN I select an older version of the documentation from the version drop-down
THEN I am taken to a rendering of the old Hugo site 
```

---

- [x] `Come back to current version`
```
GIVEN I am accessing an older version of the Knative site (rendered in Hugo) 
WHEN I click on the current version from the version drop-down
THEN I am taken back to the current version of the Knative site (rendered in mkdocs`) 
```

## Who should probably pick up this issue? 
- [x] Existing Contributors 
- [ ] New Contributors

omerbensaadon:
@omerbensaadon make sure to add old version similar to K8s in a new AC

csantanapr:
build the last 0.24 and development will be in mkdocs, 0.23, 0.22, 0.19, with hugo

PR #3888, opened 2021-06-30
Incorporate static version of old docs in to site

julz:
Built via manual ./scripts/build.sh of website and then copied over (added a version warning banner in the process, happy to iterate the css for that).

Each release we get to delete one of these docs directories (and the corresponding copy in build.sh) until they all disappear.

Also unpicked the netlify proxy redirects since we now have static html built instead.

Fixes https://github.com/knative/docs/issues/3864
Fixes https://github.com/knative/docs/issues/3801


julz:
oh annoying, the static files have some approvers files in them, lemme delete

/hold

omerbensaadon:
Am I correct in thinking that all of the static sites for the old versions will be thrown into a folder called `old docs`? 

I think everything looks good, but can we change the name of that folder to `previous versions` or something like that?

omerbensaadon:
Assuming the "latest version" link points to knative.dev on purpose and at launch it will simply link to the homepage

omerbensaadon:
tried to break it, failed! 

Will lgtm after folder name change :D

julz:
Yeah I put them in old-docs, I can make it previous-docs (or archived-docs?) instead easily enough yeah

omerbensaadon:
let's go with archived-docs, because that's really what they are

julz:
(Yeah I was lazy and didn't want to figure out how to make hugo output the right url, so I hard-coded it, should probably make it /docs/ instead actually though)

omerbensaadon:
@julz oh no there is a linting error

omerbensaadon:
Looks like you are missing a new line at the end of each file

julz:
Ugh yeah, that's a pain, those are generated files 🤔 we probably need to tell the lint robot to ignore these directories

omerbensaadon:
What Slack channel should I post this in? #productivity?

julz:
I can look at PRing it later/tomorrow, I think we just need to add archived/ to the list of ignored dirs here: https://github.com/knative-sandbox/.github/blob/main/workflow-templates/knative-style.yaml#L203

julz:
Ooh actually, looks like we might be able to just use .gitattributes to mark the dir as generated, lemme try that

julz:
.. looks like that did it \o/

omerbensaadon:
/lgtm 

🥳

PR #3913, opened 2021-07-02
Add feature stable label macro and document that DomainMapping is beta

### LINKED ISSUE #3874: Add back Feature State labels
We [previously removed](https://github.com/knative/docs/issues/2781) many of the feature state labels because they were confusing (arguably mostly due to the previous, over-concise format of the labels). Now that we plan to lower the amount of cherry-picking we do to old releases, though, we need to ensure the main docs work well for the last 4 versions, which means being clear about when certain features were introduced. This will also be useful for features like DomainMappings and MultiContainer which are beta but do not currently indicate this in a consistent way

- [ ] #3875
- [ ] Document alpha/beta features in serving with feature state label (multi-container, domain mapping, also check install docs and feature flag docs)
- [ ] Document alpha/beta features in eventing with feature state label (some things in eventing already have an api version label that does not include the knative version, we will need to update these for consistency)

julz:
Fixes #3874 

Feedback, especially on colour and formatting choices, very welcome. Before merging you may want to try `feature(alpha="1.2")`,  `feature(stable="2.3")` and (in case we ever want to be this verbose)  `feature(alpha="2.3", beta="3.4")`

/assign @omerbensaadon

### DIFF
--- a/docs/developer/serving/services/custom-domains.md
+++ b/docs/developer/serving/services/custom-domains.md
@@ -1,5 +1,7 @@
 # Configuring custom domains
 
+{{ feature(beta="0.24") }}
+
 Each Knative Service is automatically assigned a default domain name when it is created. However, you can map any custom domain name that you own to a Knative Service, by using _domain mapping_.
 
 You can create a `DomainMapping` object to map a single, non-wildcard domain to a specific Knative Service.
--- a/docs/stylesheets/extra.css
+++ b/docs/stylesheets/extra.css
@@ -216,3 +216,18 @@
 .footer-links a {
   margin: 0.4em;
 }
+
+/* Feature State Labels */
+.feature-alpha, .feature-beta, .feature-stable {
+  font-variant: small-caps;
+  padding: 0.2em;
+}
+.feature-alpha {
+  background: rgb(255, 218, 218);
+}
+.feature-beta {
+  background: rgb(251, 244, 182);
+}
+.feature-stable {
+  background: rgb(228, 255, 218);
+}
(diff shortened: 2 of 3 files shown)


omerbensaadon:
This is a great addition, thank you Julz! 

Would you be against changing the padding "0.1em" as opposed to "0.2em" ? I rendered it on my machine and I think it looks a bit cleaner. I've also removed "Feature Availability" and same idea there. 

I can commit my changes if it's cool with you?

PS: I couldn't figure out how, but what's your take on having this admonition open up to show the detail you've written in <abbr> as a collapsable admonition?

julz:
Changed the padding and put the hover text in to a twiddly-thingy

Removing "Feature Availability" does make the admonition more concise, but I worry a bit about falling in to the same trap that caused us to remove the labels previously: we don't want it to be unclear what these are to people who don't already know it's an availability label, and being over-concise seems like it might risk that?

omerbensaadon:
I'm just not sure the language "Feature Availability" is communicating anything that "Beta since v0.24" isn't? 

Plus the collapsable admonition should do some work for us in terms of explaining what these terms mean

julz:
without the feature availability text or context of what this is, something like "Stable since 1.4" seems a bit unclear, doesn't it? I'd say feature availability is communicating (1) the text is referring to the feature in question, not to Knative in general, or to the docs page about the feature the label is at the top of (2) that we're referring to availability of the feature, that is what version it became available, which seems important since otherwise it's not clear at all - without context - that the feature may not be present in previous versions just from the fact we say "alpha @ whatever".

Certainly not the most important thing ever so if you still feel strongly Im not going to bike-shed and will just drop them, but since the text will take up a line either way, it's not clear to me that dropping the words makes much difference, and they seem at least a little helpful?

csantanapr:
I'm favor of having the context "Feature ..." what about "Feature State" same as kubernetes?
image

To be frank I like better "Feature Availability"

PR #3932, opened 2021-07-02
Remove linkchecks

benmoss:
Don't have a good way to exclude the third_party dir right now

@julz


julz (approved):
Until we can figure out how to bring these back

/lgtm

PR #3933, opened 2021-07-02
Fix redirects

julz:
the new inheritance feature only works with maps, not lists, so the plugins element
needs to use map format for this to work correctly.

/assign @benmoss @omerbensaadon


omerbensaadon:
/lgtm 

Good catch! Thanks @julz  :)

PR #3934, opened 2021-07-02
Fix broken link

julz:
Discovered via @benmoss's link checking and detective work that this link is broken _even though_ the mkdocs --strict link check doesn't complain. This is because the mkdocs link checker seems to ignore links to directories, annoyingly :(. 

We should probably update the style guide to suggest using links to .md files so we get nice link checking.

/assign @omerbensaadon


snneji:
/lgtm

Added an issue to update the style guide: https://github.com/knative/docs/issues/3940

PR #3937, opened 2021-07-05
Add footer to homepage

julz:
This has two advantages:
1. We get the privacy policy (etc) linked on the homepage, which seems Right (also more consistent).
2. We can drop the duplicated cookie popup code in home.html, since it's included via the footer on all non-homepage pages.

/assign @csantanapr @omerbensaadon

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,4 +1,5 @@
 ---
 template: home.html
 title: Knative
+hide_next: true
 ---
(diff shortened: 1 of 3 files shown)


omerbensaadon:
Screen_Shot_2021-07-05_at_11_32_18_AM
What's up with this weird white line??

julz:
updated btw

csantanapr:
line is gone 
/lgtm

PR #3947, opened 2021-07-06
Linking the title, and rotate anim on main Logo

### LINKED ISSUE #3945: [mkdocs] Knative logo area behaves odly
Hovering over the Knative logo shouldn't cause it to fade out. Fading out logo is sending the signal that clicking it isn't available. That may cause confusion among the users. If we want to keep the effect, it should be subtle, not drastic.

Also, the Knative text next to the logo isn't clickable. That's a standard thing to make it clickable, so users use it to return to the homepage.

https://user-images.githubusercontent.com/540893/124590584-964dbf80-de5b-11eb-83ca-7b0fe1b47234.mp4

## Expected Behavior
Knative logo hover effect is removed or tuned down to be subtle. Knative text next to the logo is linked the same way as logo.

## Actual Behavior
Knative logo is fading out drastically, and Knative text next to it isn't clickable.

1. Visit upcoming site at https://dev-knative.netlify.app/development/ or any subpage
1. Hover the mouse on Knative logo in the header
1. Try clicking Knative next to the knative logo

## Additional Info
/kind mkdocs

cardil:
Fixes #3945 

## Proposed Changes 

- Linking the Knative title in header
- Changing the hover effect on Knative logo to slight rotate animation

Both of those are inherited from https://github.com/squidfunk/mkdocs-material. I think it's worth opening an issue/PR there as well - especially the linking the title.


julz:
looks like there's an issue discussing this in upstream MkDocs, and the reasoning behind not linking the title is that it can break the behaviour of clicking the header returning to the top on mobile: https://github.com/squidfunk/mkdocs-material/issues/2044.

I like the rotating icon vs fading, but I'm a bit torn on whether making the title clickable is worth (a) potentially breaking mobile behaviour (b) maintaining such a big override. Will defer to @omerbensaadon and you there though.

cardil:
I think it works well on mobile. I did try it. Of course, you need to scroll up, as title gets hidden by page topic, but that's quite okay in my mind.

Huge theme override for this change is painful, I agree. That's why I've opened https://github.com/squidfunk/mkdocs-material/issues/2815 (as squidfunk/mkdocs-material#2044 is closed without clear reason)

omerbensaadon:
@julz we do have that nice "back to top" button that appears in the lower right-hand side on mobile

omerbensaadon:
@cardil this has the effect of making the hamburger icon rotate slightly on mobile, is there anyway to fix that before merge?

omerbensaadon:
my take is: I don't want to be on the hook to maintain such an override (I also am imagining a can of worms opening somewhere, out of sight), do you reckon we should take the Title click bits out until there is some follow-up on your open issue @cardil?

cardil:
Okay. @omerbensaadon 

I've constrained the "rotate effect" to logo only. Now, it doesn't affect hamburger menu. I've also removed the link from the header. Let's see what maintainers of material-mkdocs say finally on https://github.com/squidfunk/mkdocs-material/issues/2815

If they will once again reject it, I think I'll another proposal there, so create smaller override region, then we will not need to maintain such long HTML just for this title link.

omerbensaadon:
Awesome!! thanks for your work on this @cardil 

/lgtm 
/approve

PR #3780, opened 2021-06-10
(Fixes #3449) Improve style in Installing Knative Serving using YAML files

RichardJJG:
Heavy style edits, as the issue requests.

PR #3950, opened 2021-07-06
Fix PR issues for #3780

abrennan89:
Follow up comments from #3780


docs/admin/install/serving/install-serving-with-yaml.md
```diff
@@ -219,106 +178,7 @@ Follow the procedure for the networking layer of your choice:
...
-        curl -H "Host: helloworld-go.default.example.com" http://192.168.39.228:32198
-        ```
-
-        !!! success "Verify the output"
-            In the case of the provided `helloworld-go` sample application, when using the default configuration the output is:
-
-            ```
-            Hello Go Sample v1!
-            ```
+{% include "dns.md" %}
```
omerbensaadon, line 181:
```suggestion
<!-- This is a snippet, to find out more about Snippets, and to see existing Snippets, see: /docs/snippets-->
--8<--  "dns.md"
```
Aligning around existing Snippets plugin which is referenced in snippets/README.md

omerbensaadon (changes requested):
Thank you for taking care of this quick-fix 💯 

Just left one suggestion for using the "snippets" formatting for includes

docs/admin/install/serving/install-serving-with-yaml.md
cardil: There's a problem with including that DNS snippet, here.

It is **only** valid if you install Knative with Operator. It advises changing KnativeServing resource (in _Real DNS_ tab). When doing installation via YAML, like this page does, this isn't valid as then there's no KnativeServing resource.

In case of installing via YAML, we need to update config map directly, like this:

```bash
kubectl patch configmap/config-domain \
  --namespace knative-serving \
  --type merge \
  --patch '{"data":{"<domain-suffix>":""}}'
```

julz: oh what a pain! I guess I goofed on this when I pulled this out into a snippet :(. Will have to split it in to smaller snippets I guess so operator can have its own Real DNS tab, tho I suggest we worry about that in another PR.

@omerbensaadon unfortunately this snippet would need to be `{%` format because it uses the artefact macro inside it so has to be included via the macro include syntax. Possibly we should actually standardise on that instead of `--8<--`.

cardil: @julz Alright, I can agree to do it in the next PR, but take care of it, please.

I'm commenting on this exact problem for the 2nd time already (Real DNS tab).

julz: I mean let's worry about moving this back to snippets in a follow-on PR (so that that can be my problem instead of @abrennan89's since it was me who messed up the snippet). IMO we should fix the text to be correct here (and I'll make sure not to break it next time!)

omerbensaadon (approved):

docs/admin/install/serving/install-serving-with-yaml.md
omerbensaadon: @julz this confuses me but I'll throw it on the agenda, would be ideal if we standardized on an extension with detailed documentation...

omerbensaadon: @julz can you open this issue? I would but I don't think I am fully grasping the context here

cardil: I should have blocked this PR, on @julz comment. Now, it's merged and the issue persist.

julz: Oops yeah, I guess I'll fix it all in one go. Will try to PR today

omerbensaadon: @cardil I figured this is something that could be fixed in a follow-on PR... was that a mistake?

PR #3955, opened 2021-07-07
Update version dropdown in old docs

julz:
Updates the drop-down on the archived docs to only include latest and dev, so the dropdown is evergreen and
will not need regenerating for future releases.

/assign @omerbensaadon


omerbensaadon:
It would be nice if it said "Pre-Release" as opposed to "Development" but I will accept this as design debt for the next 2 releases. 

/lgtm

PR #3967, opened 2021-07-07
Refactor DNS snippets

julz:
Splits in to multiple files so that Operator can have a different version for the Real DNS section.

/assign @cardil - please could you check I got this right this time 😅


cardil (approved):
Now it looks well. Thanks a lot!

/lgtm

Issue #3935, opened 2021-07-02
Move Serving developer topics to developer guide


abrennan89:
Move the docs in the developer topics of https://knative.dev/development/serving/ to the developer guide

abrennan89:
Last topic to do is load balancing: https://knative.dev/development/serving/load-balancing/
This is going to require splitting the docs into admin guide load balancing vs dev load balancing though, since there are some things that need to be configured using ConfigMaps

PR #3970, opened 2021-07-07
#3935: Moving tag resolution docs to developer guide

abrennan89:
Partial fix for #3935 , related to #3710 

## Proposed Changes

- Move and clean up tag resolution docs
- Fixed link and other minor fixes in Operator docs


docs/admin/install/operator/configuring-serving-cr.md
```diff
@@ -1,52 +1,42 @@
...
 
-## Version Configuration
+## Version configuration
 
-Cluster administrators can install a specific version of Knative Serving by using the `spec.version` field. For example,
-if you want to install Knative Serving 0.16.0, you can apply the following `KnativeServing` custom resource:
+Cluster administrators can install a specific version of Knative Serving by using the `spec.version` field.
 
-```
+For example, if you want to install Knative Serving 0.23.0, you can apply the following `KnativeServing` custom resource:
```
RichardJJG, line 19:
```suggestion
For example, if you want to install Knative Serving v0.23.0, you can apply the following `KnativeServing` custom resource:
```

docs/admin/install/operator/configuring-serving-cr.md
```diff
@@ -1,52 +1,42 @@
...
 spec:
-  version: 0.16.0
+  version: 0.23.0
 ```
 
-If `spec.version` is not specified, the Knative Operator will install the latest available version of Knative Serving.
-If users specify an invalid or unavailable version, the Knative Operator will do nothing. The Knative Operator always
-includes the latest 3 minor release versions. For example, if the current version of the Knative Operator is 0.16.x, the
-earliest version of Knative Serving available through the Operator is 0.14.0.
+If `spec.version` is not specified, the Knative Operator installs the latest available version of Knative Serving. If users specify an invalid or unavailable version, the Knative Operator will do nothing. The Knative Operator always includes the latest 3 minor release versions. For example, if the current version of the Knative Operator is v0.24.0, the earliest version of Knative Serving available through the Operator is v0.22.0.
```
RichardJJG, line 31:
```suggestion
If `spec.version` is not specified, the Knative Operator installs the latest available version of Knative Serving. If users specify an invalid or unavailable version, the Knative Operator does nothing. 

The Knative Operator always includes the latest 3 minor release versions. For example, if the current version of the Knative Operator is v0.24.0, the earliest version of Knative Serving available through the Operator is v0.22.0.

abrennan89: lol thought about doing this but wasn't sure whether to split it. Thanks.

RichardJJG (approved):
Thanks, @abrennan89, all looks good.

Should the tag-resolution doc have procedures instead of summaries of what the reader can do, though? Either way, I'd be okay with /lgtm at present. I suspect procedures would be better, but they could be tackled in a separate issue if this content is urgent.

abrennan89:
> Thanks, @abrennan89, all looks good.
> 
> Should the tag-resolution doc have procedures instead of summaries of what the reader can do, though? Either way, I'd be okay with /lgtm at present. I suspect procedures would be better, but they could be tackled in a separate issue if this content is urgent.

Yeah probably, I'll leave you to open an issue for this though if you think it's worth doing since it's not really part of the scope here for just moving docs.

PR #3979, opened 2021-07-08
Changes to nav, consolidate Getting Started README

omerbensaadon:
Early analytics are showing a 50% drop off after the first page. Based on user interviews, I noticed developers always scan for the

```
% Code Blocks %
```

I think we should consolidate the first two pages to see if this impacts drop-off rate. 

Also, changed some names in the Nav!


omerbensaadon:
/cherrypick release-0.24

Going to wait a week to get more data, but eventually would like to merge this to main site to see impact on metrics.

julz:
to be fair, I suspect a lot of the early analytics will be folks just wanting to check out the new site, so maybe not great data points (change is fine with me tho)

omerbensaadon:
@julz it's the data combined with a gut feeling, going to wait a bit to check back on the data to push this to the live site

config/nav.yml
```diff
@@ -2,9 +2,8 @@ INHERIT: redirects.yml
 
 nav:
     - Home: index.md
-    - Getting Started:
-        - Before you begin: getting-started/README.md
-        - Knative Quickstart: getting-started/install-knative-quickstart.md
```
julz, line 7: do you want to also delete this file?

omerbensaadon:
More evidence: 

Googling "Knative Quickstart" Yields "Before you Begin," ideally this would just be a page called Knative Quickstart which should improve SEO. 

https://www.google.com/search?q=knative+quickstart&rlz=1C5CHFA_enUS914US914&oq=knative+quickstart&aqs=chrome..69i57j69i60l3j69i64l3.4842j0j7&sourceid=chrome&ie=UTF-8

config/nav.yml
julz: ^ waiting on reply before lgtm-ing

julz (approved):
/lgtm

PR #3985, opened 2021-07-08
Fix blog

julz:
YAML metadata needs to be the first thing in the file. A markdown title
snuck in.

/assign @omerbensaadon


omerbensaadon:
/lgtm 

Thank you Julz. When do you sleep I am concerned (but thankful)

PR #3986, opened 2021-07-09
(Fixes #3982) Change homepage name

### LINKED ISSUE #3982: Rename homepage to `Home` instead of `Knative`
Screen Shot 2021-07-08 at 2 39 58 PM

Knative - Knative looks a bit weird

omerbensaadon:
Fixes #3982

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,5 +1,5 @@
 ---
 template: home.html
-title: Knative
+title: Home
 hide_next: true
 ---


julz (approved):
FWIW I really think we should break the habit of putting "fixes X" in the subject line -- it just makes the commit history ugly, and putting it in the description/commit message is enough to cause the issues to be linked.

/lgtm

Issue #3971, opened 2021-07-07
Feature flag docs lists are a mess


abrennan89:
The feature flag docs here: https://knative.dev/docs/serving/feature-flags/ have multiple lists that are broken and instead render as "* list item 1 * list item 2...etc"

I think this is just an issue with spacing that needs to be fixed throughout the page.

PR #3995, opened 2021-07-12
Fix format in Feature/Extension Flags

nak3:
This patch fixes the broken format in [Feature/Extension Flags](https://knative.dev/docs/serving/feature-flags/#lifecyle) section.

For example, the format is broken as:

image


nak3:
Hmm... This change does not work. https://deploy-preview-3995--knative.netlify.app/docs/serving/feature-flags/
I am not sure the format rule so will open the new issue.

nak3:
https://github.com/knative/docs/issues/3971 was already opened.

PR #4005, opened 2021-07-13
#3935: Move subroutes docs to traffic mgmt section

abrennan89:
Partial fix for #3935 

## Proposed Changes

- Remove subroutes docs, redirect and add this as a subsection for traffic management instead since the docs are actually only about the `tag` traffic spec.


config/nav.yml
```diff
@@ -113,7 +113,6 @@ nav:
       - Overview: serving/README.md
       - Developer Topics:
         - Gradually rolling out latest Revisions: serving/rolling-out-latest-revision.md
-        - Creating and using Subroutes: serving/using-subroutes.md
```
omerbensaadon, line 116: Did you mean to not have this as apart of the nav anymore?

omerbensaadon (approved):

PR #4008, opened 2021-07-14
Remove one option to make private service by DomainMapping

nak3:
https://knative.dev/docs/developer/serving/services/private-services/ says:

> Knative provides three ways to enable private services which are only available inside the cluster:
> 3. Use custom domain mappings.

But DomainMapping for private service is not expected solution.
We can map domainmapping to the local domain (`svc.cluster.local`) but
it is accessible from public.

The option should be removed.

/cc @ZhiminXiang @julz


julz (approved):
Good catch!

/lgtm

PR #4029, opened 2021-07-19
Add "Third-Party Ingress Implementations" page

nak3:
Today, Ingress implementations are listed in the docs[1] with
the installation steps.

The Ingress was added in the serving repo to run CI but that
was so hard for both Knative project and Ingress providers to maintain
as fixing/updating e2e tests are not so simple work.
So some implementations were dropped from the e2e tests as
well as from the docs.

Although I still think we don't need to maintain the e2e tests, we should
host the Knative Ingress implementations list in the docs site.

Hence, this patch adds the new "Third-Party Ingress Implementations" page.

[1] https://knative.dev/docs/admin/install/serving/install-serving-with-yaml/#install-a-networking-layer

### DIFF
--- a/docs/admin/install/serving/ingress-implementations.md
+++ b/docs/admin/install/serving/ingress-implementations.md
@@ -1,7 +1,6 @@
 # Unofficial ingress implementations
 
-In addition to the Knative Ingress implementations in the ([install a networking layer](../install-serving-with-yaml/#install-a-networking-layer)) section,
-there are other Knative Ingress implementations.
+In addition to the Knative ingress implementations in the ([install a networking layer](../install-serving-with-yaml/#install-a-networking-layer)) section, there are other Knative ingress implementations.
 
 Here is the list of the ingress implementations (alphabetically):
 
--- a/docs/admin/install/serving/ingress-implementations.md
+++ b/docs/admin/install/serving/ingress-implementations.md
@@ -1,0 +1,11 @@
+# Unofficial Ingress Implementations
+
+In addition to the Knative Ingress implementations in the ([install a networking layer](../install-serving-with-yaml/#install-a-networking-layer)) section,
+there are other Knative Ingress implementations.
+
+Here is the list of the ingress implementations (alphabetically):
+
+- [Emissary-Ingress (Ambassador API Gateway)](https://www.getambassador.io/docs/edge-stack/latest/howtos/knative/): Emissary-Ingress is an open-source Kubernetes-native API Gateway + Layer 7 load balancer + Kubernetes Ingress built on Envoy Proxy.
+- [Kong](https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/using-kong-with-knative/): Kong is an open source API Gateway built for hybrid and multi-cloud environments.
+
+These implementations are not officially supported by Knative community at this moment, but the implementation will be added to the main guide once it is ready for general use.
--- a/docs/admin/install/serving/ingress-implementations.md
+++ b/docs/admin/install/serving/ingress-implementations.md
@@ -1,4 +1,4 @@
-# Unofficial Ingress Implementations
+# Unofficial ingress implementations
 
 In addition to the Knative Ingress implementations in the ([install a networking layer](../install-serving-with-yaml/#install-a-networking-layer)) section,
 there are other Knative Ingress implementations.
--- a/docs/admin/install/serving/ingress-implementations.md
+++ b/docs/admin/install/serving/ingress-implementations.md
@@ -4,7 +4,7 @@ In addition to the Knative ingress implementations in the ([install a networking
 
 Here is the list of the ingress implementations (alphabetically):
 
-- [Emissary-Ingress (Ambassador API Gateway)](https://www.getambassador.io/docs/edge-stack/latest/howtos/knative/): Emissary-Ingress is an open-source Kubernetes-native API Gateway + Layer 7 load balancer + Kubernetes Ingress built on Envoy Proxy.
+- [Emissary ingress (Ambassador API Gateway)](https://www.getambassador.io/docs/edge-stack/latest/howtos/knative/): Emissary ingress is an open-source Kubernetes-native API Gateway + Layer 7 load balancer + Kubernetes ingress built on Envoy Proxy.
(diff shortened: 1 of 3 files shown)


nak3:
/cc @ZhiminXiang @julz @dprotaso 
Could you please take a look?

nak3:
/cc @hbagdi (Kong) @impl (Ambassador)
Could you also please take a look?

docs/admin/install/serving/knative-third-party-ingress.md
```diff
@@ -0,0 +1,8 @@
+# Third-Party Ingress Implementations
```
julz, line 1: I'm wondering if we might want to say something like "beta" or "unsupported" (or I guess "unofficial") rather than "third party", since I can imagine us adding a new ingress to this list (eg IngressV2!) that's not third party, but also isn't passing all tests/ready for general use. (And in the same way I can imagine a 3rd party ingress that _does_ pass the tests and that we're confident is working being added to the main guide rather than this list). WDYT?

nak3: Thank you! Sounds good to me. Updated "third-party" to `unofficial`.

impl:
Hey @nak3,

I wonder if it might be helpful to list the reasons why someone might want to pick each third-party/unofficial networking layer over the default? For example the service mesh and its use cases for Istio and at least for Ambassador you can e.g. put authorization (or an entire filter chain in the case of Ambassador Edge Stack) over the Knative handling. I can't speak to the unique capabilities of the other solutions.

I'm happy to give you some language for the Ambassador tab if you'd like to do that.

Otherwise, the only thing I'd note is that the Ambassador open-source project was accepted to CNCF incubation and has changed its name to Emissary, but they're still working on filling out all the details and updating things, so I don't think we need to do anything here until they release the next major version with the Emissary branding.

dprotaso:
re: @impl rather than us host that copy is there a place we can link to? I just worry about that list of benefits getting out of date etc.

hbagdi:
This is fine by us (Kong).
We still have Knative's conformance tests to sort out and that is present on our backlog (https://github.com/Kong/kubernetes-ingress-controller/issues/707).
Once we get to it, we will come back and update the docs here.

nak3:
Yes, I agree with Dave. I like @impl's idea but hosting the link on Knative site is probably better for the maintainability.

docs/admin/install/serving/install-serving-with-yaml.md
```diff
@@ -168,6 +168,9 @@ Follow the procedure for the networking layer of your choice:
         !!! tip
             Save this to use in the [Configure DNS](#configure-dns) section below.
 
+!!! info
+    You can also use other ingress implementations provdied. See [Unofficial Ingress Implementations](../ingress-implementations).
```
abrennan89, line 172:
```suggestion
    You can also use other ingress implementations provided. See [unofficial ingress implementations](../ingress-implementations).
```

docs/admin/install/serving/ingress-implementations.md
```diff
@@ -0,0 +1,11 @@
+# Unofficial Ingress Implementations
```
abrennan89, line 1:
```suggestion
# Unofficial ingress implementations
```

docs/admin/install/serving/ingress-implementations.md
```diff
@@ -0,0 +1,11 @@
...
+
+In addition to the Knative Ingress implementations in the ([install a networking layer](../install-serving-with-yaml/#install-a-networking-layer)) section,
+there are other Knative Ingress implementations.
```
abrennan89, line 4:
```suggestion
In addition to the Knative ingress implementations in the ([install a networking layer](../install-serving-with-yaml/#install-a-networking-layer)) section, there are other Knative ingress implementations.
```

This isn't capitalized below, so I'd decide on either caps or no caps and then keep it consistent. Unless it's a literal API object we're referring to I'd recommend no caps.

docs/admin/install/serving/ingress-implementations.md
```diff
@@ -0,0 +1,11 @@
...
+
+Here is the list of the ingress implementations (alphabetically):
+
+- [Emissary-Ingress (Ambassador API Gateway)](https://www.getambassador.io/docs/edge-stack/latest/howtos/knative/): Emissary-Ingress is an open-source Kubernetes-native API Gateway + Layer 7 load balancer + Kubernetes Ingress built on Envoy Proxy.
+- [Kong](https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/using-kong-with-knative/): Kong is an open source API Gateway built for hybrid and multi-cloud environments.
```
abrennan89, line 9:
```suggestion
- [Emissary ingress (Ambassador API Gateway)](https://www.getambassador.io/docs/edge-stack/latest/howtos/knative/): Emissary ingress is an open-source Kubernetes-native API Gateway + Layer 7 load balancer + Kubernetes ingress built on Envoy Proxy.
- [Kong](https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/using-kong-with-knative/): Kong is an open source API Gateway built for hybrid and multi-cloud environments.
```

docs/admin/install/serving/ingress-implementations.md
```diff
@@ -0,0 +1,11 @@
...
+
+These implementations are not officially supported by Knative community at this moment, but the implementation will be added to the main guide once it is ready for general use.
```
abrennan89, line 11: I'm not sure I would add this to main docs at all yet if it's not supported tbh; they should probably be documented for use with Knative in some other repo or on in the Kong docs, etc.

csantanapr:
bimp @nak3

nak3:
Although I fixed review comments and rebased, I am wondering if I should include this docs or not.
The Kong and Ambassador docs has not been updated for a long time.
https://www.getambassador.io/docs/edge-stack/latest/howtos/knative/
https://docs.konghq.com/kubernetes-ingress-controller/2.0.x/guides/using-kong-with-knative/
If nobody strongly demand this docs, I think we can close this without merge.

csantanapr:
I'm ok to close, if it's not officially supported by knative, it should not be in the docs. 

/close

PR #4030, opened 2021-07-19
fix the wrong value

huija:
cloudEvent  from producer should match: https://github.com/knative/docs/blob/5469d8e5cf3698b15a076773bcbf6678750dcfb6/docs/eventing/samples/helloworld/helloworld-go/sample-app.yaml#L63


huija:
@googlebot I signed it!

Issue #4058, opened 2021-07-26
Wrong version for strimzi-kafka-operator


xinydev:
## Expected Behavior

https://knative.dev/docs/eventing/samples/kafka/

There is no `apiVersion: kafka.strimzi.io/v1beta2` in strimzi-kafka-operator(0.16.2) 

strimzi spi: 
https://github.com/strimzi/strimzi-kafka-operator/blob/5ccb3a88f7060a27d579d299f520639a7be81a37/api/src/main/java/io/strimzi/api/kafka/model/Kafka.java#L33-L52

We need upgrade to  0.20.0+(at least) for strimzi-kafka-operator for using `v1beta2`

## Steps to Reproduce the Problem

1. https://knative.dev/docs/eventing/samples/kafka/
## Additional Info
**Additional context**

**Install information:**
 - Platform (GKE, IKS, AKS, etc.):
 - Knative Version: 0.24

abrennan89:
Is this covered by https://github.com/knative/docs/pull/4028 already?
I think it's better to point out to upstream Strimzi docs rather than have to keep maintaining our own separate install docs for it.

xinydev:
> Is this covered by #4028 already?
> I think it's better to point out to upstream Strimzi docs rather than have to keep maintaining our own separate install docs for it.

Thanks!
/close

PR #4028, opened 2021-07-17
General eventing and kafka updates

abrennan89:
Fixes #3783
Fixes #3078

## Proposed Changes

- Remove outdated Strimzi install and point to Strimzi quickstart docs instead
- Remove yaml sample files in samples/kafka, and references to same, and replace them with including the yaml directly in the doc instead
- Remove no-copy from verification command in Eventing installation docs
- Replace link to Kafka channel sample with link to knative-sandbox Kafka channel. Not sure on this last one, confirm with eng what's our recommended way to install a Kafka channel. We can maybe remove some of the channel sample docs altogether in this case or replace the YAML samples with what's in sandbox ⚠️  cc @devguyio I'd love your input on this.

PR #4059, opened 2021-07-26
Update strimzi-kafka-operator version

xinydev:
Fixes: https://github.com/knative/docs/issues/4058 

 Update kafka version to `2.8.0`, because strimzi(0.24.0) only suport kafka version >=2.7.0 ,see https://strimzi.io/downloads/

## Proposed Changes 

- 
-
-

### DIFF
--- a/docs/eventing/samples/kafka/README.md
+++ b/docs/eventing/samples/kafka/README.md
@@ -30,9 +30,7 @@ is to install it by using [Strimzi](https://strimzi.io).
    ```
 1. Install the Strimzi operator, like:
    ```bash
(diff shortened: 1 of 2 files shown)


abrennan89:
Is this covered by https://github.com/knative/docs/pull/4028 already?
I think it's better to point out to upstream Strimzi docs rather than have to keep maintaining our own separate install docs for it.

xinydev:
> Is this covered by #4028 already?
> I think it's better to point out to upstream Strimzi docs rather than have to keep maintaining our own separate install docs for it.

Thanks!
/close

PR #4065, opened 2021-07-28
adding Code of Conduct redirect

omerbensaadon:


omerbensaadon:
Alright this whole "still using the website repo" thing is becoming a PITA

config/redirects.yml
```diff
@@ -11,6 +11,7 @@ plugins:
       admin/install/install-serving-with-yaml.md: admin/install/serving/install-serving-with-yaml.md
       client/connecting-kn-to-your-cluster/index.md: client/README.md
       concepts/overview.md: _index.md
+      contributing/code-of-conduct/README.md: /community/contributing/code-of-conduct/README.md
```
abrennan89, line 14: not sure if the `/` is what's breaking this, or it needs to somehow be directed up a level further?

julz: This redirects outside of the mkdocs stuff, so I think we'll have to do what we did in the other case we needed to do this and use a netlify redirect

PR #4075, opened 2021-07-29
Updating knative client connection doc

vyasgun:
Fixes: [#1251](https://github.com/knative/client/issues/1251)

## Proposed Changes 

- Adding information about `KUBECONFIG` env variable
- Adding information about `--config` option in `kn`

### DIFF
--- a/docs/client/README.md
+++ b/docs/client/README.md
@@ -32,6 +32,10 @@ After you have installed `kubectl` or `kn`, these tools will search for the `kub
 
 A `kubeconfig` file is usually automatically created when you create a Kubernetes cluster.
 
+You can also specify the configuration by setting the environment variable `$KUBECONFIG` that points to the config file.
+
+`kn` has the `--config` option which can also be used for specifying a config file like `--config path/to/config`.
+
 For more information about `kubeconfig` files, see <a href="https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/" target="_blank">Organizing Cluster Access Using kubeconfig Files</a>.
 
 ### Using kubeconfig files with your platform
--- a/docs/client/README.md
+++ b/docs/client/README.md
@@ -32,9 +32,11 @@ After you have installed `kubectl` or `kn`, these tools will search for the `kub
 
 A `kubeconfig` file is usually automatically created when you create a Kubernetes cluster.
 
-You can also specify the configuration by setting the environment variable `$KUBECONFIG` that points to the config file.
+You can also specify a config file in the following ways:
 
-`kn` has the `--config` option which can also be used for specifying a config file like `--config path/to/config`.
+- Setting the environment variable `$KUBECONFIG`, and point it to the kubeconfig file.
+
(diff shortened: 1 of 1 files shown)


docs/client/README.md
```diff
@@ -32,6 +32,10 @@ After you have installed `kubectl` or `kn`, these tools will search for the `kub
 
 A `kubeconfig` file is usually automatically created when you create a Kubernetes cluster.
 
+You can also specify the configuration by setting the environment variable `$KUBECONFIG` that points to the config file.
+
+`kn` has the `--config` option which can also be used for specifying a config file like `--config path/to/config`.
+
```
snneji, line 38: Do these two new lines refer to two additional ways you can specify a configuration? If yes, it might be easier to read if these were bullet points. For example:

> You can also specify a config file with the following options:
> - Setting the environment variable `$KUBECONFIG`, which points to the config file.
> - Using the `kn` CLI `--config` option, for example, `--config <path-to-config>`.

snneji: Also, I'm wondering which `kn` command can you run with the `--config` option?

vyasgun: It's a global option so it could work with any command. I will add an improved example.

docs/client/README.md
```diff
@@ -32,6 +32,12 @@ After you have installed `kubectl` or `kn`, these tools will search for the `kub
 
 A `kubeconfig` file is usually automatically created when you create a Kubernetes cluster.
 
+You can also specify a config file in the following ways:
+
+- Setting the environment variable `$KUBECONFIG`, and point it to the kubeconfig file.
+
+- Using the `kn` CLI `--config` option, for example, `kn --config path/to/config service list`. The default config is at `~/.config/kn/config.yaml`.
```
snneji, line 39:
```suggestion
- Using the `kn` CLI `--config` option, for example, `kn service list --config path/to/config.yaml`. The default config is at `~/.config/kn/config.yaml`.
```

snneji: Hi thanks for applying my feedback! Sorry, just one more suggestion. I've usually seen the option at the end of the command and also it would be nice to have a file extension at the end of the path.

vyasgun: okay, looks good. I'll commit the suggestion :)

snneji:
Thanks @vyasgun!

/lgtm
/approve

docs/client/README.md
dsimansk: @vyasgun @snneji sorry I'm bit late to this one. However, I think the added content is mixing `kubeconfig` and kn's native config file that's outside of "Connect to cluster" scope.

There're 2 relevant flags provided by `kn` to the story:
```
     --context string      name of the kubeconfig context to use
     --kubeconfig string   kubectl configuration file (default: ~/.kube/config)

```

- `--kubeconfig` is equal to env variable `$KUBECONFIG`
- `--context` can be used to further specify name of context to use from existing `kubeconfig`, e.g. listed as `kubectl config get-contexts `

@vyasgun let me know if I should create a PR to correct the docs.

vyasgun: I will create another one with these two options covered.

PR #4080, opened 2021-07-29
add initial offset to kafka source

itsmurugappan:
Adding docs for the new feature in kafka source. [pr](https://github.com/knative-sandbox/eventing-kafka/pull/779) for reference

## Proposed Changes 

- Add docs for newly introduced field (initialOffset) in kafkasource.

### DIFF
--- a/docs/developer/eventing/sources/kafka-source/README.md
+++ b/docs/developer/eventing/sources/kafka-source/README.md
@@ -285,6 +285,31 @@ To specify it, add the label `kafkasources.sources.knative.dev/key-type` to the 
         name: event-display
    ```
 
+## (Optional) Specify the initial offset
+
+By default the `KafkaSource` starts consuming from the `latest` offset in each partition. In case you want to consume from the earliest, use this field to configure that same.
+
+   ```yaml
+   apiVersion: sources.knative.dev/v1beta1
+   kind: KafkaSource
+   metadata:
+    name: kafka-source
+   spec:
+    consumerGroup: knative-group
+    initialOffset: earliest
+    bootstrapServers:
+    - my-cluster-kafka-bootstrap.kafka:9092 # note the kafka namespace
+    topics:
+    - knative-demo-topic
+    sink:
+      ref:
+        apiVersion: serving.knative.dev/v1
+        kind: Service
+        name: event-display
+   ```
+
+*NOTE:* valid values for `initialOffset` is `earliest` or `latest`, any other value would result in an validation error. Also this field will be honored, only if there are no prior commited offsets for that consumer group.
+
 ## Connecting to a TLS enabled Kafka broker
 
 The KafkaSource supports TLS and SASL authentication methods. For enabling TLS authentication, please have the below files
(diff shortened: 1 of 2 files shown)


docs/developer/eventing/sources/kafka-source/README.md
```diff
@@ -285,6 +285,31 @@ To specify it, add the label `kafkasources.sources.knative.dev/key-type` to the
         name: event-display
    ```
 
+## (Optional) Specify the initial offset
+
+By default the `KafkaSource` starts consuming from the `latest` offset in each partition. In case, you want to consume from the earliest offset, use this field to configure the same.
```
lionelvillard, line 290: `to configure the same` does not sound right. What about `In case you want to consume from the earliest offset, set the initialOffset field to earliest`?

docs/developer/eventing/sources/kafka-source/README.md
```diff
@@ -285,6 +285,31 @@ To specify it, add the label `kafkasources.sources.knative.dev/key-type` to the
...
+    topics:
+    - knative-demo-topic
+    sink:
+      ref:
+        apiVersion: serving.knative.dev/v1
+        kind: Service
+        name: event-display
+   ```
+
+*NOTE:* valid values for `initialOffset` is `earliest` or `latest`, any other value would result in a validation error. Also this field will be honored, only if there are no prior committed offsets for that consumer group.
```
lionelvillard, line 311: I would remove `Also` and the comma after `honored`.

docs/developer/eventing/sources/kafka-source/README.md
```diff
@@ -285,6 +285,31 @@ To specify it, add the label `kafkasources.sources.knative.dev/key-type` to the
         name: event-display
    ```
 
+## (Optional) Specify the initial offset
+
+By default the `KafkaSource` starts consuming from the `latest` offset in each partition. In case you want to consume from the earliest offset, set the initialOffset field to `earliest`.
```
snneji, line 290:
```suggestion
By default the `KafkaSource` starts consuming from the latest offset in each partition. If you want to consume from the earliest offset, set the `initialOffset` field to `earliest`, for example:
```

docs/developer/eventing/sources/kafka-source/README.md
```diff
@@ -285,6 +285,31 @@ To specify it, add the label `kafkasources.sources.knative.dev/key-type` to the
...
+    topics:
+    - knative-demo-topic
+    sink:
+      ref:
+        apiVersion: serving.knative.dev/v1
+        kind: Service
+        name: event-display
+   ```
+
+*NOTE:* valid values for `initialOffset` is `earliest` or `latest`, any other value would result in a validation error. This field will be honored only if there are no prior committed offsets for that consumer group.
```
snneji, line 311:
```suggestion
!!! note
    The valid values for `initialOffset` are `earliest` or `latest`. Any other value results in a validation error. This field is honored only if there are no prior committed offsets for that consumer group.
```

snneji (changes requested):
A few formatting tweaks. Otherwise looks good.

PR #4090, opened 2021-08-02
Fix the value of the Kafka broker class to `Kafka`

rhuss:
@matzew it's only "Kafka" without a broker suffix, right? (which is a bit confusing as the other class that I know (`MTChannelBasedBroker`) uses a different scheme (e.g. should then be `KafkaBasedBroker`). Btw would have preferred just an `MTChannel` as the broker class as the `BasedBroker` is really superfluous and redundant. (cc: @lionelvillard )


rhuss:
> Browse the preview: https://deploy-preview-4090--knative.netlify.app

The preview does not reflect the change of this PR. Could it be, that the preview is always build from the main branch ?

pierDipi (approved):
/lgtm

PR #4099, opened 2021-08-04
Fixed case of cloudDNS

senthilnathan:
## Proposed Changes 
- Fixes case of cloudDNS.  Refer this example yaml from cert-manager documentation: https://cert-manager.io/docs/configuration/acme/dns01/google/#create-an-issuer-that-uses-clouddns


snneji:
Thanks @senthilnathan!

/ok-to-test

PR #4114, opened 2021-08-16
Remove outdated YAML for Hugo

### LINKED ISSUE #3882: Clean up yaml metadata
The YAML metadata was used by Hugo but isn't needed anymore since #3881 and should be cleaned up. Leaving this till after release out of an abundance of caution in case we end up needing to regenerate anything etc.

abrennan89:
Fixes  #3882
Some other minor tweaks 😉

### DIFF
--- a/config/nav.yml
+++ b/config/nav.yml
@@ -144,7 +144,6 @@ nav:
       - Code samples:
         - Overview: serving/samples/README.md
         - Cloud Events apps:
-          - Overview: serving/samples/cloudevents/README.md
           - .NET: serving/samples/cloudevents/cloudevents-dotnet/README.md
           - Go: serving/samples/cloudevents/cloudevents-go/README.md
           - Node.js: serving/samples/cloudevents/cloudevents-nodejs/README.md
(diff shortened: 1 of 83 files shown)


config/nav.yml
```diff
@@ -154,7 +153,6 @@ nav:
         - GitHub Webhook - Go: serving/samples/gitwebhook-go/README.md
         - gRPC Server - Go: serving/samples/grpc-ping-go/README.md
         - Hello world apps:
-          - Overview: serving/samples/hello-world/README.md
```
julz, line 157: just to triple check, you intended to remove these two overviews in this PR? (just because it seems different from the rest of the changes; rest lgtm)

abrennan89: Yeah, those readmes only had YAML frontmatter and nothing else in them, so there were two I removed like this IIRC

julz: aha, cool, lgtm then!

julz (approved):
/lgtm
/approve

PR #4179, opened 2021-09-03
Fix inaccurate attribute description in Broker docs

antoineco:
## Proposed Changes

- Propagate wording change from https://github.com/knative/eventing/pull/5703 to ConfigMaps cited as examples in the Broker docs.
- Improve phrasing around some explanations of how the default channel config is applied to Brokers.

PR #4182, opened 2021-09-04
Fix inaccurate description of `default-br-config`

antoineco:
## Proposed Changes

- Reflect the changes from https://github.com/knative/eventing/pull/5703 and https://github.com/knative/eventing/pull/5713 to the documentation examples.
    (Context: it appears that this comment was copied from somewhere else, and the docs happily copied it although it didn't make sense)

This is a second take on #4179, but this time I refrained from trying to clarify the sentences I found confusing and only focused on copying the changes from https://github.com/knative/eventing/pull/5703.

/assign snneji
/assign abrennan89

### DIFF
--- a/docs/admin/eventing/broker-configuration.md
+++ b/docs/admin/eventing/broker-configuration.md
@@ -13,7 +13,7 @@ metadata:
   labels:
     eventing.knative.dev/release: devel
 data:
-  # Configuration for defaulting Channels that do not specify CRD implementations.
+  # Configuration for defaulting Brokers that do not specify a spec.config or Broker class.
   default-br-config: |
     clusterDefault:
       brokerClass: MTChannelBasedBroker
@@ -56,7 +56,7 @@ metadata:
   labels:
     eventing.knative.dev/release: devel
 data:
-  # Configuration for defaulting Channels that do not specify CRD implementations.
+  # Configuration for defaulting Brokers that do not specify a spec.config or Broker class.
   default-br-config: |
     clusterDefault:
       brokerClass: MTChannelBasedBroker
@@ -85,7 +85,7 @@ metadata:
   labels:
     eventing.knative.dev/release: devel
 data:
-  # Configuration for defaulting Channels that do not specify CRD implementations.
+  # Configuration for defaulting Brokers that do not specify a spec.config or Broker class.
   default-br-config: |
     clusterDefault:
       brokerClass: MTChannelBasedBroker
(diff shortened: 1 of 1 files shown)


docs/admin/eventing/broker-configuration.md
```diff
@@ -13,7 +13,7 @@ metadata:
   labels:
     eventing.knative.dev/release: devel
 data:
-  # Configuration for defaulting Channels that do not specify CRD implementations.
+  # Configuration for defaulting Brokers that do not specify a spec.config or Broker class.
```
antoineco, line 16: @abrennan89 in response to your request for changes in the aforementioned PR:

---

```diff
- for defaulting Brokers
+ for default Brokers
```

The configuration is used for defaulting _all_ Brokers. It is not for the _default Broker_.

---

> I think use `default-br-config` throughout instead of `spec.config`

I think there is a misunderstand about what the comment says here. "Brokers that do not specify [...]": we are talking about what *Brokers* specify (or don't specify) via their *spec.config* field, not what the ConfigMap specifies.

(Here is how Eventing works: if ` spec.config` is missing, Brokers get their config from the ConfigMap. Otherwise not.)

---

tl;dr; The wording is correct here.

docs/admin/eventing/broker-configuration.md
```diff
@@ -183,18 +183,18 @@ metadata:
...
-​
+
 ### Configuring the default BrokerClass for namespaces
-​
+
 You can modify the default Broker class for one or more namespaces.
 
 For example, if you want to use a `KafkaBroker` class for all other Brokers created on the cluster, but you want to use the `MTChannelBasedBroker` class for Brokers created in `namespace-1` and `namespace-2`, you would use the following ConfigMap settings:
-​
+
```
antoineco, line 197: In case somebody wonders what these seemingly empty changes mean, the doc contained zero-width space characters which are invisible to the naked eye but were displayed by my editor. I removed them.

image

docs/admin/eventing/broker-configuration.md
snneji: Not sure if it might make sense to say something like:
"Configures the default for any Broker that does not specify a `spec.config` or Broker class."

It might make it clearer that it's not referring to the _default Broker_ but to _any_ Broker.

antoineco: That sounds clearer yes. I submitted this suggestion to Eventing and will update this PR once the change is approved: https://github.com/knative/eventing/pull/5713

PR #4199, opened 2021-09-08
Update otel-collector-config configmap

thunderboltsid:
The otel-collector-config configmap referenced in
https://knative.dev/docs/admin/collecting-metrics/#set-up-the-collector
has incorrect configuration which leads to a config loading issue causing
the otel-collector to go into a crashloop. This PR updates the config map
as per https://opentelemetry.io/docs/collector/configuration/


dprotaso:
@thunderboltsid thanks!

Issue #4178, opened 2021-09-03
A subject example for SinkBinding of CronJob uses invalid image and deprecated API version


cardil:
A subject example for SinkBinding of CronJob uses invalid image 

https://github.com/knative/docs/blob/001cd9bd29f345114fe324446cb0eb194f771d28/docs/developer/eventing/sources/sinkbinding/getting-started.md#L157

```bash
$ skopeo inspect docker://gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats
FATA[0000] Error parsing image name "docker://gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats": reading manifest latest in gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats: manifest unknown: Failed to fetch "latest" from request "/v2/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats/manifests/latest".

$ skopeo list-tags docker://gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats 
{
    "Repository": "gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats",
    "Tags": []
}
```

It also uses the deprecated API version of `batch/v1beta1`. The `batch/v1` should be used instead.

https://github.com/knative/docs/blob/001cd9bd29f345114fe324446cb0eb194f771d28/docs/developer/eventing/sources/sinkbinding/getting-started.md#L140

```
W0903 12:59:03.239558  958779 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
```

cardil:
For image problem we could link to https://knative.dev/docs/eventing/samples/container-source/#prepare-the-heartbeats-image but building that image is super cumbersome. It is used in multiple places in the docs, just like the event-display image is. 

I think we should prebuild heartbeats image and publish it the same as we are publishing event-display image.

cardil:
Hmm, This page has even more problems. 

https://github.com/knative/docs/blob/001cd9bd29f345114fe324446cb0eb194f771d28/docs/developer/eventing/sources/sinkbinding/getting-started.md#L212-L216

This command also seems not to work. There are a couple of problems with it:

- subject here uses `Job:batch/v1` but above example uses `CronJob:batch/v1beta1`
- subject uses label selector of `app=heartbeat-cron`, but above cron job example doesn't have such label
[...]

PR #4198, opened 2021-09-08
Updates `hearbeats` Source Image References

JeffNeff:
Fixes the mkdocs portion of updating issue #https://github.com/knative/docs/issues/4188
Fixes the mkdocs portion of issue #https://github.com/knative/docs/issues/4178

## Proposed Changes 

- Updates the referenced image of the heartbeats source

PR #4204, opened 2021-09-08
Example Updates in `/developer/eventing/sources/sinkbinding/getting-started.md`

JeffNeff:
Closes part of issue https://github.com/knative/docs/issues/4178

## Proposed Changes 

- Updates the example YAML API version in `/developer/eventing/sources/sinkbinding/getting-started.md` moving the batch API from `v1beta1` to `v1`

### DIFF
--- a/docs/developer/eventing/sources/sinkbinding/getting-started.md
+++ b/docs/developer/eventing/sources/sinkbinding/getting-started.md
@@ -154,7 +154,7 @@ any extra overrides given by `CE_OVERRIDES`.
               restartPolicy: Never
               containers:
                 - name: single-heartbeat
-                  image: gcr.io/knative-releases/knative.dev/eventing-contrib/cmd/heartbeats
+                  image: gcr.io/knative-nightly/knative.dev/eventing/cmd/heartbeats:latest
                   args:
                   - --period=1
                   env:
--- a/docs/developer/eventing/sources/sinkbinding/getting-started.md
+++ b/docs/developer/eventing/sources/sinkbinding/getting-started.md
@@ -186,11 +186,10 @@ Create a `SinkBinding` object that directs events from your subject to the sink.
     Create a `SinkBinding` object by running:
 
     ```bash
-    kn source binding create <name> \
-      --namespace <namespace> \
-      --subject "<subject>" \
-      --sink <sink> \
-      --ce-override "<cloudevent-overrides>"
+    kn source binding create bind-heartbeat \
+     --subject "Job:batch/v1:app=heartbeat-cron" \
+     --sink event-display \
+     --ce-override "sink=bound"
     ```
     Where:
 
--- a/docs/developer/eventing/sources/sinkbinding/getting-started.md
+++ b/docs/developer/eventing/sources/sinkbinding/getting-started.md
@@ -208,11 +208,11 @@ Create a `SinkBinding` object that directs events from your subject to the sink.
 
     For example:
     ```bash
-    $ kn source binding create bind-heartbeat \
-      --namespace sinkbinding-example \
-      --subject "Job:batch/v1:app=heartbeat-cron" \
-      --sink http://event-display.svc.cluster.local \
-      --ce-override "sink=bound"
+    $ kn source binding create <name> \
+      --namespace <namespace> \
+      --subject "<subject>" \
+      --sink <sink> \
+      --ce-override "<cloudevent-overrides>"
     ```
 
 === "YAML"
--- a/docs/developer/eventing/sources/sinkbinding/getting-started.md
+++ b/docs/developer/eventing/sources/sinkbinding/getting-started.md
@@ -137,7 +137,7 @@ any extra overrides given by `CE_OVERRIDES`.
 1. Create a YAML file for the CronJob using the following example:
 
     ```yaml
-    apiVersion: batch/v1beta1
+    apiVersion: batch/v1
     kind: CronJob
     metadata:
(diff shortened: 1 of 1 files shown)


sebgoa (changes requested):
why don't you update line 157 to `gcr.io/knative-nightly/knative.dev/eventing/cmd/heartbeats` since you just did a PR for the same thing https://github.com/knative/docs/pull/4198 it seems.

Also do run the command for ` kn source...` to make sure this is the correct syntax and that it works well in the context of the steps described on the page.

JeffNeff:
> why don't you update line 157 to `gcr.io/knative-nightly/knative.dev/eventing/cmd/heartbeats` since you just did a PR for the same thing #4198 it seems.
> 
> Also do run the command for ` kn source...` to make sure this is the correct syntax and that it works well in the context of the steps described on the page.

I thought I covered it in [this PR](https://github.com/knative/docs/pull/4198/files).. I have updated it!

Also yes, I ran the `kn source` command. It returned successfully. @sebgoa

sebgoa:
I would move the update of the heartbeat image to this PR, so that this PR can close #4178 entirely at once.

JeffNeff:
@sebgoa done.

docs/developer/eventing/sources/sinkbinding/getting-started.md
```diff
@@ -186,11 +186,10 @@ Create a `SinkBinding` object that directs events from your subject to the sink.
...
     ```bash
-    kn source binding create <name> \
-      --namespace <namespace> \
-      --subject "<subject>" \
-      --sink <sink> \
-      --ce-override "<cloudevent-overrides>"
+    kn source binding create bind-heartbeat \
+     --subject "Job:batch/v1:app=heartbeat-cron" \
+     --sink event-display \
+     --ce-override "sink=bound"
```
snneji, line 192: I think this command was valid. This command has placeholders for users to input their own values. Could you replace this command please?

The [issue](https://github.com/knative/docs/issues/4178#issuecomment-912752920) is referring to the example command on lines 211-215. If this example command is causing confusion, it might make sense to just delete it.

snneji (changes requested):
Hi @JeffNeff, I've asked a question on the issue to clarify what to do about the namespace. I'm not sure it makes sense to remove it from this command. It might make more sense to add it to the steps where it is missing.

docs/developer/eventing/sources/sinkbinding/getting-started.md
snneji, line 192: I think this command was valid. This command has placeholders for users to input their own values. Could you replace this command please?

The [issue](https://github.com/knative/docs/issues/4178#issuecomment-912752920) is referring to the example command on lines 211-215. If this example command is causing confusion, it might make sense to just delete it.

snneji (changes requested):
Hi @JeffNeff, I've asked a question on the issue to clarify what to do about the namespace. I'm not sure it makes sense to remove it from this command. It might make more sense to add it to the steps where it is missing.

docs/developer/eventing/sources/sinkbinding/getting-started.md
JeffNeff: done!.. i did not delete the 211-215 one tho.. should we? Im not positive.

snneji: The command with the placeholders needs to be before the "Where" section on line 194 because that section has the description of what the user should replace the placeholder with.

snneji: In terms of the example (`kn source binding create bind-heartbeat...`), I'm not sure whether it is confusing. 

The intention of this topic is for users to create their own sinkbinding with their own values, not just run the example commands. The reason why I suggested deleting the example is because I thought there might be some confusion about whether you are supposed to run the example command as well. 

If you don't think the example is confusing you can keep it, but it should go under the line with "For example".

snneji, line 192:
```suggestion
    kn source binding create <name> \
      --namespace <namespace> \
      --subject "<subject>" \
      --sink <sink> \
      --ce-override "<cloudevent-overrides>"
```

docs/developer/eventing/sources/sinkbinding/getting-started.md
```diff
@@ -209,11 +208,11 @@ Create a `SinkBinding` object that directs events from your subject to the sink.
...
-    $ kn source binding create bind-heartbeat \
-      --namespace sinkbinding-example \
-      --subject "Job:batch/v1:app=heartbeat-cron" \
-      --sink http://event-display.svc.cluster.local \
-      --ce-override "sink=bound"
+    $ kn source binding create <name> \
+      --namespace <namespace> \
+      --subject "<subject>" \
+      --sink <sink> \
+      --ce-override "<cloudevent-overrides>"
```
snneji, line 215:
```suggestion
    kn source binding create bind-heartbeat \
      --namespace sinkbinding-example \
      --subject "Job:batch/v1:app=heartbeat-cron" \
      --sink http://event-display.svc.cluster.local \
      --ce-override "sink=bound"
```

snneji: Either delete lines 209 to 215, or apply the example command in my suggestion above.

snneji (changes requested):

abrennan89:
@JeffNeff are you still working on this?

Issue #3762, opened 2021-06-09
Bad formatting on bentoml sample in community


julz:
See https://knative.dev/community/samples/serving/machinelearning-python-bentoml/ in current (hugo) docs.

Search for ```bash. Also notice the yaml inline and without a code block in the "Building and deploying the sample" section.

soham283:
@julz  is this issue is open or has been resolved ??

julz:
Yes, looks like this is still an issue, see e.g:

![image](https://user-images.githubusercontent.com/354013/125412506-e7f3ce00-e3b6-11eb-95ac-1d77a356e6ec.png)
(should be in code block)

and 

![image](https://user-images.githubusercontent.com/354013/125412243-9ea37e80-e3b6-11eb-81d5-16da912627b8.png)
("Run the `main.py` file to train the model" shouldn't be in a code block, and the ````bash` text should not be visible in the output)

One thing to note for anyone picking this up is the community samples are still built with Hugo rather than MkDocs, so the normal preview won't work  and to reproduce locally you'd need to do a full `./hack/build.sh serve`

abrennan89:
@julz I feel like we should either move the Community section to the new site before working on this or pull out the samples from this section and add them elsewhere in the admin or dev guides?

PR #4222, opened 2021-09-09
[WIP] Clean up formatting (community sample)

abrennan89:
Fixes https://github.com/knative/docs/issues/3762

Formatting cleanup, editing.


abrennan89:
@julz I didn't see the sense in formatting this entirely differently when we're either going to move the samples to the website or put them in their own repo eventually anyway, so I took the liberty of moving this one into the regular samples folder and formatting it for mkdocs for now, so that we're at least starting to group samples in one place with one type of formatting.
LMK if that's a problem and we can revisit if so.

docs/serving/samples/machinelearning-python-bentoml/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+### iris_classifier
+
+BentoML creates a model API server by using prediction service abstraction. In the `iris_classifier.py` file, a prediction service is defined, which:
+
+1. Requires a `scikit-learn` model.
+1. Asks BentoML to identify the required `pip` dependencies.
+1. Defines an API, which is the entry point for accessing this machine learning service.
+
+--8<-- "iris_classifier.py"
```
abrennan89, line 35: @omerbensaadon @julz howwww do I include .py files? 😭 This just doesn't show up but I wasn't sure if `include` works for non-md things. We should figure this out anyway for the whole where-do-samples-go discussion

julz: ah, right, we [currently have the base path for snippets set to `docs/snippets`](https://github.com/knative/docs/blob/mkdocs/mkdocs.yml#L36), so this would have to be:

~~~suggestion
```python
--8<-- "../serving/samples/machinelearning-python/bentoml/iris_classifier.py"
```
~~~

Maybe we should set base path to `docs/` instead (and then the snippets would be linked as `--8<-- "snippets/blah.md"`, which might be clearer anyway 🤔). Unfortunately [it looks like](https://github.com/mkdocs/mkdocs/issues/2154) mkdocs doesn't yet support using current-file-relative links in snippets :(.

abrennan89: +1 for setting the base path to `/docs`, otherwise it's going to make snippets really hard to organize because we'll have a bunch of random topics and files in one directory 🙂 
@julz can you open an issue to document this for contributing docs? I feel like this is something we should have written down so folks know how to handle it going forward.

PR #4226, opened 2021-09-10
Delete "Getting Started with App Deployment" 

omerbensaadon:
This page is confusing for users searching for the Getting Started Guide 

1. It's messing with the SEO of the Knative Getting Started guide 
Screen Shot 2021-09-10 at 1 34 57 PM

2. When people arrive at this page, they often migrate away from it to the actual Getting Started guide
Screen Shot 2021-09-10 at 1 35 21 PM

3. It's not even in the Serving nav 

4. Everything covered here is covered in the new Getting Started guide


antoineco (approved):
Agree, the page called "Deploying your first Knative Service" in the actual getting started section feels like a better getting started guide.
We also have the code samples.

PR #4229, opened 2021-09-14
Correct kubectl expected output

ikvmw:
Going through getting started guide...


snneji:
Thanks for spotting that @ikvmw!
/approve
/lgtm

PR #4234, opened 2021-09-14
Add redirect for a moved file to prevent user 404s

RichieEscarez:
Fix broken links

### DIFF
--- a/config/redirects.yml
+++ b/config/redirects.yml
@@ -1,6 +1,7 @@
 plugins:
   redirects:
     redirect_maps:
+      docs/serving/spec/knative-api-specification-1.0.md: https://github.com/knative/specs/blob/main/specs/serving/knative-api-specification-1.0.md
       serving/feature-flags.md: admin/serving/feature-flags.md
       serving/using-subroutes.md: developer/serving/traffic-management.md
       serving/rolling-out-latest-revision.md: developer/serving/rolling-out-latest-revision.md
--- a/config/redirects.yml
+++ b/config/redirects.yml
@@ -1,7 +1,7 @@
 plugins:
   redirects:
     redirect_maps:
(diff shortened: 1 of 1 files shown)


RichieEscarez:
1. I cant find in the docs contributor instructions why this redirects file excludes the base path to /docs? Should all paths exclude the root folder? https://knative.dev/docs/help/contributor/mkdocs-contributor-guide/#redirects

2. Also, its unclear where to look/test my change to HEAD in the Netlify build/site? Do all PRs get built under /development regardless of the version? ie, a change to a .22 file also gets pushed into /development - (previously, /development represented the current state of the repo at HEAD).

RichieEscarez:
In the current Netlify preview, only the /development URL works and correctly redirects to the file in the git repo:

https://614109193f91c000075166ab--knative.netlify.app/**development**/serving/spec/knative-api-specification-1.0

What **does not work** is the expected /docs URL:

https://614109193f91c000075166ab--knative.netlify.app/**docs**/serving/spec/knative-api-specification-1.0

But its unclear if this is due to the way the Netlify preview/staging builds work (and when published the /docs URL will work)?

julz:
The preview shows what happens when this merges: it'll update `/development` (bc `/development` reflects HEAD of mkdocs branch). `/docs` generally only gets updated when we do a release (because it reflects the latest release-N branch, currently 0.25), or if you cherry pick to the release branch.

julz:
Tldr we should probably cherry pick this after it merges, if we want to fix /docs as well as /development :)

RichieEscarez:
Thanks @julz!
So
/docs = latest release branch (release-0.##)
/development = mkdocs branch
(sorry I missed that before)
I think we only need to redirect /docs but come the next release ~1week from now, this file will get released and /doc will have it too, right?

julz:
Yup exactly

snneji:
@julz do you think this can be merged?

julz (approved):
thanks @RichieEscarez

/lgtm
/approve

PR #4244, opened 2021-09-15
We haven't required Istio for 2 1/2 years

evankanderson:
Updates #63

## Proposed Changes 

- For the last two and half years, we haven't required Istio, but the opening serving sentence still suggests that Istio is a hard dependency.

  Brought up by a prospective customer this week.


xtreme-sameer-vohra:
Thanks :)
I just came across that same thread !!

PR #4246, opened 2021-09-15
Update binary name to match the binary downloads

kvmware:
I don't have a mac so am unsure if doing the brew install would make it so one could invoke `kn quickstart X`. However, when I download the binary labeled `kn-quickstart` I can invoke that directly and `kn` is unaware of any command `quickstart`.

## Proposed Changes 

- Update the binary call to include a `-` instead of a space.

### DIFF
--- a/docs/snippets/quickstart-install.md
+++ b/docs/snippets/quickstart-install.md
@@ -6,9 +6,9 @@
             ```
 
     === "Using a binary"
-         You can install the `quickstart` plugin by downloading the executable binary for your system and placing it on your `PATH` (for example, in `/usr/local/bin`).
+         You can install the `kn-quickstart` plugin by downloading the executable binary for your system and placing it on your `PATH` (for example, in `/usr/local/bin`).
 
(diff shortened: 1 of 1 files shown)


docs/snippets/quickstart-install.md
```diff
@@ -36,15 +36,15 @@
...
 
-The `quickstart` plugin completes the following functions:
+The `kn-quickstart` plugin completes the following functions:
 
 1. **Checks if you have [`kind`](https://kind.sigs.k8s.io/docs/user/quick-start){target=_blank} installed,** and creates a cluster called `knative`.
 1. **Installs Knative Serving with Kourier** as the default networking layer, and nip.io as the DNS.
 1. **Installs Knative Eventing** and creates an in-memory Broker and Channel implementation.
 
-!!! todo "Install Knative and Kubernetes on a local Docker Daemon using `kn quickstart`"
+!!! todo "Install Knative and Kubernetes on a local Docker Daemon using `kn-quickstart`"
```
csantanapr, line 45: The intention of the plugin is to be use by the kn CLI and not directly the kn-quickstart binary. We want user to place kn-quickstart binary in the PATH and use "kn quickStart"

csantanapr (changes requested):
We want user to use kn<space>quickstart

PR #4253, opened 2021-09-16
Document Defaults ConfigMap

julz:
Documents the `config-defaults` config map, following the same style as the existing `config-deployment` config map.

Part of the GA work.

**Note:** because I am lazy, and because it seems worth at least exploring the idea of fully automating these eventually to prevent things getting out of sync between code and docs, this was actually generated by the super-suuuuuper-hacky toy at http://github.com/julz/gen-configmap-docs. I'm happy to touch up the results, though, it just seemed like a easy (/more fun than doing it manually) way to get these done.

If we like the results of this, I can trivially repeat it to get us started on the other config maps that need documenting.

For API review:
/assign @dprotaso

For conformance and DUX review:
/assign @omerbensaadon

### DIFF
--- a/docs/admin/serving/config-defaults.md
+++ b/docs/admin/serving/config-defaults.md
@@ -1,0 +1,374 @@
+# Configuring the Defaults ConfigMap
+
+The `config-defaults` ConfigMap, known as the Defaults ConfigMap, contains settings that determine how Knative sets default values for resources.
+
+This ConfigMap is located in the `knative-serving` namespace.
+
+You can view the current `config-defaults` ConfigMap by running the following command:
+```yaml
+kubectl get configmap -n knative-serving config-defaults -oyaml
+```
+## Example config-defaults ConfigMap
+```yaml
+apiVersion:  v1
+kind:  ConfigMap
+metadata:
+  name:  config-defaults
+  namespace:  knative-serving
+data:
+  revision-timeout-seconds: "300"
+  max-revision-timeout-seconds: "600"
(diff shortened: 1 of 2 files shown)


julz:
Preview: https://deploy-preview-4253--knative.netlify.app/development/admin/serving/config-defaults/

docs/admin/serving/config-defaults.md
```diff
@@ -0,0 +1,374 @@
...
+data:
+  allow-container-concurrency-zero: "true"
+
+```
+
+### Enable Service Links
+{% raw %}
+enable-service-links specifies the default value used for the
+enableServiceLinks field of the PodSpec, when it is omitted by the user.
+See: https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#accessing-the-service
```
dprotaso, line 349: Link isn't clickable.

docs/admin/serving/config-defaults.md
```diff
@@ -0,0 +1,374 @@
...
+
+This is a tri-state flag with possible values of (true|false|default).
+
+In environments with large number of services it is suggested
+to set this value to `false`.
+See https://github.com/knative/serving/issues/8498.
```
dprotaso, line 355: Link isn't clickable.

julz: linked

julz:
We really have to get rid of those tests (or at least move them, and the samples, to their own repo).

/retest

PR #4271, opened 2021-09-21
clean up some of the ingress.class documentation

dprotaso:


dprotaso:
/assign @abrennan89 @salaboy 

just some minor fixes

docs/developer/serving/services/ingress-class.md
```diff
@@ -1,28 +1,21 @@
 # Configuring Services custom ingress class
 
-By default, when a Knative Service is created an Ingress Class (`ingress.class`) is automatically chosen for it based on the content of the `networking-config` ConfigMap located inside the `knative-serving` namespace. 
-
-The `networking-config` ConfigMap is configured when Knative Serving is installed, and if no other ingress class is specified, this defaults to `"istio.ingress.networking.knative.dev"`. 
-
-After it is configured in the `networking-config` ConfigMap, the `ingress.class` is used for all Knative Services that do not specify an `ingress.class` annotation.
-
-!!! warning
-    The ingress class value should only be modified during Knative Serving installation to avoid issues in your deployment.
```
dprotaso, line 10: I removed this warning because changing the configmap just changes the default and doesn't affect existing installations.

abrennan89: I'd maybe add a note to call that out then though, in case a user expected it to change existing installations?

docs/developer/serving/services/ingress-class.md
```diff
@@ -1,28 +1,21 @@
...
+By default, when a Knative Service is created an Ingress Class (`ingress.class`) is automatically chosen for it based on the value in the `config-network` ConfigMap located inside the `knative-serving` namespace. This ConfigMap is part of Knative Serving installation. If the ingress class is not specified, this defaults to `istio.ingress.networking.knative.dev`. Once configured the `ingress.class` is used for all Knative Services unless it is overridden with an `ingress.class` annotation.
```
abrennan89, line 3:
```suggestion
By default, when a Knative Service is created an ingress class (`ingress.class`) is automatically assigned to it, based on the value in the `config-network` ConfigMap located inside the `knative-serving` namespace. This ConfigMap is part of Knative Serving installation. If the ingress class is not specified, this defaults to `istio.ingress.networking.knative.dev`. Once configured the `ingress.class` is used for all Knative Services unless it is overridden with an `ingress.class` annotation.
```

docs/developer/serving/services/ingress-class.md
dprotaso: good point

PR #4282, opened 2021-09-23
Enable sticky tabs

julz:
Trying out the new [sticky tabs feature](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/?h=sticky#sticky-navigation-tabs) in mkdocs. Wdyt @omerbensaadon @csantanapr?


snneji:
Hi @julz could you explain what difference I'm supposed to see? As far as I can see, according to the link you provided, the Knative site already has this behaviour.

Issue #3883, opened 2021-06-29
Clean up BYO event source docs


abrennan89:
I'm currently working on cleaning up the docs @ https://dev-knative.netlify.app/development/eventing/sources/creating-event-sources/ so just opening an issue to track the PR and give visibility.

PR #4284, opened 2021-09-23
[WIP] Custom event routing docs rewrite

abrennan89:
Fixes https://github.com/knative/docs/issues/3883

# DO NOT REVIEW - DRAFT
I'll request reviews once I need them but this is very WIP atm.

## Proposed Changes

- [ ] Rename / restructure "creating event source" section, since this doesn't accurately describe things like containersource and sinkbinding etc.; possibly name it "custom event routing" but need to brainstorm on that
- [ ] Remove all the fluff, make this more like a step by step guide for specific personas that makes it clear what the steps are for each type of custom event routing and why someone would use each of these
- [ ] Also revisit sinkbinding and containersource docs as part of this reorg, since they are loosely related to this and shouldn't really be part of the "event sources" docs


abrennan89:
@snneji made the changes, thanks. Please don't `lgtm` this until it has received some technical reviews from eventing folks, which in this case are more important than writer reviews.

/hold

PR #4298, opened 2021-09-27
Moving the blog to mkdocs

### LINKED ISSUE #4243: Migrate blog from hugo to mkdocs
With the [recent changes](https://twitter.com/squidfunk/status/1437456678753026049?s=20) and example for mkdocs-material blog. I think we can take a stab on converting the [knative blog](https://knative.dev/blog/) from hugo to mkdocs.

The mkdocs-material blog => https://squidfunk.github.io/mkdocs-material/blog/

psschwei:
Fixes #4243 

## Proposed Changes 

- Moving the blog to mkdocs

I know very little about mkdocs (and knew even less before I started this migration), so @omerbensaadon @csantanapr @julz please feel free to be quite liberal in telling me I'm doing it wrong.


julz (changes requested):
I think we need to give a bit more thought to how we do this. The blog is shared between all the docs versions (the blog doesn't get "released" every 6 weeks), so should be at /blog/, not /docs/blog (both in final URL, and in directory structure in the repo - I don't think we should need to move anything on the filesystem for this), which probably means it having its own config file and being built a bit like past releases are (but without the extra git clone). We'll also need to make sure the RSS continues to work properly.

psschwei:
Ok, will revise this to keep the blog in the original location.

Looks like we can use the mkdocs [rss plugin](https://pypi.org/project/mkdocs-rss-plugin/) for the RSS feed.

csantanapr:
Thanks @psschwei for working on this 🎉

psschwei:
Progress update: blog is being built separately and rendering nicely. 

image

Next steps: 
* figure out how to make the html blog entries play nicely with mkdocs
* clean up the build process for the blog. I just basically copied a bunch of stuff into `blog/` locally to get it working (stylesheets, images, etc), but in theory this should be done during the build rather than keeping two copies of stuff

I think we should be able to do the blog post for Knative 1.0 in mkdocs :tada:

psschwei:
Ok, this is ready to go.

/assign @csantanapr @omerbensaadon

psschwei:
To get the new tab layout (moving "Blog" with the rest of the mkdocs tabs) on knative.dev , we'll need to cherry-pick this into the latest release branch.

/cherry-pick release-0.26

If we'd rather NOT cherry-pick this, then we can simply delete this comment before merging this PR. 

FYI @omerbensaadon @csantanapr

blog/mkdocs.yml
```diff
@@ -0,0 +1,93 @@
...
+  - pymdownx.keys
+  - pymdownx.caret
+  - pymdownx.mark
+  - pymdownx.tilde
+  - pymdownx.escapeall
+  - pymdownx.tasklist:
+      custom_checkbox: true
+  - admonition
+  - toc:
+      permalink: true
```
csantanapr, line 48:
```suggestion
      permalink: true
      toc_depth: 1
```
can we do just the top level, this will benefit the release blog posts

psschwei: done

PR #4299, opened 2021-09-27
Ignore venv directory

psschwei:
## Proposed Changes 

- Updates `.gitignore` file to ignore `venv/` directory (python virtual environment, used for building mkdocs locally)


csantanapr:
/approved
If I only can spell 😮‍💨

PR #4289, opened 2021-09-24
Update docs repo for 0.26

evankanderson:
Fixes #4275 

## Proposed Changes 

- Release the 0.26 version of the docs.

/assign @julz

to verify technically

/remove-approve

Since I want someone else to actually make sure I did the right thing by the site.

PR #4300, opened 2021-09-27
Disable failing integration tests

evankanderson:
Per discussion in `#docs`, disable the integration tests for the samples, because they fail a lot and aren't adding much value to most tests.

## Proposed Changes 

- Remove integration tests. We can always add them from git history if we regret it.


evankanderson:
/assign @julz 

Because he was saying he was going to do this, and this may be blocking #4289

julz (approved):
VERY happy to see the back of these 😅. Thanks for doing this @evankanderson 

/lgtm
/approve

Issue #4287, opened 2021-09-24
The RBAC for eventing API Source doesn't seem to work


dprotaso:
The RBAC here says you only need access to `events`
https://knative.dev/docs/developer/eventing/sources/apiserversource/getting-started/

After deploying an `APISource` it actually gives me warning stating it needs access to `pods` and `apps/deployments` before the source becomes `Ready`

/assign @lionelvillard

lionelvillard:
@dprotaso lmk if this is working for you now. If not, please reopen this issue, thx.

PR #4301, opened 2021-09-28
improve ApiServerSource doc

lionelvillard:
Fixes https://github.com/knative/docs/issues/4287

## Proposed Changes 

- Fix `getting-started` page missing in the nabber
- Add event-display as an example for sink
- Document ServiceAccountName in the reference doc
- Various fixes


snneji:
Thanks for fixing these pages. There are a few formatting issues that might be easier to clean up in a separate PR. I've made a cleanup issue here #4303.

lionelvillard:
thanks @snneji.

Issue #4256, opened 2021-09-17
Explain what is different between customizing domain by DomainMapping and config-domain


nak3:
As mentioned in https://knative.slack.com/archives/C9CV04DNJ/p1631885543076700, currently we have two sections to customize domain:
- https://knative.dev/docs/developer/serving/services/custom-domains/
- https://knative.dev/docs/serving/using-a-custom-domain/

We should explain more clearly what are different between them.

I am planning to change them as:
- Add a short explanation of the use case in each section.
- Add the link for each other.
- Change the title of `config-domain` to "Customizing domain template".

/assign

PR #4319, opened 2021-10-04
Add short notes to explain about customizing domain and customizing domain template

nak3:
The following two sections have the same title and instruct how to customize domain:

https://knative.dev/docs/serving/using-a-custom-domain/
https://knative.dev/docs/developer/serving/services/custom-domains/

It should have different titles and a note about the difference.

So this patch changes to:
- Add a short explanation and the link for each section.
- Change the title for domain template to "Customizing domain template".

Fix https://github.com/knative/docs/issues/4256

### DIFF
--- a/docs/developer/serving/services/custom-domains.md
+++ b/docs/developer/serving/services/custom-domains.md
@@ -13,6 +13,11 @@ serve a Knative Service at this domain.
     If you create a domain mapping to map to a [private Knative Service](private-services.md),
     the private Knative Service is accessible from public internet with the custom domain of the domain mapping.
 
+!!! tip
+    This section instructs how to customize the domain of each service regardless of the default domain.
+    If you want to customize the domain template to assign the default domain name,
+    please refer to [setting up a custom domain template](../../../serving/using-a-custom-domain.md).
+
 ## Prerequisites
 
 - You must have access to a Kubernetes cluster, with Knative Serving and an Ingress implementation installed. For more information, see the [Installation documentation](../../../admin/install/README.md).
--- a/docs/serving/using-a-custom-domain.md
+++ b/docs/serving/using-a-custom-domain.md
@@ -1,10 +1,15 @@
-# Setting up a custom domain
+# Setting up a custom domain template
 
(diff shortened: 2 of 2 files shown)


docs/serving/using-a-custom-domain.md
```diff
@@ -1,11 +1,16 @@
-# Setting up a custom domain
+# Setting up a custom domain template
```
snneji, line 1: I'm wondering if a more direct name for this might be something like one of the following:
- Changing the default domain
- Assigning a default domain

docs/developer/serving/services/custom-domains.md
```diff
@@ -13,6 +13,11 @@ serve a Knative Service at this domain.
     If you create a domain mapping to map to a [private Knative Service](private-services.md),
     the private Knative Service is accessible from public internet with the custom domain of the domain mapping.
 
+!!! tip
+    This topic instructs how to customize the domain of each service regardless of the default domain.
+    If you want to customize the domain template to assign the default domain name,
+    see [Setting up a custom domain template](../../../serving/using-a-custom-domain.md).
```
snneji, line 19:
```suggestion
    see [Changing the default domain](../../../serving/using-a-custom-domain.md).
```

docs/developer/serving/services/custom-domains.md
```diff
@@ -13,6 +13,11 @@ serve a Knative Service at this domain.
     If you create a domain mapping to map to a [private Knative Service](private-services.md),
     the private Knative Service is accessible from public internet with the custom domain of the domain mapping.
 
+!!! tip
+    This topic instructs how to customize the domain of each service regardless of the default domain.
```
snneji, line 17:
```suggestion
    This topic instructs how to customize the domain of each service, regardless of the default domain.
```

snneji (changes requested):
Sorry, forgot to update the cross ref with the new topic title. Also, while you are here, would you be able to update the `nav.yml` file with the new title for Changing the default domain.

PR #4331, opened 2021-10-06
Adding trigger annotations

salaboy:
Fixes issue: https://github.com/knative/eventing/issues/5522

## Proposed Changes 

- Adds Trigger's annotations section to Trigger page
- Shows an example of the dependency annotation as it requires configuration
- It doesn't show an example of the injection annotation as it is just an enabled/disabled flag


lionelvillard:
@salaboy can you fix the linter errors?

salaboy:
@lionelvillard sorry.. newbie mistake.. it is passing now

PR #4391, opened 2021-10-27
Adding command-line instructions for using cloudevents player

### LINKED ISSUE #4390: Add command line instructions for CloudEvents player to getting started guide

In the "Introducing the CloudEvents Player" section of the Getting Started guide, we provide instructions for sending/receiving events using the web interface. Some users might want to use the command line instead, so it would be useful to add details on how to do that to the guide (perhaps in a tip box).

/assign

psschwei:
Fixes #4390 

## Proposed Changes 

Adds a tip box with instructions for using `curl` to send/receive cloud events to the Getting Started guide. 

Note: I put the tip box in what seemed like a logical place, but I'm totally open to placing it somewhere else on the page if others think there's a better location.

### DIFF
--- a/docs/getting-started/first-source.md
+++ b/docs/getting-started/first-source.md
@@ -76,7 +76,10 @@ Create the CloudEvents Player Service:
 
 ![CloudEvents Player Send](images/event_sent.png)
 
-??? tip "Want to send events via the command line instead?"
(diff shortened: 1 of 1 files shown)


dprotaso:
looks like it's merging

dprotaso:
or not - I think tide can't merge because of the required check is not green

PR #4434, opened 2021-11-04
fix redirect loop

snneji:
I checked all pages for broken redirects and found one more. Hopefully this should be the last one.


psschwei:
This is a bit odd... the [preview](https://deploy-preview-4434--knative.netlify.app/docs/eventing/sources/apiserversource/) still seems to be stuck in a loop, but when I built locally it works fine.

If we think it's just a glitch with the preview, then happy to stamp it, as it LGTM... @snneji let me know if you think that's ok

snneji:
As far as I can see, the redirect is fixed if you change to the development branch in the preview site. https://deploy-preview-4434--knative.netlify.app/development/eventing/sources/apiserversource/

psschwei:
:facepalm: That gets me way more often than it should...

psschwei (approved):
/lgtm

PR #4437, opened 2021-11-04
Adding command-line instructions for using cloudevents player 

### LINKED ISSUE #4390: Add command line instructions for CloudEvents player to getting started guide

psschwei:
Fixes #4390 

## Proposed Changes 

Adds a tip box with instructions for using `curl` to send/receive cloud events to the Getting Started guide. 

retry from #4391 since that one seems to be borked

### DIFF
--- a/docs/getting-started/first-source.md
+++ b/docs/getting-started/first-source.md
@@ -76,7 +76,10 @@ Create the CloudEvents Player Service:
 
 ![CloudEvents Player Send](images/event_sent.png)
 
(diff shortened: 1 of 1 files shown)


psschwei:
#4391 actually did merge, so this isn't needed

PR #4454, opened 2021-11-08
Change the format of the section of installing eventing sources

houshengbo:
## Proposed Changes 

- This PR changed the format of the section installing eventing with different eventing sources.


csantanapr:
Thank you so much @houshengbo for addressing this, I hit this problem yesterday on not finding this section

PR #4458, opened 2021-11-09
Add Activator capacity to navigation

snneji:
Noticed that this file was on the site but could not be found in the navigation


abrennan89 (approved):
/lgtm

PR #4467, opened 2021-11-11
Fix sample output

matthiasgubler:
Fix deprecated sample output documentation

## Proposed Changes 

- Use domain-mapping and domainmapping-webhook insted of the net-istio in sample output
- Use new sample output for knative eventing installation
-

### DIFF
--- a/docs/install/operator/knative-with-operators.md
+++ b/docs/install/operator/knative-with-operators.md
@@ -183,7 +183,7 @@ autoscaler             1/1     1            1           18s
 autoscaler-hpa         1/1     1            1           14s
 controller             1/1     1            1           18s
 domain-mapping         1/1     1            1           12s
-domainmapping-webhoop  1/1     1            1           12s
+domainmapping-webhook  1/1     1            1           12s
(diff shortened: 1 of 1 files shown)


docs/install/operator/knative-with-operators.md
```diff
@@ -182,8 +182,8 @@ activator              1/1     1            1           18s
 autoscaler             1/1     1            1           18s
 autoscaler-hpa         1/1     1            1           14s
 controller             1/1     1            1           18s
-net-istio-webhook      1/1     1            1           12s
-net-istio-controller   1/1     1            1           12s
+domain-mapping         1/1     1            1           12s
+domainmapping-webhoop  1/1     1            1           12s
```
matthiasgubler, line 186:
```suggestion
domainmapping-webhook  1/1     1            1           12s
```

nak3:
LGTM

Depending on the configuration, we can see `net-*` or some other deployments. But this is a __sample__ output so I think this change would be fine.

abrennan89 (approved):
/lgtm

PR #4480, opened 2021-11-12
Move doc about network layer up before verifying

### LINKED ISSUE #4468: Installationguide for Knative Serving via operator is not working
## Expected Behavior
I can follow the guideline step by step and the result is a nice and clean installation of Knative Serving with an arbitrary network layer.

## Actual Behavior
We installed it up until this part of the documentation:
https://knative.dev/docs/install/operator/knative-with-operators/#:~:text=Check%20the%20status%20of%20Knative%20Serving%20Custom%20Resource%3A

And we got an error instead of a ready instance. By checking the logs, we saw the message that istio is missing. For me this implies, that istio should be installed in beforehand.

```
{"severity":"INFO","timestamp":"2021-11-11T07:26:45.306424703Z","logger":"knative-operator.event-broadcaster","caller":"record/event.go:282","message":"Event(v1.ObjectReference{Kind:\"KnativeServing\", Namespace:\"knative-serving\", Name:\"knative-serving\", UID:\"1bd34580-0143-4799-9d80-bd5fd1635bc4\", APIVersion:\"operator.knative.dev/v1alpha1\", ResourceVersion:\"2163\", FieldPath:\"\"}): type: 'Warning' reason: 'InternalError' please install istio or disable the istio ingress plugin: no matches for kind \"Gateway\" in version \"networking.istio.io/v1alpha3\"","knative.dev/pod":"knative-operator-57887499b8-nxkd7"}
```

1. Follow the manual on a fresh kubernetes instance (1.20+) up until the mentioned text
2. Check the status of Knative Serving with `kubectl get KnativeServing knative-serving -n knative-serving` 
3. See an error and validate it via `stern knative`

## Additional Info
**Additional context**
I created a PR #4467 because of a different output we got. But maybe this is because something in the order of the installation changed

**Install information:**
 - Platform (GKE, IKS, AKS, etc.): Kubernetes local in Docker
 - Knative Version: latest

matthiasgubler:
Fixes #4468

## Proposed Changes 

- Move the network layer installation guide part up, right before the verification of Knative serving. Else you will get an error by following the installation manual step by step
-
-


matthiasgubler:
Following discussion on slack is open: https://knative.slack.com/archives/C9CV04DNJ/p1636730362092700
This by this the operator-based documentation could be aligned with the yaml-based documentation, depending on the decision

houshengbo:
@matthiasgubler This can be what I would like to suggest:
https://github.com/houshengbo/docs/blob/modify-operator-network-layer/docs/install/operator/knative-with-operators.md#installing-the-knative-serving-component
Just add a small section before create and apply the CR for networking layers.

houshengbo:
/lgtm
Waiting for the comments from others.

matthiasgubler:
@houshengbo I cherry-picked your change and aligned the yaml and operator doc to have both the same "default" for the networking-layer.

docs/install/operator/knative-with-operators.md
```diff
@@ -49,6 +49,13 @@ kubectl logs -f deploy/knative-operator
 
 ## Installing the Knative Serving component
 
+### Prepare the networking layer
+
+Knative Serving relies on the networking layer to manage the ingress and egress traffic. Istio is the default network layer
+for Knative Operator configuration. Please follow [Install Istio](../serving/installing-istio.md) before proceeding.
+
+If you would like to leverage networking layers other than Istio, directly go to [this section](#installing-knative-serving-with-different-networking-layers).
```
snneji, line 57: I have a few questions about this new section:

1. Do users have to install the networking layer before applying the Serving CR? Or do they just have to install Istio at this point if they want to use the default networking layer. 
2. If they want to use other networking layers, should they install and configure them now?
3. I find it a bit confusing that there are two sections for installing the networking layer. Would it make sense to move the section "Installing Serving with different networking layers" before the section for applying the Serving CR instead of having this short section that links to another one below? 
That way users can do the sections in the order they appear on the page instead of skipping down to another section and then having to go back up to apply the Serving CR.

cc: @houshengbo

docs/install/serving/install-serving-with-yaml.md
```diff
@@ -31,27 +31,28 @@ Follow the procedure for the networking layer of your choice:
 <!-- TODO: Link to document/diagram describing what is a networking layer.  -->
 <!-- This indentation is important for things to render properly. -->
 
-=== "Kourier (Choose this if you are not sure)"
```
snneji, line 34: I think it's still useful for users to know that we recommend that they use Kourier if they don't know which layer to install. Could you replace this information?

docs/install/serving/install-serving-with-yaml.md
```diff
@@ -31,27 +31,28 @@ Follow the procedure for the networking layer of your choice:
...
+=== "Istio (default)"
```
snneji, line 34: Is Istio the default layer when installing with YAML Istio? I don't know if the default networking layer here is the same as when installing with the Operator.

snneji (changes requested):
I've left some questions to consider. One other comment is that the reply to your Slack message from @nak3 said that it would be useful to mention "Choose Kourier if you are not sure" in the Operator doc as well, because it would be better for new users to use Kourier instead of Istio.

docs/install/serving/install-serving-with-yaml.md
nak3: Samia is correct. Istio is not default when installing with YAML.

abrennan89:
@matthiasgubler are you still working on this one?

matthiasgubler:
@abrennan89 yes I am, but currently I'm on vacation. I will review the comments in the next one to two weeks.

snneji (approved):

PR #4481, opened 2021-11-12
Switch to snake-casing for config map properties

dprotaso:
- update casing for serving's config-deployment
- update casing for serving's config-network

Part of https://github.com/knative/serving/issues/7300

### DIFF
--- a/docs/serving/using-auto-tls.md
+++ b/docs/serving/using-auto-tls.md
@@ -217,7 +217,7 @@ in the `knative-serving` namespace to reference your new `ClusterIssuer`.
 
 Update the
 [`config-network` ConfigMap](https://github.com/knative/serving/blob/main/config/core/configmaps/network.yaml)
-in the `knative-serving` namespace to enable `auto-tls`and specify how HTTP
+in the `knative-serving` namespace to enable `auto-tls` and specify how HTTP
 requests are handled:
 
 1.  Run the following command to edit your `config-network` ConfigMap:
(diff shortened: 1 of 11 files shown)


docs/serving/using-auto-tls.md
```diff
@@ -217,7 +217,7 @@ in the `knative-serving` namespace to reference your new `ClusterIssuer`.
 
 Update the
 [`config-network` ConfigMap](https://github.com/knative/serving/blob/main/config/core/configmaps/network.yaml)
-in the `knative-serving` namespace to enable `autoTLS`and specify how HTTP
+in the `knative-serving` namespace to enable `auto-tls`and specify how HTTP
```
snneji, line 220:
```suggestion
in the `knative-serving` namespace to enable `auto-tls` and specify how HTTP
```

snneji (changes requested):
Looks good. Just a few minor things I noticed for consistency. 

1. Should `autoTLS` always be `auto-tls`? I noticed a few of the former still remain in the docs. 
1. Just to confirm, this is only for  ConfigMaps right. There are a few occasions, for example, with `rollout-duration` where it is in camel case for the operator, but not for the config map.

dprotaso:
> Should autoTLS always be auto-tls? I noticed a few of the former still remain in the docs.

I'll defer to you - the style guide has `autoTLS` as the preferred use. I just updated the PR so it's consistent.

> Just to confirm, this is only for ConfigMaps, right? 

Yes

> There are a few occasions, for example, with rollout-duration where it is in camel case for the operator, but not for the config map.

`spec` properties in the operator YAML should be `camelCase` - that's the general convention.

snneji (approved):

PR #4508, opened 2021-11-18
Update ContainerSource README.md

xtreme-sameer-vohra:
## Proposed Changes 
- Add resolved url to heartbeats example. I wasn't sure if there is automation for resolving `[gcloud-project]` to a value or this was just an oversight 🤔  
- Add missing env vars required for heartbeats example to work when using kn


abrennan89 (approved):
/lgtm

PR #4510, opened 2021-11-19
Use `redirected` instead of `enabled` in the HTTPS redirection example

nak3:
In [HTTPS redirection](https://knative.dev/docs/serving/services/http-option/) doc, the example uses `enabled` but it is a default value and so it doe not change anything actually.

The section is talking about `HTTPS redirection` so we should use
`redirected` in the example configuration.

/cc @abrennan89 @mgencur


abrennan89 (approved):
/lgtm

mgencur:
Looks good to me.

PR #4514, opened 2021-11-22
add stable window documentatino to "scale-bounds"

enoodle:
Documents: https://github.com/knative/serving/pull/12286

## Proposed Changes 

- Document "Stable window" in scal bounds page as it is now supported with both auto scalers.


docs/serving/autoscaling/kpa-specific.md
```diff
@@ -264,3 +264,59 @@ This setting determines the maximum ratio of existing to desired pods. For examp
         autoscaler:
           max-scale-down-rate: "4.0"
     ```
+
+### Scale Down Delay
```
julz, line 268: I'm not sure we should hide this feature inside the KPA page since it's a pretty useful feature. It seems more confusing than helpful for a reader to have to guess to look in the KPA page rather than seeing it with the other scale related properties. The other things on the KPA page are very specific to KPA, rather than just not implemented in HPA yet -- I'd prefer we just mark that this does not currently work with HPA.

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -157,23 +157,17 @@ When the Revision is created, the larger of initial scale and lower bound is aut
           allow-zero-initial-scale: "true"
     ```
 
+## Stable window
 
+The stable window defines minimal time that will pass between descisions to change the scale of a revision.
```
julz, line 162: Is this true? What if we're in panic mode?

docs/serving/autoscaling/kpa-specific.md
enoodle: :+1: I will move this back.

docs/serving/autoscaling/scale-bounds.md
enoodle: you are right, I will add reference to panic mode.

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -217,6 +213,59 @@ delay period.
           scale-down-delay: "15m"
     ```
 
+## Stable window
+
+The stable window defines minimal time that will pass between descisions to change the scale of a revision. [Panic mode](kpa-specific.md) can override this setting.
```
snneji, line 218:
```suggestion
The stable window defines minimal time that passes between decisions to change the scale of a revision. [Panic mode](kpa-specific.md) can override this setting.
```

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -157,18 +157,14 @@ When the Revision is created, the larger of initial scale and lower bound is aut
           allow-zero-initial-scale: "true"
     ```
 
-
-
-
-
-## Scale Down Delay
+### Scale Down Delay
```
snneji, line 160:
```suggestion
## Scale Down Delay
```

snneji (changes requested):
Just a few minor typos, other than the text looks good to me.

docs/serving/autoscaling/scale-bounds.md
enoodle: :+1:

docs/serving/autoscaling/scale-bounds.md
enoodle: :+1:

snneji:
Hi @julz, if you are happy with the technical information here, could you add an LGTM?

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -157,18 +157,14 @@ When the Revision is created, the larger of initial scale and lower bound is aut
...
-
 ## Scale Down Delay
 
 Scale Down Delay specifies a time window which must pass at reduced concurrency
 before a scale-down decision is applied. This can be useful, for example, to
 keep containers around for a configurable duration to avoid a cold start
 penalty if new requests come in. Unlike setting a lower bound, the revision
 will eventually be scaled down if reduced concurrency is maintained for the
-delay period.
+delay period. *only supported for KPA*
```
julz, line 167: Maybe convert this to something like:

`!!! note "Only Supported for the default KPA autoscaler class"`

(so we get nice formatting)

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -217,6 +213,59 @@ delay period.
           scale-down-delay: "15m"
     ```
 
+## Stable window
+
+The stable window defines minimal time that passes between decisions to change the scale of a revision. [Panic mode](kpa-specific.md) can override this setting.
```
julz, line 218: I think this isn't quite right. We average over the stable and panic windows to get the _input_ for decisions, but we actually make decisions [every two seconds by default](https://github.com/knative/serving/blob/61178b2e8bf8d8afe870151ff59ee9724739b496/pkg/autoscaler/scaling/multiscaler.go#L36) . I'd say something like:

```suggestion
The stable window defines the sliding time window over which metrics are averaged to provide the input for scaling decisions when the autoscaler is not in [Panic mode](kpa-specific.md).
```

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -217,6 +213,59 @@ delay period.
...
+
+* **Global key:** `stable-window`
+* **Per-revision annotation key:** `autoscaling.knative.dev/window`
+* **Possible values:** Duration, `6s` <= value <= `1h`
+* **Default:** `60s`
+
+!!! note
+    During scale down, the last Replica will only be removed after there has not been any traffic to the Revision for the entire duration of the stable window.
```
julz, line 226: I _think_ this is true, in that if there hasn't been any traffic during the stable window the average metric will not be 0, so we won't scale down, but I'm not completely sure that we guarantee this e.g. if [ExponentialWeightedBuckets](https://github.com/knative/serving/blob/main/pkg/autoscaler/aggregation/bucketing.go#L150) were used I wonder if the average might go to zero faster. I'm probably being too pedantic here, though 😅.

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -157,18 +157,14 @@ When the Revision is created, the larger of initial scale and lower bound is aut
...
-
 ## Scale Down Delay
 
 Scale Down Delay specifies a time window which must pass at reduced concurrency
 before a scale-down decision is applied. This can be useful, for example, to
 keep containers around for a configurable duration to avoid a cold start
 penalty if new requests come in. Unlike setting a lower bound, the revision
 will eventually be scaled down if reduced concurrency is maintained for the
-delay period.
+delay period. !!! note "Only Supported for the default KPA autoscaler class"
```
snneji, line 167:
```suggestion
delay period. 
!!! note 
    Only supported for the default KPA autoscaler class.
```

snneji (changes requested):
Correct note formatting

docs/serving/autoscaling/scale-bounds.md
enoodle: :+1:

docs/serving/autoscaling/scale-bounds.md
snneji: @julz Do you think slightly changing the wording here would make sense - maybe to something less definite like:
> During scale down, the last Replica is usually removed after there has not been any traffic to the Revision for the entire duration of the stable window

Or do you think what is here now is OK?

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -217,4 +215,58 @@ delay period.
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            autoscaling.knative.dev/window: "40s"
```
csantanapr, line 243: I think the annotation is on the Service Resource not on the pod spec right?

docs/serving/autoscaling/scale-bounds.md
julz: sorry for slow response @snneji I think I might've replied to this on mobile and forgot to hit submit 🤦; yeah I think that looks good

docs/serving/autoscaling/scale-bounds.md
```diff
@@ -217,4 +215,58 @@ delay period.
...
+
+The stable window defines the sliding time window over which metrics are averaged to provide the input for scaling decisions when the autoscaler is not in [Panic mode](kpa-specific.md).
+
+* **Global key:** `stable-window`
+* **Per-revision annotation key:** `autoscaling.knative.dev/window`
+* **Possible values:** Duration, `6s` <= value <= `1h`
+* **Default:** `60s`
+
+!!! note
+    During scale down, the last Replica will only be removed after there has not been any traffic to the Revision for the entire duration of the stable window.
```
snneji, line 228:
```suggestion
    During scale down, in most cases the last Replica is removed after there has been no traffic to the Revision for the entire duration of the stable window.
```

docs/serving/autoscaling/scale-bounds.md
snneji: @csantanapr do you have a suggested change that should be made here? I'd like to get this PR merged.

csantanapr: >I think the annotation is on the Service Resource not on the pod spec right?

I double-check, I don't know what I was thinking when I wrote the comment

The annotation is in the correct space under `spec.metadata.annotations` 
Ignore my comment, the way you have it is correct.

snneji (approved):

PR #4523, opened 2021-11-29
Update details in the tracing blog post

benmoss:
I was doing another pass at using tracing and going through my [blog post](https://knative.dev/blog/articles/distributed-tracing/). I noticed that some things no longer work because of upstream changes. This hard-codes the versions of the OTEL and Jaeger operator so that they continue to work with the examples as written.


evankanderson (commented):
/lgtm
/approve

PR #4524, opened 2021-11-29
Redirect to Go package docs from a real browser

imjasonh:
Opening this to start a conversation, feel free to close if this has been proposed and rejected before.

Basically, I'd love to be able to browse to https://knative.dev/serving, https://knative.dev/pkg, etc., and not get a 404. With this change, these URLs should be processed by real browsers with a client-side redirect to https://pkg.go.dev/... where any available Go package docs will be served.

If the repo doesn't have Go packages, this will redirect to a 404 over there. But for knative.dev/pkg, etc., this would be more useful. We can talk about adding some configuration to only redirect if the package has Go code, maybe? 🤷 

The only real code change is in `tools/redir-gen/main.go` (at the bottom of the list of changed files in this PR), the rest is generated by running that script. I've also deleted `tools/redir-gen/redirtemplate.gohtml` which appears to be unused.

### DIFF
--- a/golang/_redirects
+++ b/golang/_redirects
@@ -60,13 +60,14 @@
 /networking/* go-get=1 /golang/networking.html 200
 /operator/* go-get=1 /golang/operator.html 200
 /pkg/* go-get=1 /golang/pkg.html 200
+/pkg/* https://pkg.go.dev/knative.dev/pkg/:splat
 /reconciler-test/* go-get=1 /golang/reconciler-test.html 200
 /release/* go-get=1 /golang/release.html 200
 /sample-controller/* go-get=1 /golang/sample-controller.html 200
 /sample-source/* go-get=1 /golang/sample-source.html 200
 /serving/* go-get=1 /golang/serving.html 200
+/serving/* https://pkg.go.dev/knative.dev/serving/:splat
 /specs/* go-get=1 /golang/specs.html 200
 /test-infra/* go-get=1 /golang/test-infra.html 200
 /ux/* go-get=1 /golang/ux.html 200
-/website/* go-get=1 /golang/website.html 200
 /wg-repository/* go-get=1 /golang/wg-repository.html 200
--- a/tools/redir-gen/main.go
+++ b/tools/redir-gen/main.go
@@ -29,6 +29,11 @@ var (
 	knativeOrgs        = []string{"knative", "knative-sandbox"}
 	allowedRepoRe      = regexp.MustCompile("^[a-z][-a-z0-9]+$")
 	archivedExceptions = []string{"eventing-contrib"}
+
+	// Repos known to contain Go libraries, so that requesting
+	// https://knative.dev/<repo> in a browser should redirect to
+	// https://pkg.go.dev/knative.dev/<repo>
+	redirectingGoRepos = []string{"pkg", "serving"}
 )
 
 // repoInfo provides a simple holder for GitHub repo information needed to
@@ -102,6 +107,8 @@ const (
 `
 	redirText = `/{{.Repo}}/* go-get=1 /golang/{{.Repo}}.html 200
 `
+	godocRedirText = `/{{.Repo}}/* https://pkg.go.dev/knative.dev/{{.Repo}}/:splat
+`
 
 	autogenPrefix = `
 # This file is AUTO-GENERATED
@@ -125,8 +132,13 @@ func appendRedirs(ris []repoInfo) error {
 		return fmt.Errorf("unable to write to %q: %w", redirFilename, err)
 	}
 	for _, ri := range ris {
-		if redirTemplate.Execute(redirFile, ri) != nil {
+		if err := redirTemplate.Execute(redirFile, ri); err != nil {
 			return fmt.Errorf("unable to write %s to %q: %w", ri.Repo, redirFilename, err)
+		}
+		if isGoRepo(ri.Repo) {
+			if err := godocRedirTemplate.Execute(redirFile, ri); err != nil {
+				return fmt.Errorf("unable to write godoc redirect for %s to %q: %w", ri.Repo, redirFilename, err)
+			}
 		}
 	}
 	return nil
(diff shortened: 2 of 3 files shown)


imjasonh:
This seems to have not worked, because of the `go-get=1` requirement -- https://deploy-preview-4524--knative.netlify.app/pkg?go-get=1 redirects as expected.

I'm pushing a revision without the `go-get=1` directive, let's see how it goes. 🤞

imjasonh:
This works now for https://deploy-preview-4524--knative.netlify.app/pkg

But not for subdirectories: https://deploy-preview-4524--knative.netlify.app/pkg/apis redirects to the root. I'm close, I can fix it.

imjasonh:
> This works now for https://deploy-preview-4524--knative.netlify.app/pkg
> 
> But not for subdirectories: https://deploy-preview-4524--knative.netlify.app/pkg/apis redirects to the root. I'm close, I can fix it.

Both links work now ✨ 
- https://deploy-preview-4524--knative.netlify.app/pkg
- https://deploy-preview-4524--knative.netlify.app/pkg/apis

And the Go import redirect is unchanged: 
```
$ curl https://deploy-preview-4524--knative.netlify.app/pkg\?go-get\=1
<html><head>
    <meta name="go-import" content="knative.dev/pkg git https://github.com/knative/pkg">
    <meta name="go-source" content="knative.dev/pkg     https://github.com/knative/pkg https://github.com/knative/pkg/tree/main{/dir} https://github.com/knative/pkg/blob/main{/dir}/{file}#L{line}">
</head></html>
```

evankanderson:
I need to check that knative.dev/docs is unaffected by this when I'm at a computer...

imjasonh:
> I need to check that knative.dev/docs is unaffected by this when I'm at a computer...



https://deploy-preview-4524--knative.netlify.app/docs/ seems to do what you want it to do, and not redirect to https://pkg.go.dev/knative.dev/docs

julz:
In general I love this, but there's a small problem: `docs/` doesn't redirect to `pkg.go.dev/knative.dev/docs` (which is good), but [docs/potato](https://deploy-preview-4524--knative.netlify.app/docs/potato) _does_ try to redirect to `pkg.go.dev/knative.dev/docs/potato` which means it ends up giving a pkg-go-dev style 404 rather than a netlify one. I don't know that that's the most important thing ever, but it'd be nice if there was a way to fix it.

dprotaso:
I think automation is pulling too many non-go repos into the redirects.

ie. we don't want knative.dev/release & community etc.

dprotaso:
It's easy to filter out

ie.

`curl proxy.golang.org/knative.dev/serving/@v/list`

imjasonh:
Ooh, nice. This does make it quite a bit slower since it has to hit ~70-some proxy.golang.org URLs, so I did all the HTTP checking concurrent. This might be getting a bit overengineering-ey, maybe we should just maintain a manual blocklist for non-Go repos? I think it's just `docs` and `community`...

imjasonh:
As an added bonus, https://deploy-preview-4524--knative.netlify.app/pkg/apis#Conditions even maintains the URL fragment (in my Chrome at least)

dprotaso:
> maybe we should just maintain a manual blocklist for non-Go repos? I think it's just docs and community...

It's probably easier to make it opt-in for the repos you want
serving
eventing
pkg
networking

imjasonh:
> It's probably easier to make it opt-in for the repos you want

That seems a bit overly restrictive to me. Why should `knative.dev/pkg` get all the fun and leave out `knative.dev/eventing-awssqs`, `knative.dev/net-kourier`, etc.? People might not import those as libraries very often, but people working on them might want to browse around the godoc.

dprotaso:
> People might not import those as libraries very often, but people working on them might want to browse around the godoc.

Then they'd make a PR to add it?

imjasonh:
> Then they'd make a PR to add it? I don't really have a preference

Man, I've been uselessly typing "knative.dev/pkg/webhook" into browsers for _years_, and having just recently found the activation energy to make it do something useful, I just don't want anybody to ever have to think about it ever again. 😆

mattmoor:
(drive by comment) I think if this redirected to source, I'd opt repos in by default.  Given that it redirects to godoc, and only a handful of repos are actually used for library code, I kinda agree with @dprotaso the more I think about it.

imjasonh:
> only a handful of repos are actually used for library code

Depends how you define that.

Taking a totally arbitrary example, [`knative.dev/eventing-awssqs` defines some API types](https://pkg.go.dev/knative.dev/eventing-awssqs/pkg/apis/sources/v1alpha1) that are [imported by `knative.dev/eventing-autoscaler-keda`](https://pkg.go.dev/knative.dev/eventing-awssqs@v0.27.0/pkg/apis/sources/v1alpha1?tab=importedby).

`knative.dev/eventing-autoscaler-keda` [imports `knative.dev/eventing-kafka` and `knative.dev/eventing-redis`](https://pkg.go.dev/knative.dev/eventing-autoscaler-keda/pkg/reconciler/keda?tab=imports).

`knative.dev/eventing-kafka` is also [imported by its CLI plugin, and `knative.dev/eventing-kafka-broker`](https://pkg.go.dev/knative.dev/eventing-kafka/pkg/apis/sources/v1beta1?tab=importedby).

I'd posit that developers of any of those repos would benefit from having an easier way to find and browse godoc for these packages. Should we make them ask for it first? I've already done the work.

dprotaso:
Majority of the updated list just 404's in pkg.go.dev now - there's probably more utility to redirect to the GitHub repos

csantanapr:
What is the latest status for this PR still work in progress and being reviewed?

imjasonh:
> What is the latest status for this PR still work in progress and being reviewed?

I'd still like to do it. If an allowlist for pkg and serving is preferred I can make that change.

csantanapr:
My 2 cents: less magic and make an allow list even if it's long and manual to add another one in the future.

dprotaso:
> > What is the latest status for this PR still work in progress and being reviewed?
> 
> I'd still like to do it. If an allowlist for pkg and serving is preferred I can make that change.

let's do it!

dprotaso:
@csantanapr do we need to cherry pick for this to take effect?

dprotaso:
Looks like it's live now

PR #4536, opened 2021-12-02
Fixed wrong docs for DomainMapping API

lenalebt:
DomainMapping API uses `secretName`, not `secret`, see https://knative.dev/docs/reference/api/serving-api/#serving.knative.dev/v1alpha1.SecretTLS


julz (approved):
thanks for the fix @lenalebt!

/lgtm
/approve

PR #4537, opened 2021-12-02
Document Kafka Broker commit offset interval

### LINKED ISSUE #4460: Document Kafka Broker commit offset interval
## What is changing? (Please include as many details as possible.)

The committed offset is a pointer to the last record that a
the consumer has successfully processed.

We're committing the offset every [`auto.commit.interval.ms`]
milliseconds.

Users can change the interval by changing the
`config-kafka-broker-data-plane` ConfigMap
that is introduced in:
https://knative.dev/docs/eventing/broker/kafka-broker/#kafka-producer-and-consumer-configurations

Note: this is using the Kafka consumer configuration
with the same name [`auto.commit.interval.ms`] but we use it
even without setting `enable.auto.commit` to true.

[`auto.commit.interval.ms`]: https://kafka.apache.org/documentation/#consumerconfigs_auto.commit.interval.ms

## In what release will this take happen (to the best of your knowledge)?

- 0.26
- 1.0
- 1.1

## Context
Link to associated PRs or issues from other repos here.

1. https://github.com/knative-sandbox/eventing-kafka-broker/pull/1405

## Additional info

/kind eventing

pierDipi:
Fixes #4460

## Proposed Changes 

- Document Kafka Broker commit offset interval

### DIFF
--- a/docs/eventing/broker/kafka-broker/README.md
+++ b/docs/eventing/broker/kafka-broker/README.md
@@ -227,6 +227,40 @@ kubectl create secret --namespace <namespace> generic <my_secret> \
 !!! note
     `ca.crt` can be omitted to fallback to use system's root CA set.
 
+## Consumer Offsets Commit Interval
+
+Kafka consumers keep track of the last successfully sent events by committing offsets.
+
+For performance reasons, it's not recommended committing offsets every time an event is successfully sent to a
+subscriber.
+
+Knative Kafka Broker commits the offset every `auto.commit.interval.ms` milliseconds.
+
+`auto.commit.interval.ms` can be changed by changing the `config-kafka-broker-data-plane` `ConfigMap`
+in the `knative-eventing` namespace by changing the parameter `auto.commit.interval.ms` as follows:
+
+```yaml
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-kafka-broker-data-plane
+  namespace: knative-eventing
+data:
+  # Some configurations omitted ...
+  config-kafka-broker-consumer.properties: |
+    # Some configurations omitted ...
+
+    # Commit the offset every 5000 millisecods (5 seconds)
+    auto.commit.interval.ms=5000
+```
+
+!!! note
+    While Knative Kafka Broker guarantees at least once delivery, hence your applications may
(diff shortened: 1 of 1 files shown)


docs/eventing/broker/kafka-broker/README.md
```diff
@@ -227,6 +227,40 @@ kubectl create secret --namespace <namespace> generic <my_secret> \
...
+    # Commit the offset every 5000 millisecods (5 seconds)
+    auto.commit.interval.ms=5000
+```
+
+!!! note
+    While Knative Kafka Broker guarantees at least once delivery, hence your applications may
+    receive duplicate events, the higher the commit interval is the higher is the probability
+    of receiving duplicate events since when a Consumer restarts, it starts again from last
+    committed offsets.
+
```
abrennan89, line 263:
```suggestion
## Consumer offsets commit interval

Kafka consumers keep track of the most recent, successfully sent events by committing offsets.

The Knative Kafka Broker commits offsets every `auto.commit.interval.ms` milliseconds.

!!! note
    To prevent negative impacts to performance, it is not recommended to commit offsets every time an event is successfully sent to a subscriber.

This interval can be modified in the `config-kafka-broker-data-plane` ConfigMap in the `knative-eventing` namespace, by modifying the `auto.commit.interval.ms` parameter as follows:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-kafka-broker-data-plane
  namespace: knative-eventing
data:
  # Some configurations omitted ...
  config-kafka-broker-consumer.properties: |
    # Some configurations omitted ...

    # Commit the offset every 5000 millisecods (5 seconds)
    auto.commit.interval.ms=5000
```
!!! note
    Knative Kafka Broker guarantees at least once delivery, which means that your applications may receive duplicate events. A higher commit interval means that there is a higher probability of receiving duplicate events, because when a Consumer restarts, it restarts from the last committed offset.
```

abrennan89: For some reason the suggestion is formatting weird, so you don't need to commit this - you can see my suggestions for changes and feel free to use ones that make sense though

pierDipi: Thank you, that's much better!

Issue #2108, opened 2020-01-16
Support for networking protocols is not explained anywhere


ahmetb:
Upon visiting https://knative.dev/docs/serving/
I read what Knative serving is.

I wonder, does it support HTTP? Ctrl-f 'http':
![image](https://user-images.githubusercontent.com/159209/72502073-57eabd80-37ed-11ea-8d92-07ed7985ba71.png)

I wonder: does it support grpc? 
Ctrl-F 'grpc': [no results]

I wonder: does it support any TCP application? How about UDP?
Ctrl-F 'tcp', 'protocol', 'network': [no relevant results]

abrennan89:
@ahmetb is this issue still relevant? I tried searching grpc and saw some results but idk if it's what you're looking for.
I'd recommend using the actual site search rather than Ctrl+F until docs are organized better...

ahmetb:
I think it’s still very much relevant. Almost like it should be in project introduction. A very fundamental question a visitor might ask.

evankanderson:
This should probably be in some very introductory (first 2-3 pages of the site) content that both HTTP and HTTPS are supported, and that Knative has tools to help with HTTPS setup if you want that and are internet-connected with a DNS name.

abrennan89:
@evankanderson I see this one's still assigned to you, do you plan to add a PR or something for this? Thanks

PR #4539, opened 2021-12-03
Add note about HTTPS

abrennan89:
Fixes https://github.com/knative/docs/issues/2108 
Remove fluff content

### DIFF
--- a/docs/serving/README.md
+++ b/docs/serving/README.md
@@ -1,15 +1,13 @@
 # Knative Serving
 
-Knative Serving builds on Kubernetes to support deploying and serving
-of serverless applications and functions. Serving is easy to get started with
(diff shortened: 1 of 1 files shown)


docs/serving/README.md
```diff
@@ -1,16 +1,14 @@
 # Knative Serving
 
-Knative Serving builds on Kubernetes to support deploying and serving
-of serverless applications and functions. Serving is easy to get started with
```
snneji, line 4: The second sentence is definitely fluff, but I feel like we should keep the first sentence (or something similar). I like having a sentence introducing what Knative is, especially for new users who won't already know this. Other than that, LGTM.

snneji (changes requested):
Just one comment

PR #4566, opened 2021-12-13
Remove old archive folder

snneji:
Removes archive folder containing old version (0.23). The contents folder is no longer on the site, but was causing tests to fail.


psschwei (approved):
/lgtm

PR #4572, opened 2021-12-14
Fix links community docs

snneji:
Found a few more links that needed updating after moving the community docs.


psschwei (approved):
/lgtm

csantanapr:
Let's add a redirect in netlist this way we cover links outside this git repo

csantanapr:
Also I think we should redirect to /development/ not /docs/ this way changes to any files on community don't need to wait 6 weeks for the next release

PR #4602, opened 2021-12-17
Adds namespace to Role and RoleBinding in ApiServerSource example

gab-satchi:
"Fixes #issue-number" or "Add description of the problem this PR solves":
Following the `Creating an ApiServerSource object` docs leads to a permission issue with the ApiServerSource.

## Proposed Changes 
Adds namespace to the Role and RoleBinding. They would be tied to the default namespace otherwise and the ApiServerSource wouldn't start


snneji:
Hi @gab-satchi, thanks for the PR!

pierDipi (approved):
/lgtm

PR #4608, opened 2021-12-29
Fix bugs: delete indented code block in installing istio

Shunpoco:
## Proposed Changes 
- [In the part of install istio](https://knative.dev/docs/install/serving/installing-istio/#choosing-an-istio-installation), unnecessary back quotes and string `sh` is shown (see the image below).
- This bug is caused by the rule of Markdown. If we indent 4 spaces or more, markdown makes a code block ([ref](https://dosenbachlab.wustl.edu/wiki/markdown-guide/#:~:text=To%20produce%20a%20code%20block,This%20is%20a%20code%20block.&text=A%20code%20block%20continues%20until,the%20end%20of%20the%20article)).
- To solve the bug, I just deleted indents.
- I already checked by my forked [github-pages](https://shunpoco.github.io/docs/install/serving/installing-istio/#choosing-an-istio-installation).

Screenshot from 2021-12-29 22-06-18

### DIFF
--- a/docs/install/serving/installing-istio.md
+++ b/docs/install/serving/installing-istio.md
@@ -51,9 +51,9 @@ Enter the following command to install Istio:
 
 To install Istio without sidecar injection:
 
-    ```sh
-    istioctl install -y
-    ```
+```sh
+istioctl install -y
+```
 
 #### Installing Istio with sidecar injection
 


dprotaso:
/ok-to-test

Oddly it still shows up after dropping the indent - so it might be something else

dprotaso:
I also don't think it's the indentation but something else

snneji:
Hi @Shunpoco, thank you for the PR! 

Looking at the [preview](https://deploy-preview-4608--knative.netlify.app/development/install/serving/installing-istio) it looks like the PR has fixed the issue. 

/lgtm
/approve

dprotaso:
oh my bad - I was looking at the 1.1 version

PR #4614, opened 2022-01-04
Add documentation for freezer to knative docs

jwcesign:
Documents: https://github.com/knative-sandbox/container-freezer/issues/20

## Proposed Changes
Document "freezer service"  knative serving config-deployment page.

### DIFF
--- a/docs/serving/configuration/deployment.md
+++ b/docs/serving/configuration/deployment.md
@@ -52,6 +52,9 @@ data:
   # for the queue proxy sidecar container.
   # If omitted, no value is specified and the system default is used.
   queue-sidecar-ephemeral-storage-limit: "1024Mi"
+  # concurrency-state-endpoint is the endpoint that queue-proxy calls when its traffic drops to zero or
+  # scales up from zero.
+  concurrency-state-endpoint: ""
 ```
 
 ## Configuring progress deadlines
@@ -105,4 +108,26 @@ metadata:
 data:
   # List of repositories for which tag to digest resolving should be skipped
   registries-skipping-tag-resolving: registry.example.com
+```
+
+## Enable container-freezer service
+
+You can configure queue-proxy to pause pods when not in use by enabling the `container-freezer` service. It calls a stand-alone service (via a user-specified endpoint) when a pod's traffic drops to zero or scales up from zero. To enable it, set `concurrency-state-endpoint` to a non-empty value. With this configuration, you can achieve some features like freezing running processes in pods or billing based on the time it takes to process the requests.
+
+Before you configure this, you need to implement the endpoint API. The official implementation is container-freezer. You can install it by following the installation instructions in the [container-freezer README](https://github.com/knative-sandbox/container-freezer).
+
+The following example shows how to enable the container-freezer service. When using `$HOST_IP`, the container-freezer service inserts the appropriate value for each node at runtime:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-deployment
+  namespace: knative-serving
+  labels:
+    serving.knative.dev/release: devel
+  annotations:
+    knative.dev/example-checksum: "fa67b403"
+data:
+  concurrency-state-endpoint: "http://$HOST_IP:9696"
 ```


jwcesign:
cc @julz @psschwei

docs/serving/configuration/deployment.md
```diff
@@ -106,3 +109,27 @@ data:
...
   registries-skipping-tag-resolving: registry.example.com
 ```
+
+## Enable Freezer Service
+
+You can configure queue-proxy to call a endpoint when it's traffic drops to zero or scales up from zero by modifying `concurrency-state-endpoint` ConfigMap setting. With this configuration, you can achieve some features like freezing runing process in user-contianer or charing calculation based on time of handling requests。
+
+Before you configure this, your need to implement this endpoint API. The official implementation is [container-freezer](https://github.com/knative-sandbox/container-freezer).
+
+The following example show how to enable Freezer Service, $HOST_IP is node's IP:
```
psschwei, line 119: We actually want users to use `$HOST_IP` (and not a specific host's IP) here -- the freezer will swap in the appropriate value for each node at runtime

jwcesign: fixde

jwcesign:
cc @julz @psschwei

docs/serving/configuration/deployment.md
```diff
@@ -106,3 +109,27 @@ data:
   # List of repositories for which tag to digest resolving should be skipped
   registries-skipping-tag-resolving: registry.example.com
 ```
+
+## Enable Freezer Service
```
psschwei, line 113:
```suggestion
## Enable Container-Freezer Service
```

docs/serving/configuration/deployment.md
```diff
@@ -106,3 +109,27 @@ data:
...
+  namespace: knative-serving
+  labels:
+    serving.knative.dev/release: devel
+  annotations:
+    knative.dev/example-checksum: "fa67b403"
+data:
+  # concurrency-state-endpoint is the endpoint that queue-proxy calls when its traffic drops to zero or
+  # scales up from zero.
+  concurrency-state-endpoint: "http://$HOST_IP:9696"
+```
```
psschwei, line 135:
```suggestion
data:
  concurrency-state-endpoint: "http://$HOST_IP:9696"
```
```

**Review by psschwei — COMMENTED** [review]
couple of nits, otherwise LGTM

File: docs/serving/configuration/deployment.md [hunk 4]
```diff
@@ -106,3 +109,27 @@ data:
   # List of repositories for which tag to digest resolving should be skipped
   registries-skipping-tag-resolving: registry.example.com
 ```
+
+## Enable Freezer Service
+
+You can configure queue-proxy to call a endpoint when it's traffic drops to zero or scales up from zero by modifying `concurrency-state-endpoint` ConfigMap setting. With this configuration, you can achieve some features like freezing runing process in user-contianer or charing calculation based on time of handling requests。
```
snneji, line 115:
```suggestion
You can configure queue-proxy to call an endpoint when its traffic drops to zero or scales up from zero by modifying the `concurrency-state-endpoint` ConfigMap setting. With this configuration, you can achieve some features like freezing a running process in a user-container or charing calculation based on time of handling requests.
```

snneji: I'm not sure what is meant by: 
> charing calculation based on time of handling requests.

Could you clarify this please?

docs/serving/configuration/deployment.md
```diff
@@ -106,3 +109,27 @@ data:
...
+
+Before you configure this, your need to implement this endpoint API. The official implementation is [container-freezer](https://github.com/knative-sandbox/container-freezer).
```
snneji, line 117:
```suggestion
Before you configure this, you need to implement the endpoint API. The official implementation is [container-freezer](https://github.com/knative-sandbox/container-freezer).
```

docs/serving/configuration/deployment.md
```diff
@@ -106,3 +109,27 @@ data:
...
+
+The following example show how to enable Freezer Service, when using `$HOST_IP`, the freezer service will swap in the appropriate value for each node at runtime:
```
snneji, line 119:
```suggestion
The following example shows how to enable the container-freezer service. When using `$HOST_IP`, the container-freezer service inserts the appropriate value for each node at runtime:
```

docs/serving/configuration/deployment.md
snneji: Does this mean you need to follow the instructions to install the container-freezer daemon in [README](https://github.com/knative-sandbox/container-freezer) for the container-freezer repo?

docs/serving/configuration/deployment.md
snneji: I think this can be lower case: 

```suggestion
## Enable container-freezer service
```

snneji (changes requested):
Thanks for the PR! I've suggested a few typo fixes and clarifications.

docs/serving/configuration/deployment.md
jwcesign: yes, if developer want to use official implementation.

docs/serving/configuration/deployment.md
jwcesign: fixed

docs/serving/configuration/deployment.md
jwcesign: fixed

docs/serving/configuration/deployment.md
jwcesign: fixed, I change it to `billing based on the time it takes to process the request`

docs/serving/configuration/deployment.md
jwcesign: fixed

docs/serving/configuration/deployment.md
jwcesign: fixed

jwcesign:
cc @psschwei @snneji

docs/serving/configuration/deployment.md
```diff
@@ -106,3 +109,25 @@ data:
   # List of repositories for which tag to digest resolving should be skipped
   registries-skipping-tag-resolving: registry.example.com
 ```
+
+## Enable container-freezer service
+
+You can configure queue-proxy to call an endpoint when its traffic drops to zero or scales up from zero by modifying the `concurrency-state-endpoint` ConfigMap setting. With this configuration, you can achieve some features like freezing a running process in a user-container or billing based on the time it takes to process the request.
```
psschwei, line 115: On a second read-through, I think it might be clearer to readers to start off with what the freezer does and then get into how it does it. Something along the lines of this (though feel free to tweak as needed, it probably can be better than what I came up with):

```suggestion
You can configure queue-proxy to pause pods when not in use by enabling the `container-freezer` service. It calls a stand-alone service (via a user-specified endpoint) when a pod's traffic drops to zero or scales up from zero. To enable it, set `concurrency-state-endpoint` to a non-empty value. With this configuration, you can achieve some features like freezing a running process in a user-container or billing based on the time it takes to process the request.
```

jwcesign:
cc @psschwei

docs/serving/configuration/deployment.md
```diff
@@ -106,3 +109,25 @@ data:
   # List of repositories for which tag to digest resolving should be skipped
   registries-skipping-tag-resolving: registry.example.com
 ```
+
+## Enable container-freezer service
+
+You can configure queue-proxy to pause pods when not in use by enabling the `container-freezer` service. It calls a stand-alone service (via a user-specified endpoint) when a pod's traffic drops to zero or scales up from zero. To enable it, set `concurrency-state-endpoint` to a non-empty value. With this configuration, you can achieve some features like freezing running processes in pods or billing based on the time it takes to process the requests.
+
+Before you configure this, you need to implement the endpoint API. The official implementation is [container-freezer](https://github.com/knative-sandbox/container-freezer).
```
snneji, line 117:
```suggestion
Before you configure this, you need to implement the endpoint API by following the installation instructions in the [container-freezer README](https://github.com/knative-sandbox/container-freezer).
```

snneji (changes requested):
Just one more small clarification. Otherwise LGTM. 🙂

docs/serving/configuration/deployment.md
psschwei: Technically, and this is very much a nit, users do have the option of implementing their own service (as long as it could handle the endpoint calls) and use that instead of the container-freezer. So the freezer isn't strictly necessary, although it's probably what we want to recommend. Not sure exactly the best way to phrase that for the docs though.

snneji: That wording sounds fine to me, but I would keep the first sentence the same as it was before because we prefer the to use active voice so it is clear who is implementing the endpoint API. Also tweak the location of the link:

```suggestion
Before you configure this, you need to implement the endpoint API. The official implementation is container-freezer. You can install it by following the installation instructions in the [container-freezer README](https://github.com/knative-sandbox/container-freezer).
```

snneji:
Thanks for the PR @jwcesign! 

/lgtm
/approve

snneji (approved):

PR #4623, opened 2022-01-10
Build client docs

### LINKED ISSUE #4430: Add kn client commands to the docs
This came up in [Slack](https://knative.slack.com/archives/C9CV04DNJ/p1635978757051300) today -- essentially, it'd be good to get the kn commands (i.e. [this info](https://github.com/knative/client/blob/main/docs/cmd/kn.md) ) into the docs.

Probably related to #2810

psschwei:
## Proposed Changes 

Add documentation for the `kn` client to the docs. 

Fixes #4430


psschwei:
Update: looks like I need to make sure I'm using go1.16 for all these builds...

psschwei:
#4624 fixes the Verify failure

config/nav.yml
```diff
@@ -223,6 +223,101 @@ nav:
...
+            - kn service update: reference/client/kn_service_update.md
+          - kn source:
+            - kn source: reference/client/kn_source.md
+            - kn source create: reference/client/kn_source_apiserver_create.md
+            - kn source delete: reference/client/kn_source_apiserver_delete.md
+            - kn source describe: reference/client/kn_source_apiserver_describe.md
+            - kn source apiserver:
+              - kn source apiserver: reference/client/kn_source_apiserver.md
+              - kn source apiserver list: reference/client/kn_source_apiserver_list.md
+              - kn source apiserver update: reference/client/kn_source_apiserver_update.md
```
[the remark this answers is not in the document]
snneji: APIServerSource create, delete, and describe links need to be moved down under the `kn source apiserver` section.

config/nav.yml
```diff
@@ -223,6 +223,101 @@ nav:
...
+              - kn source binding: reference/client/kn_source_binding.md
```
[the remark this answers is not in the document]
snneji: Create a section for `kn source binding`

snneji (changes requested):
Thanks for adding this! Just a few items in the nav need to be tweaked.

snneji:
Is it possible to have some of these on the same page? For example, grouping all `kn broker` commands on one page with sections for create, delete, describe, list. Maybe doing the same for all of the nested groups of commands?

I don't know if this is possible since all the files in the client repo are on separate pages.

psschwei:
> I'm wondering if it is possible to have some of these on the same page? For example, grouping all `kn broker` commands on one page with sections for create, delete, describe, list. Maybe doing the same for all of the nested groups of commands like `kn channel`, `kn source` etc?
> 
> I don't know if this is possible since all the files in the client repo are on separate pages.

@rhuss can probably answer better than I can, but my understanding is that the change would need to be made on the source files, unless we wanted to try concatenating them in the build script (though I'd really rather not)

psschwei:
Discussed in WG: using a single page that links to the client/docs/cmd/kn.md in the reference section (with a variable for the branch to be filled when building the site)

psschwei:
Linking to Github proved to be a lot easier than I thought... @snneji let me know if the page I added looks ok or if the text should be tweaked a bit.

rhuss:
The preview referenced above does not reflect the recent changes. Do we need to regenerate the preview or is this done automatically ?

psschwei:
Looks like the preview didn't default to the development branch... here's a proper: https://deploy-preview-4623--knative.netlify.app/development/reference/client/

rhuss:
I think this is a first step in the right direction (so happy to merge this), but eventually, I really would love a tighter and more seamless integration that is easier to consume for the reader.

Let me check if I can consolidate the markdown generation into fewer pages and generate a nice drill-down navigation.

rhuss:
FYI, if I select 1.1 in the rendered preview, I don't see a "Client" item in the "Reference" navigation bar (guess this will work only for future versions ?)

psschwei:
> I think this is a first step in the right direction (so happy to merge this), but eventually, I really would love a tighter and more seamless integration that is easier to consume for the reader.

+1

> FYI, if I select 1.1 in the rendered preview, I don't see a "Client" item in the "Reference" navigation bar (guess this will work only for future versions ?)

It needs to get cherry picked into the different release branches, which we can do once we merge.

docs/reference/client/README.md
```diff
@@ -0,0 +1,6 @@
+# Knative Client Reference
+
+Please see the `kn` documentation available at:
+
+[{{ clientdocs() }}]({{ clientdocs() }}){target=_blank}
```
snneji, line 5:
```suggestion
See the [`kn` documentation]({{ clientdocs() }}){target=_blank} in GitHub.
```

snneji: I think it is better for accessibility purposes to have link text instead of a URL. Also it is in our style guide.

snneji: Also, should this link go to the 1.1 branch? At the moment, from the [preview](https://deploy-preview-4623--knative.netlify.app/development/reference/client/) it looks like it's going to main.

snneji:
@rhuss 
> I think this is a first step in the right direction (so happy to merge this), but eventually, I really would love a tighter and more seamless integration that is easier to consume for the reader.
> 
> Let me check if I can consolidate the markdown generation into fewer pages and generate a nice drill-down navigation.

That would be great!

docs/reference/client/README.md
psschwei: > Also, should this link go to the 1.1 branch? At the moment, from the [preview](https://deploy-preview-4623--knative.netlify.app/development/reference/client/) it looks like it's going to main.

It should get updated automatically depending on which branch is being built, similar to how the serving/eventing yaml links in the getting started guide are handled.

snneji: Ah I see now! Since we're in main it'll go to main.

snneji (approved):

Issue #2050, opened 2019-12-13
Remove Ambassador from docs in 0.12


mattmoor:
It is unclear whether this integration still works, and we need to land e2e testing for this integration to remain.  We shouldn't just remove these eagerly, but if it looks like we won't have them in 0.12, then we should remove the docs to avoid them becoming a hazard for users.

See also: https://github.com/datawire/ambassador/issues/1871

samodell:
@mattmoor  can you confirm if the e2e testing has landed and these docs are good to stay, or if we should take them out?

mattmoor:
I don't see a tab, though I believe some of the pieces have landed.  @concaf where are we on getting this into testgrid.knative.dev?

concaf:
@samodell @mattmoor e2e tests have landed, I'm now looking at what the next steps are, will be pinging on Slack for the same.

Tests have landed at https://github.com/knative/serving/pull/6268 and https://github.com/knative/test-infra/pull/1598. I've updated the docs at https://github.com/knative/docs/pull/1893 as well. Will be updating to the latest Ambassador which was released last week.

https://github.com/datawire/ambassador/issues/1871 has been fixed as well, so instructions work fine without any panic.

abrennan89:
@concaf can this issue be closed or is there any additional docs work to do? Thanks :slightly_smiling_face:

concaf:
@abrennan89 the docs are fine, there are tests as well which are not perfect but being worked on as and when time permits - I'm OK with closing this issue.

thecodacus:
Issue is still happening if installed via helm chart

abrennan89:
> Issue is still happening if installed via helm chart

What issue?

thecodacus:
> > Issue is still happening if installed via helm chart
> 
> What issue?

[datawire/ambassador#1871](https://github.com/datawire/ambassador/issues/1871)

if I install Ambassador via helm chart, its throwing same error
Found a temporary solution I have added as comment [datawire/ambassador/issues/1871#issuecomment-669095130](https://github.com/datawire/ambassador/issues/1871#issuecomment-669095130l)

PR #4670, opened 2022-01-26
Remove unsupported Ambassador networking layer from docs

psschwei:
Per this [Slack thread](https://knative.slack.com/archives/C0186KU7STW/p1637152982231400), we should probably remove Ambassador as a supported networking layer. 

See also #2050

### DIFF
--- a/docs/install/operator/knative-with-operators.md
+++ b/docs/install/operator/knative-with-operators.md
@@ -303,69 +303,6 @@ Knative Serving with different ingresses:
 
         Save this for configuring DNS later.
 
-=== "Ambassador"
-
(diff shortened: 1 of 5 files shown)


csantanapr:
Thank you @psschwei , I thought we already removed Ambassador before

alexanderGalushka:
@psschwei @csantanapr I've been using emissary ingress with knative for almost a year with no issues (except for gRPC), is the decision to drop support for emissary final? I don't see any release notes associated with dropping emissary ingress support.

csantanapr:
We can't document something we don't have tests and can't support. We will not remove any current code on purpose.

PR #4683, opened 2022-01-31
More explicit net-istio-controller controller registry override in Operator CR

xfu83:
There's a pitfall when following the doc to use Operator CR to override container registry. 
In the CR, when directly using `net-istio-controller` as the key, it somehow falls back to the knative serving `controller` image without any warning, thus we got `IngressNotConfigured` error when starting an ksvc.

There has been some discussions about it but the doc is still somewhat misleading.
- https://github.com/knative/operator/issues/519#issuecomment-827447017
- https://github.com/knative/serving/issues/12413#issuecomment-989966245
- https://github.com/knative/operator/issues/860

Therefore, it would be better to use explicit `net-istio-controller/controller` instead.

## Proposed Changes 

- Replace `net-istio-controller` with `net-istio-controller/controller`


houshengbo (commented):
/lgtm
/approve

snneji:
Thanks for the PR @xfu83! We're merging it now. 🙂

PR #4695, opened 2022-02-03
[main] update actions main

pierDipi:
As per title + fix security workflow

Replaces https://github.com/knative/docs/pull/4686
Fixes https://github.com/knative/docs/pull/4686


snneji:
Thanks for helping on this @pierDipi . Is this still a work in progress?

pierDipi:
@snneji no, it's good to go

PR #4708, opened 2022-02-07
adding PVC and fixing TOC table

csantanapr:
Can't "PVC" in search, also the TOC (Table of Content) was empty the flags where not showing up.

**Before**
image

**After**
image

### DIFF
--- a/docs/serving/configuration/feature-flags.md
+++ b/docs/serving/configuration/feature-flags.md
@@ -78,8 +78,6 @@ Beta stage
 GA stage
 : The feature is allowed by default.
 
-# Available Flags
-
 ## Multiple containers
 
 * **Type**: Feature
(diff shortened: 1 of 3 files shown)


docs/serving/configuration/feature-flags.md
```diff
@@ -78,8 +78,6 @@ Beta stage
 GA stage
 : The feature is allowed by default.
 
-# Available Flags
```
snneji, line 81: Maybe we should keep this heading but change it to an H2. The heading provides separation between the information about how flags are configured, and the flags you can use. 

The "Available flags" heading should be an H2 header and the flags such as "Multiple containers" under it should be H3 headings.

snneji (dismissed):

docs/serving/configuration/feature-flags.md
csantanapr: made the suggested change

csantanapr:
this is how it looks now using H3
@snneji PTAL

csantanapr:
image

snneji (commented):
LGTM, but one small issue - should the final commit be in another PR? There is a change to the `extra.css` file on this PR that maybe shouldn't be here.

csantanapr:
Yes let remove the CSS file 👍

csantanapr:
@snneji I removed the exta.css file, PTAL

PR #4742, opened 2022-02-15
Add a note about otel and prometheus being mutually exclusive

upodroid:
## Proposed Changes 

- Added a warning to the docs about a recent [query](https://knative.slack.com/archives/C964EFQAW/p1644776122603229). 

/cc @skonto

### DIFF
--- a/docs/snippets/collecting-metrics.md
+++ b/docs/snippets/collecting-metrics.md
@@ -9,6 +9,9 @@ Knative supports different popular tools for collecting metrics:
 
 You can also set up the OpenTelemetry Collector to receive metrics from Knative components and distribute them to other metrics providers that support OpenTelemetry.
 
+    !!! warning
+        You can't use OpenTelemetry Collector and Prometheus at the same time. The default metrics backend is Prometheus. You will need to remove `metrics.backend-destination` and `metrics.request-metrics-backend-destination` keys from the config-observability Configmap to enable Prometheus metrics.
+
 ## About Prometheus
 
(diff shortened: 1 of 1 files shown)


docs/snippets/collecting-metrics.md
```diff
@@ -9,6 +9,9 @@ Knative supports different popular tools for collecting metrics:
 
 You can also set up the OpenTelemetry Collector to receive metrics from Knative components and distribute them to other metrics providers that support OpenTelemetry.
 
+    !!! warning
+        You can't use OpenTelemetry Collector and Prometheus at the same time. The default metrics backend is Prometheus. You will need to remove `metrics.backend-destination` and `metrics.request-metrics-backend-destination` keys from the config-observability Configmap to enable Prometheus metrics.
```
csantanapr, line 13:
```suggestion
!!! warning
    You can't use OpenTelemetry Collector and Prometheus at the same time. The default metrics backend is Prometheus. You will need to remove `metrics.backend-destination` and `metrics.request-metrics-backend-destination` keys from the config-observability Configmap to enable Prometheus metrics.
```

csantanapr: The indentation is off and doesn't render correctly
image

csantanapr (changes requested):

docs/snippets/collecting-metrics.md
csantanapr: With the suggestion it will look like this
image

PR #4746, opened 2022-02-15
fix: first-service.md uses `bash` rather than `yaml` for YAML

starpit:
## Proposed Changes 

- This PR fixes a blockquote that erroneously specifies a `bash` as the language, when the content is `yaml`.


csantanapr:
Good catch

PR #4772, opened 2022-02-18
sync main with default home page branch

csantanapr:
Trying to fix the recent discrepancy between main branch and default home page (release-1.2) that might be affecting cherry-picking from main branch


csantanapr:
@gabo1208 
I don't get why the github actions are failing with SVG images, they should be ignoring *.svg but the github action in `main` branch. I don't think github actions from release-1.2 are the ones running?

csantanapr:
@dprotaso would you have an idea why would the github be failing, is it running the action from the release-1.2 branch, and running the version before getting updated with the ones in this PR?

csantanapr:
I checked it looks like we need to ignore svg for more checks https://github.com/knative-sandbox/.github/pull/179

gabo1208 (approved):

Issue #4716, opened 2022-02-08
Update client docs on current path-lookup behavior


psschwei:
We need to update the [client configuration doc](https://github.com/knative/docs/blob/main/docs/install/client/configure-kn.md?plain=1#L25) to include the proper behavior of the `path-lookup` option: 
* default value is `true`
* is deprecated and will be removed in the future

[source](https://github.com/knative/client/blob/8a5e349f97fa0bd06e237284d6afdcfb35a615d6/docs/README.md?plain=1#L80-L81)

psschwei:
Fixed via #4793

PR #4793, opened 2022-03-01
Updating client configuration details

psschwei:
## Proposed Changes 

As discussed in WG in Feb, updating the client configuration details to note that `path-lookup` is enabled by default. Also some minor syncing between the docs text and what's in https://github.com/knative/client/tree/main/docs#options


psschwei:
This fixed #4716

PR #4798, opened 2022-03-02
fix comma on cncf blog

csantanapr:
fix comma on cncf blog


csantanapr:
/assign @psschwei @lance 
addressed feedback from @psschwei

PR #4826, opened 2022-03-15
update docker image with extra python dependencies

dprotaso:
This PR sets up the correct dependencies for the upcoming mkdocs macro updates


psschwei (approved):
hack script builds the image just fine
/lgtm
/approve

PR #4831, opened 2022-03-15
update macro to lookup point releases for artifacts

dprotaso:
Note - the way the build script works now we'll need to cherry pick this change to older release branches

We should also confirm the netlify builds use a valid GITHUB_TOKEN so we don't hit any rate limiting

### DIFF
--- a/hack/macros.py
+++ b/hack/macros.py
@@ -65,7 +65,7 @@ def define_env(env):
         if version == None:
             return f'https://storage.googleapis.com/knative-nightly/{repo}/latest/{file}'
 
-        version = version.removeprefix('v')
+        version = version.removeprefix('knative-v')
 
         try:
             v = semver.VersionInfo.parse(version)
--- a/hack/macros.py
+++ b/hack/macros.py
@@ -14,7 +14,7 @@ class GithubReleases:
         self.tags_for_repo = {}
         self.client = Github(os.getenv("GITHUB_TOKEN"))
 
-    def latest_release(self, major_minor, org, repo):
+    def get_latest(self, major_minor, org, repo):
         key = f'{org}/{repo}'
 
         if key not in self.tags_for_repo:
@@ -34,7 +34,7 @@ class GithubReleases:
             return None
 
 def define_env(env):
-    g = GithubReleases()
+    releases = GithubReleases()
 
     @env.macro
     def feature(alpha="", beta="", stable=""):
@@ -69,7 +69,7 @@ def define_env(env):
 
         try:
             v = semver.VersionInfo.parse(version)
-            latest_version = g.latest_release(f'{v.major}.{v.minor}', org, repo)
+            latest_version = releases.get_latest(f'{v.major}.{v.minor}', org, repo)
 
             if latest_version is None:
                 print_to_stdout(f'repo "{org}/{repo}" has no tags using latest release for file "{file}"')
--- a/hack/macros.py
+++ b/hack/macros.py
@@ -12,14 +12,14 @@ def print_to_stdout(*vargs):
 class GithubReleases:
     def __init__(self):
         self.tags_for_repo = {}
-        self.g = Github(os.getenv("GITHUB_TOKEN"))
+        self.client = Github(os.getenv("GITHUB_TOKEN"))
 
     def latest_release(self, major_minor, org, repo):
         key = f'{org}/{repo}'
 
         if key not in self.tags_for_repo:
             tags = []
-            for release in self.g.get_repo(key, lazy=True).get_releases():
+            for release in self.client.get_repo(key, lazy=True).get_releases():
                 tags.append(release.tag_name)
             tags = map(lambda tag: tag.removeprefix("knative-v"), tags)
             self.tags_for_repo[key] = list(tags)
--- a/hack/macros.py
+++ b/hack/macros.py
@@ -6,8 +6,8 @@ import traceback
 from github import Github
 
 # By default mkdocs swallows print() messages from macros
-def print_to_stdout(*a):
-    print(*a, file = sys.stdout)
+def print_to_stdout(*vargs):
+    print(*vargs, file = sys.stdout)
 
 class GithubReleases:
     def __init__(self):
--- a/hack/macros.py
+++ b/hack/macros.py
@@ -65,7 +65,8 @@ def define_env(env):
         if version == None:
             return f'https://storage.googleapis.com/knative-nightly/{repo}/latest/{file}'
 
-        version = version.removeprefix('knative-v')
+        version = version.removeprefix('knative-')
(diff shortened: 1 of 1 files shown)


dprotaso:
/hold 

Looking for reviews

cc @psschwei @snneji for docs
cc @kvmware for python
cc @csantanapr since we need to make sure the netlify builds have a `GITHUB_TOKEN` so we aren't rate limited

hack/macros.py
```diff
@@ -25,24 +59,32 @@ def artifact(repo, file, org="knative"):
...
+                return f'https://github.com/{org}/{repo}/releases/latest/download/{file}'
+            elif version.startswith("1."):
+                return f'https://github.com/{org}/{repo}/releases/download/knative-v{latest_version}/{file}'
             else:
-                return 'https://github.com/{org}/{repo}/releases/download/{version}/{file}'.format(
-                    repo=repo,
-                    file=file,
-                    version=version,
-                    org=org)
+                return f'https://github.com/{org}/{repo}/releases/download/{latest_version}/{file}'
```
psschwei, line 80: As of v1.3, we're only supporting docs version of the pattern `knative-v{version}`. Should we remove this block?

psschwei (commented):
I tested this by making a couple of edits locally:

L65:

```python
if version == None:
      #return f'https://storage.googleapis.com/knative-nightly/{repo}/latest/{file}'
      version = "v1.3.0"
version = version.removeprefix('v')
```

and L77:

```python
elif version.startswith("1."):
     print(f'https://github.com/{org}/{repo}/releases/download/knative-v{latest_version}/{file}')
     return f'https://github.com/{org}/{repo}/releases/download/knative-v{latest_version}/{file}'
```

When running `hack/build/sh`, building the main branch had a stdout that showed: 

```
...
https://github.com/knative-sandbox/eventing-kafka-broker/releases/download/knative-v1.3.1/eventing-kafka-controller.yaml
https://github.com/knative-sandbox/eventing-kafka-broker/releases/download/knative-v1.3.1/eventing-kafka-sink.yaml
https://github.com/knative-sandbox/eventing-kafka-broker/releases/download/knative-v1.3.1/eventing-kafka-controller.yaml
https://github.com/knative-sandbox/eventing-kafka-broker/releases/download/knative-v1.3.1/eventing-kafka-source.yaml
https://github.com/knative/eventing/releases/download/knative-v1.3.1/eventing-sugar-controller.yaml
https://github.com/knative/serving/releases/download/knative-v1.3.0/serving-hpa.yaml
https://github.com/knative/net-certmanager/releases/download/knative-v1.3.0/release.yaml
https://github.com/knative/net-http01/releases/download/knative-v1.3.0/release.yaml
https://github.com/knative/net-contour/releases/download/knative-v1.3.0/net-contour.yaml
https://github.com/knative/net-contour/releases/download/knative-v1.3.0/contour.yaml
https://github.com/knative/net-istio/releases/download/knative-v1.3.0/net-istio.yaml
https://github.com/knative/net-istio/releases/download/knative-v1.3.0/istio.yaml
https://github.com/knative/net-kourier/releases/download/knative-v1.3.0/kourier.yaml
https://github.com/knative-sandbox/eventing-kafka-broker/releases/download/knative-v1.3.1/eventing-kafka-sink.yaml
https://github.com/knative-sandbox/eventing-kafka-broker/releases/download/knative-v1.3.1/eventing-kafka-controller.yaml
...
```

I did a quick spot check of the most recent releases on some of the repos, and they match what  shows in Github (for example, `net-istio` is on v1.3.0 whereas `eventing-kafka-broker` is on v1.3.0).

/lgtm
/hold (for other reviewers)

hack/macros.py
```diff
@@ -1,6 +1,40 @@
 import os
+import semver
+import sys
+import traceback
+
+from github import Github
+
+# By default mkdocs swallows print() messages from macros
+def print_to_stdout(*a):
```
kvmware, line 9: usually *x should be *vargs and **kwargs for **

hack/macros.py
```diff
@@ -1,6 +1,40 @@
...
+    print(*a, file = sys.stdout)
+
+class GithubReleases:
+    def __init__(self):
+        self.tags_for_repo = {}
+        self.g = Github(os.getenv("GITHUB_TOKEN"))
```
kvmware, line 15: self.token or self.github_token would be more descriptive

hack/macros.py
```diff
@@ -1,6 +1,40 @@
...
+
+        if key not in self.tags_for_repo:
+            tags = []
+            for release in self.g.get_repo(key, lazy=True).get_releases():
+                tags.append(release.tag_name)
+            tags = map(lambda tag: tag.removeprefix("knative-v"), tags)
+            self.tags_for_repo[key] = list(tags)
+
+        tags = self.tags_for_repo[key].copy()
+        tags = list(filter(lambda tag: tag.startswith(major_minor), tags))
```
kvmware, line 28: https://www.askpython.com/python/dictionary/sort-a-dictionary-in-python might want to just keep the sorted values in the dict itself. Before 3.6 this was done with a sorted dict type, but 3.6+ dicts are sorted types and thus remain sorted.

hack/macros.py
```diff
@@ -25,24 +59,32 @@ def artifact(repo, file, org="knative"):
...
+
+        version = version.removeprefix('v')
+
+        try:
+            v = semver.VersionInfo.parse(version)
+            latest_version = g.latest_release(f'{v.major}.{v.minor}', org, repo)
+
+            if latest_version is None:
+                print_to_stdout(f'repo "{org}/{repo}" has no tags using latest release for file "{file}"')
+                return f'https://github.com/{org}/{repo}/releases/latest/download/{file}'
```
kvmware, line 76: shouldn't this raise an error?

hack/macros.py
dprotaso: Good point - dropped

hack/macros.py
dprotaso: done

hack/macros.py
dprotaso: named this `self.client`

hack/macros.py
dprotaso: we link some third party repos `vmware-tanzu/sources` and `google/knative-gcp` - we don't control the releases so it's best effort

kvmware:
LGTM, using the sorted dict is optional

dprotaso:
I thought of a bug so don't merge this yet

hack/macros.py
kvmware: One can sort by value as well, depends on lambda you pass along (https://datagy.io/python-sort-a-dictionary-by-values/), but I think since we are not really taking advantage of an ordered dict it is not worth the trouble and can be skipped.

PR #4841, opened 2022-03-17
netlify - use python 3.9

dprotaso:


dprotaso:
looks like the build image doesn't have python 3.9

dprotaso:
python 3.9 is not available
https://github.com/netlify/build-image/blob/focal/included_software.md

Issue #4855, opened 2022-03-20
`ingress.class` or `ingress-class` in config-network config map


Legion2:
The "Install a networking layer" documentation describes how to set the correct ingress class in the config-network config map for the different network options (kourier, istio and contour). However, the documentation is inconsistent:
kourier: 
```
'{"data":{"ingress.class":"kourier.ingress.networking.knative.dev"}}'
```
contour:
```
'{"data":{"ingress-class":"contour.ingress.networking.knative.dev"}}'
```
`ingress.class` vs `ingress-class`, what is the correct key?

**Describe the change you'd like to see**
The documentation should be consistent. If both options are valid, choose one for the documentation. If not both options are valid, replace the invalid option with the correct one.

May also add this to the istio section, to make the default value explicit.

nak3:
Thank you for your report.
https://github.com/knative/docs/pull/4858 and https://github.com/knative/docs/pull/4859 will fix it.

PR #4857, opened 2022-03-21
only build main branch for deploy preview

psschwei:
Signed-off-by: Paul S. Schweigert <paulschw@us.ibm.com>

Sets `BUILD_VERSIONS="no"` for deploy previews to only build `main` branch and not the various release versions

Fixes #4525 
(see also [slack discussion](https://knative.slack.com/archives/C9CV04DNJ/p1647874975497899))

PR #4858, opened 2022-03-22
Replace `ingress.class` with `ingress-class`

nak3:
As per title, this patch replace `ingress.class` with `ingress-class`.

The style was updated by https://github.com/knative/networking/pull/522. The old `ingress.class` is still available but it should be consistent in the docs.
Note, only Kong uses `ingress.class` in the code tree so I leave it as it is after opening https://github.com/Kong/kubernetes-ingress-controller/issues/2344.

Fix https://github.com/knative/docs/issues/4855


psschwei:
Using the preview to test the changes made in #4857 , should only take a few minutes...
/hold

psschwei:
Dev preview working as expected!
/hold cancel

psschwei (commented):
/lgtm
/approve

PR #4865, opened 2022-03-22
Update Node.js code samples w/ latest CloudEvents

lance:
This commit updates two of the examples in this repository that use the Node.js CloudEvents module. A newer version of the module has been released and the examples have been updated to use it. The newer version allows us to eliminate a dependency on axios, and simplify the example code.

IMO there is no need for a backport, as the existing code still works. This is just clearer.

Signed-off-by: Lance Ball <lball@redhat.com>


code-samples/eventing/writing-event-source-easy-way/index.js
```diff
@@ -49,6 +38,6 @@ function registerGracefulExit() {
 
 function logExit(message = 'Exiting...') {
   // Handle graceful exit
-  console.log(message);
+  process.stdout.write(`${message}\n`);
```
csantanapr, line 41: what's wrong with using `consle.log` ?

lance: It's a small thing, but `console.log()` is asynchronous. Using `process.stdout.write()` ensures the message is actually written before exit.

csantanapr (changes requested):
Please update the code in the markdown to match the new code changes here https://github.com/knative/docs/tree/main/code-samples/eventing/writing-event-source-easy-way#example

code-samples/eventing/writing-event-source-easy-way/index.js
csantanapr: sounds good

lance:
> Please update the code in the markdown to match the new code changes here https://github.com/knative/docs/tree/main/code-samples/eventing/writing-event-source-easy-way#example

Good catch - done.

PR #4871, opened 2022-03-25
Update kpa-specific.md

deliahu:
## Proposed Changes

Clarify the docs regarding the window's impact on scale-to-zero (assuming my understanding is correct!)

### DIFF
--- a/docs/serving/autoscaling/kpa-specific.md
+++ b/docs/serving/autoscaling/kpa-specific.md
@@ -24,7 +24,7 @@ Stable mode is used for general operation, while panic mode by default has a muc
 * **Default:** `60s`
 
 !!! note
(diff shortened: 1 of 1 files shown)


docs/serving/autoscaling/kpa-specific.md
```diff
@@ -24,7 +24,7 @@ Stable mode is used for general operation, while panic mode by default has a muc
 * **Default:** `60s`
 
 !!! note
-    During scale down, the last Replica will only be removed after there has not been any traffic to the Revision for the entire duration of the stable window.
+    When scaling to 0 Replicas, the last Replica will only be removed after there has not been any traffic to the Revision for the entire duration of the stable window.
```
snneji, line 27:
```suggestion
    When scaling to zero Replicas, the last Replica will only be removed after there has not been any traffic to the Revision for the entire duration of the stable window.
```

PR #4872, opened 2022-03-26
Pin jinja version

psschwei:

Looks like [Jinja released a new version recently](https://pypi.org/project/Jinja2/#history) and it's breaking the build. It gets pulled in by mkdocs as a dependency, but because we don't directly use it we aren't pinning it's version in [requirements.txt](https://github.com/knative/docs/blob/main/requirements.txt).

This can be fixed by pinning the Jinja2 version to 3.0.3 (which was the latest version up until yesterday), but we'll also need to track upstream to see if/when this gets fixed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,5 @@ mkdocs-redirects>=1.0.3
 mkdocs-rss-plugin>=0.18.0
 pygithub==1.55
 semver==2.13.0
+jinja2==3.0.3
 


psschwei:
/hold
Was fixed in [mkdocs-material v8.2.7](https://github.com/squidfunk/mkdocs-material/releases/tag/8.2.7), so let's upgrade mkdocs instead (leaving this open in case this is an easier fix to get in quickly, in which case just unhold it)

Issue #4440, opened 2021-11-04
Move Eventing getting started page to getting started guide


snneji:
## Context

The Eventing docs has an extra [Eventing getting started](https://knative.dev/docs/eventing/getting-started/) page (location in repo: [`eventing/getting-started.md`](https://github.com/knative/docs/blob/main/docs/eventing/getting-started.md)). 
This page provides a `kubectl` command line based getting started tutorial, whereas the current [getting started](https://knative.dev/docs/getting-started/getting-started-eventing/) guide seems to be GUI based.

## Acceptance criteria

- Move the page to the [Getting started](https://knative.dev/docs/getting-started/getting-started-eventing/) guide. Maybe split into multiple pages.

(update WG meeting Feb 15, 2022) 
- Ensure the tutorial uses the `kn` CLI
- Simplify to filter on something like fruit
- Remove current Eventing tutorial (CloudEventsPlayer).

_(updated after discussing at the Docs and UX working group meeting)_

csantanapr:
I think it should be deleted, and anything useful from it incorporate into the main eventing getting started https://knative.dev/docs/getting-started/getting-started-eventing/

csantanapr:
Let's discuss on wg meeting and decide which option to move forward

abrennan89:
Duplicate page is removed, but there's a follow up action here to remove the cloud player app thing and replace it with a nice simple eventing tutorial that uses kn and maybe just shows how to add a couple of different triggers

PR #4908, opened 2022-04-05
Updated a few more Loadbalancer IPs

snneji:
As part of #4815 we removed the IP address from links and commands because it did not work for minikube. 

This PR fixes a few locations that were missed out of that PR. Should hopefully make this page easier to use until we replace the Eventing tutorial as part of https://github.com/knative/docs/issues/4440.


psschwei (commented):
/lgtm

PR #4914, opened 2022-04-08
removed step 1 of creating service

### LINKED ISSUE #4891: in doc-->eventing-->developer topics-->Event source-->kafka source, there is a error
in the command :cd knative-docs/code-samples/eventing/Kafka/source. there actually exists no directory called source inside kafka.Instead there is a directory called binding where already a file event-display.YAML is present.So there are 3 problems
1. in the instructions It is instructed to make a file but actually not needed.
2. The file in event _display.yaml there is no namespace: default
3. even after adding namespace its not working
Screenshot 2022-04-01 031900
 
plz comment if my issue is valid

Subhasish-Behera:
fixes #4891

### DIFF
--- a/docs/eventing/sources/kafka-source/README.md
+++ b/docs/eventing/sources/kafka-source/README.md
@@ -89,17 +89,7 @@ If you are using Strimzi:
 
 ## Create a Service
 
-1. Clone the sample code GitHub repository, and navigate to the local directory of the repository:
-
-    ```bash
-    git clone -b "{{ branch }}" https://github.com/knative/docs knative-docs
-    ```
-
-    ```bash
-    cd knative-docs/code-samples/eventing/kafka/source
-    ```
-    <!--  does this folder exist? -->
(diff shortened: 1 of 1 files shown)


docs/eventing/sources/kafka-source/README.md
```diff
@@ -116,7 +106,7 @@ If you are using Strimzi:
                image: gcr.io/knative-releases/knative.dev/eventing/cmd/event_display
      ```
 
-1. Apply the YAML file by running the command:
+2. Apply the YAML file by running the command:
```
psschwei, line 109: Markdown / mkdocs should convert the numbers to the correct ones on when the site is published, so we should leave all these as `1. xxxxx` (or update the ones that weren't to `1. xxxxx`

psschwei (changes requested):
Thanks for the PR! See the comment below about numbering...

psschwei (commented):
/lgtm
/approve

PR #4916, opened 2022-04-08
Remove Knative-supported AWS SQS source

lionelvillard:

## Proposed Changes 

- Remove Knative-supported AWS SQS source as it has been deprecated.
-
-


psschwei (commented):
Should we remove the golang redirects too?
https://github.com/knative/docs/blob/8f742c9f4800ab88a086ba21ee814d55b7c4e952/golang/eventing-awssqs.html
https://github.com/knative/docs/blob/d1af870e70fbccd2bf88fdd7a47568e1fe023858/golang/_redirects#L22

lionelvillard:
> Should we remove the golang redirects too? https://github.com/knative/docs/blob/8f742c9f4800ab88a086ba21ee814d55b7c4e952/golang/eventing-awssqs.html
> 
> https://github.com/knative/docs/blob/d1af870e70fbccd2bf88fdd7a47568e1fe023858/golang/_redirects#L22

I don't think we should as this will break existing builds.

psschwei (commented):
/lgtm
/approve

PR #4950, opened 2022-05-03
Change warning to note

snneji:
Small PR to change a warning to a note because I don't think this is important enough to be a warning. 

See this comment thread: https://github.com/knative/docs/pull/4893#discussion_r845964217


psschwei (commented):
/lgtm
/approve

PR #4962, opened 2022-05-16
Bumped the CPU requirements

### LINKED ISSUE #4944: Update minimum CPU recommendation
We currently list the minimum system requirements as 2 CPU / 4 GB RAM in the [prereqs snippet](https://github.com/knative/docs/blob/main/docs/snippets/prerequisites.md?plain=1#L5), but as we've [seen with the quickstart plugin](https://github.com/knative-sandbox/kn-plugin-quickstart/issues/224#issuecomment-1086754010), 2 CPUs is not enough to install both Serving and Eventing. 

We should consider bumping the min recommended number of CPUs to 3. Open question as to whether we should only bump it in the "prototyping" section (see link above) or also update the ["multiple nodes" line in the "production" section](https://github.com/knative/docs/blob/main/docs/snippets/prerequisites.md?plain=1#L14).

hamees-sayed:
Update minimum CPU recommendation
Fix #4944 

## Proposed Changes 

-
-
-

### DIFF
--- a/docs/snippets/prerequisites.md
+++ b/docs/snippets/prerequisites.md
@@ -11,7 +11,7 @@ Before installing Knative, you must meet the following prerequisites:
 - **For production purposes**, it is recommended that:
 
     - If you have only one node in your cluster, you need at least 6&nbsp;CPUs, 6&nbsp;GB of memory, and 30&nbsp;GB of disk storage.
-    - If you have multiple nodes in your cluster, for each node you need at least 3&nbsp;CPUs, 4&nbsp;GB of memory, and 20&nbsp;GB of disk storage.
+    - If you have multiple nodes in your cluster, for each node you need at least 2&nbsp;CPUs, 4&nbsp;GB of memory, and 20&nbsp;GB of disk storage.
 - You have a cluster that uses Kubernetes v1.22 or newer.
 - You have installed the [`kubectl` CLI](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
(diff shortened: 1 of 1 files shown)


docs/snippets/prerequisites.md
```diff
@@ -11,7 +11,7 @@ Before installing Knative, you must meet the following prerequisites:
 - **For production purposes**, it is recommended that:
 
     - If you have only one node in your cluster, you need at least 6&nbsp;CPUs, 6&nbsp;GB of memory, and 30&nbsp;GB of disk storage.
-    - If you have multiple nodes in your cluster, for each node you need at least 2&nbsp;CPUs, 4&nbsp;GB of memory, and 20&nbsp;GB of disk storage.
+    - If you have multiple nodes in your cluster, for each node you need at least 3&nbsp;CPUs, 4&nbsp;GB of memory, and 20&nbsp;GB of disk storage.
```
psschwei, line 14: I think making this update for production too is probably the right choice, but let's confirm with @csantanapr

csantanapr: You can leave it at 2CPUs, since the 2CPUs is per worker and this line is about multiple nodes, so at the minimum cluster will have 4 CPUs in total.

csantanapr: TLDR; don't change this line

psschwei (commented):
Thanks for fixing this!

/lgtm
/approve

PR #4966, opened 2022-05-19
Add per-revision examples to serving config

### LINKED ISSUE #4732: Serving Defaults configmap should document the field names being defaulted.
I noticed that the default entries like `revision-timeout-seconds` here: https://knative.dev/docs/serving/configuration/config-defaults/ to not actually specify the fields being defaulted.

**Additional context**
I really like the way this was solved in autoscaling where it talks about global configuration as well as per-revision: https://knative.dev/docs/serving/autoscaling/scale-bounds/#lower-bound

abrennan89:
Fixes #4732 

## Proposed Changes

- Add per-revision config settings where they exist, or explain that they are N/A
- Clean up formatting

### DIFF
--- a/docs/serving/configuration/config-defaults.md
+++ b/docs/serving/configuration/config-defaults.md
@@ -435,7 +435,7 @@ the pod receives.
 
 {% raw %}
 
-The `container-concurrency-max-limit` is an Operator setting that disables arbitrary large concurrency values, or autoscaling targets, for individual revisions. The `container-concurrency` default setting must be at or below this value. The value of the `container-concurrency-max-limit` setting must be greater than 1.
+The `container-concurrency-max-limit` setting disables arbitrary large concurrency values, or autoscaling targets, for individual revisions. The `container-concurrency` default setting must be at or below this value. The value of the `container-concurrency-max-limit` setting must be greater than 1.
 
 !!! note
     Even with this set, a user can choose a `containerConcurrency` value of zero (unbounded), unless `allow-container-concurrency-zero` is set to `"false"`.
@@ -448,6 +448,17 @@ The `container-concurrency-max-limit` is an Operator setting that disables arbit
 * **Default:** `"1000"`
 
 **Example:**
+
+=== "Global (ConfigMap)"
+    ```yaml
+    apiVersion:  v1
+    kind:  ConfigMap
+    metadata:
+      name:  config-defaults
+      namespace:  knative-serving
+    data:
+      container-concurrency-max-limit: "1000"
+    ```
 
 === "Global (Operator)"
     ```yaml
--- a/docs/serving/configuration/config-defaults.md
+++ b/docs/serving/configuration/config-defaults.md
@@ -68,9 +68,11 @@ The revision timeout value determines the default number of seconds to use for t
       name: helloworld-go
       namespace: default
       spec:
-        timeout: 300
-        containers:
-          - image: gcr.io/knative-samples/helloworld-go
+        template:
+          spec:
+            timeout: 300
+            containers:
+            - image: gcr.io/knative-samples/helloworld-go
     ```
 
 ### Max revision timeout seconds
@@ -81,7 +83,7 @@ If this value is increased, the activator's `terminationGraceTimeSeconds` should
 {% endraw %}
 
 * **Global key:** `max-revision-timeout-seconds`
-* **Per-revision annotation key:** `timeout`
+* **Per-revision annotation key:** N/A
 * **Possible values:** integer
 * **Default:** `"600"` (10 minutes)
 
@@ -96,19 +98,6 @@ If this value is increased, the activator's `terminationGraceTimeSeconds` should
       namespace:  knative-serving
     data:
       max-revision-timeout-seconds: "600"
-    ```
-
-=== "Per Revision"
-    ```yaml
-    apiVersion: serving.knative.dev/v1
-    kind: Service
-    metadata:
-      name: helloworld-go
-      namespace: default
-      spec:
-        timeout: 600
-        containers:
-          - image: gcr.io/knative-samples/helloworld-go
     ```
 
 ### Revision CPU request
--- a/docs/serving/configuration/config-defaults.md
+++ b/docs/serving/configuration/config-defaults.md
@@ -451,13 +451,14 @@ The `container-concurrency-max-limit` is an Operator setting that disables arbit
 
 === "Global (Operator)"
     ```yaml
-    apiVersion: operator.knative.dev/v1alpha1
+    apiVersion: operator.knative.dev/v1beta1
     kind: KnativeServing
     metadata:
       name: knative-serving
+      namespace: knative-serving
     spec:
       config:
-        data:
+        defaults:
           container-concurrency-max-limit: "1000"
     ```
 
--- a/docs/serving/configuration/config-defaults.md
+++ b/docs/serving/configuration/config-defaults.md
@@ -37,11 +37,9 @@ See below for a description of each property.
 
 ## Properties
 
-### Revision Timeout Seconds
+### Revision timeout seconds
 {% raw %}
-revision-timeout-seconds contains the default number of
-seconds to use for the revision's per-request timeout, if
-none is specified.
+The `revision-timeout-seconds` value determines the default number of seconds to use for the revision's per-request timeout if none is specified.
 {% endraw %}
 
 * **Global key:** `revision-timeout-seconds`
@@ -50,6 +48,17 @@ none is specified.
 * **Default:** `"300"` (5 minutes)
 
 **Example:**
+
+=== "Global (ConfigMap)"
+    ```yaml
+    apiVersion:  v1
+    kind:  ConfigMap
+    metadata:
+      name:  config-defaults
+      namespace:  knative-serving
+    data:
+      revision-timeout-seconds: "300"
+    ```
 
 === "Per Revision"
     ```yaml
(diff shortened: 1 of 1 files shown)


docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/revision-timeout-seconds: "300"
+        spec:
+          containers:
+            - image: gcr.io/knative-samples/helloworld-go
```
dprotaso, line 77: This annotation doesn't exist - the config map property configures the default for `Spec.Timeout` on the `Revision`

```suggestion
        spec:
          timeout: 300
          containers:
            - image: gcr.io/knative-samples/helloworld-go
```

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/max-revision-timeout-seconds: "600"
```
dprotaso, line 116: Likewise this isn't an annotation but an upper bound on `Spec.Timeout` on the `Revision.Spec`

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/revision-cpu-request: "400m"
```
dprotaso, line 156: This annotation doesn't exist - the config map property configures the  default value for  `Spec.Containers[*].Requests.CPU` of a `Revision`

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/revision-memory-request: "400m"
```
dprotaso, line 197: Likewise this annotation doesn't exist - the config map property configures the  default value for `Spec.Containers[*].Requests.Memory` of a `Revision`

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/revision-ephemeral-storage-request: "500M"
```
dprotaso, line 238: Likewise this annotation doesn't exist - the config map property configures the  default value for `Spec.Containers[*].Requests.EphemeralStorage` of a `Revision`

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/revision-cpu-limit: "1000m"
```
dprotaso, line 278: Likewise this annotation doesn't exist - the config map property configures the  default value for `Spec.Containers[*].Limits.CPU` of a `Revision`

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/revision-memory-limit: "200M"
```
dprotaso, line 318: Likewise this annotation doesn't exist - the config map property configures the   default value for `Spec.Containers[*].Limits.Memory` of a `Revision`

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/revision-ephemeral-storage-limit: "750M"
```
dprotaso, line 358: Likewise this annotation doesn't exist - the config map property configures the default value for `Spec.Containers[*].Limits.EphemeralStorage` of a `Revision`

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/container-name-template: "user-container"
```
dprotaso, line 399: This annotation doesn't exist - users can set the `Spec.Containers[*].Name` attribute on the `Revision` explicitly

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    metadata:
+      name: helloworld-go
+      namespace: default
+    spec:
+      template:
+        metadata:
+          annotations:
+            serving.knative.dev/container-name-template: "0"
```
dprotaso, line 440: typo in the annotation - but it doesn't exist

The container-concurrency annotation maps to `Spec.ContainerConcurrency` on the `Revision`

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
-**Key**: `container-concurrency-max-limit`
-
-**Default**: `"1000"`
+* **Global key:** `container-concurrency-max-limit`
+* **Per-revision annotation key:** N/A
+* **Possible values:** integer
+* **Default:** `"1000"`
 
 **Example:**
-```yaml
```
dprotaso, line 292: Why drop the config map example?

docs/serving/configuration/config-defaults.md
```diff
@@ -30,319 +32,497 @@ data:
...
+=== "Global (Operator)"
+    ```yaml
+    apiVersion: operator.knative.dev/v1alpha1
+    kind: KnativeServing
+    metadata:
+      name: knative-serving
+    spec:
+      config:
+        data:
+          container-concurrency-max-limit: "1000"
```
dprotaso, line 473: cc @houshengbo can you confirm how to configure this for the operator

docs/serving/configuration/config-defaults.md
abrennan89: Can you give an example of this in the YAML or can it just not be set for individual revisions?
I don't have examples for a lot of these, so just had to work with the assumption of annotations as placeholders for documenting the revisions bit 🙂

docs/serving/configuration/config-defaults.md
abrennan89: see previous comments

docs/serving/configuration/config-defaults.md
abrennan89: see previous comments

docs/serving/configuration/config-defaults.md
abrennan89: see previous comments

abrennan89:
@dprotaso can you give examples for the rest of these as you have for `timeout` please? I have no frame of reference for some of them.

docs/serving/configuration/config-defaults.md
houshengbo: This is a key in the CM called config-defaults, so it should be 
```
    apiVersion: operator.knative.dev/v1beta1
    kind: KnativeServing
    metadata:
      name: knative-serving
      namespace: knative-serving
    spec:
      config:
        defaults:
          container-concurrency-max-limit: "1000"
```

In addition, add the field `namespace` and change the CR version into v1beta1.

docs/serving/configuration/config-defaults.md
dprotaso: You would update the revision template of a Knative Service as follows:

```yaml
kind: Service
metadata:
  name: helloworld-go
  namespace: default
spec:
  template:
    spec:
      containers:
        - image: gcr.io/knative-samples/helloworld-go
          resources:
            requests:
              cpu: "400m"
```

docs/serving/configuration/config-defaults.md
dprotaso: similar to above but "memory" instead of "cpu" ie.  https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#example-1

docs/serving/configuration/config-defaults.md
dprotaso: similar to above but "ephemeral-storage" instead of "cpu" 

ie.  https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#setting-requests-and-limits-for-local-ephemeral-storage

docs/serving/configuration/config-defaults.md
dprotaso: similar to above

docs/serving/configuration/config-defaults.md
dprotaso: similar to above

docs/serving/configuration/config-defaults.md
dprotaso: similar to above

docs/serving/configuration/config-defaults.md
dprotaso: You can set the name of the container like this

```yaml
kind: Service
metadata:
  name: helloworld-go
  namespace: default
spec:
  template:
    spec:
      containers:
        - name: some-name
          image: gcr.io/knative-samples/helloworld-go
```

docs/serving/configuration/config-defaults.md
dprotaso: If some user is editing these ConfigMaps we have been calling them the 'operator'
> unless this was supposed to mean operator as in user?

dprotaso:
> @dprotaso can you give examples for the rest of these as you have for `timeout` please? I have no frame of reference for some of them.

done

dprotaso (commented):
some minor stuff

dprotaso:
/lgtm
/approve

thanks!

Issue #4700, opened 2022-02-03
gRPC sample does not appear to build


sriumcp:
## Expected Behavior
The `gRPC` knative sample (Docker image) should build. It is documented [here](https://github.com/knative/docs/tree/main/code-samples/serving/grpc-ping-go), which in turn is linked to, from [here](https://knative.dev/docs/samples/serving/).

## Actual Behavior
I get a Docker build error.

## Steps to Reproduce the Problem

1.  `cd knative-docs/code-samples/serving/grpc-ping-go`
2.  `docker build --tag "sriumcp/grpc-ping-go" .`

## Additional Info
The Docker build output looks like this for me.

```shell
[+] Building 24.9s (9/15)                                                                                                  
 => [internal] load build definition from Dockerfile                                                                  0.0s
 => => transferring dockerfile: 1.82kB                                                                                0.0s
 => [internal] load .dockerignore                                                                                     0.0s
 => => transferring context: 2B                                                                                       0.0s
 => [internal] load metadata for gcr.io/distroless/static:latest                                                      0.2s
 => [internal] load metadata for docker.io/library/golang:1.13                                                        0.6s
 => [auth] library/golang:pull token for registry-1.docker.io                                                         0.0s
 => CACHED [builder 1/6] FROM docker.io/library/golang:1.13@sha256:8ebb6d5a48deef738381b56b1d4cd33d99a5d608e0d03c5fe  0.0s
 => CACHED [stage-1 1/3] FROM gcr.io/distroless/static@sha256:8ad6f3ec70dad966479b9fb48da991138c72ba969859098ec689d1  0.0s
 => [internal] load build context                                                                                     0.0s
 => => transferring context: 2.07kB                                                                                   0.0s
 => ERROR [builder 2/6] RUN go get google.golang.org/grpc                                                            24.3s
------                                                                                                                     
 > [builder 2/6] RUN go get google.golang.org/grpc:                                                                        
#8 20.61 # golang.org/x/net/http2
#8 20.61 src/golang.org/x/net/http2/transport.go:417:45: undefined: os.ErrDeadlineExceeded
------
executor failed running [/bin/sh -c go get google.golang.org/grpc]: exit code: 2
```

snneji:
@yanweiguo since you're on call on Knative.party this week, would you be able to help with this issue?

snneji:
@sriumcp is this issue related to PR https://github.com/knative/docs/pull/4701?

sriumcp:
[...]

PR #4967, opened 2022-05-20
Fix Dockerfile for gRPC sample

nak3:
Fix https://github.com/knative/docs/issues/4700

## Proposed Changes

This patch updates Dockerfile for gRPC sample.

Currently the builder image uses old golang so it causes an error as reported by https://github.com/knative/docs/issues/4700.

/cc @sriumcp @snneji @abrennan89


psschwei (commented):
/lgtm
/approve

PR #4971, opened 2022-05-20
Fix eventing spec links

odacremolbap:
## Fixes

- 3 link pointing to the spec docs fixed.
- There is a [slightly related PR at spec](https://github.com/knative/specs/pull/99) that should improve further navigation.


odacremolbap:
Thanks @psschwei 

Yep, no need to merge specs first.

Issue #4727, opened 2022-02-11
kn install snippet suggests both `brew install` and `brew upgrade`


starpit:
## Expected Behavior

I wasn't sure whether to categorize this as a bug or a feature request... 

The documentation suggests that users should try both `brew install kn` *and* `brew upgrade kn`. Whereas `brew install` will check for upgrades by default. From the standpoint of understandability, the current site renders these as two consecutive code blocks, giving the appearance the *both* should be executed.

Screenshot of current knative docs web site:

<img width="631" alt="Screen Shot 2022-02-10 at 7 38 01 PM" src="https://user-images.githubusercontent.com/4741620/153520585-19d44ca4-d0f6-40a7-9b3f-3733625004ce.png">

From the `brew install -h` output:

```
Usage: brew install [options] formula|cask [...]

Install a formula or cask. Additional options specific to a formula may be
appended to the command.

Unless HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK is set, brew upgrade or brew
reinstall will be run for outdated dependents and dependents with broken
linkage, respectively.
```

## Actual Behavior

I think we can reduce clutter and potential for confusion by suggesting only `brew install kn`. Thoughts?

## Steps to Reproduce the Problem

n/a

## Additional Info
**Additional context**

**Install information:**
 - Platform (GKE, IKS, AKS, etc.): any
 - Knative Version: any

snneji:
I can add "Do one of the following" before the command to show that you don't have to do both. 

The second bullet point was added to fix an issue where running `brew install` while having an existing install caused the new version to install as well as leaving the old version available.

starpit:
Howdy! That makes sense. What about putting it under an expandable section? To me, this would make it even more clear that the instal/upgrade dischotomy is a discrete choice to be made?
<img width="723" alt="Screen Shot 2022-02-15 at 12 53 59 PM" src="https://user-images.githubusercontent.com/4741620/154120603-c40d1175-985c-403e-a6da-a93d51942178.png">

snneji:
Hi @starpit, we made the decision to not use a note because notes are easy to ignore. 

Also, we would like users to use the upgrade command instead of the install command in the case of having an existing install. Having the note after the install command might mean that users only read the note _after_ running the install command.

Do you think this suggestion would fix the issue:

[...]

PR #4973, opened 2022-05-20
kn CLI install should use tap instead of compile for brew install

### LINKED ISSUE #4972: kn CLI install should use tap instead of compile for brew install
According to @rhuss users should be using the command `brew install knative/client/kn` this way they get the release binary and not use `brew install kn` which required the user to have all the compile tools and compile from source code getting their own edition of the build.

@rhuss suggested to update the docs to use the preferred method.

csantanapr:
Fixes #4972 #4727


psschwei (commented):
/lgtm
/approve

PR #4980, opened 2022-05-27
Fix values and general cleanup

### LINKED ISSUE #4352: Use general example domain in doc

Some example uses a specific domain for example:

- [Setting up a custom domain](https://knative.dev/docs/serving/using-a-custom-domain/) uses `mydomain.com`.
- [Enabling automatic TLS certificate provisioning](https://knative.dev/docs/serving/using-auto-tls/) uses `myemail@gmail.com`.

The mydomain.com is already used by someone so we should not use it in our doc as an example. (I haven't confirmed but `myemail@gmail.com` might be used by someone.)
Instead, [IANA's reserved domain](https://www.iana.org/domains/reserved) like `example.com`, `example.org` or knative's `knative.dev` should be used.

abrennan89:
Fixes #4352 and general cleanup of sections

Preview links:
- https://deploy-preview-4980--knative.netlify.app/docs/serving/using-auto-tls/
- https://deploy-preview-4980--knative.netlify.app/docs/serving/using-a-custom-domain/

### DIFF
--- a/docs/serving/using-a-custom-domain.md
+++ b/docs/serving/using-a-custom-domain.md
@@ -1,29 +1,32 @@
-# Default domain name settings
+# Configuring domain names
 
-The fully qualified domain name for a route by default is `{route}.{namespace}.{default-domain}`. Knative Serving routes use `example.com` as the default domain.
+You can customize the domain of an individual Knative Service, or set a global default domain for all Services created on a cluster. The fully qualified domain name for a route by default is `{route}.{namespace}.example.com`.
 
-You can change the default domain by modifying the [`config-domain` ConfigMap](https://github.com/knative/serving/blob/main/config/core/configmaps/domain.yaml).
+## Configuring a domain for a single Knative Service
 
-!!! warning
-    Customizing a domain using this ConfigMap affects your cluster globally. If you want to customize the domain of an individual Service, use [`DomainMapping`](services/custom-domains.md) instead.
+If you want to customize the domain of an individual Service, see the documentation about [`DomainMapping`](services/custom-domains.md).
 
-## Procedure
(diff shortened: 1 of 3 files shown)


docs/serving/using-a-custom-domain.md
```diff
@@ -1,29 +1,32 @@
...
 
-     ```bash
-     kubectl edit configmap config-domain -n knative-serving
-     ```
+    ```bash
+    kubectl edit configmap config-domain -n knative-serving
+    ```
 
-1. Edit the file to replace `example.com` with the domain you want to use, then remove the `_example` key and save your changes. In this example, `mydomain.com` is configured as the domain for all routes:
+1. Edit the file to replace `example.com` with the domain you want to use, then remove the `_example` key and save your changes. In this example, `knative.dev` is configured as the domain for all routes:
```
csantanapr, line 21: I think we should use a different domain than `knative.dev` for this page. why `mydomain.com` bad to use?

abrennan89: It was the specific request of the issue https://github.com/knative/docs/issues/4352 - I thought it would be confusing to use changing from `example.com` to `example.org` since they're so similar, so went with `knative.dev` since this was the other option. It shouldn't matter since a user shouldn't be using any of the example values anyway, but we should make a decision here about what we're using, because we end up with back and forth every time I try to fix this issue 😂

psschwei: First thing that popped in my head was something like `customdomain.example.com`, but then adding a subdomain might not be any less confusing...  that said, all else being equal, I think it's probably better to use a domain we own in our docs as opposed to one we don't.

PR #5005, opened 2022-06-06
[release-1.4] Correct the instruction on system resource settings

houshengbo:
## Proposed Changes 

- This is a critical doc issue in 1.4 for knative operator. We need to drop the incorrect system resource configuration for knative operator.


abrennan89 (approved):
/lgtm
/approve

PR #5018, opened 2022-06-08
Remove duplicate getting started, tidy up info

abrennan89:
Fixes https://github.com/knative/docs/issues/4440

## Proposed Changes

- Remove duplicate getting started eventing page (same steps, very different implementation, not much reusable content IMO)
- Move examples to main eventing landing page - these just seem distracting at the start of what is supposed to be a simplified tutorial, and this is more consistent with serving: https://knative.dev/development/serving/#more-samples-and-demos


csantanapr:
We discussed this in Slack
The content needs to be use to replace the tutorial for eventing and the cloud player and can be done in a different PR

/lgtm

PR #5032, opened 2022-06-09
Rename home page buttons

### LINKED ISSUE #4905: Clarify the text on the homepage Developer and Cluster Admin buttons
## Background
The buttons on the homepage currently say "Developer? Start Here!" and "Cluster Administrator? Start Here!". We have received feedback that this might be confusing if a user does not know which persona they are. Also, it does not tell the user where the buttons will take them.

## Task
Rename the buttons to describe were they will take the user: 
- Developer Button → Reword to say something like “Try the tutorial”
- Cluster Administrator Button → Reword to say something like “Install Knative on your cluster”

(Note: the URLs should stay the same.)

For example:
 
Screen Shot 2022-04-05 at 18 34 34

csantanapr:
Fixes #4905

This is a change that we discussed in the WG back on March 22; see #4905 for more details

### DIFF
--- a/overrides/home.html
+++ b/overrides/home.html
@@ -12,10 +12,8 @@
           </h1>
           <h2>Serverless Containers in Kubernetes environments.</h2>
           <p style="display: block">
-            <a href="{{ 'getting-started/' | url }}" class="md-button white md-button--primary">
-              Developer? Start Here!
-            </a>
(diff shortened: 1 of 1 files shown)


snneji:
Thanks @csantanapr 
/lgtm

PR #5073, opened 2022-07-01
Add github token to local docker build

### LINKED ISSUE #4930: Can't run full build locally to test github API rate limit exceeded
We have a script to perform a full build with all versions of the website including the blog, then this build can be served locally to check out any problems.

This is the method that I use to preview the blog edits

When running the command we document
```
./hack/docker/test.sh
```
The user gets errors about github rate limit.

This problem was recently introduced because a GITHUB_TOKEN is now required to do a full build of the website.

We should update the instructions and script to require the env variable GITHUB_TOKEN with the minimal access in the `test.sh` and pass it to the docker container in case someone wants to do a full build locally.

@psschwei 
As for an easy way to preview blog locally, what should we do here?
Maybe create a `blog.sh` or update `run.sh` to add the blog content and we get live preview for blog also

psschwei:

Fixes #4930 

## Proposed Changes 

Updates the local full site build using Docker to include a required github token (provided by user)

/assign @csantanapr

### DIFF
--- a/contribute-to-docs/getting-started/previewing-docs-locally.md
+++ b/contribute-to-docs/getting-started/previewing-docs-locally.md
@@ -37,6 +37,8 @@ The build output is the entire static site located in `./site`.
 
 You can preview the website locally by running a webserver using this directory like `npx http-server site -p 8000` if you have Node.js or `python3 -m http.server 8000` if you have Python 3.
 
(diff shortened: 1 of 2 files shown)


csantanapr:
Thanks, @psschwei I remember discussing this. 
/lgtm

Issue #4881, opened 2022-03-30
Custom Certs for Tag to Digest Resolution needs a passover


dprotaso:
Some feedback on https://knative.dev/docs/serving/tag-resolution/#custom-certificates

We should mention that the controller deployment you are modifying is the one serving provides. We can probably do this with a `kubectl patch` command

The follow copy makes it sound like you can apply the supplied example deployment which isn't valid 

> For example, if you are using a custom-certs secret that contains your CA certificates, the Deployment object is as follows:
seems like you can apply the supplied example 

Additionally, including steps to create the secret could be useful - see this blog post:

https://williamlam.com/2022/03/how-to-configure-knative-and-containerd-in-vmware-event-broker-appliance-veba-to-use-a-custom-registry.html

PR #5075, opened 2022-07-05
Update custom certs docs

abrennan89:
Fixes https://github.com/knative/docs/issues/4881

@dprotaso I took a stab at including the info about secrets but wasn't sure about it so please feel free to add comments or corrections. Thanks.

### DIFF
--- a/docs/serving/tag-resolution.md
+++ b/docs/serving/tag-resolution.md
@@ -8,37 +8,46 @@ helps to provide consistency for Deployments. For more information, see the docu
 
 ## Custom certificates
 
-If you are using a registry that has a self-signed certificate, you must configure the Knative Serving controller to trust that certificate.
+If you are using a registry that has a self-signed certificate, you must configure the default Knative Serving `controller` Deployment to trust that certificate. You can configure trusting certificates by mounting your certificates into the `controller` Deployment, and then setting the environment variable appropriately.
 
-Knative Serving accepts the [`SSL_CERT_FILE` and `SSL_CERT_DIR`](https://golang.org/pkg/crypto/x509/#pkg-overview) environment variables.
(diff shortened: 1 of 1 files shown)


abrennan89:
cc @knative/serving-wg-leads @knative/serving-reviewers

docs/serving/tag-resolution.md
```diff
@@ -8,37 +8,46 @@ helps to provide consistency for Deployments. For more information, see the docu
 
 ## Custom certificates
 
-If you are using a registry that has a self-signed certificate, you must configure the Knative Serving controller to trust that certificate.
+If you are using a registry that has a self-signed certificate, you must configure the default Knative Serving `controller` Deployment to trust that certificate. You can configure trusting certificates by mounting your certificates into the `controller` Deployment, and then setting the environment variable appropriately.
```
skonto, line 11: I think this should be aligned with the [operator's section](https://knative.dev/docs/install/operator/configuring-serving-cr/#ssl-certificate-for-controller.). We should describe both config map and secret based custom certs. The knative operator essentially modifies the controller deployment when the CR contains a custom-certs field and sets `SSL_CERT_DIR`.

abrennan89: Is this required to complete the work requested by https://github.com/knative/docs/issues/4881 though?

I would like to be careful to keep this within scope of what was asked for in that issue vs delving too much into adding additional content or changing docs structure - if there's something missing that isn't part of that original issue, it should be opened as a new issue.

dprotaso: Yeah I'd probably open an separate issue

abrennan89:
@dprotaso PTAL 🙂

PR #5102, opened 2022-07-14
Add the instruction on how to install and configure with the operator CLI plugin

houshengbo:
## Proposed Changes 

- This PR added the instruction on how to install and use the operator CLI plugin to install operator, serving and eventing.

### DIFF
--- a/docs/install/operator/knative-with-operator-cli.md
+++ b/docs/install/operator/knative-with-operator-cli.md
@@ -8,6 +8,8 @@ resources.
 
 ## Install the Knative Operator CLI Plugin
 
+Before you install the Knative Operator CLI Plugin, first install the [Knative CLI](../../client/install-kn.md).
+
 === "MacOS"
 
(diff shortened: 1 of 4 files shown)


docs/install/operator/knative-with-operator-cli.md
```diff
@@ -0,0 +1,220 @@
...
+```bash
+mv kn-operator /usr/local/bin
+```
+
+## Verify the installation of the Knative Operator CLI Plugin
+
+You can run the following command to verify the installation:
+
+```bash
+kn-operator -h
```
psschwei, line 48: In the past, we had generally opted for usage as a plugin to `kn` rather than as a standalone binary in the docs (i.e `kn operator` vs. `kn-operator`).  At least, that was the case when we added quickstart, not sure if that's still the case.

houshengbo: I asked about the same question in the client channel, but no one confirmed that. If we use the plugin based on `kn`, I will change the docs accordingly.

psschwei (commented):
/lgtm
/approve

PR #5113, opened 2022-07-18
Add the instruction on how to configure podDisruptionBudget with the knative operator

houshengbo:
## Proposed Changes 

- Added the instruction on how to configure the podDisruptionBudget with the knative operator.


abrennan89 (commented):
I think this is actually correct as it is - the command here aren't YAML, so it should remain `bash`.

abrennan89:
See the style guide for info: https://github.com/knative/docs/blob/main/contribute-to-docs/style-guide/documenting-code.md#documenting-commands

PR #5121, opened 2022-07-20
🧹 remove defunct markdown checking varibles from scripts

kellyaa:
Two markdown checking variables are no longer valid, removing them from associated scripts.

Addresses: https://github.com/knative/hack/issues/200


psschwei (commented):
Thanks @kellyaa !
/lgtm
/approve

PR #5133, opened 2022-07-26
Update the version of CLI plugin for operator to 1.6.2

houshengbo:


docs/install/operator/knative-with-operator-cli.md
```diff
@@ -12,7 +12,7 @@ Before you install the Knative Operator CLI Plugin, first install the [Knative C
 
 === "MacOS"
 
-    1. Download the binary `kn-operator-darwin-amd64` for your system from the [release page](https://github.com/knative-sandbox/kn-plugin-operator/releases/tag/knative-v1.6.0).
+    1. Download the binary `kn-operator-darwin-amd64` for your system from the [release page](https://github.com/knative-sandbox/kn-plugin-operator/releases/tag/knative-v1.6.2).
```
psschwei, line 15: Might be worth just linking to the main release page (rather than a specific version) so that this won't require an update on each new release

docs/install/operator/knative-with-operator-cli.md
```diff
@@ -22,7 +22,7 @@ Before you install the Knative Operator CLI Plugin, first install the [Knative C
 
 === "Linux"
 
-    1. Download the binary `kn-operator-linux-amd64` for your system from the [release page](https://github.com/knative-sandbox/kn-plugin-operator/releases/tag/knative-v1.6.0).
+    1. Download the binary `kn-operator-linux-amd64` for your system from the [release page](https://github.com/knative-sandbox/kn-plugin-operator/releases/tag/knative-v1.6.2).
```
psschwei, line 25: same as above

docs/install/operator/knative-with-operator-cli.md
houshengbo: well we have different version . for example, if people click on the knative-1.6 docs, I hope that it will link them to the 1.6 version instead of the latest released version, which can be 1.7, 1.8, or something.

psschwei (commented):
/lgtm
/approve

PR #5134, opened 2022-07-26
Istio supported versions change

AngeloDanducci:
Link to the net-istio releases github page as the source of truth for supported versions per release instead of having a hard coded value.

**Description of problem the PR solves:**
Received messages on slack asking which versions of Istio we support and the docs page is out of date.
Linking to the net-istio releases page seems a better strategy in combination with [this issue](https://github.com/knative-sandbox/net-istio/issues/958) as opposed to updating the documentation every time. 

Worst case scenario the release notes can be updated manually instead of new documentation PRs.

## Proposed Changes 

- Link to the net-istio release page to show supported Istio version instead of hard coding.

### DIFF
--- a/docs/install/installing-istio.md
+++ b/docs/install/installing-istio.md
@@ -16,8 +16,7 @@ You need:
 
 ## Supported Istio versions
 
-You can see which version is the latest to have been tested [in the most recent net-istio releases](https://github.com/knative-sandbox/net-istio/releases).
-
+You can view the latest tested Istio version on the [Knative Net Istio releases page](https://github.com/knative-sandbox/net-istio/releases).
 
 ## Installing Istio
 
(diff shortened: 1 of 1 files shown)


dprotaso:
LGTM

deferring to docs folks for the copy

dprotaso:
also note: we'll want to cherry pick this change back to prior release branches - since I believe we adopted newer versions of istio in prior releases

docs/install/installing-istio.md
```diff
@@ -16,8 +16,8 @@ You need:
 
 ## Supported Istio versions
 
-The current known-to-be-stable version of Istio tested in conjunction with Knative is **v1.12**.
-Versions in the 1.12 line are generally fine too.
+You can see which version is the latest to have been tested [in the most recent net-istio releases](https://github.com/knative-sandbox/net-istio/releases).
+
```
abrennan89, line 20:
```suggestion
You can view the latest tested Istio version on the [Knative Net Istio releases page](https://github.com/knative-sandbox/net-istio/releases).
```

suggest something like this for readability

PR #5142, opened 2022-07-27
Add documentation for active-min-scale

psschwei:

## Proposed Changes 

Adds documentation for the `active-min-scale` annotation added in https://github.com/knative/serving/pull/13136

There's an ongoing discussion over the name of this annotation on [slack](https://knative.slack.com/archives/C9CV04DNJ/p1658938603476139), so this may need an update. But since the serving PR has already merged, may as well start getting the docs in for it.


psschwei:
/hold
We're updating the name

PR #5156, opened 2022-08-01
Updating docs for Revisions

abrennan89:
### Proposed Changes

- Created Resources > Revisions topic under Concepts
- Created Revisions topic under Serving

Additional PRs will follow to clean up / move revision config options

### Direct preview links:
- https://deploy-preview-5156--knative.netlify.app/docs/concepts/resources/revisions/
- https://deploy-preview-5156--knative.netlify.app/docs/serving/revisions/

### DIFF
--- a/docs/concepts/resources/revisions.md
+++ b/docs/concepts/resources/revisions.md
@@ -7,7 +7,7 @@
 * **[Autoscaling](../../serving/autoscaling/README.md){target=_blank}**
     - Revisions can be automatically scaled up and down according to incoming traffic.
 
-* **[Traffic routing](../../serving/traffic-management.md){target=_blank}**
+* **[Configuring gradual rollout of traffic to Revisions](../../serving/rolling-out-latest-revision.md){target=_blank}**
     - Revisions enable progressive roll-out and roll-back of application changes by changing the HTTP routing between Service names and Revision instances.
 <!--
 - **Garbage collection**
--- a/docs/concepts/resources/revisions.md
+++ b/docs/concepts/resources/revisions.md
@@ -8,7 +8,7 @@
     - Revisions can be automatically scaled up and down according to incoming traffic.
 
 * **[Configuring gradual rollout of traffic to Revisions](../../serving/rolling-out-latest-revision.md){target=_blank}**
-    - Revisions enable progressive roll-out and roll-back of application changes by changing the HTTP routing between Service names and Revision instances.
+    - Revisions enable progressive roll-out and roll-back of application changes.
 <!--
 - **Garbage collection**
     - TODO: add Something about how this works w/ revisions; only apply for Operator installs? GC docs need a general cleanup before adding something / links here-->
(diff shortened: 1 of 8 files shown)


docs/concepts/resources/revisions.md
```diff
@@ -0,0 +1,29 @@
...
+
+--8<-- "about-revisions.md"
+
+## Related concepts
+
+* **[Autoscaling](../../serving/autoscaling/README.md){target=_blank}**
+    - Revisions can be automatically scaled up and down according to incoming traffic.
+
+* **[Configuring gradual rollout of traffic to Revisions](../../serving/rolling-out-latest-revision.md){target=_blank}**
+    - Revisions enable progressive roll-out and roll-back of application changes by changing the HTTP routing between Service names and Revision instances.
```
nader-ziada, line 11: I think routing for a specific service can be changed between different revisions, but the wording `Service names and Revision instances` implied to me that you can route to a different service

nader-ziada (commented):
connect mostly good, but I don't get the website structure from this PR by itself so will not comment on that.

docs/concepts/resources/revisions.md
abrennan89: I think this can be simplified actually to just the first part of the sentence. Thanks!

docs/snippets/about-revisions.md
```diff
@@ -0,0 +1,13 @@
+<!-- Snippet used in the following topics:
+- /docs/concepts/resources/revisions.md
+- /docs/serving/revisions/README.md
+-->
+Revisions are stateless Knative Serving resources that contain point-in-time snapshots of the application code and configuration for each change made to a Knative Service.
```
csantanapr, line 5:
```suggestion
Revisions are Knative Serving resources that contain point-in-time snapshots of the application code and configuration for each change made to a Knative Service.
```
I would remove the word stateless to add confusion.

abrennan89: This came from the API spec https://github.com/knative/specs/blob/main/specs/serving/knative-api-specification-1.0.md#revision so if we remove it here it should also be updated there IMO?

csantanapr: The spec is correct
>The Knative Revision represents a stateless, autoscaling snapshot-in-time of application code and configuration. 

Is saying that a revision is a representation of code and configuration of a stateless application. 
The thing that is stateless if the application code&config itself.

abrennan89: Ah, OK. I think it's misleading how it's worded in the spec still, but I understand what you're saying now.

csantanapr: What about these options?
1. Revisions represent point-in-time snapshots of the application code and configuration for each change made to a Knative Service.
2. Revisions represent point-in-time snapshots of the stateless application code and configuration for each change made to a Knative Service.

PR #5161, opened 2022-08-04
Adding Security-Guard repo

davidhadas:
Adding Security-Guard repo
The tool added other changes


golang/_redirects
```diff
@@ -5,10 +5,8 @@
 #
 # To regenerate, run:
 #   go run ./tools/redir-gen/
-/actions-downstream-test/* go-get=1 /golang/actions-downstream-test.html 200
```
evankanderson, line 8: This got archived, but we forgot to run the tool.

golang/_redirects
```diff
@@ -5,10 +5,8 @@
...
-/actions-kind/* go-get=1 /golang/actions-kind.html 200
```
evankanderson, line 9: Removed, again didn't run the tool:

https://github.com/knative/community/commit/9cf199f9a0bbbd0b258fdec039c324de19cdc310

evankanderson (commented):
I didn't check all the others, but it looks like we haven't put this into CI / knobots to auto-generate the PRs, so it only happens when a human _cares_ about the changes; actions and func-tastic don't look like they affect go imports, so /shrug.

/lgtm
/approve

PR #5162, opened 2022-08-06
Fix & Simplify ApiServerSource docs

marniks7:
## Proposed Changes 

- issue: pingsource-example is from another tutorial, it is `<namespace>`
- confusion: `<sink-name>` used for kn is not a name, use `<sink>` instead
- simplify: for yaml - 3 values can be provided directly, for details there is a link to additional documentation

### DIFF
--- a/docs/eventing/sources/apiserversource/getting-started.md
+++ b/docs/eventing/sources/apiserversource/getting-started.md
@@ -186,14 +186,14 @@ command:
               --mode "Resource" \
               --resource "Event:v1" \
               --service-account <service-account> \
-              --sink <sink-name>
+              --sink <sink>
             ```
             Where:
 
             - `<apiserversource>` is the name of the source that you want to create.
             - `<namespace>` is the name of the namespace that you created in step 1 earlier.
             - `<service-account>` is the name of the ServiceAccount that you created in step 2 earlier.
-            - `<sink-name>` is the name of your sink, for example, `http://event-display.pingsource-example.svc.cluster.local`.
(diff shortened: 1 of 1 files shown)


docs/eventing/sources/apiserversource/getting-started.md
```diff
@@ -186,14 +186,14 @@ command:
...
-              --sink <sink-name>
+              --sink <sink>
             ```
             Where:
 
             - `<apiserversource>` is the name of the source that you want to create.
             - `<namespace>` is the name of the namespace that you created in step 1 earlier.
             - `<service-account>` is the name of the ServiceAccount that you created in step 2 earlier.
-            - `<sink-name>` is the name of your sink, for example, `http://event-display.pingsource-example.svc.cluster.local`.
+            - `<sink>` is the sink you created in step 5 earlier, for example, Service url: `http://event-display.<namespace>.svc.cluster.local`
```
abrennan89, line 196: We prefer to use more descriptive names like `sink-name` for variables in docs generally
@snneji any thoughts on this?

abrennan89:
The intention of the sample code is not to provide a copy/paste example but to explain the fields, which is why we document it this way with variables.

marniks7:
I am providing feedback as a user, as for such a trivial getting-started guide it has too much weird things in one way or another.

More over (in addition to initial description)
- `kn` and `yaml` options don't have the same experience. `kn` has almost all predefined values, `yaml` has non-predefined values
- In case if you don't click `yaml` - you will not see the field explanation at all

Issue #4937, opened 2022-04-26
Subscription doc lacks reply channel feature


matzew:
The yaml version of a subscription does lack the optional `reply` feature.

The subscription has two `Destination`s, `subscriber` and `reply`:
https://github.com/knative/eventing/blob/main/pkg/apis/messaging/v1/subscription_types.go#L97-L102

The `kn` client doc does have it, but the yaml not.. would be nice if doc would be changed to:

    ```yaml
    apiVersion: messaging.knative.dev/v1
    kind: Subscription
    metadata:
      name: <subscription-name> # Name of the Subscription.
      namespace: default
    spec:
      channel:
        apiVersion: messaging.knative.dev/v1
        kind: Channel
        name: <channel-name> # Configuration settings for the Channel that the Subscription connects to.
      delivery:
        deadLetterSink:
          ref:
            apiVersion: serving.knative.dev/v1
            kind: Service
            name: <service-name>
            # Configuration settings for event delivery.
            # This tells the Subscription what happens to events that cannot be delivered to the Subscriber.
            # When this is configured, events that failed to be consumed are sent to the deadLetterSink.
            # The event is dropped, no re-delivery of the event is attempted, and an error is logged in the system.
            # The deadLetterSink value must be a Destination.
      reply:
        ref:
          apiVersion: messaging.knative.dev/v1
          kind: InMemoryChannel
          name: <service-name> # Configuration settings for the reply event. This is the event Sink that events replied from the subscriber are delivered to.
      subscriber:
        ref:
          apiVersion: serving.knative.dev/v1
          kind: Service
          name: <service-name> # Configuration settings for the Subscriber. This is the event Sink that events are delivered to from the Channel.
    ```

PR #5166, opened 2022-08-10
Add reply feature to subscription docs

abrennan89:
Fixes https://github.com/knative/docs/issues/4937

### DIFF
--- a/docs/eventing/channels/subscriptions.md
+++ b/docs/eventing/channels/subscriptions.md
@@ -15,7 +15,7 @@ For more information about Subscription objects, see
 === "kn"
     Create a Subscription between a Channel and a Sink by running:
 
-    ```
+    ```bash
     kn subscription create <subscription-name> \
       --channel <Group:Version:Kind>:<channel-name> \
       --sink <sink-prefix>:<sink-name> \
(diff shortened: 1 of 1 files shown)


docs/eventing/channels/subscriptions.md
```diff
@@ -74,23 +74,28 @@ For more information about Subscription objects, see
...
+              kind: InMemoryChannel
+              name: <service-name>
+              # Configuration settings for the reply event. This is the event Sink that events replied from the subscriber are delivered to.
           subscriber:
             ref:
               apiVersion: serving.knative.dev/v1
               kind: Service
-              name: <service-name> # Configuration settings for the Subscriber. This is the event Sink that events are delivered to from the Channel.
+              name: <service-name>
+              # Configuration settings for the Subscriber. This is the event Sink that events are delivered to from the Channel.
```
pierDipi, line 98: Can we move the comments on top of the field like:
```yaml
          # Configuration settings for the reply event. This is the event Sink that events replied from the subscriber are delivered to.
          reply:
            ref:
              apiVersion: messaging.knative.dev/v1
              kind: InMemoryChannel
              name: <service-name>
```
```yaml
          # Required configuration settings for the Subscriber. This is the event Sink that events are delivered to from the Channel.
          subscriber:
            ref:
              kind: Service
              name: <service-name> # Configuration settings for the Subscriber. This is the event Sink that events are delivered to from the Channel.
              name: <service-name>
```

pierDipi: added also notes to signal whether a field is optional or required

abrennan89:
@pierDipi thanks for the comments, made some updates, PTAL

pierDipi (approved):
/lgtm

Issue #5065, opened 2022-06-29
How to list all plugins?


abrennan89:

abrennan89:
Reference: https://github.com/knative/client/blob/main/docs/cmd/kn_plugin_list.md

PR #5167, opened 2022-08-10
Add list plugins command docs

abrennan89:
Fixes https://github.com/knative/docs/issues/5065

### DIFF
--- a/docs/client/kn-plugins.md
+++ b/docs/client/kn-plugins.md
@@ -39,8 +39,6 @@ You can install some plugins can be installed using the [Knative plugins Homebre
 
 ## List available plugins
 
-Available plugins are plugins that are executable, begin with `kn-`, are included in the `kn` plugin directory, and are located somewhere in the execution `$PATH`.
-
 You can list all available (installed) plugins by entering the command:
 
 ```bash
--- a/docs/client/kn-plugins.md
+++ b/docs/client/kn-plugins.md
@@ -36,3 +36,13 @@ You can manually install all plugins. To manually install a plugin:
 ## Install a plugin by using Homebrew
 
 You can install some plugins can be installed using the [Knative plugins Homebrew Tap](https://github.com/knative-sandbox/homebrew-kn-plugins/). For example, you can install the `kn-admin` plugin by running `brew install knative-sandbox/kn-plugins/admin`.
+
+## List available plugins
+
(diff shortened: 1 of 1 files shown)


docs/client/kn-plugins.md
```diff
@@ -36,3 +36,13 @@ You can manually install all plugins. To manually install a plugin:
 ## Install a plugin by using Homebrew
 
 You can install some plugins can be installed using the [Knative plugins Homebrew Tap](https://github.com/knative-sandbox/homebrew-kn-plugins/). For example, you can install the `kn-admin` plugin by running `brew install knative-sandbox/kn-plugins/admin`.
+
+## List available plugins
+
+Available plugins are plugins that are executable, begin with `kn-`, are included in the `kn` plugin directory, and are located somewhere in the execution `$PATH`.
```
rhuss, line 42: * The `kn` plugin directory is `~/.config/kn/plugins` on Linux and macOs, and on `%APPDATA%\kn\plugins` (I think this information is also missing on https://knative.dev/docs/client/kn-plugins/#manually-install-a-plugin where it is only mentioned that it can be put on the $PATH
* It's an *or*, not an *and* (located in the plugin directory *or* in the Path, with the Plugin directory taking precedence)

abrennan89: RE the plugins directory part, can you open a new issue if we need to update the docs to talk about using a plugin directory please, it's out of scope for this issue, which is just adding a quick PR for the list command

rhuss: without knowing where the "kn plugin directory" is, this information here raises more questions than it solves. Maybe just leave out the plugin dir until it is explained elsewhere ?

abrennan89: It's mentioned here already: https://knative.dev/docs/client/configure-kn/
Let me take another look at this when I have more time. RE #5171 please open issues with an actual user story that explains why someone needs to know about this, I feel like right now we have a bunch of disconnected info about _where_ the plugin directory is etc but no info about what it is or why use it

abrennan89: I'm going to just remove this line and leave the bit that mentions the list command

docs/client/kn-plugins.md
```diff
@@ -36,3 +36,13 @@ You can manually install all plugins. To manually install a plugin:
...
+
+You can list all available (installed) plugins by entering the command:
```
abrennan89, line 44:
```suggestion

You can list all available (installed) plugins by entering the command:
```

PR #5176, opened 2022-08-15
Go security guard

davidhadas:
Adding a  go repo security-guard to the list of go repositories to get it to redirect from knative.dev/security-guard


davidhadas:
If that works, we need to consider adding other repositories... and/or documenting it as another step in getting a repo on board

PR #5209, opened 2022-09-01
Using RPS as scaling metric always requires an annotation

psschwei:
Signed-off-by: Paul S. Schweigert <paul@paulschweigert.com>

## Proposed Changes 

Making it clear in the docs that when using RPS for autoscaling, users need to set the scaling metric annotation regardless of if the target is set per-revision or via the global configmap.

Issue surfaced in [slack discussion](https://knative.slack.com/archives/C93E33SN8/p1662019087164369)


psschwei:
whoops, missed the whitespace failure... fixed now

PR #5220, opened 2022-09-16
Fix issues in broker docs

creydr:
PR fixes:
* some markdown syntax in the RabbitMQ Broker doc, so the code example is displayed properly (048ba6d8ae5beff5a01f6fc3ec84d223eca3224c)
* some broker spec indentation in the Broker configuration example (e9eb2d51d4a43f29d364ab419a2e13453884eebf)
* use correct target namespace for broker in broker configuration example, as the current namespace depends on the used kubeconfig context (6e41f0645d48cb2d4b381570c4b91d885c280e79)
* states support for RabbitMQ broker in configuration example (009f6d1bc8f06c8d45ee2e01a953eacbd7c3ac33)

### DIFF
--- a/docs/eventing/brokers/broker-developer-config-options.md
+++ b/docs/eventing/brokers/broker-developer-config-options.md
@@ -32,7 +32,7 @@ spec:
 ```
 
 - You can specify any valid `name` for your broker. Using `default` will create a broker named `default`.
-- The `namespace` must be an existing namespace in your cluster. Using `default` will create the broker in the current namespace.
+- The `namespace` must be an existing namespace in your cluster. Using `default` will create the broker in the `default` namespace.
 - You can set the `eventing.knative.dev/broker.class` annotation to change the class of the broker. The default broker class is `MTChannelBasedBroker`, but Knative also supports use of the `Kafka` broker class. For more information about Kafka brokers, see the [Apache Kafka Broker](../brokers/broker-types/kafka-broker/README.md) documentation.
 - `spec.config` is used to specify the default backing channel configuration for MT channel-based broker implementations. For more information on configuring the default channel type, see the documentation on [Configure Broker defaults](../configuration/broker-configuration.md).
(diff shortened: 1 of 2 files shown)


docs/eventing/brokers/broker-developer-config-options.md
```diff
@@ -8,31 +8,31 @@ The following is a full example of a multi-tenant (MT) channel-based Broker obje
...
+      uri: example-uri
+    retry: 5
+    backoffPolicy: exponential
+    backoffDelay: "2007-03-01T13:00:00Z/P1Y2M10DT2H30M"
 ```
 
 - You can specify any valid `name` for your broker. Using `default` will create a broker named `default`.
-- The `namespace` must be an existing namespace in your cluster. Using `default` will create the broker in the current namespace.
-- You can set the `eventing.knative.dev/broker.class` annotation to change the class of the broker. The default broker class is `MTChannelBasedBroker`, but Knative also supports use of the `Kafka` broker class. For more information about Kafka brokers, see the [Apache Kafka Broker](../brokers/broker-types/kafka-broker/README.md) documentation.
+- The `namespace` must be an existing namespace in your cluster. Using `default` will create the broker in the `default` namespace.
```
abrennan89, line 35: Is this correct re the broker namespace?
@knative/eventing-wg-leads

abrennan89 (commented):
One comment - please check with someone in the Eventing WG that this is true about the broker, I always thought it was the current namespace the user is in, not the literal `default` namespace, but I might be wrong.

Otherwise lgtm! Happy to approve once this is confirmed.

docs/eventing/brokers/broker-developer-config-options.md
lionelvillard: This is correct. Here `default` is just an example.

pierDipi:
> One comment - please check with someone in the Eventing WG that this is true about the broker, I always thought it was the current namespace the user is in, not the literal `default` namespace, but I might be wrong.
> 

Yes, this is generally true, it's not the literal `default` but it is the default namespace in the context which is usually `default` but it might be differerent.

> Otherwise lgtm! Happy to approve once this is confirmed.

PR #5230, opened 2022-09-22
Add SASL for KafkaSource docs

abrennan89:
Fixes https://github.com/knative/docs/issues/3583

- Added SASL docs
- Moved "clean up" to end of section (source deletion should be after all the other options)

PR #5244, opened 2022-09-23
Fix markdown rendering issue with brackets

psschwei:
xref #5230 
/assign @abrennan89


psschwei:
check the preview to make sure :smile:

psschwei:
doesn't actually fix the issue
/hold

abrennan89:
Looks like it fixes it OK?
/lgtm
/approve

The build check I guess is still failing and unrelated, but at least this fixes that one page

psschwei:
> Looks like it fixes it OK? 

first fix showed the backslashes, new one uses `{% raw %}` jinja tags so that macros plugin will ignore it

PR #5251, opened 2022-09-26
Fix build test

lionelvillard:
## Proposed Changes 

- Remove code same vendor directory causing the build test to fail due to the `js`build tag conflicting with the default os tag
-
-


psschwei (commented):
/lgtm
/approve

psschwei:
Thanks @lionelvillard !!

PR #5252, opened 2022-09-26
Add badges to sources and minor editing

lionelvillard:
## Proposed Changes 

- Add new "stage" badge to better communicate the source current status
- Minor KafkaSource editing, trying to highlight that KafkaSources connect to existing topics
-


lance (commented):
/lgtm

PR #5260, opened 2022-10-03
 Provide explicit example of bash command, similar to `kn` quickstart

bobfuru:
It would be nice to have the example command that the user could copy, similar to how it's done in the previous procedure for installing the `kn` CLI in step 3 (https://knative.dev/docs/install/quickstart-install/#install-the-knative-cli)


lance (commented):
/lgtm

psschwei (commented):
/approve

Issue #423, opened 2018-09-27
Openshift: Knative namespace change for monitoring


matzew:
## Expected Behavior

Correct usage of the `knative-monitoring` namespace

## Actual Behavior

using the old name for the namespace (`monitoring`)

RichieEscarez:
This was done in #416

PR #5261, opened 2022-10-04
Remove note about requirement to use http-protocol enabled with http01 challenge

izabelacg:
## Proposed Changes

- Remove note about requirement to use http-protocol enabled with http01 challenge. 
  It seems to imply that http-01 challenge will work only with http-protocol set to enabled, but it does work with "Redirected". This seems to be have been an issue in the past [#423](https://github.com/knative-sandbox/net-contour/issues/423) before  being solved by this PR #610: 
- Typo in cluster

_PS_: There is also a mention to a deprecated value (Disabled) for http-protocol, but it seems to be still valid to use. Not sure if it should be removed or not from line 252.

### DIFF
--- a/docs/serving/using-auto-tls.md
+++ b/docs/serving/using-auto-tls.md
@@ -41,7 +41,7 @@ Knative supports the following Auto TLS modes:
 
 1.  Using HTTP-01 challenge
 
-    - In this type, your cluster does not need to be able to talk to your DNS server. You must map your domain to the IP of the cluser ingress.
+    - In this type, your cluster does not need to be able to talk to your DNS server. You must map your domain to the IP of the cluster ingress.
     - When using HTTP-01 challenge, **a certificate will be provisioned per Knative Service.**
     - **HTTP-01 does not support provisioning a certificate per namespace.**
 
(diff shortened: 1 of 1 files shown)


nak3:
> PS: There is also a mention to a deprecated value (Disabled) for http-protocol, but it seems to be still valid to use. Not sure if it should be removed or not from line 252.

Could you remove `Disabled` from line 252? It has been [deprecated](https://github.com/knative/networking/blob/main/config/config-network.yaml#L129) and we should not mention about it in the doc.

nak3:
LGTM other than that.

izabelacg:
> > PS: There is also a mention to a deprecated value (Disabled) for http-protocol, but it seems to be still valid to use. Not sure if it should be removed or not from line 252.
> 
> Could you remove `Disabled` from line 252? It has been [deprecated](https://github.com/knative/networking/blob/main/config/config-network.yaml#L129) and we should not mention about it in the doc.

Cool, I updated according to your recommendation. Thanks

abrennan89 (approved):
/lgtm
/cherrypick release-.17

PR #5273, opened 2022-10-13
chore: multi-arch support with buildx in all Dockerfiles

kahirokunn:
## Proposed Changes 

I deployed a docker image built on M1 mac to an intel cluster and got a cpu architecture error.
Now that support for both amd and arm is expanding, I think it is better for users to use buildx to build both to try out the sample.

### DIFF
--- a/code-samples/serving/grpc-ping-go/README.md
+++ b/code-samples/serving/grpc-ping-go/README.md
@@ -32,11 +32,8 @@ for production containers.
   Replace `{username}` with your Docker Hub username then run the commands:
 
   ```bash
-  # Build the container on your local machine.
-  docker build --tag "{username}/grpc-ping-go" .
-
-  # Push the container to docker registry.
-  docker push "{username}/grpc-ping-go"
+  # Build and push the container on your local machine.
+  docker buildx build --platform linux/arm64,linux/amd64 -t "{username}/grpc-ping-go" --push .
   ```
 
 3. Update the `service.yaml` file in the project to reference the published image from step 1.
--- a/code-samples/community/serving/helloworld-clojure/README.md
+++ b/code-samples/community/serving/helloworld-clojure/README.md
@@ -109,11 +109,8 @@ folder) you're ready to build and deploy the sample app.
    username:
 
    ```bash
-   # Build the container on your local machine
-   docker build -t {username}/helloworld-clojure .
-
-   # Push the container to docker registry
-   docker push {username}/helloworld-clojure
+   # Build and push the container on your local machine.
+   docker buildx build --platform linux/arm64,linux/amd64 -t "{username}/helloworld-clojure" --push .
    ```
 
 1. After the build has completed and the container is pushed to docker hub, you
--- a/code-samples/community/serving/helloworld-clojure/index.md
+++ b/code-samples/community/serving/helloworld-clojure/index.md
@@ -109,11 +109,8 @@ folder) you're ready to build and deploy the sample app.
    username:
 
    ```bash
-   # Build the container on your local machine
-   docker build -t {username}/helloworld-clojure .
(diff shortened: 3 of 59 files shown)


abrennan89:
@knative/serving-reviewers

code-samples/serving/grpc-ping-go/README.md
```diff
@@ -32,11 +32,8 @@ for production containers.
   Replace `{username}` with your Docker Hub username then run the commands:
 
   ```bash
-  # Build the container on your local machine.
-  docker build --tag "{username}/grpc-ping-go" .
-
-  # Push the container to docker registry.
-  docker push "{username}/grpc-ping-go"
+  # Build and push the container on your local machine.
+  docker buildx build --platform linux/arm64,linux/amd64 -t "{username}/grpc-ping-go" --push .
```
psschwei, line 36: The new command failed for me with the following output:

```console
[+] Building 0.0s (0/0)                                                                                  
ERROR: multiple platforms feature is currently not supported for docker driver. 
Please switch to a different driver (eg. "docker buildx create --use")
```

whereas the original commands worked without issue. If we're going to update it to use `buildx`, we'll need some more set up info since it doesn't work out of the box.

psschwei (changes requested):

code-samples/serving/grpc-ping-go/README.md
csantanapr: I think adding a link to the [buildx](https://docs.docker.com/build/buildx/install/) should be enough.

psschwei: Personally, I think I'd rather leave the default as-is and provide a link to buildx for those who need to do a multiarch build (which won't be everyone)

kahirokunn: If you let people use buildx from the beginning, the container image will run on many people's PCs.
You will not have to worry about CPU architecture errors.
Therefore, buildx is also a good choice.

kahirokunn:
keep

dprotaso:
The Dockerfile needs to be updated to support multi arch first

ie.
https://github.com/knative/docs/commit/098a9da831db14c380de1a04484103a09a5f4b49

skonto:
@kahirokunn gentle ping, are you still planning to update it?

kahirokunn:
@dprotaso @skonto I have addressed the points raised. How about now?

skonto:
I will test the updates thanks @kahirokunn. We do not want this for the rest of the samples?

kahirokunn:
Do you wish to buildx all images in this repository?

skonto:
It would be nice to have build options or maybe a separate page for building alternatives instead of copying the same command everywhere.

skonto:
@dprotaso wdyth? Should we re-write docs also to include builds with docker buildx for the other samples too?

psschwei:
> @dprotaso @psschwei wdyth? Should we re-write docs also to include builds with docker buildx for the other samples too?

I think I'm still slightly in favor of leaving things as is, with maybe a link out to buildx and/or other alternatives (with the view towards minimizing changes as these samples aren't really actively maintained). But, that said, I don't feel particularly strongly about it (I also don't use a Mac), so if we do decide we should update, that's fine too.

But if we do update, we should update all, I think.

kahirokunn:
@psschwei Okay, I will update all.

kahirokunn:
done. how about now?

kahirokunn:
done. how about now?

kahirokunn:
@psschwei I would appreciate a review before conflicts arise. thx.

psschwei:
Sorry about that, I missed the notification for this one one...

/lgtm
/approve

Issue #5265, opened 2022-10-06
Create a Functions section 


lance:
**Describe the change you'd like to see**

Evan in https://github.com/knative/docs/pull/5243#discussion_r989327583

We'll need to add more content here (fine in a subsequent PR):

- [x] What is it / how does it work?
- [ ] How do I use it / what's the intended workflow?
  - [ ] Create
  - [ ] Iterate
  - [ ] Integrate in build system
  - [ ] Update buildpack / other maintenance needed?
- [ ] What's a language pack / how can I get more of them?
  - [ ] Language pack inventory page
  - [ ] Writing your own language pack

lance:
No I mean language pack. https://github.com/knative/func/blob/main/docs/language-pack-providers/language-pack-contract.md

It's how companies and vendors extend Functions

abrennan89:
Moved advanced questions about lang packs to separate issue.

abrennan89:
@lance per Evan's original comment, is there enough info about buildpacks yet to warrant opening a follow up issue for advanced docs about that?

PR #5286, opened 2022-10-19
Follow up to add advanced func docs

abrennan89:
Fixes https://github.com/knative/docs/issues/5265
Fixes https://github.com/knative/docs/issues/5287

Add docs for:
- invoke command
- splits up build, run, and deploy docs sections
- split on-cluster builds vs local builds docs
- added language pack docs

Preview links:
- https://deploy-preview-5286--knative.netlify.app/docs/getting-started/create-a-function/
- https://deploy-preview-5286--knative.netlify.app/docs/functions/building-functions/
- https://deploy-preview-5286--knative.netlify.app/docs/functions/running-functions/
- https://deploy-preview-5286--knative.netlify.app/docs/functions/deploying-functions/
- https://deploy-preview-5286--knative.netlify.app/docs/functions/invoking-functions/
- https://deploy-preview-5286--knative.netlify.app/docs/functions/language-packs/


docs/snippets/create-a-function.md
```diff
@@ -5,12 +5,24 @@ After you have installed Knative Functions, you can create a function project by
```
lance, line 5: I think the snippet usage here is probably a copy/paste error

docs/functions/building-functions.md
```diff
@@ -0,0 +1,42 @@
...
+
+--8<-- "proc-building-function.md"
+
+## On-cluster Builds
+
+If you do not have a local Docker daemon running, or you are using a CI/CD pipeline, you might want to build your function on the cluster instead of using a local build. You can create an on-cluster build by using the `func deploy --remote` command.
+
+### Prerequisites
+
+- The function must exist in a Git repository.
```
lance, line 23: This is not entirely true for the current release. We have just added the ability to do direct source upload so that you don't have to have your function in a git repo. The procedure is almost identical, but instead of the cluster pulling function source code from github, the source is pushed to the cluster directly from the developer's laptop. This is the behavior when issuing the command `func deploy --remote` with no `--git-url` flag provided.

docs/functions/building-functions.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Prerequisites
+
+- You have a Docker daemon on your local machine. This is already provided if you have used the Quickstart installation.
+
+### Procedure
+
+--8<-- "proc-building-function.md"
+
+## On-cluster Builds
```
lance, line 17: I wonder if this actually makes more sense in the deploy section. These are technically on-cluster builds, but the end result is a deployment, and the command used for these examples is `func deploy`.  wdyt?

docs/functions/building-functions.md
```diff
@@ -0,0 +1,42 @@
...
+    func deploy --remote --registry <registry> --git-url <git-url> -p hello
+    ```
+
+=== "kn func"
+
+    ```{ .console }
+    kn func deploy --remote --registry <registry> --git-url <git-url> -p hello
+    ```
+
+After you have specified the Git URL for your function once, you can omit it in subsequent commands.
```
lance, line 42: This is the tricky part of the direct-upload change that was recently made. If you have specified the Git URL in a previous command, and then subsequently run `func deploy --remote` as you describe here, the source will be pulled from the git repo. If you later want to perform a direct upload deployment, you need to run `func deploy --remote --git-url=""` which is pretty terrible, and probably a point in favor of not yet documenting the direct-upload method until we get the UX improved.

docs/functions/invoking-functions.md
```diff
@@ -0,0 +1,10 @@
+# Invoking functions
+
+You can use the `func invoke` command to send a test request to invoke a
+function either locally or on your Knative cluster.
+
+This command can be used to test that a function is working and able to receive HTTP requests and CloudEvents correctly.
+
+If your function is running locally, `func invoke` sends a test request to the local instance.
```
lance, line 8: If the function is running remotely, `func invoke` sends a test request to the remote instance on the cluster, unless it is also running locally.

lance (commented):
Looks great, @abrennan89 - really excited to see this coming together!

Just a few comments...

docs/functions/building-functions.md
abrennan89: I think in terms of use cases, because it's referred to as "on cluster builds", it makes more sense that a user would look for it here. Maybe down the line the CLI command could be reworked so that it makes more sense and doesn't use `deploy`, or we could also add docs to a deploy section, but I think from a user perspective for now it makes sense that it's here.

docs/functions/building-functions.md
abrennan89: By current release do you mean 1.7 or 1.8?

docs/functions/building-functions.md
abrennan89: Can you open a new issue maybe for this so that we can merge it as is and update it later? This is all based on what was provided in the other PR, so technically making changes at this point we're starting to add scope creep to what was originally meant to be published in the docs for kubecon this week, and it's going to end up delaying the whole PR.

abrennan89 (commented):
comment responses @lance

docs/functions/building-functions.md
lance: 1.8

abrennan89: In that case I'd add it as an additional PR so that this one can be CP'd back to older versions

abrennan89:
@lance I think it would be good to go ahead and get this merged so that the changes can be cherrypicked to 1.7 and then open a new issue / PR to add the 1.8 changes. It's probably a good habit to get into anyway to prevent scope creep.

PR #5291, opened 2022-10-19
Add blog post for KnativeCon/KubeCon NA 2022

lance:
A hastily written blog post so that we can do a little promotion of the upcoming talks. Please let me know if I have missed anything, and apologies for the boilerplate language!

Signed-off-by: Lance Ball <lball@redhat.com>

PR #5298, opened 2022-10-21
Improve overall structure of kubecon blog post

lance:
Basically, make it easy for readers to see when an event is occurring.

Signed-off-by: Lance Ball <lball@redhat.com>

/cc @evankanderson - thanks for your belated input on https://github.com/knative/docs/pull/5291. I have addressed your concerns in a follow up PR. I did not include information about vendor booths. If you want to help out with some wording on that, I'd be happy to include it. As far as Red Hat goes, we'll have Serverless demos at our booth, but those will be part of a larger set of demos that feature other tech and I don't really have a schedule for that. Not sure what IBM will be doing.

### DIFF
--- a/blog/docs/events/knative-at-kubecon-na-2022.md
+++ b/blog/docs/events/knative-at-kubecon-na-2022.md
@@ -15,27 +15,37 @@ its progress in the past year.
 
 ## Monday, October 24 KnativeCon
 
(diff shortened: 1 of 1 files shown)


evankanderson (commented):
/lgtm
/approve

PR #5300, opened 2022-10-24
Remove `.github/workflows/knative-mkdocs.yaml`

nak3:
As per title, this patch removes `.github/workflows/knative-mkdocs.yaml`.

The action is triggered when PR was sent against `web` branch but the branch is already gone and so this action is not used at all.

https://github.com/knative/docs/blob/main/.github/workflows/knative-mkdocs.yaml#L5


lance (commented):
/lgtm

PR #5191, opened 2022-08-26
fix build bc command not found on 1.7 release

csantanapr:
fix build bc command not found on 1.7 release

PR #5302, opened 2022-10-25
Fix build bc command not found

nak3:
This is similar fix with https://github.com/knative/docs/pull/5191 but another place.

/cc @psschwei

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -71,7 +71,7 @@ else
 
     echo "Building for previous version $version"
     git clone --depth 1 -b ${DOCS_BRANCHES[$i+1]} https://github.com/${GIT_SLUG} "$TEMP/docs-$version"
-    if (( $(echo "$version > 1.6" | bc -l) )); then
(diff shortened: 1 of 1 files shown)


nak3:
Here is my tested script:
```bash
#!/usr/bin/env bash

VERSIONS=("1.8" "1.7" "1.6" "1.5") # Docs version, results in the url e.g. knative.dev/docs-0.23/..
previous=("${VERSIONS[@]:1}")

for i in "${!previous[@]}"; do
  version=${previous[$i]}
  if [ ${version#*1.} -gt 6 ]; then
    echo -n "$version"
    echo " is greater than 1.6"
  else
    echo -n "$version"
    echo " is not greater than 1.6"
  fi
done
```

```
$ bash a.sh
1.7 is greater than 1.6
1.6 is not greater than 1.6
1.5 is not greater than 1.6
```

psschwei (commented):
/lgtm
/approve

psschwei:
It worked!

PR #5319, opened 2022-11-03
Update README.md

Kaiass:
KafkaSource sasl type is specified as 'saslType' config parameter but should be 'type'.

## Proposed Changes 

-
-
-


pierDipi (approved):
Yes, this is correct, reference: https://github.com/knative-sandbox/eventing-kafka/blob/3992a225fc57927c9b0dcfc4fcdb6b4e17db2a11/pkg/apis/bindings/v1beta1/kafka_types.go#L60-L62

Thanks for the fix @Kaiass!

/lgtm

Kaiass:
@pierDipi  Glad to help!

PR #5332, opened 2022-11-18
Remove information about strict-subscriber flag again

creydr:
The strict-subscriber feature flag was removed in https://github.com/knative/eventing/pull/6608 so the validation cannot be disabled anymore. This commit adjusts the docs too.

## Proposed Changes
* remove information about the `strict-subscriber` flag from the docs again


pierDipi (approved):
/lgtm
/approve

psschwei (commented):
/approve

PR #5360, opened 2022-12-09
Add Debezium as an event source

cab105:
Add Debezium as a Knative event source

## Proposed Changes
Based on feedback from KnativeCon EU May 2022, it was suggested that an entry be made to the list of Knative event sources for Debezium. This PR adds the entry to the list along with a link to the Debezium project and the repo I created for the KnativeCon talk that exposes Debezium as a Knative event source along with some configuration notes.


psschwei (commented):
/approve

/assign @pierDipi  
for Eventing signoff

evankanderson (commented):
/lgtm
/approve

It would be awesome to get this documentation in a more official place, but it's even more awesome to have documentation at all.

Note that this catalog may be moving to ArtifactHub at some point, but whoever is doing that migration will be responsible for dragging this over.

PR #5362, opened 2022-12-13
docs(code-samples/go): fix indentation for code and package missing

maxgio92:
This PR fixes small documentation format for Golang sample code.

### DIFF
--- a/code-samples/eventing/helloworld/helloworld-go/README.md
+++ b/code-samples/eventing/helloworld/helloworld-go/README.md
@@ -34,50 +34,49 @@ cd knative-docs/code-samples/eventing/helloworld/helloworld-go
    code creates a basic web server which listens on port 8080:
 
    ```go
+   package main
    import (
-      "context"
-      "log"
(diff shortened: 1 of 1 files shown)


psschwei:
@maxgio92  Thanks for the PR!

Seems to be an issue with the remote reference:

```bash
$ git fetch upstream pull/5362/head:pr-5362 --force
fatal: couldn't find remote ref pull/5362/head
```

Seeing a similar error in netlify that's causing the deploy preview to fail (`git ref pull/5362/head does not exist or you do not have permission`)

FWIW, I'm able to access https://github.com/maxgio92/docs/tree/docs/code-samples-go , so not exactly sure what's causing the problem here...

maxgio92:
Is there something I can do @psschwe? I just rebased on `main` branch`.

psschwei:
It looks like the indentation was fixed in #5443 ... were there things missed in that PR that still need to be fixed?

maxgio92:
Don't think so @psschwei. Actually the diff shows one against older head :)
I think we can close this

maxgio92:
Thanks

PR #5380, opened 2023-01-13
Eventing istio golang redirect

pierDipi:
Part of https://github.com/knative/community/issues/1239


evankanderson (commented):
/lgtm
/approve

I think we have a tool that will do this, not sure if you ran it.

Anyway, you can `/hold cancel` when ready.

PR #5392, opened 2023-01-25
Updates slack URLs to CNCF instance

psschwei:
Signed-off-by: Paul S. Schweigert <paul@paulschweigert.com>

Updates to the website, docs repo, and the blog. I skipped the old release notices, but we can add them if someone really, really wants them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ To help you get started, see the following resources:
 
 ### Getting help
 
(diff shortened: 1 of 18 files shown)


evankanderson (commented):
/lgtm
/approve

We should make a snippet for this in the future.

PR #5394, opened 2023-01-26
Guard script fix - for 1.9

davidhadas:
Signed-off-by: David Hadas <david.hadas@gmail.com>

1. Found indentation issues with the example script for installing guard with TLS using operator
2. script worked only if one first installed guard without TLS... change the order of the script such that it can be installed on vanilla Knative.


psschwei (commented):
/lgtm
/approve

PR #5444, opened 2023-02-23
Twitter link style was broken.

nainaz:
"Fixes #issue-number" or "Follow us on Twitter link was breaking the page"

## Proposed Changes 

-
-
-


psschwei:
Fixed in #5449

PR #5448, opened 2023-02-25
Guard functions

davidhadas:
1. Add documentation for protecting Guard with functions
2. do a small fix in guard documentation - highlighting the per namespace config, and making sure it is there for knative operator


psschwei (commented):
@davidhadas Unrelated to this PR, but the bottom of the quickstart page has an errant bash block quote

/lgtm
/approve

PR #5449, opened 2023-02-27
Restores missing sections from homepage

psschwei:
See [slack discussion](https://cloud-native.slack.com/archives/C04LGEU2T9B/p1677004777478829)


csantanapr:
@psschwei thank you for the fix

PR #5454, opened 2023-03-02
First draft of Event Mesh definition

matzew:
## Proposed Changes 

- The knative.dev webstite talks about "event mesh", but not going any further. The term is generally a common term to describe "eventing infrastructure". Instead of leaving it just to different vendors we are trying to get some knative community definition for the term "Event Mesh".

A bit of discussion on the subject is already here: https://github.com/knative/eventing/issues/6719 and in the Eventing WG was also some discussion about this.

This PR is a first iteration (yes image and text may change), but we are getting started. It would be good to get this over the line in the next couple of weeks


docs/eventing/event-mesh.md
```diff
@@ -0,0 +1,17 @@
+# Event Mesh
+
+An Event Mesh is a dynamic infrastructure that is designed to smoothly transport events from producers to consumers via a network of interconnected _event brokers_ across any environment, and even between clouds in a seamless and loosely coupled way. In an Event Mesh both, producing and consuming application, can leverage their platform native event APIs. For instance a Phyton or Golang application can produce an event, using their native APIs, while a consuming application written in Java or Node.js can subscribe to it with their own native APIs as well.
```
gab-satchi, line 3: The italicized _event brokers_ in this case, is it talking more about things like Kafka/RabbitMQ SaaS solutions. Or the `Broker` API type?

matzew: Good question. I think generally the term is kept abstract on intention. An Event Mesh is not exclusive to Knative. So I guess things like Rabbit, Kafka, or any other SaaS broker is possible to be understood here.

However, I used the italicized _event brokers_ in this case, to have a hint forward to our own `Broker` API that we define in Knative Eventing!

gab-satchi: Yeah, I think that wording is better to highlight the distinction.

matzew: Sweet! Do you think this item/point is resolved, @gab-satchi ?

gab-satchi: Yep, resolved for me

matzew:
Anyone else ?  @aliok @creydr @pierDipi @evankanderson 

Or good for now and iterate on it with follow-up PRs?

psschwei (changes requested):
You'll need to add an entry in [`config/nav.yml`](https://github.com/knative/docs/blob/1b02d1e96e694738eded5a69e631f7c1e98ff236/config/nav.yml#L178) so that it gets added to the navigation

matzew:
@psschwei thanks, I added that

psschwei (commented):
/approve

PR #5456, opened 2023-03-06
Missing step in knative functions tutorial

### LINKED ISSUE #5308: Missing step in knative functions tutorial

While working through the knative functions tutorial (https://knative.dev/docs/getting-started/build-run-deploy-func/)

```
~ $ func create -l go hello
Created go function in /home/user1/hello

~ $ func run
Error: stat /home/user1/func.yaml: no such file or directory
```

It looks like I need to change directory  - this may be obvious to knative users, but for beginners, it should be added to the steps in the tutorial.

```
~ $ ls ~/hello/
func.yaml  go.mod  handle.go  handle_test.go  README.md

~ $ cd hello

~/hello $ func run
   Building function image
```

**Additional context**
N/A

manoelmarques:
Fixes #5308

## Proposed Changes 

- Adds a `cd hello` step to instructions

### DIFF
--- a/docs/snippets/proc-running-function.md
+++ b/docs/snippets/proc-running-function.md
@@ -9,6 +9,10 @@ The `run` command builds an image for your function if required, and runs this i
     Run the function locally by running the command inside the project directory. If you have not yet built the function you will need to provide the `--registry` flag:
 
     ```bash
+    cd hello
+    ```
+
+    ```bash
     func run [--registry <registry>]
     ```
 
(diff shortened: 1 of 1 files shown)


matejvasek:
Isn't it implied that user should be in the dir? Build step also do not mention `cd`.

manoelmarques:
@matejvasek yes, it is implied in the comment but it seems some people are forgetting the `cd` step. I thought just adding one `cd` statement combined with the comment would be suffiicient. Another option would be like below, with `cd` embedded above each command. Is that what you mean by `cd` missing in build step ?
Screenshot 2023-03-08 at 9 53 11 AM

zroubalik:
I aggree, it is better to be explicit in these examples, so users avoid confusion :)

PR #5461, opened 2023-03-10
Update the usage of Apache Kafka term 

matzew:
## Proposed Changes 

- The PR underlines that we offer implementation/apis for Apache Kafka, not the other way around


pierDipi (approved):
/lgtm
/approve

PR #5467, opened 2023-03-11
Update NOT and CESQL which are not working 

debasishbsws:
`not` do not expect a list of filters, it only expects only one filter. [not_filter.go](https://github.com/knative/eventing/blob/f6ca59b33fb42ff981367b43dc0280841ac324b6/pkg/eventfilter/subscriptionsapi/not_filter.go#L27-L29)

The equality parameter in `cesql` should be `=` instead of `==` to work. [INFO](https://github.com/knative/eventing/discussions/6717#discussioncomment-4859920)

## Proposed Changes 

- change the code part with those fix

### DIFF
--- a/docs/eventing/experimental-features/new-trigger-filters.md
+++ b/docs/eventing/experimental-features/new-trigger-filters.md
@@ -134,7 +134,7 @@ spec:
   filters:
       - not:
           exact:
-              type: com.github.push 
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
Thanks @debasishbsws !

/lgtm
/approve

PR #5475, opened 2023-03-16
Add link to MTChannelBasedBroker developer docs

creydr:
## Proposed Changes

- :gift: Add link to MTChannelBasedBroker developers overview to docs for a deeper view (follow up on https://github.com/knative/eventing/pull/6791)

### DIFF
--- a/docs/eventing/brokers/broker-types/README.md
+++ b/docs/eventing/brokers/broker-types/README.md
@@ -8,6 +8,8 @@ Knative Eventing provides a multi-tenant (MT) channel-based broker implementatio
 
 Before you can use the MT channel-based broker, you must install a channel implementation.
 
+For a more in deep view, you can check out the [developers docs](https://github.com/knative/eventing/blob/main/docs/mt-channel-based-broker/README.md) of the MT channel-based broker.
+
 ## Alternative broker implementations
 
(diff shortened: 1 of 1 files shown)


docs/eventing/brokers/broker-types/README.md
```diff
@@ -8,6 +8,8 @@ Knative Eventing provides a multi-tenant (MT) channel-based broker implementatio
 
 Before you can use the MT channel-based broker, you must install a channel implementation.
 
+For a more in deep view, you can check out the [developers docs](https://github.com/knative/eventing/blob/main/docs/mt-channel-based-broker/README.md) of the MT channel-based broker.
+
```
pierDipi, line 12: Would it be more appropriate to link this in the eventing contributing.md?

I don't think serving is linking https://github.com/knative/serving/blob/main/docs/scaling/SYSTEM.md on the website.

pierDipi: As a user that content is hard to digest, especially if you're getting started, it might be overwhelming to see lots of moving parts

creydr: I agree, contributing.md would be a good spot too. The initial intention for adding this here was, that through the diagram on the page some parts became more clear to me (e.g. the connection between some different config maps), what _could_ be useful for users too. But I what ever fits best...

pierDipi: Is there any page specific to channel-based broker? Can we add it at the end of such page as additional information section and make it clear that this is internal implementation details?

PR #5492, opened 2023-03-18
Document Request Logging Feature on the Website

naveenrajm7:
"Fixes https://github.com/knative/serving/issues/13669" 

## Proposed Changes 

- Add a new page under knative serving observability for documenting request logging options

To start with I have simply copied the content of [observability.yaml](https://github.com/knative/serving/blob/main/config/core/configmaps/observability.yaml) 

Happy to take feedback for changes, if I need to give an example commands on how to configure and see the logs , Also if I need to separate request logging keys from other observability options (like `profiling.enable` )


docs/serving/observability/logging/request-logging.md
```diff
@@ -0,0 +1,21 @@
+# Configuring Request Log Settings
+
+The request logging for knative serving is managed through the `config-observability` ConfigMap in `knative-serving` namespace. The request logs will be printed by the queue-proxy sidecar.
+
+Mentioned below are the flags used to configure request logging and other obervability features.
+
+
+| ConfigMap key                     | Description                                                                                                                                                      |
```
dprotaso, line 8: The only relevant properties for request logging are

                                                                                         
ConfigMap key | Description
-|-
| `logging.enable-request-log` | If true, the request logging will be enabled. |
| `logging.enable-probe-request-log`| If true the probes will be logged
 `logging.request-log-template`      | The value determines the shape of the request logs and it must be a valid go text/template. It is important to keep this as a single line. Multiple lines are parsed as separate entities by most collection agents and will split the request logs into multiple records.

dprotaso:
/lgtm
/approve

thanks for the PR

Issue #5083, opened 2022-07-07
PVC support 


giladk1221:
Hi,
I would like any sample examples for PVC support in Service YAML.
I want to configure the PV/PVC for SMB/NFS shares and I didn't see any references for those configurations.

It will be very helpful to see some examples(step by step) of how to configure it well like your desired configuration.

waiting for any update.

Thanks
Gilad

abrennan89:
@giladk1221 I don't have bandwidth to pick this up right now, but there is some information in our Red Hat documentation here about enabling PVC for Knative Serving: https://docs.openshift.com/container-platform/4.10/serverless/admin_guide/serverless-configuration.html#serverless-enabling-pvc-support_serverless-configuration

Hope this helps for now.

@skonto @knative/serving-wg-leads anyone have any additional info that can be added here to form a full docs example beyond what we have in our docs? I'll write this up properly for knative.dev when I have a chance but might need more info unless this solves the issue.

abrennan89:
I actually found this right after also: https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-persistentvolumeclaim-pvc

The RH docs have an example of the Service etc though which might be more useful.

zzuchen:
> @giladk1221 I don't have bandwidth to pick this up right now, but there is some information in our Red Hat documentation here about enabling PVC for Knative Serving: https://docs.openshift.com/container-platform/4.10/serverless/admin_guide/serverless-configuration.html#serverless-enabling-pvc-support_serverless-configuration
> 
> Hope this helps for now.
> 
> @skonto @knative/serving-wg-leads anyone have any additional info that can be added here to form a full docs example beyond what we have in our docs? I'll write this up properly for knative.dev when I have a chance but might need more info unless this solves the issue.

I want to mount the NFS shares for my pod in knative v0.23.0，but the pvc is supported after knative v1.2，could you please tell me how to solve the problem?

PR #5507, opened 2023-03-23
[WIP] Add PVC docs

abrennan89:
Fixes https://github.com/knative/docs/issues/5083

## Proposed Changes

- Moves services docs to resources section
- Adds PVC info for services

Direct preview link: https://deploy-preview-5507--knative.netlify.app/docs/serving/services/pvc/

### DIFF
--- a/config/nav.yml
+++ b/config/nav.yml
@@ -100,7 +100,7 @@ nav:
           - Creating a Service: serving/services/creating-services.md
           - Configuring Services:
             - Configuring private Services: serving/services/private-services.md
-            - Configuring Persistent Volume Claims for Services: serving/services/pvc.md
+            - Permanent data storage: serving/services/pvc.md
             - Service metrics: serving/services/service-metrics.md
             - Configuring ingress class: serving/services/ingress-class.md
             - Configuring certificate class: serving/services/certificate-class.md
--- a/config/nav.yml
+++ b/config/nav.yml
@@ -95,6 +95,21 @@ nav:
     - Serving:
       - Knative Serving overview: serving/README.md
       - Resources:
+        - Services:
+          - About Knative Services: serving/services/README.md
+          - Creating a Service: serving/services/creating-services.md
+          - Configuring Services:
+            - Configuring private Services: serving/services/private-services.md
+            - Configuring Persistent Volume Claims for Services: serving/services/pvc.md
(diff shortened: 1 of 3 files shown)


docs/serving/services/pvc.md
```diff
@@ -0,0 +1,46 @@
+# Permanent data storage
+
+You can provide permanent data storage for Knative Services by configuring `PersistentVolumeClaims` (PVCs).
```
evankanderson, line 3: Nit: I'd describe this as "shared filesystem storage".  Given the description below, you'd probably also want to highlight that your volume claim will need an `accessMode` of `ReadWriteMany` if you want read-write access.  If you want read-only shared data, you can set the `readOnly` property of the `volume` entry to `true` and use `ReadOnlyMany`.

`ReadWriteOnce` and `ReadWriteOncePod` will not work with Knative autoscaling.

abrennan89: Great feedback, thanks!

config/nav.yml
```diff
@@ -95,6 +95,21 @@ nav:
     - Serving:
       - Knative Serving overview: serving/README.md
       - Resources:
+        - Services:
+          - About Knative Services: serving/services/README.md
+          - Creating a Service: serving/services/creating-services.md
+          - Configuring Services:
+            - Configuring private Services: serving/services/private-services.md
+            - Permanent data storage: serving/services/pvc.md
```
skonto, line 103: I would recommend something like: "Data Storage Access" or "Storage Access", storage can be non permanent. 
In any case to be precise: "A PersistentVolumeClaim (PVC) is a request for storage by a user" that is why I am more inclined to change this.
Unless we are going to split storage to permanent and ephemeral.
For the latter we support [empty dir ]( https://deploy-preview-5507--knative.netlify.app/docs/serving/configuration/feature-flags/#kubernetes-emptydir-volume) which is part of the ephemeral storage volume types [provided by K8s](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#types-of-ephemeral-volumes).

docs/serving/services/pvc.md
skonto: I also provided some naming alternatives above. Besides the access mode, depending on the volume type users may have to adjust ksvc security context wrt to file related properties eg. FSGroup.

PR #5520, opened 2023-03-31
Relabel tab to 'No DNS'

dprotaso:
'Temporary DNS' is confusing since we are not actually setting up DNS in this instance


dprotaso:
/hold

This might be better in a section

dprotaso:
/hold cancel

Actually I think it might be fine

PR #5523, opened 2023-04-02
Add conceptual documentation on Knative Serving request flow

### LINKED ISSUE #5100: Document behavior of Kservice readinessProbe and queue-proxy

Document the behavior of the readinessProbe on a Knative Kservice, and how that is translated over to the Pod queue-proxy and service container.

**Additional context**

The KNative controller will turn the Knative service template into a Pod where the readinessProbe is applied to the queue-proxy, and the service container has no probe. The queue-proxy then re-implements the http-probe of the service container. This can lead to confusing behavior.

### LINKED ISSUE #5173: Add docs on probes
We have some info on what types of probes the user can add to the Services [here](https://github.com/knative/specs/blob/main/specs/serving/runtime-contract.md#meta-requests), but it'd be nice to have a doc (possibly at something like `https://knative.dev/docs/serving/services/user-probes`) that describe what types of probes users can set (readiness/liveness), which ones they can't right now (startup, grpc), and an example of how to configure one.

**Additional context**
Came up in a [slack discussion](https://knative.slack.com/archives/C0186KU7STW/p1660245013942819)

evankanderson:
Fixes #5173
Fixes #5100

Adds architecture / conceptual documentation

## Proposed Changes 

- Added high-level architecture description of the flow of HTTP requests through Knative.
  - (this has been on a to-do list for at least a year)

I'm not sure if this is the best place to add this; I figured I'd start by adding this here, but we may want to link to it from other places.

/cherrypick release-1.9


psschwei (changes requested):
LGTM, just needs to be added to the navigation: https://github.com/knative/docs/blob/d94720c164621b266880efbdb534f45b5b48359e/config/nav.yml#L173

evankanderson:
> LGTM, just needs to be added to the navigation:
> 
> https://github.com/knative/docs/blob/d94720c164621b266880efbdb534f45b5b48359e/config/nav.yml#L173

Thanks, updated!

psschwei (approved):
/lgtm
/approve

PR #5531, opened 2023-04-12
fix go redirect

dprotaso:
we've been getting 404s with netlify with the go get redirect.

I opened an netlify support issue and they suggested that the cause might be because the `_redirect` format isn't valid (`go-get=1`) 

Part of https://github.com/knative/docs/pull/5532


dprotaso:
Going to close this - netlify support said

> We detected a small bug on the way we handled these requests within the same CDN node. Mind trying to hit the 404's are still seen?

dprotaso:
Actually we should still do some cleanup

golang/_redirects
```diff
@@ -5,75 +5,73 @@
...
-/networking/* go-get=1 /golang/networking.html 200
-/operator/* go-get=1 /golang/operator.html 200
-/pkg/* go-get=1 /golang/pkg.html 200
+/actions/* go-get=:go-get /golang/actions.html 200
+/async-component/* go-get=:go-get /golang/async-component.html 200
+/caching/* go-get=:go-get /golang/caching.html 200
+/client/* go-get=:go-get /golang/client.html 200
+/client-pkg/* go-get=:go-get /golang/client-pkg.html 200
+/community/* go-get=:go-get /golang/community.html 200
+/container-freezer/* go-get=:go-get /golang/container-freezer.html 200
```
psschwei, line 14: should we delete this one since we're archiving the repo whenever I remember to actually do it?

dprotaso: When you archive the repo - we can regenerate this config that will remove it.

psschwei (approved):
/lgtm

PR #5561, opened 2023-05-11
Update next-steps.md, remove knative.tips link

rhuss:
Knative.tips is history (not reachable anymore).

Thanks, @ahmetb, for all your support at that time!


ahmetb:
@rhuss please Don't mention user handles in commit messages (practically ever). They cause extra emails every time the repo is forked/rebased/re-pushed, or simply stuff like this: 
image

Issue #5099, opened 2022-07-14
Update the blog template to remove Hugo frontmatter


abrennan89:
## Expected Behavior
Blog template should reflect the format used here https://github.com/knative/docs/pull/5059/files

## Actual Behavior
Blog template still includes old Hugo frontmatter. Leaving this in increases work for docs either during content reviews or afterwards if something is merged using this template and then needs to be cleaned up.

snneji:
@abrennan89  Do you mean this template here: https://github.com/knative/docs/blob/main/contribute-to-docs/templates/template-blog-entry.md#copy-the-template

abrennan89:
No idea tbh, thought I saw an older one at one point with the hugo frontmatter but maybe it was somewhere else.

Either way, the one you're linking to looks like it's missing the author info, date etc and is different to what's [actually being used in the docs](https://knative.dev/blog/articles/eventing-rabbitmq-announcement/) as we decided a while back on the WG call, so either all the existing blogs need to be updated again to match this template, or the template should be updated to match what's actually used in practice.

I don't mind what we use as long as it's consistent and folks have a template that actually aligns with what's on the site.

snneji:
@abrennan89, Do you think we should keep that table with tracking info about reviews and approvals? I'm not sure it's needed since reviews will be logged in GH.

abrennan89:
Nah I think it can be removed

abrennan89:
@snneji it looks like the PR is opened against your fork instead of the docs repo. Would you mind fixing it and we can get this merged then? 🙂

AmanSarraf:
Hi @abrennan89, I want to work on this issue

dprotaso:
@AmanSarraf are you still working on this?

PR #5583, opened 2023-05-30
chore: remove front matter from blog template

lance:
Hugo is no longer used, so this is not needed.

Fixes: https://github.com/knative/docs/issues/5099

/kind cleanup

### DIFF
--- a/blog/docs/events/install-fest-04-2022.md
+++ b/blog/docs/events/install-fest-04-2022.md
@@ -1,6 +1,12 @@
-# Knative InstallFest 2022/04/26
+---
+title: "Knative InstallFest 2022/04/26"
+type: "blog"
+date: 2022-04-06
+linkTitle: "Knative InstallFest 2022/04/26"
+description: "Join the webinar Knative InstallFest on 2022/04/26"
+---
 
-**Date: 2022-04-26**
+
 ## _Leading developers in the community get you up and running fast_
 
 [![Knative](https://global-uploads.webflow.com/5f683649f57c921f6db67087/623216bbbcf4107b28345769_Knative%20InstallFest%20(3).png)](https://triggermesh.zoom.us/webinar/register/WN_QrPd39pPT3uapuq-ttaU0g)
--- a/blog/docs/events/install-fest-04-2022.md
+++ b/blog/docs/events/install-fest-04-2022.md
@@ -1,12 +1,6 @@
----
-title: "Knative InstallFest 2022/04/26"
-type: "blog"
-date: 2022-04-06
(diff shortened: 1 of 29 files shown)


dprotaso:
Should we remove it from existing blog pages?

ie. https://deploy-preview-5583--knative.netlify.app/blog/releases/announcing-knative-v1-7-release/ ?

lance:
> Should we remove it from existing blog pages?
> 
> ie. https://deploy-preview-5583--knative.netlify.app/blog/releases/announcing-knative-v1-7-release/ ?

Sure - I'll do that

dprotaso:
The nav is a bit off - 
Screenshot 2023-05-30 at 15 23 21

lance:
@dprotaso what do you mean? It looks the same as knative.dev as far as I can tell (although I made it consistently prefixed with "Announcing knative"). Here is the published nav 
image

dprotaso:
Is it me or is 1.6 etc missing the period in the middle ?

dprotaso:
The PR deployment is on the right - and it seems like it's making it worse
Screenshot 2023-05-31 at 11 26 10

dprotaso:
maybe we do need the front matter?

lance:
> maybe we do need the front matter?

I just pushed a commit that uses front matter on the 1.9 release and it's definitely better than without.

image

dprotaso:
> But it seems to me like the release announcements are better when front matter is used. I'm inclined to revert the changes to the release announcements, and to add front matter to the handful of them which did not use it (1.6 and up). WDYT?

I agree

lance:
> > But it seems to me like the release announcements are better when front matter is used. I'm inclined to revert the changes to the release announcements, and to add front matter to the handful of them which did not use it (1.6 and up). WDYT?
> 
> I agree

OK - I've done this. There are two articles in the Releases blog posts that I think should retain the new style of heading/title and be moved to the Articles blog posts, since they differ in form and style from all of the other release posts.

dprotaso:
/lgtm
/approve

> Also, it's interesting - I realized when looking at these two outliers that the first top-level heading is used for the menu nav, and set in the meta tags. But there is no author information in meta tags, and the description in the tag is just "Knative Blog".

Wanna make a follow up issue?

lance:
> /lgtm /approve
> 
> > Also, it's interesting - I realized when looking at these two outliers that the first top-level heading is used for the menu nav, and set in the meta tags. But there is no author information in meta tags, and the description in the tag is just "Knative Blog".
> 
> Wanna make a follow up issue?

https://github.com/knative/docs/issues/5588

PR #5598, opened 2023-06-11
docs: fix eventing gitlab source readme

arkadiusjonczek:
Add fix for eventing gitlab source readme

## Proposed Changes

- Use project access token instead of personal access token for gitlab source (which is more secure from security perspective)
-  Use right breadcrumb for gitlab webhook (Settings >> Webhooks)


pierDipi (approved):
Thanks!

/lgtm
/approve

PR #5610, opened 2023-06-27
docs: update instructions on client code-sign verification

BobyMCbobs:
the client binary is code-sign verified with cosign in keyless mode and cosign has changed the way it's keyless code signing works since the docs were made here

## Proposed Changes 

- updates the docs for code-sign verification for client binary

### DIFF
--- a/docs/reference/security/README.md
+++ b/docs/reference/security/README.md
@@ -35,10 +35,9 @@ Our releases from 1.9 are signed with [cosign](https://docs.sigstore.dev/cosign/
     ```
 
 !!! note
-    `COSIGN_EXPERIMENTAL=1` is used to allow verification of images signed
-    in `KEYLESS` mode. To learn more about keyless signing, please refer to
+    Knative images are signed in `KEYLESS` mode. To learn more about keyless signing, please refer to
     [Keyless Signatures](https://github.com/sigstore/cosign/blob/main/KEYLESS.md#keyless-signatures)
-    Our signing identity(Subject) for our releases is `signer@knative-nightly.iam.gserviceaccount.com` and the Issuer is `https://accounts.google.com`
+    Our signing identity(Subject) for our releases is `signer@knative-releases.iam.gserviceaccount.com` and the Issuer is `https://accounts.google.com`
 
 ### Apple macOS
 
--- a/docs/reference/security/README.md
+++ b/docs/reference/security/README.md
@@ -14,15 +14,17 @@ Our releases from 1.9 are signed with [cosign](https://docs.sigstore.dev/cosign/
 
 1. Download the files you want, and the `checksums.txt`, `checksum.txt.pem` and `checksums.txt.sig` files from the releases page:
     ```sh
-    # this example verifies the 1.9.0 kn cli from the knative/client repository
-    wget https://github.com/knative/client/releases/download/knative-v1.9.0/checksums.txt
-    wget https://github.com/knative/client/releases/download/knative-v1.9.0/kn-darwin-amd64
-    wget https://github.com/knative/client/releases/download/knative-v1.9.0/checksums.txt.sig
-    wget https://github.com/knative/client/releases/download/knative-v1.9.0/checksums.txt.pem
+    # this example verifies the 1.10.0 kn cli from the knative/client repository
+    wget https://github.com/knative/client/releases/download/knative-v1.10.0/checksums.txt
+    wget https://github.com/knative/client/releases/download/knative-v1.10.0/kn-darwin-amd64
+    wget https://github.com/knative/client/releases/download/knative-v1.10.0/checksums.txt.sig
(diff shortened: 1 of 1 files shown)


docs/reference/security/README.md
```diff
@@ -14,15 +14,17 @@ Our releases from 1.9 are signed with [cosign](https://docs.sigstore.dev/cosign/
 
 1. Download the files you want, and the `checksums.txt`, `checksum.txt.pem` and `checksums.txt.sig` files from the releases page:
     ```sh
-    # this example verifies the 1.9.0 kn cli from the knative/client repository
-    wget https://github.com/knative/client/releases/download/knative-v1.9.0/checksums.txt
-    wget https://github.com/knative/client/releases/download/knative-v1.9.0/kn-darwin-amd64
-    wget https://github.com/knative/client/releases/download/knative-v1.9.0/checksums.txt.sig
-    wget https://github.com/knative/client/releases/download/knative-v1.9.0/checksums.txt.pem
+    # this example verifies the 1.10.0 kn cli from the knative/client repository
```
psschwei, line 17: nit: the notarized mac binaries section below is still using v1.9 ... I'd either update that one too or just leave this one on v1.9

docs/reference/security/README.md
```diff
@@ -14,15 +14,17 @@ Our releases from 1.9 are signed with [cosign](https://docs.sigstore.dev/cosign/
...
+    wget https://github.com/knative/client/releases/download/knative-v1.10.0/checksums.txt
+    wget https://github.com/knative/client/releases/download/knative-v1.10.0/kn-darwin-amd64
+    wget https://github.com/knative/client/releases/download/knative-v1.10.0/checksums.txt.sig
+    wget https://github.com/knative/client/releases/download/knative-v1.10.0/checksums.txt.pem
     ```
 1. Verify the signature:
     ```sh
-    COSIGN_EXPERIMENTAL=1 cosign verify-blob \
```
psschwei, line 25: We should probably also edit the note starting on L37 now that we've dropped the `COSIGN_EXPERIMENTAL=1` bit (either to note that it's still needed for cosign v1 or just drop that line entirely).

psschwei (commented):
Thanks for the update!

Couple of nits from me, but otherwise looks good from a docs perspective.

/assign @davidhadas @evankanderson for security WG signoff

docs/reference/security/README.md
BobyMCbobs: I can't see anything specific to 1.9 here other than cosign begins with 1.9
https://github.com/knative/docs/pull/5610/files#diff-c61d6cfb7b0c5e5b53a34351906fb92b1f73fdcd3879cf0c5e4d520f7fa58f35L41-L59

docs/reference/security/README.md
BobyMCbobs: updated in 164d5804

docs/reference/security/README.md
psschwei: Yeah, I don't think it's much, just that it's a little jarring to have 1.10 in one place and 1.9 in the other (and maybe there'd be a new timestamp), but it's a pretty minor thing.

psschwei (commented):
/approve
cc @knative/security-wg-leads

docs/reference/security/README.md
BobyMCbobs: To me, it reads as two parts: 
- a note that image are signed from a particular version 
- how to download all the assets to prove a signature

Potentially it could come down to whether the download part is always at the latest or if it stays at one version and folks need to figure out what the latest version is.

I'm OK to revert it to 1.9.0 in the download section if need be.

psschwei:
gentle nudge @davidhadas  @evankanderson

skonto:
@evankanderson @davidhadas gentle ping.

davidhadas:
Do we really need to maintain duplicated documentation?
Can't we instead just point here to the [knative docs](https://knative.dev/docs/client/install-kn/#verifying-cli-binaries) that seem to include this information?

IMHO the Readme can be used to dive into aspects not in the docs, but should not be a copy of the docs forcing us to maintain the information twice.

psschwei:
If the info is used in multiple places, then it should be converted into a snippet

davidhadas:
I am guessing it was first added as a Readme, and later added to Knative docs. 
I think it is enough to point from the Readme to the Knative docs + review the Knative docs and change them if we see a gap there.

psschwei:
Could we merge this as is for now, and then fix as needed afterwards (would be nice to get this in before the next version of the docs is cut next week)

davidhadas:
I will go with @psschwei suggestion - merge it now (to fix wrong documentation). 
Long term, I prefer to avoid duplicates where we need to maintain twice the documentation. 

/lgtm

BobyMCbobs:
Thanks for the merge @psschwei @davidhadas

PR #5619, opened 2023-07-05
Document current behaviour of scaling down with activation scale

xtreme-vikram-yadav:
The changes in this PR only reflect the existing behaviour of scale down operation by autoscaler when activation scale is set until the behaviour is changed in future.

Related Issue: https://github.com/knative/serving/issues/14017


psschwei (approved):
/lgtm
/approve

PR #5628, opened 2023-07-13
Fix nits in fuzzing audit blogpost

AdamKorcz:
Changes the date in the blogpost and report to 13th July which is when they were published.

Adds a line over lists in the blogpost which I believe should fix it on the website (they are not currently displayed as lists).


ReToCode (approved):
/lgtm
/approve

PR #5639, opened 2023-07-26
Added mention of channels to event type autocreate docs

Cali0707:
This adds mention of eventtype auto create supporting the channel as well as the broker.

## Proposed Changes 

- Mention the channel when discussing event type autocreate working with brokers
- Make the Broker section subsections of the example, so that it is clear that this is only an example, and you don't need to use a Broker
-

### DIFF
--- a/docs/eventing/experimental-features/eventtype-auto-creation.md
+++ b/docs/eventing/experimental-features/eventtype-auto-creation.md
@@ -11,7 +11,7 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


Cali0707:
@matzew do you think it is worth adding an example of using this with a channel as well, or is this enough to show that it works with channels as well?

Cali0707:
> Have we done channel auto-create implementation yet? Because my PR was only on broker ingress.

@dsimansk I added it here: https://github.com/knative/eventing/pull/7089

dsimansk (approved):
Thanks @Cali0707!

/lgtm

PR #5644, opened 2023-08-01
fix: rm deprecated step

vincent0426:

## Proposed Changes 

- RabbitMQBroker is deprecated as mentioned here https://github.com/knative-sandbox/eventing-rabbitmq/tree/main/samples/quick-setup


dsimansk:
@knative/eventing-reviewers PTAL if the step is OK to be removed. Thanks!

gabo1208:
What is deprecated is to create a RabbitMQ Broker without a RabbitMQConfig but not the Broker, so this is completely valid still :)

vincent0426:
Oh, my bad. May I ask what makes the change to create both brokerConfig and broker instead of putting brokerConfig right into broker? Thanks!

PR #5651, opened 2023-08-02
Finalize pvc support

skonto:
Fixes https://github.com/knative/docs/issues/5083

## Proposed Changes 

- Continues the work in https://github.com/knative/docs/pull/5507

### DIFF
--- a/docs/serving/services/storage.md
+++ b/docs/serving/services/storage.md
@@ -1,12 +1,35 @@
 # Volume Support for Knative services
 
-By default Serving supports the mounting the [volume types](https://kubernetes.io/docs/concepts/storage/volumes): `emptyDir`, `secret`, `configMap` and `projected`. [PersistentVolumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) are supported but require a [feature flag](../configuration/feature-flags.md) to be enabled.
+You can provide data storage for Knative Services by configuring different types of volumes.
+Serving supports mounting the [volume types](https://kubernetes.io/docs/concepts/storage/volumes): `emptyDir`, `secret`, `configMap` and `projected`.
+[PersistentVolumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) are supported but require a [feature flag](../configuration/feature-flags.md) to be enabled.
 
 !!! warning
     Mounting large volumes may add considerable overhead to the application's start up time.
 
 
-Here is an example of using a persistent volume claim with a Knative Service:
+Bellow there is an example of using a persistent volume claim with a Knative Service.
+
+## Prerequisites
+
+Before you can configure PVCs for a Service, this feature must be enabled in the `KnativeServing` ConfigMap:
+
+```yaml
+...
+spec:
+  config:
+    features:
+      "kubernetes.podspec-persistent-volume-claim": enabled
+      "kubernetes.podspec-persistent-volume-write": enabled
+...
+```
+
+* The `kubernetes.podspec-persistent-volume-claim` extension controls whether persistent volumes (PVs) can be used with Knative Serving.
+* The `kubernetes.podspec-persistent-volume-write` extension controls whether PVs are available to Knative Serving with the write access.
+
+## Procedure
+
+* Modify the PVC configuration for your Service:
 
 ```yaml
 apiVersion: serving.knative.dev/v1
--- a/docs/serving/services/storage.md
+++ b/docs/serving/services/storage.md
@@ -26,7 +26,7 @@ kubectl patch --namespace knative-serving configmap/config-features \
 
 !!! note
    
-    If you have installed Serving via the Knative operator then you need to set the above feature flags ** only ** at the corresponding Serving CR.
+    If you have installed Serving via the Knative operator then you need to set the above feature flags **only** at the corresponding Serving CR.
 
 ## Procedure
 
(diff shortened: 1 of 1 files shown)


config/nav.yml
```diff
@@ -100,6 +100,20 @@ nav:
         - Architecture: serving/architecture.md
         - Request Flow: serving/request-flow.md
       - Resources:
+        - Services:
```
ReToCode, line 103: What is the idea about moving this? I (personally) do not get what why "Resources" and "Developer Topics" are two chapters in the first place. Maybe merge them all to getter?

docs/serving/services/storage.md
```diff
@@ -1,12 +1,35 @@
 # Volume Support for Knative services
 
-By default Serving supports the mounting the [volume types](https://kubernetes.io/docs/concepts/storage/volumes): `emptyDir`, `secret`, `configMap` and `projected`. [PersistentVolumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) are supported but require a [feature flag](../configuration/feature-flags.md) to be enabled.
+You can provide data storage for Knative Services by configuring different types of volumes.
```
ReToCode, line 3: `four` volume types or do you see projected as another definition of the three?

docs/serving/services/storage.md
```diff
@@ -1,12 +1,35 @@
...
 !!! warning
     Mounting large volumes may add considerable overhead to the application's start up time.
 
 
-Here is an example of using a persistent volume claim with a Knative Service:
+Bellow there is an example of using a persistent volume claim with a Knative Service.
+
+## Prerequisites
+
+Before you can configure PVCs for a Service, this feature must be enabled in the `KnativeServing` ConfigMap:
```
ReToCode, line 15: Do we also want to document without the operator? For the yaml install a modification of the config-map?

skonto: Actually we can have a note about the operator but this is the first feature described separately, most are handled in the feature-flags page and there is no mentioning of the operator there. It is only about the config map.

docs/serving/services/storage.md
skonto: Actually I see them as four given that at the K8s api level we have:
```
Volume -> (name, VolumeSource)

VolumeSource -> {EmptyDir, Secret, ConfigMap, Projected, DownwardAPI,...}

Projected ->  {DownwardAPIProjection, ConfigMapProjection, SecretProjection, ServiceAccountProjection }

```
Note that volumeSource `DownwardAPI` is [not supported directly it seems](https://github.com/knative/serving/blob/main/pkg/apis/serving/k8s_validation.go#L187-L202).
The last [PR](https://github.com/knative/serving/pull/13896) on this area was btw the addition of the projected downwardapi.

config/nav.yml
skonto: This was an idea coming from the other PR. Having just revisions didnt make much sense, Services fall under the same category. I will take a look and see if we can have a better grouping but this could be something to handle in a different PR as well.

docs/serving/services/storage.md
ReToCode: Ah, but then the listing is a bit misleading. 

> different types of volumes

and in the next sentence:

> [volume types](https://kubernetes.io/docs/concepts/storage/volumes): `emptyDir`, `secret`, `configMap` and `projected`

Can we change one of those "types" then?

docs/serving/services/storage.md
ReToCode: > in the `KnativeServing` ConfigMap:

But then this is not correct, right? It is either the Custom Resource `KnativeServing` or the `config-features` ConfigMap.

docs/serving/services/storage.md
skonto: These are different type of volumes according to the api.

docs/serving/services/storage.md
skonto: Listing one option does not exclude the other. For completeness sake we could add both options.
So when you deploy Serving with the Operator your API is the Serving CR so config should be changed there.
Operator will propagate the change to the configmap as usual.
When you dont deploy Serving with the Operator configmap is your place where to changed stuff.

skonto:
@ReToCode hi I updated the PR, the re-organization should de done in a follow up PR.

docs/serving/services/storage.md
```diff
@@ -1,12 +1,37 @@
...
```
ReToCode, line 15: Sorry, this is still a bit confusing

> this feature must be enabled in the `KnativeServing` ConfigMap

It's either `KnativeServing` CR OR the config-map

Also then the hint

> Note that if you have installed Serving via the Knative operator then you need to set the above feature flags at the corresponding Serving CR.

Is unnecessary when the example is with the KnativeServing CR already.


Seems like in other pages we assume the YAML installation was used, so maybe it it easier to have a command here to patch the config-map, as we have [here](https://knative.dev/docs/install/yaml-install/serving/install-serving-with-yaml/#install-a-networking-layer)

> kubectl patch configmap/config-network \
  --namespace knative-serving \
  --type merge \
  --patch '{"data":{"ingress-class":"kourier.ingress.networking.knative.dev"}}'

Then we can also keep the note. Ideally we'll also have a section in the operator, describing the config-overwrite mechanism in general.

skonto: > Is unnecessary when the example is with the KnativeServing CR already.

@ReToCode the yaml file you see is not with the Serving CR it is with the configmap I will use the patch just to clarify.

skonto:
@ReToCode pls take a look.

ReToCode (approved):
/lgtm
/approve

PR #5696, opened 2023-09-23
Update on Kind Local Registry Setup in knative-quickstart-plugin Doc

Priyansurout:

## Proposed Changes 
[Concept] : I have added a note to quickstart-install.md. It states that for 
[KinD](https://kind.sigs.k8s.io/), a local registry will no longer be created by default. Users will now need to pass the --registry flag if they wish to create a local registry. 
This can be done using the command: `kn quickstart kind --registry`.
-
-
-


ReToCode:
See https://github.com/knative/docs/issues/5778.

/close

PR #5715, opened 2023-10-13
Source cloudevents revision

Micah-Gafford:
When doing the `Creating your first source` section of the `Using a Knative Service as a source` tutorial I noticed that the service revision created was consistently labeled with `00001` rather than being randomly generated as the doc would imply. 

## Proposed Changes 

I updated the expected output to have a revision of `00001` rather than `vwybw-1`. I also removed the `Why is my Revision name something different!` notice since the generated revision is no longer variable.


nak3:
/assign @pierDipi @creydr @Leo6Leo @Cali0707 
Could you please take a look?

creydr (commented):
/lgtm

PR #5726, opened 2023-10-21
Helloworld rust/bump dependencies

tatodorov:
This PR updates the dependencies and fixes the following error when trying to build the sample code:
```bash
package `socket2 v0.5.5` cannot be built because it requires rustc 1.63 or newer, while the currently active rustc version is 1.62.1
```
In addition, it updates the steps to recreate the sample code as it is outdated and following it, leads to the following errors:
```bash
  error: edition 2021 is unstable and only available with -Z unstable-options.

  error: could not compile `proc-macro2
```

## Proposed Changes 

- Bump the versions of all dependencies to the latest ones
- Bump Rust to version 1.73.0
- Update the README.md with up-to-date steps to reproduce sample code

### DIFF
--- a/code-samples/community/serving/helloworld-rust/Cargo.lock
+++ b/code-samples/community/serving/helloworld-rust/Cargo.lock
@@ -3,23 +3,27 @@
 version = 3
 
 [[package]]
-name = "aho-corasick"
-version = "0.7.18"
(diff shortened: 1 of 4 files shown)


nak3:
/lgtm
/approve

Thank you!

PR #5753, opened 2023-11-09
Add support s390x and ppc64le archs

### LINKED ISSUE #5718: Code samples should support `s390x` and `ppc64le` archs
https://github.com/knative/docs/blob/c522018a2474384778c970b3487778207105fe3e/.github/workflows/knative-docs-image.yaml#L40-L54

As the code samples are used in our tests, they should support the same range or architectures as we do the tests with. We should add support for `s390x` and `ppc64le` archs.

/kind feature

lysliu:
"Fixes #5718 " or "Add support s390x and ppc64le archs"

## Proposed Changes 

-
-
-

### DIFF
--- a/.github/workflows/knative-docs-image.yaml
+++ b/.github/workflows/knative-docs-image.yaml
@@ -41,7 +41,7 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           push: true
-          platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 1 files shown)


dprotaso:
/hold 

Currently our OSS releases don't work on these architectures

See for a discussion: https://github.com/knative/infra/issues/344

PR #5773, opened 2023-11-25
Documentation - progress deadline annotation on per revision basis

### LINKED ISSUE #5767: Document progressDeadline annotation on per revision basis
**Documentation request**

The feature to set progressDeadline on a per-revision basis has been recently introduced. Previously, this setting was only configurable on a global level. However, there appears to be a lack of documentation regarding this new capability.
- https://github.com/knative/serving/issues/12743

Current documentation for global progress deadline can found [here](https://knative.dev/docs/serving/configuration/deployment/#configuring-progress-deadlines).

Possibly documentation can be similar to target RPS: https://knative.dev/docs/serving/autoscaling/rps-target/#configuring-the-requests-per-second-rps-target

here is an example of the annotation:

```
kind: Service
metadata:
  name: hello
spec:
  template:
    metadata:
      annotations:
         serving.knative.dev/progress-deadline: "45s" 
    spec:
      containers:
        - image: ghcr.io/knative/helloworld-go:latest
          ports:
            - containerPort: 8080
          env:
            - name: TARGET
              value: "World"

```

navinag1989:
Fixes #5767

### DIFF
--- a/docs/serving/configuration/deployment.md
+++ b/docs/serving/configuration/deployment.md
@@ -73,21 +73,45 @@ You may want to configure this setting as a higher value if any of the following
 
 See the [Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#progress-deadline-seconds) for more information.
 
-The following example shows a snippet of an example Deployment Config Map that sets this value to 10 minutes:
+Progress deadline setting can be configured at global level through a ConfigMap or at the per-revision level using an annotation.
 
-```yaml
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: config-deployment
-  namespace: knative-serving
-  labels:
-    serving.knative.dev/release: devel
-  annotations:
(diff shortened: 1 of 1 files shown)


psschwei (commented):
/approve

Probably not a bad idea to include the operator config option here too (that would be a global change, so use a heading like `Global (Operator)`), the details of how to do so can be found here: https://knative.dev/docs/install/operator/configuring-with-operator/

Note: this will need to be cherry-picked into the current release branch in order to show up on the main site

ReToCode (approved):
> Probably not a bad idea to include the operator config option here too (that would be a global change, so use a heading like Global (Operator)), the details of how to do so can be found here: https://knative.dev/docs/install/operator/configuring-with-operator/

Agreed, if we want to do this, we should have it on all examples or reference the page you linked. I don't think it makes sense to introduce it just here, so

/lgtm
/approve

/cherry-pick release-1.12

Issue #5759, opened 2023-11-14
Errors when deploying newest version of Knative CRDs


gunpuz:
## Expected Behavior

Knative CRDS would deploy without STDERRs

## Actual Behavior

Standart error is reported about incorrect CRD: 

```ps1
> kubectl apply -f https://github.com/knative/serving/releases/download/knative-v1.12.0/serving-crds.yaml
customresourcedefinition.apiextensions.k8s.io/certificates.networking.internal.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/configurations.serving.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/clusterdomainclaims.networking.internal.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/ingresses.networking.internal.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/metrics.autoscaling.internal.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/podautoscalers.autoscaling.internal.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/revisions.serving.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/routes.serving.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/serverlessservices.networking.internal.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/services.serving.knative.dev unchanged
customresourcedefinition.apiextensions.k8s.io/images.caching.internal.knative.dev unchanged
The CustomResourceDefinition "domainmappings.serving.knative.dev" is invalid: status.storedVersions[0]: Invalid value: "v1alpha1": must appear in spec.versions
```

## Steps to Reproduce the Problem

1. `kubectl apply -f https://github.com/knative/serving/releases/download/knative-v1.12.0/serving-crds.yaml`

jasonsmithio:
I just deployed on a GKE cluster without issue. Tell me more about your setup

gunpuz:
What info would be usefull?

jasonsmithio:
What version of kubernetes? Where is it hosted (MiniKube, GKE, EKS, etc.)? What else is installed on the cluster (specifically at the time of deploying knative)

gunpuz:
Its hosted on Ubuntu: `microk8s`
Kubernetes version: `1.28`

There is already some custom services, helm charts... , including Knative Serving, Knative Eventing

ReToCode:
[...]

PR #5785, opened 2023-12-04
Use create instead of apply

ReToCode:
- Use `kubectl create` instead of `kubectl apply` because of https://github.com/kubernetes/kubernetes/issues/44501
- Follow-up for https://github.com/knative/docs/issues/5759#issuecomment-1838550871

/assign @skonto 

/cherry-pick release-1.12


skonto:
I verified the PR: 

```
kubectl apply -f https://github.com/knative/serving/releases/download/knative-v1.11.2/serving-post-install-jobs.yaml
error: from storage-version-migration-serving-: cannot use generate name with apply
```
create works as expected.

PR #5798, opened 2023-12-19
Fix command kn quickstart to be kn-quickstart in install docs

schokocaylee:
The command `kn quickstart` cannot be found as the binary is named `kn-quickstart`.
This PR renames the command to match the binary name.

## Proposed Changes 

- Rename the command `kn quickstart` to `kn-quickstart`

### DIFF
--- a/docs/snippets/quickstart-install.md
+++ b/docs/snippets/quickstart-install.md
@@ -33,7 +33,7 @@ To get started, install the Knative `quickstart` plugin:
     1. Verify that the plugin is working by running the command:
 
         ```bash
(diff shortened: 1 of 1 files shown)


ReToCode:
`kn quickstart` is correct, you need to make sure it is correctly named and in your path, then `kn` can find the plugin, check https://knative.dev/docs/getting-started/quickstart-install/#install-the-knative-quickstart-plugin.

/hold

schokocaylee:
Oh, I missed the step to install Knative CLI, sorry...works flawlessly now

PR #5806, opened 2024-01-04
Enable copy for code blocks

ReToCode:
Context: https://cloud-native.slack.com/archives/C04LY5G9ED7/p1704358780773299

## Proposed Changes
- Enables copy buttons on code (shell) blocks

/assign @skonto

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -19,6 +19,7 @@ theme:
     - navigation.tracking
     - navigation.tabs.sticky
     - navigation.top
+    - content.code.copy
 
 markdown_extensions:
   # - mdx_include:


skonto:
Out of curiosity do we want to have this button everywhere or just with snippets (starting with ` ```bash`)? Right now it is everywhere. Does not hurt imho (checked other projects eg. Istio which have the same).

ReToCode:
> Out of curiosity do we want to have this button everywhere or just with bash commands (starting with  ```bash)? Right now it is everywhere. Does not hurt imho (checked other projects eg. Istio which have the [same](https://istio.io/latest/docs/ops/diagnostic-tools/istioctl-analyze/)). WDYTH?

I think it would be sufficient for bash, but seems this is the way mkdocs works. Not sure if we could tweak every place we have a code block somehow.

PR #5867, opened 2024-02-15
queue-sidecar annotations should be defined under spec.template.metad…

mgencur:
…ata.annotations

Otherwise they're not propagated to the Revision.

## Proposed Changes 

- Move the annotations to the right place


skonto:
/lgtm

Wrong pasting :)

dprotaso:
Should we cherry pick this back to 1.13 and 1.12 branch?

mgencur:
> Should we cherry pick this back to 1.13 and 1.12 branch?

Yes. I think so.

PR #5878, opened 2024-02-23
Implemented sliding animation for the adopter logos

prajjwalyd:
Addresses - https://github.com/knative/community/issues/1416

## Proposed Changes 

I've been working on implementing the sliding animation for the adopters' logos.

[scrnli_2_19_2024_1-16-44 AM.webm](https://github.com/knative/community/assets/111794524/13d91cc6-8396-423a-83cf-19181c9dd184)

I haven't used any external libraries for this...
Also, the logos along with their links are dynamically rendered as per the instructions.
Public preview link: https://prajjwalyd.github.io/docs/


overrides/home.html
```diff
@@ -331,20 +319,60 @@ <h2>Follow us on Twitter / X</h2>
...
-      logos[i].style.order = Math.floor(Math.random() * 99);
+  document.addEventListener("DOMContentLoaded", function() {
+    const adopters = [
+      { logoPath: "images/corporate-logos/Google.svg", url: "https://www.google.com" },
+      { logoPath: "images/corporate-logos/vmware.svg", url: "https://www.vmware.com" },
+      { logoPath: "images/corporate-logos/IBM.svg", url: "https://www.ibm.com" },
+      { logoPath: "images/corporate-logos/Redhat.svg", url: "https://www.redhat.com" },
+      { logoPath: "images/corporate-logos/triggermesh_logo.svg", url: "https://www.triggermesh.com" }
+      
+    ];
```
aliok, line 330: Can you also add the adopters here?

https://github.com/knative/community/blob/main/ADOPTERS.MD

Unfortunately, we don't have their logos handy :(

aliok: And, can you make the whole list randomized? Not statically randomized, but randomized in the runtime?

I mean, it would be nice if we see the adopters+vendors in random order every time we refresh this page.

aliok (commented):
Thanks @prajjwalyd this looks great already!

I added some comments inline.

I just realized something though. If we add the adopters too (on top of current list of vendors), we might need a 2nd row. Ideally scrolling in the opposite direction, like CNCF does.

What do you think?

overrides/home.html
dprotaso: We should drop trigger mesh - they're no longer in business

prajjwalyd: The latest changes are as follows:
1. Added more corporate logos by following - https://github.com/knative/community/blob/main/ADOPTERS.MD
2. Mixed the adopters and vendors under the 'Trusted by' heading.
3. Added a second row that scrolls to the right.
4. Randomized the logo display order on each refresh.
5. Dropped Trigger mesh.

prajjwalyd: The current list is not yet fully completed, and there are some aspect ratio fixes needed for certain images...
however, the logic has been implemented, and we simply need to continue adding the remaining logos while ensuring that the aspect ratio is maintained for all the images. 
Let me know if this approach seems acceptable, and then I can proceed with the final adjustments!

aliok: Looks great!

As you said, some images need some work (aspect ratio, whitespace padding), but the overall logic looks good to me.

Special thanks for keeping the commit history clean!

I just noticed that padding is a little bit too much on mobile screens, but that's another story. Our website is not properly responsive.

Cali0707 (commented):
/lgtm
/cc @aliok

aliok:
/lgtm
/approve

Thanks a lot @prajjwalyd 

As we talked in Slack, let's create tickets for the missing logos.

PR #5896, opened 2024-03-07
Sample Bookstore App: DB setup for Book Review Comments

### LINKED ISSUE #5881: Sample App: Database Service: Schema design and database creation
### Problem

Create a SQL file that contains the SQL commands to create the table as describe in [this doc.](https://docs.google.com/document/d/1dYvNdHFf4qoiIbWWOWMLAXDAFpEjjRsa9lP27tpqPTI/edit)

Additionally, we should include 1-2 fake data as the sample.

### [Persona:](https://github.com/knative/eventing/blob/main/docs/personas.md)
Which persona is this feature for?
Administrators and developers

### Exit Criteria
- A SQL file that contains the SQL command to create the table and insert the fake data
- A detailed documentation that explains it

**Time Estimate** (optional):
How many developer-days do you think this may take to resolve? 
1

Zazzscoot:
Fixes #5881 

## Proposed Changes
* Added the SQL file for creating a table for BookReviews according to the schema described [here](https://docs.google.com/document/d/1dYvNdHFf4qoiIbWWOWMLAXDAFpEjjRsa9lP27tpqPTI/edit)
* Added dummy data into BookReviews table
* Wrote documentation for implemented schema and added dummy rows

### DIFF
--- a/code-samples/eventing/bookstore-sample-app/db/sample.sql
+++ b/code-samples/eventing/bookstore-sample-app/db/sample.sql
@@ -4,15 +4,13 @@ CREATE TABLE BookReviews(
   content TEXT NOT NULL,
   sentiment TEXT,
   CONSTRAINT sentiment_check CHECK (sentiment IN ('positive', 'negative', 'neutral')),
-  book_id SERIAL NOT NULL,
-  
 );
 
-INSERT INTO BookReviews VALUES(1, '2020-01-01 00:00:00', 'This book is great!', 'positive', 100);
-INSERT INTO BookReviews VALUES(2, '2020-01-02 00:02:00', 'This book is terrible!', 'negative', 2);
-INSERT INTO BookReviews VALUES(3, '2020-01-03 00:01:30', 'This book is okay.', 'neutral', 3);
-INSERT INTO BookReviews VALUES(4, '2020-01-04 00:00:00', 'Meh', 'neutral', 45);
-INSERT INTO BookReviews VALUES(5, '2020-01-04 00:04:00', 'I HATE THIS BOOK!', 'negative', 77);
-INSERT INTO BookReviews VALUES(6, '2020-01-06 00:00:00', 'Best thing I have ever read!', 'positive', 100);
-INSERT INTO BookReviews VALUES(7, '2020-01-07 00:12:00', 'ABSOLUTELY LOVELY!', 'positive', 50);
-INSERT INTO BookReviews VALUES(8, '2020-01-08 00:00:00', 'Most boring book I picked up', 'negative', 42);
+INSERT INTO BookReviews VALUES(1, '2020-01-01 00:00:00', 'This book is great!', 'positive');
+INSERT INTO BookReviews VALUES(2, '2020-01-02 00:02:00', 'This book is terrible!', 'negative');
+INSERT INTO BookReviews VALUES(3, '2020-01-03 00:01:30', 'This book is okay.', 'neutral');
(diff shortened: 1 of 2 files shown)


code-samples/eventing/bookstore-sample-app/db/sample.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE BookReviews(
```
Cali0707, line 1: We should verify that the table does not exist before creating it. Also small nit on the naming of the table

```suggestion
CREATE TABLE IF NOT EXISTS book_reviews(

```

code-samples/eventing/bookstore-sample-app/db/sample.sql
```diff
@@ -0,0 +1,12 @@
...
+  post_time datetime NOT NULL,
+  content TEXT NOT NULL,
+  sentiment TEXT,
+  CONSTRAINT sentiment_check CHECK (sentiment IN ('positive', 'negative', 'neutral')),
+);
+
+INSERT INTO BookReviews VALUES(1, '2020-01-01 00:00:00', 'This book is great!', 'positive');
+INSERT INTO BookReviews VALUES(2, '2020-01-02 00:02:00', 'This book is terrible!', 'negative');
+INSERT INTO BookReviews VALUES(3, '2020-01-03 00:01:30', 'This book is okay.', 'neutral');
+INSERT INTO BookReviews VALUES(4, '2020-01-04 00:00:00', 'Meh', 'neutral');
```
Cali0707, line 12: This is probably better as a single set of inserts, and also when you have a `SERIAL` type in postgres it is best to let postgres increment that field rather than doing it yourself otherwise you can run into scenarios where postgres doesn't know about the newest value you have
```suggestion
INSERT INTO book_reviews (post_time, content, sentiment) VALUES 
    ('2020-01-01 00:00:00', 'This book is great!', 'positive'),
    ('2020-01-02 00:02:00', 'This book is terrible!', 'negative'),
    ('2020-01-03 00:01:30', 'This book is okay.', 'neutral'),
    ('2020-01-04 00:00:00', 'Meh', 'neutral');
```

Cali0707 (changes requested):

code-samples/eventing/bookstore-sample-app/db/sample.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS book_reviews(
+
+  id SERIAL PRIMARY KEY,
+  post_time datetime NOT NULL,
```
Cali0707, line 4: Postgres doesn't support datetime as a type, it uses timestamp instead
```suggestion
  post_time timestamp NOT NULL,
```

code-samples/eventing/bookstore-sample-app/db/sample.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS book_reviews(
+
```
Cali0707, line 2:
```suggestion

```

Cali0707 (changes requested):
Two more small changes then this is good on my end @Zazzscoot !

code-samples/eventing/bookstore-sample-app/db/sample.sql
Zazzscoot: Thank you so much for the reviews!

Cali0707 (approved):
/lgtm
/approve

Thanks for your work on this @Zazzscoot !!

PR #5901, opened 2024-03-09
The min-scale annotation, which is deprecated, has been moved to minScale.

kahirokunn:
The annotation of these two kebab cases will follow the Camel case since it is the current official support
https://github.com/knative/client/pull/1498

## Proposed Changes 

-
-
-


ReToCode:
/hold

it is the other way around, see https://github.com/knative/serving/issues/12103. The new annotations are the ones with kebab case.

PR #5933, opened 2024-04-01
Docs: Change namespace-wildcard-cert-selector value type from object to string

hubwoop:
This PR proposes changes the docs on automatic TLS certificate provisioning. In the examples given under "Provisioning certificates per namespace (wildcard certificates)" objects are provided as values for `namespace-wildcard-cert-selector`, I think only strings can be values for ConfigMap data values. In the next section (Configure config-certmanager ConfigMap) a string is provided. 

## Proposed Changes 

- ConfigMap's data values can only be of type String


ReToCode (approved):
I think you are right, the existing method does not work. Thanks for the fix.

/lgtm
/approve

PR #5937, opened 2024-04-07
fix correct display

waitstory:
"kafka-source-dispatcher" is a statefulset app, not a deployment app, so it needs to be modified. #


Leo6Leo:
@waitstory Thanks for catching that! 
/lgtm

pierDipi (approved):
/lgtm
/approve

PR #5949, opened 2024-04-25
Intro to open source with knative blog series p3

Leo6Leo:
This PR contains a draft of the first post in the blog series we proposed recently to make it easier for new contributors to contribute code. The proposal can be seen [here](https://docs.google.com/document/d/1RZIeFEuwXU7PfvS0zr79AIs4TurhRTnVFsvldA46zlc/edit?usp=sharing)
## Proposed Changes 

- Create post 3: API Changes


code-samples/eventing/bookstore-sample-app/db/README.md
```diff
@@ -10,7 +10,7 @@ Our goal is to deploy a PostgreSQL pod within Kubernetes, loaded with the sample
 
 ## Overview
 ### The Database Schema
-The BookReviews table contains all reviews made on the bookstore website. 
+The BookReviews table https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#custom-resources all reviews made on the bookstore website.
```
Cali0707, line 13: @Leo6Leo why are these changes here? This seems unrelated to the PR

Leo6Leo: Removed.

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,152 @@
...
+
+![](/blog/images/getting-started-blog-series/post3/001.png)
+
+Hello and welcome back to the blog series! In this blog post we will start making actual code changes to some Knative code. Specifically, we will be learning what an API is in Knative, and how we can change them. To facilitate this and the future coding focused blog posts, we will be building a simple event source based on the Knative sample source code. Our source will function similarly to the sample source and send an event at an interval, except it will allow you to create a text template which will be populated with variables on every event that is sent. By the end of this blog post you should be able to understand what APIs are in Knative, and have modified the sample source API to support our new functionality.
+
+## What is an API?
+![](/blog/images/getting-started-blog-series/post3/002.png)
+An API is an “Application Programming Interface”. You can think of it as a contract, where an application provides an interface to users, which the users can use to programmatically specify what they want the application to do for them
```
Cali0707, line 11: nit: this comma is unnecessary
```suggestion
An API is an “Application Programming Interface”. You can think of it as a contract, where an application provides an interface to users which the users can use to programmatically specify what they want the application to do for them
```

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,152 @@
...
+
+## How can we change an API?
+
+![](/blog/images/getting-started-blog-series/post3/004.png)
+Now that we know what APIs are (generally, and in the context of Knative), we are ready to explore how we can make changes to them. When we want to update an API there are two, maybe three steps that you need to take.
+Update the struct for the object. In our case, we want to update the SampleSource struct. In Knative, these are often found in pkg/apis/<groupname>/<version>/<resourcename>_types.go. Looking at this struct, we can see that there is a structure to it that all resources generally follow:
```
Cali0707, line 23: This should have inline code styling. Also, for some reason it renders as pkg/apis///_types.go in the rendered site. We should verify it is correct once the inline code styling is applied, if not we may need to look at this area more carefully to figure how to render the placeholder names
```suggestion
Update the struct for the object. In our case, we want to update the SampleSource struct. In Knative, these are often found in `pkg/apis/<groupname>/<version>/<resourcename>_types.go`. Looking at this struct, we can see that there is a structure to it that all resources generally follow:
```

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,152 @@
...
+
+![](/blog/images/getting-started-blog-series/post3/005.png)
```
Cali0707, line 25: Is there any way to make this image shorter? It is VERY large on the rendered site

Cali0707: Also, we should give some kind of context to this image in the text. It feels very out of place when reading - you are expecting to see a block of code, and then for some reason you see a large image of a file directory, and it is unclear what the relation is of this image to the code block below

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,152 @@
...
+    // Spec holds the desired state of the SampleSource (from the client).
+    Spec SampleSourceSpec `json:"spec"`
+
+    // Status communicates the observed state of the SampleSource (from the controller).
+    // +optional
+    Status SampleSourceStatus `json:"status,omitempty"`
+}
+```
+
+Looking at this code, we can see that there is a metav1.TypeMeta embedded struct, a metav1.ObjectMeta embedded struct, and then a Spec and a Status embedded struct. This is how Knative objects (and most k8s resources) are structured. The TypeMeta and ObjectMeta provide metadata about the type and the object, and generally all you need to know about them is that the struct needs to include them. The Spec and the Status are the structs that you will actually be modifying. The Spec is where all the options we give the users of Knative will live while the Status is where all the information we give to users about the status of a specific object is stored.
```
Cali0707, line 45:
```suggestion
Looking at this code, we can see that there is a `metav1.TypeMeta` embedded struct, a `metav1.ObjectMeta` embedded struct, and then a `Spec` and a `Status` embedded struct. This is how Knative objects (and most k8s resources) are structured. The TypeMeta and ObjectMeta provide metadata about the type and the object, and generally all you need to know about them is that the struct needs to include them. The Spec and the Status are the structs that you will actually be modifying. The `Spec` is where all the options we give the users of Knative will live while the `Status` is where all the information we give to users about the status of a specific object is stored.
```

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,152 @@
...
+    // MessageTemplate is the text/template which is used for every event sent.
+    //
+    // The string format is any valid go text/template template string where
+    // the fields are any variables set in the config map
+    MessageTemplate string `json:"messageTemplate"`
+}
+
+```
+
+1. All we added was the MessageTemplate string to the spec. Note the JSON tag - this is very important! Without a JSON tag, this field will not be read out of a JSON object when the controller receives info about it from the API Server, and the value will not be written to the JSON we send back to the API Server, so it would not be stored in etcd.
```
Cali0707, line 85: nit: let's not use numbers for these sections, since we already have the numbers in the images. Also, the images break up the list formatting so only the first one is formatted properly
```suggestion
All we added was the MessageTemplate string to the spec. Note the JSON tag - this is very important! Without a JSON tag, this field will not be read out of a JSON object when the controller receives info about it from the API Server, and the value will not be written to the JSON we send back to the API Server, so it would not be stored in etcd.
```

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,152 @@
...
+
+![](/blog/images/getting-started-blog-series/post3/007.png)
+2. After updating the struct, we normally want to update the [codegen](https://www.redhat.com/en/blog/kubernetes-deep-dive-code-generation-customresources){:target="_blank"}. Knative uses custom code generators to automatically implement parts of the reconciler (more on that in coming blog posts), as well as [deep copy](https://stackoverflow.com/questions/184710/what-is-the-difference-between-a-deep-copy-and-a-shallow-copy){:target="_blank"} functions for the structs and auto-generated API documentation. In our case, we want to update the DeepCopy function for the SampleSourceSpec struct. To update the generated code in Knative, you just need to run ./hack/update-codegen.sh. There may be specific versions of dependencies you need to run this in a given repository, so always check the DEVELOPMENT.md file when setting up your repository for development.
```
Cali0707, line 88:
```suggestion
After updating the struct, we normally want to update the [codegen](https://www.redhat.com/en/blog/kubernetes-deep-dive-code-generation-customresources){:target="_blank"}. Knative uses custom code generators to automatically implement parts of the reconciler (more on that in coming blog posts), as well as [deep copy](https://stackoverflow.com/questions/184710/what-is-the-difference-between-a-deep-copy-and-a-shallow-copy){:target="_blank"} functions for the structs and auto-generated API documentation. In our case, we want to update the DeepCopy function for the SampleSourceSpec struct. To update the generated code in Knative, you just need to run ./hack/update-codegen.sh. There may be specific versions of dependencies you need to run this in a given repository, so always check the DEVELOPMENT.md file when setting up your repository for development.
```

Cali0707 (changes requested):
Nice start @Leo6Leo !

Just a couple changes needed for clarity/better rendering and then we should be good to go here!

blog/docs/articles/getting-started-blog-p3.md
Cali0707: Whatever you think is best, maybe we could also not use an image but just mention the file path in the actual text? I don't feel like this image is currently providing a lot of value

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,152 @@
...
+
+![](/blog/images/getting-started-blog-series/post3/003.png)
+
+Once Kubernetes is aware of the CRD (for example, by kubectl apply -f mycrd.yaml), users can interact with the custom resources as if they were normal kubernetes resources. For example, if I wanted to get all the brokers in namespace my-namespace, I could run kubectl get brokers -n my-namespace. This is a really powerful concept, and is a large part of the philosophy of Knative: we provide Kubernetes Native resources to users so that they can more easily build applications on Kubernetes. In Knative we provide APIs to users by using the core Kubernetes API of [CustomResourceDefinitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#custom-resources){:target="_blank"} so that we can realize our philosophy of being Kubernetes Native.
+
+## How can we change an API?
+
+![](/blog/images/getting-started-blog-series/post3/004.png)
+Now that we know what APIs are (generally, and in the context of Knative), we are ready to explore how we can make changes to them. When we want to update an API there are two, maybe three steps that you need to take.
```
nainaz, line 22: make it either 2 steps or 3 steps and not both. Let's remove the ambiguity. :)

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,152 @@
...
+Update the struct for the object. In our case, we want to update the SampleSource struct. In Knative, these are often found in `pkg/apis/<groupname>/<version>/<resourcename>_types.go`. Looking at this struct, we can see that there is a structure to it that all resources generally follow:
```
nainaz, line 23: While reading it seems Update the struct is the first step. I would start this as a new para and also it is hard to follow where one step ended and second began.

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,152 @@
...
+    // MessageTemplate is the text/template which is used for every event sent.
+    //
+    // The string format is any valid go text/template template string where
+    // the fields are any variables set in the config map
+    MessageTemplate string `json:"messageTemplate"`
+}
+
+```
+
+All we added was the MessageTemplate string to the spec. Note the JSON tag - this is very important! Without a JSON tag, this field will not be read out of a JSON object when the controller receives info about it from the API Server, and the value will not be written to the JSON we send back to the API Server, so it would not be stored in etcd.
```
nainaz, line 85: this spec should `Spec`.

nainaz:
Great start and left some comments. 
Maybe that was the intention of the blog post but to me it felt that it ended abruptly. I didn't "see" any result or output. 
I just changed a file, got suggestion to change the other file but then to what end.

Cali0707:
>  Maybe that was the intention of the blog post but to me it felt that it ended abruptly. I didn't "see" any result or output.

@nainaz that's a great point, we should probably improve the conclusion a bit

The idea is that readers would then continue to the future blog posts we are working on and make the rest of the required changes there. But, I can see how that could be unclear or leave someone unsatisfied reading the current post.

nainaz:
Making it more clear would be helpful.
I probably missed the last piece of the blog. I still think adding what
would this "change" ultimately accomplish would be better
Thank you,
-N

blog/docs/articles/getting-started-blog-p3.md
```diff
@@ -0,0 +1,166 @@
...
+            status:
+              type: object
+              properties:
+                status:
+                  type: string
+                sinkUri:
+                  type: string
+```
+
+By following these steps, you have successfully modified the SampleSource API to support a new feature: text templates. This change allows users to create a text template populated with variables for each event sent by the source. This demonstrates the power and flexibility of Knative's API system and how you can extend it to meet your specific needs.
```
Cali0707, line 160:
```suggestion
By following these steps, you have successfully modified the SampleSource API to support a new feature: text templates. This change allows users to provide a text template which we will use in future blog posts to populate variables for each event sent by the source. This demonstrates the power and flexibility of Knative's API system and how you can extend it to meet your specific needs.
```

Cali0707 (approved):
/lgtm

Thanks @Leo6Leo !!

PR #5804, opened 2024-01-03
Add documentation for Serving Encryption

ReToCode:
Fixes #5805
Fixes https://github.com/knative/docs/issues/5800

## Proposed Changes
- Adds documentation for new Serving encryption features: https://github.com/orgs/knative/projects/63/views/1
- Consolidates existing pages to new structure
- Adds an overview of Serving encryption parts
- Renames the features to align to the parts in overview
- Delegates the documentation to certbot, certmanager and trust-manager instead of repeating it inline
- Fixes broken documentation links

PR #5955, opened 2024-05-01
Improve Serving encryption docs

ReToCode:
Addressing https://github.com/knative/docs/pull/5804#discussion_r1584701059

### DIFF
--- a/docs/serving/encryption/configure-certmanager-integration.md
+++ b/docs/serving/encryption/configure-certmanager-integration.md
@@ -70,14 +70,14 @@ As this is also not possible via ACME protocol (DNS01/HTTP01), you need to confi
 
     ```yaml
     apiVersion: cert-manager.io/v1
-    kind: ClusterIssuer
+    kind: ClusterIssuer # this issuer is used by cert-manager to sign all certificates
     metadata:
-      name: selfsigned-cluster-issuer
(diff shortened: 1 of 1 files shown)


rhuss (commented):
Thanks @ReToCode for adressing. Just a minor nit, but looks good to me and I think it makes it easier for peoples to understand.

You could also use something like callouts, which would render nicely in asciidoc, but could make sense, here, too. Maybe even our HTML rendering framework does support something like this ? 🤔 

```
kind:ClusterIssuer    
metadata:
  name: cluster-selfsigned-user  # <1>
....
....
# <1>: Issuer used by cert-manager to ...
#      .... <next line>
# <2>: Knative specific issuer that will uses the CA stored
       in the secret "<name>" which has been created by the
       certificate "<name>"
```

But that's just some icing on the cake, so I approve it with a `/hold`. Feel free to adjust or merge.

/approve
/lgtm
/hold

ReToCode:
@rhuss I don't think it is supported with the config used here:
image
Checked with and without the #.

ReToCode:
/unhold

@rhuss PTAL

ReToCode:
/unhold

Let's revisit this as a separate topic. If we do that, we should probably do it as a general pattern :)

PR #5962, opened 2024-05-10
Apply New Designs to Case Studies Section

rohitsrma:
In this PR I've implemented the new designs for the **case studies** section, as provided in the [Figma design](https://www.figma.com/file/3arTeN4tR3byPKSkrzTibp/Knative-Website?type=design&node-id=389%3A52&mode=design&t=83FbvBnMeqYt7lIe-1) shared by @mmejia02

### Before

Untitled design (1)

### After

new (2)


shivamgupta2020:
@rohitsrma, I think you should fix the following:-
1. The font style should be changed to Roboto.
2. In mobile mode, the margin should be increased between two blocks.
3. the "case study" font and style do not match the design in mobile mode.

PR #5963, opened 2024-05-11
fixed the aligment of sponsor section

TenzDelek:
Changes: the current alignment of the sponsor seems to have no margin given resulting in default left and a negative white space.
image

Changes made: made the sponsor section center to the screen for better white space.
image

### DIFF
--- a/overrides/assets/stylesheets/home.css
+++ b/overrides/assets/stylesheets/home.css
@@ -124,7 +124,7 @@ h2.secondary-headline, h3.trusted-by {
 
 .scroller {
   max-width: 60vw;
-  
+  margin: auto;
 }
 
 .scroller__inner {


aliok:
@TenzDelek I see the adopters are listed fine on https://knative.dev/docs/ page.

Can you check again please?

TenzDelek:
> @TenzDelek I see the adopters are listed fine on https://knative.dev/docs/ page.
> 
> Can you check again please?

hey @aliok , I have raised that issue in #5996  and from what I have seen, it is only happening for my device(window). but for my  mac, it is working perfectly fine. I guess that problem resides from my end.. Thank you for the concern, this pr can be close :)

aliok:
@TenzDelek 

Hm, interesting. Do you have anything special on your Windows machine? That shouldn't happen on Windows too

TenzDelek:
> @TenzDelek
> 
> Hm, interesting. Do you have anything special on your Windows machine? That shouldn't happen on Windows too

@aliok ,  I don't recall having any specific installation that might raise this behavior. but as per my discussion in the above issue section, other member who are having windows doesn't seem to have this kind of issue. so I guess technically this comes down to my machine. I will try to look further into it. but i find it funny how it only the marquee that is not working?  is there any case regarding this in the pass? :)

PR #5967, opened 2024-05-13
Update docs for Eventing TLS Beta

pierDipi:
## Proposed Changes 

- Rename experimental features to just features
- Add CA trust bundle configuration guide
- Add setup selfsigned issuer guide for getting started


docs/eventing/features/transport-encryption.md
```diff
@@ -67,6 +156,10 @@ The possible values for `transport-encryption` are:
     - Addressables must not accept events to non-HTTPS endpoints
     - Addressables must only advertise HTTPS endpoints
 
+!!! important
+    The `strict` is only enforced on the Broker and Channel receiver/ingress. Senders can still send
+    events over plaintext HTTP.
```
Cali0707, line 161: @pierDipi could we reword this? I find it a little confusing. To me, `Senders can still send events over plaintext HTTP` sounds like I can still send my events over http to a broker. But, my understanding is that we are trying to say that when a broker or channel sends events to a subscriber, if that subscriber only has a HTTP address, the broker or channel will send the event with HTTP instead of HTTPS.

docs/eventing/features/transport-encryption.md
```diff
@@ -80,21 +173,50 @@ data:
   transport-encryption: "strict"
 ```
 
-## Trusting CA for a specific event sender
+## Configure additional CA trust bundles
+
+By default, Eventing clients trusts the system root CA (public CA).
+
+If you need to add additional CA bundles for Eventing, you can do so by creating ConfigMaps in the
+`knative-eventing` namespace with label `networking.knative.dev/trust-bundle: true`:
```
Cali0707, line 181: Maybe we can mention trust-manager here as a good way to do this?

docs/eventing/sources/kafka-source/README.md
```diff
@@ -208,7 +208,7 @@ Alternatively, if you are using a GitOps approach, you can add the `consumers` k
 
 ### Automatic Scaling with KEDA
 
-Kafka Sources have experimental (Alpha) support for serverless scaling with KEDA, including scale to zero. If you want Knative and KEDA to scale your Kafka source for you,
+Kafka Sources have (Alpha) support for serverless scaling with KEDA, including scale to zero. If you want Knative and KEDA to scale your Kafka source for you,
```
Cali0707, line 211:
```suggestion
Kafka Sources and Brokers have (Alpha) support for serverless scaling with KEDA, including scale to zero. If you want Knative and KEDA to scale your Kafka source for you,
```

pierDipi:
@Cali0707 thanks, see latest commits!

Cali0707 (approved):
/lgtm

The eventing features looks a lot clearer now, thanks @pierDipi !

PR #5916, opened 2024-03-19
Add SubscriptionsAPI filters to APIServerSource

rh-hemartin:
This MR adds a page for the experimental feature that add `filters` to the APIServerSource kind.

Related to [#7799](https://github.com/knative/eventing/pull/7799) and [#7791](https://github.com/knative/eventing/issues/7791)

## Proposed Changes 

- Add a page for the experimental feature that adds `filters` to `APIServerSource`

PR #5968, opened 2024-05-13
Add WG leads to docs approvers

pierDipi:
Every time there is a new page added, I cannot approve PRs (example https://github.com/knative/docs/pull/5916), so I'm adding WG leads to the top-level directory approvers

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -6,6 +6,12 @@ approvers:
 - steering-committee
 - knative-release-leads
 - ux-wg-leads
+# WG leads
+- client-wg-leads
+- eventing-wg-leads
+- functions-wg-leads
+- operations-wg-leads
+- serving-wg-leads
 reviewers:
 - docs-writers
 - docs-reviewers


Cali0707 (commented):
+1 from me!

/lgtm

psschwei (commented):
If the issue is approving page changes, would it make sense to add the working group leads as owners on the config directory rather than for all of docs? Though given that there's no docs team right now, perhaps it's better to have more folks able to approve for the whole repo?

dprotaso:
> If the issue is approving page changes, would it make sense to add the working group leads as owners on the config directory rather than for all of docs? Though given that there's no docs team right now, perhaps it's better to have more folks able to approve for the whole repo?

Yeah - let's subdivide the owners to the various sections

dprotaso:
@pierDipi can you update the PR?

skonto:
@pierDipi gentle ping

pierDipi:
/reopen

I still think WG leads need access to the whole project to change nav, etc, since most of the time WG leads have also contributed to docs extensively to be able to be approver of the entire docs

pierDipi:
and obviously, WG leads will approve PRs in their area of expertise or expedite/review PRs they know are ok

dprotaso:
> and obviously, WG leads will approve PRs in their area of expertise or expedite/review PRs they know are ok

Yeah I agree - it's just the owner file right now will just tag doc approvers/reviewers first. So in addition to what you have we want to make sure we tag eventing approvers for eventing docs etc.

You can do that with separate owner files in nested directories

/lgtm
/approve

can you do the follow up in another PR?

PR #5970, opened 2024-05-14
Added the golang redirect for autoscaler-keda and func-python

skonto:

## Proposed Changes 

- Similar to https://github.com/knative/docs/pull/5691/
- Needed for https://github.com/knative/community/issues/1527
- I run the tool for generating redirects but I hit a rate limit with anonymous github API requests (50). I upgraded the github client locally, in order to pass a token to bypass the rate limit. Client is very old btw. If we want to upgrade that part we will need a re-write of the paging part, because it didnt work and I had to increase the items return from 50 to 100: `ListOptions: github.ListOptions{PerPage: 100},`. The actual problem is that
 

> if resp.NextPage == 0 {
> 	    break }

never gets updated and never reaches 0.


golang/_redirects
```diff
@@ -49,7 +48,6 @@
 /kn-plugin-source-kamelet/* go-get=:go-get /golang/kn-plugin-source-kamelet.html 200
 /knobots/* go-get=:go-get /golang/knobots.html 200
 /monitoring/* go-get=:go-get /golang/monitoring.html 200
-/net-certmanager/* go-get=:go-get /golang/net-certmanager.html 200
```
dprotaso, line 52: we shouldn't drop some of these old redirects - since it will break older clients

I wonder if we should just not do this for archived repos

matzew:
and for `func-python`, not just KEDA

golang/_redirects
skonto: Ok I can fix it in another PR.

PR #5983, opened 2024-05-21
fix ML example

mattn:
Fix machinelearning-python-bentoml to work in latest Python and bentoml.


aliok:
cc @knative/serving-approvers

dprotaso:
thanks for the fix!

/cherry-pick release-1.16

PR #5991, opened 2024-05-23
Add UX leads as docs reviewers

Cali0707:
Currently, there are many PRs which are being opened to the docs repo regarding design changes (which I should review). But, since the UX leads group are only approvers and not reviewers on this repository, I don't have my review requested.

This PR adds the UX leads as docs reviewers (on top of the already existing approver permissions) so that we can get our reviews requested


psschwei:
+1

PR #5998, opened 2024-05-28
Add Lazy Loading for Homepage Images and Check script

### LINKED ISSUE #5981: Improve Site Performance with Lazy Loading and WEBP Images
Implement two changes to enhance site performance:
1. Add loading="lazy" to image tags to improve Cumulative Layout Shift (CLS).
2. Convert PNG/JPG/JPEG images to WEBP format for faster loading times.

**Additional context**
These optimizations will reduce initial page load time and bandwidth usage, improving user experience and SEO.

shivamgupta2020:
Fixes #5981 

## Proposed Changes 

- added lazy load in homepage images

### DIFF
--- a/check_img_lazy.py
+++ b/check_img_lazy.py
@@ -3,14 +3,6 @@ import sys
 import subprocess
 from bs4 import BeautifulSoup
 
-def build_website():
-    """Build the website using the specified command."""
-    try:
-        subprocess.check_call(['./hack/docker/test.sh'])  
-    except subprocess.CalledProcessError as e:
-        print(f"Website build failed: {e}")
-        sys.exit(1)
-
 def check_img_tags(base_dir):
     """Recursively check HTML files for <img> tags without the lazy loading attribute."""
     issues_found = False
@@ -46,9 +38,6 @@ def find_line_number(content, tag):
 
 if __name__ == "__main__":
 
-    build_website()
-
-
     base_dir = 'site'  
     issues_found = check_img_tags(base_dir)
 
@@ -58,4 +47,3 @@ if __name__ == "__main__":
     else:
         print("All img tags have lazy loading attribute.")
         sys.exit(0)
-
--- a/check_img_lazy.py
+++ b/check_img_lazy.py
@@ -1,0 +1,61 @@
+import os
+import sys
+import subprocess
+from bs4 import BeautifulSoup
+
+def build_website():
+    """Build the website using the specified command."""
+    try:
+        subprocess.check_call(['./hack/docker/test.sh'])  
+    except subprocess.CalledProcessError as e:
+        print(f"Website build failed: {e}")
+        sys.exit(1)
+
+def check_img_tags(base_dir):
+    """Recursively check HTML files for <img> tags without the lazy loading attribute."""
+    issues_found = False
+
+    for root, _, files in os.walk(base_dir):
+        for file in files:
+            if file.endswith('.html'):
+                file_path = os.path.join(root, file)
+                with open(file_path, 'r', encoding='utf-8') as f:
+                    content = f.read()
+                    soup = BeautifulSoup(content, 'html.parser')
+                    for img in soup.find_all('img'):
+                        if 'loading' not in img.attrs or img.attrs['loading'] != 'lazy':
+                            issues_found = True
+                            line_number = find_line_number(content, img)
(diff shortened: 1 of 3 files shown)


aliok:
@shivamgupta2020 

The check script should do these:
- Build the website
- Parse HTML pages that are built
- Check for `img` tags without `lazy`
- Report the locations (file, line number) of the tags w/o `lazy` attribute in the script logs (stdout)
- Fail (by exiting with a code that's not zero)

This script should be running in a Github action.

I would recommend this approach:
1. Write that script (can be a Python script)
2. Include it in the PR, with documentation and dependencies (`requirements.txt` that can be used by `pip`)
3. Get feedback, make sure it works
4. Then create a GitHub action that runs the script

shivamgupta2020:
Thanks, @aliok , for the suggestions. I’m currently working on the script part and will include all your suggestions in a new commit as soon as possible.

shivamgupta2020:
> @shivamgupta2020
> 
> The check script should do these:
> 
> * Build the website
> * Parse HTML pages that are built
> * Check for `img` tags without `lazy`
> * Report the locations (file, line number) of the tags w/o `lazy` attribute in the script logs (stdout)
> * Fail (by exiting with a code that's not zero)
> 
> This script should be running in a Github action.
> 
> I would recommend this approach:
> 
> 1. Write that script (can be a Python script)
> 2. Include it in the PR, with documentation and dependencies (`requirements.txt` that can be used by `pip`)
> 3. Get feedback, make sure it works
> 4. Then create a GitHub action that runs the script

Should the script also check versions v1.13 and v1.12?

check_img_lazy.py
```diff
@@ -0,0 +1,61 @@
+import os
+import sys
+import subprocess
+from bs4 import BeautifulSoup
```
aliok, line 4: Is it a possibility to use standard Python libraries for parsing HTML files?

What's your opinion here @shivamgupta2020 ? We need to instruct people to install some dependencies if we go with BeautifulSoup. However, we might have some non-standard HTML content that might not be parseable with standard Python libraries.

Let's ask @knative/productivity-leads and @cardil specifically.

check_img_lazy.py
```diff
@@ -0,0 +1,61 @@
...
+
+def build_website():
+    """Build the website using the specified command."""
+    try:
+        subprocess.check_call(['./hack/docker/test.sh'])  
+    except subprocess.CalledProcessError as e:
+        print(f"Website build failed: {e}")
+        sys.exit(1)
```
aliok, line 12: Please leave this part out of the script.

This script will eventually run in a GitHub Actions workflow:
- Build the website
- Run the script
  - If there are images w/o `loading=lazy`, script should fail
  - If there are no images, script should exit successfully

Can you put the script in a GitHub actions workflow please?

Also, we need to document this change with instructions. In the relevant section here: https://github.com/knative/docs/blob/main/contribute-to-docs/README.md

check_img_lazy.py
```diff
@@ -0,0 +1,61 @@
...
+            return i + 1
+    
+    return 'unknown'
+
+if __name__ == "__main__":
+
+    build_website()
+
+
+    base_dir = 'site'
```
aliok, line 52: Please make this configurable. Script should receive the directories to check HTML files.

check_img_lazy.py
shivamgupta2020: Yeah, sure, I will remove the build function from the script.
I will look into the GitHub action workflow and add the script as soon as possible.

check_img_lazy.py
aliok: Sounds good. Let's leave as is. I need to check this later.

check_img_lazy.py
shivamgupta2020: @aliok, I experimented with HTMLParser and IXML for parsing HTML files, but encountered numerous errors and incomplete parsing. Therefore, I switched to BeautifulSoup. I also think we should gather opinions from others on this approach.

aliok: cc @pierDipi @cardil

shivamgupta2020:
I've created the workflow file as @aliok suggested. Currently, most of the images on the website are not lazy-loaded, resulting in many errors. Could you please check the script output on your local machine to verify it's working as expected? Also, please review the workflow code to ensure it's implemented correctly.
Thank you!

PR #6004, opened 2024-06-03
Fix footer styling issue

rohitsrma:
Currently, in the footer of our website's documentation section, the 'Next' and 'Previous' links are shown in front of the page's title, and they have different font sizes. I'm sure this wasn't done on purpose. The problem is because the title and the navigation links are in the same class, which messes up the styles.

To fix this, I moved the footer text outside of that class. This way, the existing styles won't get overridden by the class styles.

### Before

image

### After

image

### DIFF
--- a/overrides/partials/footer.html
+++ b/overrides/partials/footer.html
@@ -23,10 +23,10 @@
             {% include ".icons/material/arrow-left.svg" %}
           </div>
           <div class="md-footer__title">
+            <span class="md-footer__direction">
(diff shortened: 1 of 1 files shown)


Cali0707:
/remove-lifecycle stale

Adding this PR to the next UX WG meeting agenda for review

Cali0707:
Moving this to the next UX meeting time

Cali0707 (approved):
/lgtm
/approve

PR #6022, opened 2024-06-18
feat: add docs for runtime-class-name feature

BobyMCbobs:
adds docs for configuring runtime-class-name

## Proposed Changes 

- related: https://github.com/knative/serving/issues/12729


Cali0707 (commented):
/cc @dprotaso

PR #6023, opened 2024-06-19
Add link to community roles to contributor page

psschwei:
## Proposed Changes 

It was mentioned on the TOC call today that some folks were finding it difficult to find the different contributor roles, so this PR adds an explicit link to them from the "contributor" page.


pierDipi (approved):
/lgtm

/cc @Cali0707

PR #6027, opened 2024-06-26
Remove as_updated field in rss plugin

dsimansk:
It should be defaulting to `False` anyway. Otherwise we have "expected string got bool error".

https://guts.github.io/mkdocs-rss-plugin/configuration/#categories_example


matzew (approved):
/lgtm
/approve

ship it!

PR #6049, opened 2024-07-09
Update the github action versions, as reviewdog is complaining

Leo6Leo:
## Proposed Changes 

- Update docker/setup-buildx-action, docker/login-action, docker/build-push-action, and actions/checkout


Leo6Leo:
/cc @Cali0707
I think the only way to test this is merging it right?

Cali0707:
>  I think the only way to test this is merging it right?

probably?

PR #6060, opened 2024-07-18
feat: documented new trigger filters are GA

Cali0707:
This PR updates the trigger documentation to reflect the fact that the new trigger filters are GA soon. Note: this _should not be merged_ until after https://github.com/knative-extensions/eventing-kafka-broker/pull/3964 has merged

/hold for https://github.com/knative-extensions/eventing-kafka-broker/pull/3964


matzew (approved):
/lgtm
/approve

config/nav.yml
```diff
@@ -296,7 +296,6 @@ nav:
         - About Eventing features: eventing/features/README.md
         - DeliverySpec.Timeout field: eventing/features/delivery-timeout.md
         - DeliverySpec.RetryAfterMax field: eventing/features/delivery-retryafter.md
-        - New Trigger Filters: eventing/features/new-trigger-filters.md
```
pierDipi, line 299: We need to remove that file `eventing/features/new-trigger-filters.md`

Cali0707: When I removed the file, it broke links in the old release note pages. So, I decided to:
1. leave it in
2. remove it from the nav so that you can only access it through the old release notes

WDYT @pierDipi ?

pierDipi: or we could update the old release note pages to point to the new place?

pierDipi: Leaving it there means we need to update/maintain two places for this and it's not really worth it

Cali0707:
/cc @pierDipi 
I've removed the old file now

config/nav.yml
Cali0707: Should be fixed now

PR #6097, opened 2024-08-14
design changes in about and community section

ShravaniAK:
1. in the community section i have replaced the image with the new designed image
2. in the about section made the font color , weight and image size changes .


Cali0707 (changes requested):
For this image:
Screenshot 2024-09-17 at 1 53 25 PM

Would it be possible to:
1. Replace the central box background with a lighter shade of blue from the design system color scheme
2. Change the font for Knative/Kubernetes to match the other font in the image

PR #6109, opened 2024-09-10
Adding WebSocket Go example

matzew:
As per title


code-samples/serving/websockets-go/config/service.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: websockets-go
+  namespace: default
+spec:
+  template:
+    spec:
+      containers:
+      - image: ko://github.com/knative/docs/code-samples/serving/websockets-go/cmd/server
```
skonto, line 10: The other samples have docker.io/{username}/grpc-ping-go, so let's keep consitency.

code-samples/serving/websockets-go/Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Use the official Golang image to create a build artifact.
+# This is based on Debian and sets the GOPATH to /go.
+FROM golang:latest as builder
+
+ARG TARGETOS
```
skonto, line 5: Tested locally works as epxected:
```
$ wscat --connect 127.0.0.1:8080/ws
Connected (press CTRL+C to quit)
> hello
< hello
> hi
< hi
> 
```

skonto: To be consistent with the other samples eg. grpc we need to add something like
```
docker run --rm {username}/grpc-ping-go \
  /client \
  -server_addr="grpc-ping.default.1.2.3.4.sslip.io:80" \
  -insecure
```

matzew: never done that before. I mostly do never use `docker`, I just run things w/ `ko` 

feel free to add things to this post.

code-samples/serving/websockets-go/config/service.yaml
skonto: In the past we added [buildx](https://github.com/knative/docs/pull/5273) and since it touches many files any other addition should be more visible and get proper reviews. If something does not work we can open another issue and fix it (migrate to ko etc).

code-samples/serving/websockets-go/Dockerfile
matzew: @skonto 

```
docker run --rm docker.io/matzew/ws-goss \      
  /server \                                                 
  -server_addr="websockets-go.default.1.2.3.4.sslip.io:80" \
  -insecure
```

I am not getting this to work, please feel free to add the correct content. Otherwise we merge this and deal with this later?

The section above shows how to the binary w/in kube already. Not really adding too much value

matzew: I could do `go run cmd/server/main.go` ... 

and localhost - that is simple enought - and has similar value

code-samples/serving/websockets-go/config/service.yaml
skonto: @rhuss @pierDipi wdyth? Any objections to stamp?

code-samples/serving/websockets-go/Dockerfile
skonto: It is ok let's fix this in #6110

PR #6142, opened 2024-10-21
fix: remove next/previous text from blog footer

### LINKED ISSUE #6124: Remove next/previous text from blog footer
Currently in the blog there is text on the footer above each link saying next/previous. As the blog posts are reverse chronological order this can be confusing, so it would be better to remove this altogether (but only for the blog, not the main docs)

**Additional context**
Screenshot 2024-09-17 at 1 45 21 PM

prushh:
Fixes #6124

### Proposed Changes

- remove next/previous text in blog footer, based on `config.site_url` variable


Cali0707 (approved):
Thanks @prushh !
/lgtm

PR #6154, opened 2024-10-31
doc: Mark TriggerMesh as Defunct

zacheryph:
## Proposed Changes 

- Marks All TriggerMesh Eventing Sources as _Defunct_.

Sadly Triggermesh shuttered a year ago, last fall. It is sad to see them gone, though these probably should not be marked as active anymore.

I wonder if these should simply be removed or possibly add a 3rd section for.. `Historic`? `Defunct`? not sure what to name it 🤔 

> Reference: https://github.com/triggermesh/triggermesh/issues/1544#issuecomment-2314595965


pierDipi:
> I wonder if these should simply be removed or possibly add a 3rd section for.. `Historic`? `Defunct`? not sure what to name it

Thanks @zacheryph, I would remove them.

In the meantime, we're working on replacements and we would love feedback on which one you would be using here: https://github.com/knative/eventing/issues/8185

zacheryph:
@pierDipi apologize, I missed notification for this one. Updated to remove Triggermesh. I will have to take a look at that issue.

zacheryph:
So much for that...

PR #6165, opened 2024-11-27
Update Sink Index page

matzew:
## Proposed Changes 

- Adding new integration sink options to sink index


matzew:
@Cali0707 wanna take a look here?

PR #6177, opened 2024-12-16
Add support to ignore request in metrics

jan-stanek:
## Proposed Changes 

* Added possibility to mark request to be ignored in metrics using header `K-Ignore-Metrics`

### DIFF
--- a/docs/serving/services/service-metrics.md
+++ b/docs/serving/services/service-metrics.md
@@ -19,6 +19,9 @@ Requests endpoint.
 !!! note
     The `revision_queue_depth` metric will be exported only if the revision concurrency hard limit is set to a value greater than 1.
 
+!!! note
+    To exclude request from metrics (e.g. health check requests), set the `K-Ignore-Metrics` header to non-empty value.
+
 ## Exposing Queue proxy metrics
 
 Queue proxy exports metrics for the requests endpoint on port 9091. The metrics can be scraped by Prometheus when `metrics.request-metrics-backend-destination` is set to `prometheus` (default) in the configmap `observability`. The backend can be changed to `opencensus` which uses a push model and requires a destination


docs/serving/services/service-metrics.md
```diff
@@ -19,6 +19,9 @@ Requests endpoint.
 !!! note
     The `revision_queue_depth` metric will be exported only if the revision concurrency hard limit is set to a value greater than 1.
 
+!!! note
+    To exclude request from metrics (e.g. health check requests), set the `K-Ignore-Metrics` header to non-empty value.
```
evankanderson, line 23: What are the security implications of this header?  I'm concerned that someone could use this header to sneak in a DDoS attack that wouldn't show up on monitoring.

jan-stanek: Good point, would it be better to have possibility to specify additional label for metrics?

evankanderson: I think a single additional label would be safe; extracting a label into a header has a different DoS risk by expanding the cardinality of the metric.

dprotaso: I think this feature needs to go through a feature track document to outline possible approaches and to confirm the problem we're trying to address. In the serving repo all this work seems to go towards external health checks but we need to outline the behaviour when we've scaled to zero etc.

Details on the feature track process is here:
https://github.com/knative/community/blob/main/mechanics/FEATURE-TRACKS.md

evankanderson: I think this can be fairly short, but it would be good to explain why external health checks should be treated differently from other traffic -- I'm _assuming_ this is because the artificial traffic would perturb dashboards and other monitoring, but I don't have full context.

If you have external health checks that don't execute the application, it seems like you could just deploy (for example) a dummy Revision and use tag-routed headers or hostnames for the external probe.  If the health checks _do_ execute some part of the application, it seems like you'd want to scale them in the same way as other requests (which may mean you always have one instance live).

jan-stanek:
Feature request created https://github.com/knative/serving/issues/15795, closing this

PR #5775, opened 2023-11-27
Add security audit report and blogpost

AdamKorcz:
Adds the Knative security audit report and blogpost.

cc @davidhadas @evankanderson

PR #6193, opened 2025-01-14
Add documentation on fetching Knative supply-chain security attestations

evankanderson:
Adds documentation on how to consume Knative supply chain attestations.  Captures discussion from [this comment](https://github.com/knative/docs/pull/5775#discussion_r1414296163), where they are unlikely otherwise to be found.

## Proposed Changes 

- Adds initial security documentation, focused on supply chain security.  Future security configuration documents should probably be added to the same location.

### DIFF
--- a/config/nav.yml
+++ b/config/nav.yml
@@ -343,7 +343,9 @@ nav:
       - Eventing code samples: samples/eventing.md
     # Reference docs
     - Reference:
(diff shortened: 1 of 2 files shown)


evankanderson:
Good idea, I'll send a follow-up PR.  I also want to revise some of the security documentation to give some more holistic guidance.

PR #6197, opened 2025-01-17
Add interconnected as an attribute of the Event Mesh definition

matzew:
## Proposed Changes 

- as per title


matzew:
CC @pierDipi @dsimansk

PR #6199, opened 2025-01-19
Link to supply-chain details rather than including inline.

evankanderson:
Followup to https://github.com/knative/docs/pull/6193

## Proposed Changes 

- Reduces the text on the YAML/operator install pages, and makes clear that signature verification is an _optional_ step, and not a required prerequisite

### DIFF
--- a/docs/reference/security/verifying-images.md
+++ b/docs/reference/security/verifying-images.md
@@ -25,7 +25,12 @@ cosign verify-attestation \
   $IMAGE
 ```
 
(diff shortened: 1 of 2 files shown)


Cali0707 (approved):
/lgtm

I like this, IMO definitely makes the install instructions clearer

/hold in case @pierDipi or @dprotaso have opinions (since this changes the install docs for Serving/Eventing)

dprotaso:
/hold cancel

thanks for the follow up Evan

PR #6205, opened 2025-01-25
sample-app: Add serviceName to postgresql manifest

guzalv:
To fix this error:

`error: error validating "db-service/200-create-postgre.yaml": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec; if you choose to ignore these errors, turn validation off with --validate=false`

### DIFF
--- a/code-samples/eventing/bookstore-sample-app/solution/db-service/200-create-postgre.yaml
+++ b/code-samples/eventing/bookstore-sample-app/solution/db-service/200-create-postgre.yaml
@@ -7,6 +7,7 @@ spec:
   selector:
     matchLabels:
       app: postgresql
+  serviceName: postgresql
   template:
     metadata:
       labels:
--- a/code-samples/eventing/bookstore-sample-app/start/db-service/200-create-postgre.yaml
+++ b/code-samples/eventing/bookstore-sample-app/start/db-service/200-create-postgre.yaml
@@ -7,6 +7,7 @@ spec:
   selector:
     matchLabels:
       app: postgresql
+  serviceName: postgresql
   template:
     metadata:
       labels:


Leo6Leo:
@guzalv Thanks for the PR! I have tried the original yaml file to create the StatefulSet in the cluster, it doesn't seem to have the error you have mentioned above. Could you please provide some of your tech spec (e.g your k8s version, minikube/kind, knative version etc...) so that we can figure out the reason why the error happened too you?

guzalv:
Hi @Leo6Leo, thanks for your comment. Good that you noticed, this was an interesting issue: `serviceName` is marked as `required` in the [Kubernetes' StatefulSet reference](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/stateful-set-v1/#StatefulSetSpec):

- serviceName (string), required

I went to reproduce it again, and it failed, as I expected. However I noticed I was using a rather old kubectl version (1.22.5). When I upgraded to 1.30.0, to match my Kubernetes cluster, it didn't fail anymore. So even if the docs still state a serviceName is required, that is actually not enforced by recent kubectl versions:

```console
$ kubectl-1.30.0 apply -f ./code-samples/eventing/bookstore-sample-app/solution/db-service/200-create-postgre.yaml
statefulset.apps/postgresql created

$ ./kubectl-1.22.5 apply -f ./code-samples/eventing/bookstore-sample-app/solution/db-service/200-create-postgre.yaml
error: error validating "./code-samples/eventing/bookstore-sample-app/solution/db-service/200-create-postgre.yaml": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1.StatefulSetSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

(Could it be that [the StatefulSets' limitation of requiring a headless service](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#limitations) is being removed?) 

I also checked that this PR works fine with the latest kubectl too, but I'm not familiar enough with the project to know whether it's ok/recommended/discouraged to make a change to support an older version of a tool. What do you think?

dprotaso:
I just tested this out and it worked without setting the `serviceName` - it defaults to an empty string

@guzalv what version of k8s are you using?

```
 % k version
Client Version: v1.32.1
Kustomize Version: v5.5.0
Server Version: v1.32.0
```

dprotaso:
This might be a Kubernetes bug in newer version.

/lgtm
/approve

guzalv:
> I just tested this out and it worked without setting the `serviceName` - it defaults to an empty string
> 
> @guzalv what version of k8s are you using?
> 
> ```
>  % k version
> Client Version: v1.32.1
> Kustomize Version: v5.5.0
> Server Version: v1.32.0
> ```

Hi @dprotaso,

```console
$ kubectl version
Client Version: v1.30.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.30.0
```

In case the Kubernetes folks want to pick this up: the conclusion from my earlier trials was that the issue lies in kubectl: 1.22.5 complains about the missing serviceName, as expected according to the docs, and 1.30.0 doesn't.

Thanks for checking and merging the PR!

PR #6274, opened 2025-05-28
Phase 1: Adding documentation metadata tags

evankanderson:
## Proposed Changes 

In preparation for a [CNCF TechDocs Assessment](https://github.com/cncf/techdocs/issues/315), start labelling Knative documentation with metadata to assist with re-organizing content to different information architecture options (we will still need manual review, but this may simplify that review).

I'm using a schema with 3 fields:

| field | allowed values | explanation |
|---|---|---|
| `audience` | `developer`, `administrator`, `buyer`, `contributor` | **Who** the target audience is; ensure documents have a clear, single audience |
| `components` | `eventing`, `functions`, `serving` (list) | **What** components are featured on the page. |
| `function` | `marketing`, `community`, `tutorial`, `how-to`, `reference`, `explanation` | A [diataxis](https://diataxis.fr/) tag or "marketing" or "community" indicating the needs addressed by the documentation. |

### DIFF
--- a/docs/about-analytics-cookies.md
+++ b/docs/about-analytics-cookies.md
@@ -2,6 +2,9 @@
 title: "Learn about Google Analytics cookies"
 linkTitle: "Learn about cookies"
 type: "docs"
+audience: developer
+components: []
+function: community
 ---
 
 By using this website, you are confirming that you accept our use of cookies for
--- a/docs/about/case-studies/README.md
+++ b/docs/about/case-studies/README.md
@@ -1,6 +1,11 @@
 ---
 hide:
   - toc
+audience: buyer
+components:
+  - eventing
+  - serving
+function: marketing
 ---
 
 # Knative Case Studies
--- a/docs/about/case-studies/deepc.md
+++ b/docs/about/case-studies/deepc.md
@@ -1,6 +1,10 @@
 ---
 hide:
   - toc
+audience: buyer
+components:
+  - eventing
+function: marketing
 ---
 <h1 style="color:#0071c7;font-size: 3em;">deepc Case Study</h1>
 <table style="border: 0;">
--- a/docs/about/case-studies/ibm.md
+++ b/docs/about/case-studies/ibm.md
@@ -1,6 +1,10 @@
 ---
 hide:
   - toc
+audience: buyer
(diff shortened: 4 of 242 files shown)


evankanderson:
https://diataxis.fr/complex-hierarchies/#two-dimensional-problems may be helpful when figuring out how to structure these eventually.

evankanderson:
@dwelsch-esi -- PTAL (please take a look), as this should help (hopefully) with the initial phase of techdocs assessment.  If this makes sense to you, then you can reply with `/lgtm` at the start of a line, and Prow will merge this PR (and it will be pushed to Netlify / go live).  You can also take a look at the Netlify preview, though it should be roughly identical to the current live site, as the new metadata is only in the source code.

dwelsch-esi:
@evankanderson, Just one minor question regarding audience:

The same page can be relevant to more than one audience. Looks like you've inserted comments to that effect in some of the files, for example  `docs/eventing/accessing-traces.md`:
```
audience: administrator
# And audience: developer for accessing traces
```
Is there a reason you didn't make this field a list like you did for components? Regardless, I don't think this is a big deal -- I'm sure the role you've identified in each case is the primary audience for the page. If the meta tags become super important later, you could open another PR to update them, but I don't think this question is worth holding up the merge.

/lgtm

evankanderson:
> @evankanderson, Just one minor question regarding audience:
> 
> The same page can be relevant to more than one audience. Looks like you've inserted comments to that effect in some of the files, for example  `docs/eventing/accessing-traces.md`:
> ```
> audience: administrator
> # And audience: developer for accessing traces
> ```
> Is there a reason you didn't make this field a list like you did for components? Regardless, I don't think this is a big deal -- I'm sure the role you've identified in each case is the primary audience for the page. If the meta tags become super important later, you could open another PR to update them, but I don't think this question is worth holding up the merge.

Generally, I'm suspicious of trying to target more than one audience or purpose in a single document.  In cataloguing our current docs, we **do** mix these audiences, but I would probably prefer to move a bunch of the installation docs separate from the usage docs, rather than encourage this mixing.

Cali0707 (approved):
/lgtm
/approve

PR #6281, opened 2025-06-19
Remove domain-mapping pods from console outputs

Bo0km4n:

## Proposed Changes 
- Remove domain-mapping pod name from stdout result
  - domain-mapping controllers are included in controller pod.


dprotaso:
/lgtm
/approve

thanks @Bo0km4n !  yeah we've merged the domain mapping webhook & controller into the main serving controller

PR #6294, opened 2025-07-26
Fix: wrong filters attribute in trigger document

mexeniz:
I tried to create a  Trigger object by following the guideline at https://knative.dev/docs/eventing/triggers/#trigger-filtering .

```yaml
apiVersion: eventing.knative.dev/v1
kind: Trigger
metadata:
  name: my-service-trigger
spec:
  broker: default
  filters:
    exact:
      type: dev.knative.foo.bar
    suffix:
      myextension: -value
  subscriber:
    ref:
      kind: Service
      name: my-service
```

However, I got the error due to a wrong syntax since `spec.filters` expects an array of filters.

```
strict decoding error: unknown field "spec.filters.exact", unknown field "spec.filters.suffix"
```

## Proposed Changes 

- Fix a wrong attribute in Trigger filtering's example manifest file

### DIFF
--- a/docs/eventing/triggers/README.md
+++ b/docs/eventing/triggers/README.md
@@ -88,10 +88,10 @@ with `-extensions.
     spec:
       broker: default
       filters:
-        exact:
-          type: dev.knative.foo.bar
-        suffix:
-          myextension: -value
+        - exact:
+            type: dev.knative.foo.bar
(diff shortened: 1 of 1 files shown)


evankanderson (commented):
/lgtm
/approve

Issue #6310, opened 2025-07-29
Review graphics


dwelsch-esi:
## Overview

Graphics should enhance understanding by illustrating concepts and processes.
Many graphics on the Knative site detract from, rather than enhance, the
usability of documentation. 

Review the graphics on the site to make sure they meet these criteria:

- Does the graphic enhance understanding? Does it add information that can't be
  conveyed in a few words?
- Is the graphic placed in context by the surrounding text?
- Does the graphic's size and aspect ratio enable proper placement and text flow
  on a variety of screens sizes?
- Is the graphic maintainable: Will it require frequent updating for
  localization, software updates, or interface changes?

Review the graphics that are on the site. Remove graphics that don't justify the
space they take up. Scale the graphics to a more appropriate size if necessary.

Audience: All

Type: Conceptual

## Context

This issue tracks recommended changes resulting from an analysis of the Knative
documentation commissioned by CNCF. The analysis and supporting documents are
here: https://github.com/cncf/techdocs/tree/main/analyses under `0015-knative`.

## Possible Implementation

Related material in the current doc:

- https://knative.dev/docs/concepts/
- https://knative.dev/docs/concepts/eventing-resources/brokers/
- https://knative.dev/docs/getting-started/tutorial/
- https://knative.dev/docs/getting-started/next-steps/
- https://knative.dev/docs/bookstore/page-0/welcome-knative-bookstore-tutorial/
- https://knative.dev/docs/bookstore/disclaimer/
- https://knative.dev/docs/bookstore/* (all bookstore pages)
- https://knative.dev/docs/serving/architecture/
- https://knative.dev/docs/serving/request-flow/
- https://knative.dev/docs/serving/encryption/encryption-overview/

Suggested changes:

- Briefly introduce the system diagram in the [Concepts overview]. This can be a
  single phrase added to the sentence below the graphic: "The primary Knative
  Serving resources are Services, Routes, Configurations, and Revisions, as
  shown in the previous figure:".
- Similarly for other conceptual graphics on the site: Refer to them from the
  text.

evankanderson:
We should establish a standard set of icons under https://github.com/knative/docs/tree/main/contribute-to-docs/style-guide; see [this thread](https://cloud-native.slack.com/archives/C05MW1AT1T8/p1755021149045219) for more details.  A few other highlights:

[...]

PR #6327, opened 2025-08-12
Support Mermaid.js for diagrams

evankanderson:
Supports #6310

Currently, we have a wide variety of different graphic styles on our website.  I'm going to suggest narrowing our art sources to using either SVGs from a real drawing program, or generating them from [Mermaid.js markup](https://mermaid.js.org/), which is [supported (and recommended)](https://squidfunk.github.io/mkdocs-material/reference/diagrams/) by the Material MkDocs tool we use for the site

## Proposed Changes 

- Enable Mermaid code blocks
- Use these to create a top-level diagram for "what eventing is"

Once we have an icon library, we should look at [customizing mermaid](https://squidfunk.github.io/mkdocs-material/reference/diagrams/#customization) to [use a custom icon pack](https://mermaid.js.org/config/icons.html) with our icons.

### DIFF
--- a/docs/snippets/about-eventing.md
+++ b/docs/snippets/about-eventing.md
@@ -4,6 +4,21 @@
 -->
 Knative Eventing is a collection of APIs that enable you to use an [event-driven architecture](https://en.wikipedia.org/wiki/Event-driven_architecture){target=_blank} with your applications. You can use these APIs to create components that route events from event producers (known as sources) to event consumers (known as sinks) that receive events. Sinks can also be configured to respond to HTTP requests by sending a response event.
 
+``` mermaid
+architecture-beta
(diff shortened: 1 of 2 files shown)


dprotaso:
The titles overlap a bit with the arrows - unsure if you can tweak that. Ideally you could put the text on top of the arrow with some border.

Screenshot 2025-08-12 at 6 48 08 PM

evankanderson:
The [Mermaid.js documentation](https://mermaid.js.org/syntax/architecture.html) does the same thing, so I think that's something we may have to live with or get fixed upstream

evankanderson:
Probably the "correct" solution would be to include the text box in the arrow target, rather than just the icon.

PR #6367, opened 2025-09-06
Knative quickstart - content tab fixes, added success output for kn func

iRaindrop:
iRaindrop = Bruce Hamilton bhamilton@expertsupport.com

The top two content tab sections in the quickstart, https://knative.dev/docs/getting-started/build-run-deploy-func/, needed attention. In the first, a note interrupted the tab functionality and in the second I added the "kn func invoke" example and moved the existing REST response the "func invoke" tab.

## Proposed Changes: 

Formatting fixes applied for content tabs with correct output examples for func invoke and kn func invoke.
-
-
-

### DIFF
--- a/docs/snippets/proc-running-function.md
+++ b/docs/snippets/proc-running-function.md
@@ -16,7 +16,7 @@ The `run` command builds an image for your function if required, and runs this i
     func run [--registry <registry>]
     ```
 
-!!! note
+    !!! note
     The coordinates for the **image registry** can be configured through an environment variable (`FUNC_REGISTRY`) as well.
 
     Using this command also builds the function if necessary.
@@ -67,13 +67,7 @@ You can verify that your function has been successfully run by using the `invoke
     func invoke
     ```
 
-=== "kn func"
-
-    ```bash
-    kn func invoke
-    ```
-
-!!! Success "Expected output"
+    !!! Success "Expected output"
     ```{ .bash .no-copy }
     Received response
     POST / HTTP/1.1 hello.default.127.0.0.1.sslip.io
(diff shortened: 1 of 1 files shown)


iRaindrop:
CLA is signed.

docs/snippets/proc-running-function.md
```diff
@@ -88,3 +82,15 @@ You can verify that your function has been successfully run by using the `invoke
...
+
+=== "kn func"
+
+    ```bash
+    kn func invoke
+    ```
+
+    !!! Success "Expected output"
+    ```{ .bash .no-copy }
+    INFO:root:OK: Request Received
```
evankanderson, line 95: I'm surprised this produces different output

evankanderson (commented):
/approve

I'll give a `/lgtm` and this will auto-merge once the one question on results from `func` vs `kn func` is answered.

docs/snippets/proc-running-function.md
iRaindrop: Here's my output in VS Code. The previous command before `kn func invoke` was `kn func run` and it successfully built the function. Running `kn func invoke` from another terminal just shows `OK`.

(Running 'kn func invoke' after the function is stopped with Ctrl-C results in a namespace error.)

kn-func-invoke

iRaindrop: I repeated this a few times and I did have pip 25.2 installed even though it says to upgrade.

evankanderson: It's probably because the container is building with an older version of pip.

PR #6348, opened 2025-08-20
Drop 1.19 from version drop-down due to OTel issues & warn folks on the blog

dprotaso:

PR #6370, opened 2025-09-12
Revert "Drop 1.19 from version drop-down due to OTel issues & warn folks on the blog (#6348)"

dprotaso:
This reverts commit ebd4cfefbdea484d39cd67a89471f9fc4dffcf5d.

We've fixed documentation and various issues with our OTel integration. There's one minor bug with metrics but it should be out in the next patch releases. I think we're good to advertise the 1.19 release again.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,8 +5,8 @@
   publish = "site"
 
 [build.environment]
-  NODE_VERSION= "14"
-  PYTHON_VERSION = "3.8"
+  NODE_VERSION= "22"
+  PYTHON_VERSION = "3.9"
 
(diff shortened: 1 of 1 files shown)


dprotaso:
huh - that's a weird failure

/retest

netlify.toml
```diff
@@ -5,8 +5,8 @@
   publish = "site"
 
 [build.environment]
-  NODE_VERSION= "14"
-  PYTHON_VERSION = "3.8"
+  NODE_VERSION= "22"
+  PYTHON_VERSION = "3.9"
```
dprotaso, line 9: Dunno why this fixed the build in this PR and other PRs didn't need the change

PR #6319, opened 2025-08-07
feat: route knative docs base URL from knative.dev/docs to knative.dev

rohan-019:
# Migrate documentation from /docs/* to root domain with 301 redirects

## Overview
This PR restructures the Knative documentation URL hierarchy by moving content from `https://knative.dev/docs/*` to `https://knative.dev/*`, creating a cleaner and more intuitive navigation experience.

fixes: https://github.com/knative/docs/issues/6292

### Configuration Updates
- **mkdocs.yml**: Updated `site_url` from `https://knative.dev/docs` to `https://knative.dev`
- **blog/mkdocs.yml**: Updated homepage reference to point to root domain
- **netlify.toml**: Added comprehensive redirect rules for seamless URL migration

### Redirect Strategy
```toml
# Redirect all /docs/* paths to root-relative equivalents
[[redirects]]
  from = "/docs/*"
  to = "/:splat"
  status = 301

# Serve root path as the documentation homepage  
[[redirects]]
  from = "/"
  to = "/index.html"
  status = 200
```

## Benefits
- **Improved UX**: Shorter, cleaner URLs that are easier to share and remember
- **SEO Preservation**: 301 redirects maintain search engine rankings and link equity
- **Zero Downtime**: Backward compatibility ensures no broken links or service interruption
- **Future-Proof**: Simplified URL structure supports better site organization

## Testing Completed
- ✅ Verified `/docs/*` → `/*` redirects function correctly
- ✅ Confirmed root domain serves documentation properly
- ✅ Validated internal link functionality
- ✅ Tested backward compatibility with existing bookmarks
- ✅ Confirmed SEO-friendly 301 redirect behavior

## Impact
- **Immediate**: Documentation accessible at cleaner root URLs
- **Gradual**: Search engines will re-index URLs over time
- **No Breaking Changes**: All existing `/docs/*` URLs remain functional via redirects

Issue #6302, opened 2025-07-29
Separate technical documentation from the project page


dwelsch-esi:
## Overview

The Knative project [landing page] redirects to the doc page:
`https://knative.dev/docs/`. The project landing page contains all the
elements you'd expect on an open-source software page, including case studies,
user endorsements, and links to the community. 

It also includes the technical documentation in the banner menu. This
arrangement conflates the project and the technical documentation, mixing
marketing and technical information. 

It's helpful to users if there's a clear separation between technical
documentation and other information. The technical documentation is goal-driven
and should be factual, specific, and purposeful.

The request: Separate the technical documentation from the project landing page.

Audience: All

Type: All

## Context

This issue tracks recommended changes resulting from an analysis of the Knative
documentation commissioned by CNCF. The analysis and supporting documents are
here: https://github.com/cncf/techdocs/tree/main/analyses under `0015-knative`.

## Possible Implementation

Related material in the current doc:

- https://github.com/knative/
- https://github.com/knative/docs/

Suggested changes:

The simplest solution to untangling the documentation from the landing page is
probably to:

- Reconfigure the site to use the base URL `https://knative.dev` as the project
  landing page.
- Add a "Documentation" tab on the right alongside "Blog," "About," and
  "Community" on the project page that links to the docs page,
  `https://knative.dev/docs`.
- Move the tech docs items from the banner to the left-hand sidebar (LHS) on the
  tech docs page.
- Remove the version drop-down from the project landing page (and the other
  non-documentation project pages).

evankanderson:
This looks related to #6292, would you agree?

evankanderson:
Completed this in #6371 and #6392 .

PR #6371, opened 2025-09-12
Attempt to rebuild docs build process, inspired by #6319

### LINKED ISSUE #6292: Make knative.dev to be the base url
The base url for knative docs should route to [knative.dev](knative.dev) instead of [knative.dev/docs](knative.dev/docs)

**Additional context**
The main concern here is that this change might cause a breaking change in crawled search results. How do we mitigate that?

evankanderson:
Fixes #6292  #6302 

## Proposed Changes 

Rewrites the documentation building system to run `mkdocs build` exactly once, with all the files in the tree.  The current build runs `mdocs build` N+2 times for the number of releases we have (N releases, plus "development" and "blog"), and forces our homepage, about, and community sites to live under `https://knative.dev/docs` because versioning.

I believe this is relatively complete, but we are stretching the capabilities of `mkdocs`.  I think it's still worth it, but I wanted to get this generally-working checkpoint in before going further.  (See also the unfortunate breakages that @rohan-019 had to deal with in #6319)

- I discovered that `awesome-pages-plugin` had been replaced by `awesome-nav`, which has a few more features. One feature is being able to build up global nav from `.nav.yaml` files in directories. I used that to build up the nav for each subtree of the site separately.
- I replaced a global environment variables with page metadata for things like Knative version and whether there is a newer version.  The global environment variables weren't going to work well with a single `mkdocs` run.
- Eventually, we probably want to move the documentation content to a sub-directory, but I used `mv` for testing. This is probably wrong; I ended up copying some images for now to avoid making weird break/fix changes.
- I removed the extra "blog" build, which removed a lot of duplicated (or even never-referenced) files and images.  That's the majority of the file status of this PR.
   - This necessitated fixing some of the blog image URLs, which meant that I noticed a bunch didn't have alt text.  I fixed those.
- Because the top-level page got moved up to `/`, I needed to redo the `redirects.yml`.  We also currently use this to redirect the top-level docs page to `concepts/`, but when we move the docs under a dedicated sub-directory, we should move that page up to the top level.
- I needed to tweak the navigation templates in a few places.  Logically, we have four levels of navigation:
  - Content section at the top-most bar
  - (docs-specific) Topic-specific tabs, under the topmost bar
  - Between-page navigation on the left
  - TOC / within-page navigation on the right.

  I *think* this is generally correct, but there is definitely a lot.  We might relocate the TOC navigation within the between-page nav (Material-Mkdocs has an option for this), but I didn't want to mess with it here.  There's probably also some additional fixing to be done -- I think we may need to re-do the top nav bar to get it working well on mobile.  I _also_ didn't work on that, because this PR is nearly-unreviewable as is.


evankanderson:
@rohan-019, I'd love your review.

@Cali0707 or @dprotaso -- if you have some cycles, please take a look.

evankanderson:
The `mkdocs build --strict` GitHub Action is failing because it's not running all our lovely custom `build.sh` logic that we use to produce a release.  I'll see what I can do to fix that.

docs/install/operator/knative-with-operator-cli.md
```diff
@@ -156,7 +156,7 @@ you can configure Knative Serving with different ingresses:
     1. Install a properly configured Contour:
 
         ```bash
-        kubectl apply --filename {{artifact(repo="net-contour",file="contour.yaml")}}
+        kubectl apply --filename {{artifact(repo="net-contour",org="knative-extensions",file="contour.yaml")}}
```
evankanderson, line 159: I noticed some build complaints trying to fetch these from the wrong place.  If they offend, I can revert and offer these separately.

docs/stylesheets/extra.css
```diff
@@ -211,11 +211,6 @@ a.md-header__button.md-logo img {
   color: gray;
 }
 
-#sub-nav li:nth-last-child(3), #sub-nav li:nth-last-child(2), #sub-nav li:nth-last-child(1) {
-  display: none;
-}
-
-
 /* bit of a hack to stop the banner taking space when empty */
```
evankanderson, line 214: This is how we hid the "blog" "about" and "community" nav items from the bar in the redesign.  Sorry I missed this!

hack/build.sh
```diff
@@ -106,29 +125,9 @@ mkdir site/golang
...
-<body>
-  Redirecting to <a href="docs/">docs/</a>...
-</body>
-</html>
-EOF
-
 # Clean up
-rm -rf $TEMP
+# rm -rf $TEMP
+echo "Temp dir was: $TEMP"
```
evankanderson, line 137: I can put back this removal, but I found it helpful sometimes to poke into these temp directories while debugging.

hack/macros.py
```diff
@@ -110,12 +110,18 @@ def artifact(repo, file, org="knative"):
...
+            version = env.page.meta.get("knative_version")
 
         if version == None:
             return f'https://storage.googleapis.com/knative-nightly/{repo}/latest/{file}'
 
         version = drop_prefix(version)
 
+        # Don't try to version-match outside Knative-managed repos
+        if org not in ("knative", "knative-extensions"):
+            return f'https://github.com/{org}/{repo}/releases/latest/download/{file}'
```
evankanderson, line 123: Another attempt to avoid a bunch of build time complaints (though there are still a bunch of others that seem harmless but annoying).

mkdocs.yml
```diff
@@ -61,12 +62,22 @@ plugins:
...
+    feeds_filenames:
+      rss_created: "blog/feed_rss_created.xml"
+      rss_updated: "blog/feed_rss_updated.xml"
+    date_from_meta:
+      as_creation: "date"
+      datetime_format: "%Y-%m-%d"
+  # Allow `.meta.yml` to define metadata for all pages below that directory.
+  # See https://squidfunk.github.io/mkdocs-material/plugins/meta/
+  meta:
+    enabled: true
```
evankanderson, line 90: We use this to set the version / outdated / sample branch fields for the versioned documentation in the shell script.

netlify.toml
```diff
@@ -6,7 +6,7 @@
 
 [build.environment]
   NODE_VERSION= "22"
-  PYTHON_VERSION = "3.9"
+  PYTHON_VERSION = "3.13"
```
evankanderson, line 9: `awesome-pages` needs at least 3.10, but I didn't see a good reason not to move past that point.

overrides/main.html
```diff
@@ -22,3 +22,42 @@ <h1>⚠ You are viewing an archived snapshot of the documentation for Knative ve
...
+    {%- set mike = config.plugins.mike -%}
+    {%- if not mike or mike.config.version_selector -%}
+      {%- set _.version = config.extra.version -%}
+    {%- endif -%}
+  {%- endif -%}
+
+  <!-- Configuration -->
+  <script id="__config" type="application/json">
+    {{- {
+      "base": base_url | replace("../", "", 1),
```
evankanderson, line 45: This is the only line changed from the upstream template.

The `mike` versioning plugin (which seems the only one supported) builds a new version of the documentation for each release and checks it in to the `gh-pages` branch under a sub-directory name.  This leads to the `docs/` at the root of the tree problem.

overrides/partials/header.html
```diff
@@ -97,7 +97,7 @@
         </a>
       </li>
       <li class="md-tabs__item">
-        <a href="/docs/about/testimonials/" class="md-tabs__link">
+        <a href="/about/testimonials/" class="md-tabs__link">
```
evankanderson, line 100: We may want to do something more sophisticated / reach out to the material mkdocs team to figure out a better solution for these top-level menu tabs.  I'm not sure they were designed or thought through on mobile.

overrides/partials/nav.html
```diff
@@ -0,0 +1,78 @@
...
+    </div>
+  {% endif %}
+
+  <!-- Navigation list -->
+  {# Create nav from the "section" of docs that we are in.
+     Some sections (about, community) are not versioned, while docs is. #}
+  {% set section = nav | selectattr("active") | first %}
+  {% if section %}
+    {% set nav = section %}
+  {% endif %}
```
evankanderson, line 68: This is the new bit, the rest is imported from upstream.

dprotaso:
/lgtm
/approve
/hold for any other reviews

I did a pass through the preview and it looked fine (eg. the URLs paths work as expected). Feel free to unhold 

As a follow up we might want to drop the the version picker on the home page (and other pages) when you're not looking at the docs

evankanderson:
> /lgtm /approve /hold for any other reviews
> 
> I did a pass through the preview and it looked fine (eg. the URLs paths work as expected). Feel free to unhold
> 
> As a follow up we might want to drop the the version picker on the home page (and other pages) when you're not looking at the docs

Oh, good point.  Unfortunately, I think that may be tricky given that the version picker is added via TypeScript, which is built into a minified bundle: https://github.com/squidfunk/mkdocs-material/blob/master/src/templates/assets/javascripts/integrations/version/index.ts#L151

We may be able to manually set `config.version.provider` to a string other than `mike` in the `{% block config %}` override, which looks like it should disable the selector.  Let me see what I can do.

evankanderson:
I fixed the GitHub Action and the version picker appearing on unversioned pages.  I also chased down a number of Hugo shortcodes (in blog and `smoketest.md`) that never worked with `mkdocs`.

overrides/main.html
evankanderson: (This comment is no longer true, as I also changed the `version` field in the config to only appear on pages that have `samples_branch` metadata set.)

PR #6372, opened 2025-09-15
Advertise qpoption maturity and provide a call to action to inform the maintainers of usage

dprotaso:
I don't want to advertise qpoptions in the docs. It's really an implementation detail for security guard at this point.

### DIFF
--- a/docs/versioned/serving/queue-extensions.md
+++ b/docs/versioned/serving/queue-extensions.md
@@ -5,6 +5,12 @@ components:
 function: how-to
 ---
 
+!!! note
+
+    This integration is current [Alpha](https://github.com/knative/community/blob/main/mechanics/MATURITY-LEVELS.md#alpha-stage).
+    We're trying to guage usage of this so if you're using or interested please reach out to us in
+    [#knative-serving](https://cloud-native.slack.com/archives/C04LMU0AX60) slack.
+
 # Extending Queue Proxy image with QPOptions
 
 Knative service pods include two containers:
--- a/docs/versioned/serving/services/using-queue-extensions.md
+++ b/docs/versioned/serving/services/using-queue-extensions.md
@@ -5,6 +5,12 @@ components:
 function: explanation
 ---
 
+!!! note
+
+    This integration is current [Alpha](https://github.com/knative/community/blob/main/mechanics/MATURITY-LEVELS.md#alpha-stage).
+    We're trying to guage usage of this so if you're using or interested please reach out to us in
+    [#knative-serving](https://cloud-native.slack.com/archives/C04LMU0AX60) slack.
+
 # Using extensions enabled by QPOptions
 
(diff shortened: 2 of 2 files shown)


davidhadas:
What is the benefit for the community in not publishing this extendibility?
I can understand why if such publication was not available you would not spend the effort to publish it, but remove documentation when it is already there?

dprotaso:
> What is the benefit for the community in not publishing this extendibility?

It's only being used by security guard at the moment.

In the future I'd like to revisit how this mechanic works (eg. I like how envoy gateway patches work) but having it on the website will make that harder to do.  eg. I'm expecting an influx of traffic from our graduation announcement.

davidhadas:
This does not make much sense to me. 

Either we are expecting an influx of traffic from our graduation announcement. In which case we may see renewed interest in the project, including in its various features such as security guard.  If indeed graduation will result in teams adopting knative to production projects, they may find security features such as security guard extremely helpful. 

Or that we expect a continued decline in interest in which case I can understand why we would start decreasing the project as a whole and archive parts which are not core - very possibly including security guard. 

In both cases, I would suggest to wait till we see which of the two options occur to decide if it is justified to narrow the project scope and remove its features and documentation related to such features.

dprotaso:
You're conflating security-guard with qpoptions.

I'm saying I want to revisit the mechanic of how qpoptions works thus I don't want people to write new extensions using it.

davidhadas:
Of course we have full control over people writing new extensions using it, as such extensions would be evaluated by the community and suspended based on a new plan to do the extendability, rather than approved in serving.

We can add a sentence to the documentation to state that a new extendability feature is planned and no further use of this extendability is presently expected. 

At the same time removing it will result in making it hard for people to follow how the extension presently used by security guard works. I am quite sure that security guard docs also link to these pages and rely on their content. We could with some effort clean all such references and ensure to copy the data to the security guard docs such that it will not appear in serving (although reffering to serving configs). I again suggest that this is not a productive move at this point in time, but if you insist on removing it, I would think it is wise to move it (with some rephrasing) rather than to remove it.

evankanderson:
Sorry about the merge -- in https://github.com/knative/docs/pull/6398 , I moved all the versioned documentation under versioned, to separate it from the blog, testimonials, community pages that are not version-specific. You should be able to just move these files to the new location; I didn't need to change any contents.

dprotaso:
/hold cancel

I change the PR. Instead of removing the content I added an admonition that indicates the alpha maturity and a call to action for those interested to reach out. 

If we don't hear from anybody in a few months then that might be a good enough signal.

evankanderson:
We looked at download numbers for security-guard, and they are generally <12 for the most recent release from a year ago, but we don't know if other teams or projects might have taken advantage of this hook without mentioning it to us.

evankanderson (commented):
/lgtm
/approve

PR #6381, opened 2025-09-17
fix homepage links

dprotaso:


Cali0707:
There are a lot of warnings throughout the deployment logs about other missing links, we should probably go through and verify them all throughout the website

evankanderson (approved):
/lgtm

PR #6392, opened 2025-09-17
Fix site search

evankanderson:
This is gross, but should fix site search temporarily, while I [negotiate a better fix upstream in the built mkdocs TypeScript bundle](https://github.com/squidfunk/mkdocs-material/issues/8453).

## Proposed Changes 

- Vendor a typescript bundle built from https://github.com/evankanderson/mkdocs-material/tree/versioning
- Adjust the build process to store version info in metadata
- Adjust the build process to remove all documentation except the current release from the search index

This fixes search, and also fixes the version-picking nav (which will try to go to the equivalent page in the target version, if it exists, and otherwise navigate to the docs root, e.g. http://knative.dev/development/reference/security/verifying-images/ ==> http://knative.dev/v1.18-docs/reference/security/verifying-images/)

Note that (for the first time) search works across both the blog and the technical documentation -- previously, they were separate builds, so you wouldn't find blog posts searching the main documentation, and you wouldn't find documentation pages when searching from the blog.


evankanderson:
This is not working correctly.  Let me try a worse approach.

evankanderson:
It turns out search uses `config.base` for the output URLs.  My next attempt will be to fix the version-picker logic, which seems slightly simpler (though inside the minified JS bundle currently).

dprotaso:
Should search work in the preview?

evankanderson:
Not yet, I broke a second thing and it turns out the fix is more complex

dprotaso:
> Not yet, I broke a second thing and it turns out the fix is more complex

Do you want to merge this now and then provide a follow up?

evankanderson:
I believe this is now ready for review.

evankanderson:
Hmm, the cross-version nav worked locally, but not on the site.  This still fixes search, so I think it's an improvement, but I'll need to see what's going on with cross-version navigation later.

evankanderson:
> Hmm, the cross-version nav worked locally, but not on the site. This still fixes search, so I think it's an improvement, but I'll need to see what's going on with cross-version navigation later.

The cross-version deep-linked navigation seems to be being defeated by a change to the `sitemap.xml` when built by netlify.  Comparing built locally vs netlify:

```
$ grep -o '/v1.17-docs/\?<' netlify-sitemap.xml site/sitemap.xml
netlify-sitemap.xml:/v1.17-docs<
site/sitemap.xml:/v1.17-docs/<
```

evankanderson:
It turns out we're using the netlify sitemap plugin (probably a holdover from Hugo days).  I reconfigured it to include the trailing slash until I can get into the console to disable it.

evankanderson:
Okay, this is ready for review; search and between-version nav are both working correctly, AFAICT.

dprotaso:
as a follow up is there a way to build mkdocs-material and patch it - in case we need to make further edits?

or alternatively fork the project given your upstream contribution is paused?

evankanderson:
> as a follow up is there a way to build mkdocs-material and patch it - in case we need to make further edits?
> 
> or alternatively fork the project given your upstream contribution is paused?

My pull request is in a fork -- I was able to clone the fork using GitHub Codespaces in order to produce a bundle with the same checksum, so I think we can probably reproduce it (though we can certainly produce more forks of my fork if we want).

Issue #6305, opened 2025-07-29
Update Installation to better orient users


dwelsch-esi:
## Overview

The installation instructions are generally very good, but there are a few speed
bumps to finding and performing the right installation procedure.

Add "roadmap" information to orient readers in the Installation section of the
documentation.

Audience: Admins and developers

Type: Instructional

## Context

This issue tracks recommended changes resulting from an analysis of the Knative
documentation commissioned by CNCF. The analysis and supporting documents are
here: https://github.com/cncf/techdocs/tree/main/analyses under `0015-knative`.

## Possible Implementation

Related material in the current doc:

- https://github.com/knative/install

Suggested changes:

On the [Installing Knative] page, provide a roadmap at the top to direct users
to the correct installation. Aside from installing the different components
(Eventing and Serving) using different modalities (YAML, Kubernetes Operator,
and upgrading), disclose what operating systems are supported, and differentiate
between local and server installations.

skmahe1077:
I’d like to take on this issue. Could someone please assign it to me?

allexistence:
/assign

I’ve raised a PR to address this. Could you please assign it to me and take a look? I’m happy to make any changes based on feedback.

PR #6403, opened 2025-09-24
Move install docs to administration

evankanderson:
Related to #6305 -- today, the "Install" section is the third section in the docs (after "Overview" and "Tutorial").  This suggests that most of the audience should read that next, but 90% of the content only applies to administrators.

## Proposed Changes 

- Moves "Install" topic to an "Administration" section.
- Removes install/quickstart-install in favor of getting-started/quickstart-install, with almost exactly the same content.
- Removes install/install-kn in favor of client/install-kn, with the same content
- Cleans up some homepage content after checking the homepage nav and noticing that the primary CTA button styling actually made the secondary button more visible.  I ended up adding some CSS selectors because the `home.css` interacts with the material theme CSS and sometimes needs to be more specific.


overrides/home.html
```diff
@@ -17,12 +17,12 @@ <h2 class="secondary-headline">
...
             <a
               href="{{ 'docs/getting-started/tutorial' | url }}"
-              class="md-button black md-button--primary"
+              class="md-button md-button--primary"
               >Try the tutorial</a
             >
             <a
               href="{{ 'docs/install/' | url }}"
-              class="md-button white md-button--secondary"
+              class="md-button"
```
dprotaso, line 25: When should we update this href to the newer location?

PR #6404, opened 2025-09-25
Add a note that Apache Kafka is required to use EKB

creydr:
On the overview page we don't say that Apache Kafka is required to use EKB (only on a subpage). This PR adds a quick note about it


creydr:
@twoGiants FYI

PR #6406, opened 2025-09-25
docs: explain KEDA/no KEDA scaling more

### LINKED ISSUE #6402: Explain that KEDA Kafka Scaling for EKB scales consumers, not pods
## Expected Behavior
In https://knative.dev/docs/eventing/configuration/keda-configuration/#configure-autoscaling-for-a-resource it is clear to users that the different min-scale, max-scale parameters are used to scale the number of consumers, not the number of pods.

## Actual Behavior
The docs don't mention whether or not they are for consumers or pods

Cali0707:
Resolves #6402 

This PR adds much more explanation for how the EKB scaling works with and without KEDA

### DIFF
--- a/docs/versioned/eventing/configuration/keda-configuration.md
+++ b/docs/versioned/eventing/configuration/keda-configuration.md
@@ -34,9 +34,77 @@ data:
 
 ```
 
+## Configure KEDA scaling globally
+
+To configure the KEDA scaling behaviour globally for all resources you can edit the configmap `config-kafka-autoscaler` in the `knative-eventing` namespace.
+
+Note that the `min-scale` and `max-scale` parameters both refer to the number of Kafka consumers, not the number of dispatcher pods. The number of dispatcher pods in the StatefulSet is determined by `scale / POD_CAPACITY`, where `POD_CAPACITY` is an environment variable you can configure on the `kafka-controller` deployment, and defaults to `20`.
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-kafka-autoscaler
+  namespace: knative-eventing
+data:
+  # What autoscaling class should be used. Can either be keda.autoscaling.knative.dev or dispabled.
+  class: keda.autoscaling.knative.dev
+  # The period in seconds the autoscaler waits until it scales down
+  cooldown-period: "30"
+  # The lag that is used for scaling (1<->N)
+  lag-threshold: "100"
+  # The maximum number of replicas to scale up to
+  max-scale: "50"
+  # The minimum number of replicas to scale down to
+  min-scale: "0"
+  # The interval in seconds the autoscaler uses to poll metrics in order to inform its scaling decisions
+  polling-interval: "10"
+```
+
+## Understanding KEDA Scaling Patterns
+
+KEDA uses different lag thresholds to determine when to scale your Kafka components:
+
+### Scaling from 0 to 1 (Activation)
+- **Trigger**: When consumer lag exceeds the `activation-lag-threshold` (default: 0)
+- **Behavior**: KEDA activates the first replica when any messages are waiting in the Kafka queue
+- **Use case**: Quick response to incoming events when the system is idle
+
+### Scaling from 1 to N (Scale Up)
+- **Trigger**: When consumer lag exceeds the `lag-threshold` (default: 100)
+- **Behavior**: KEDA adds more consumer replicas based on the lag-to-threshold ratio
+- **Formula**: `desired_replicas = min(ceil(current_lag / lag_threshold), max_scale)`
+
+### Scaling Down and to Zero
+- **Trigger**: When consumer lag falls below thresholds
+- **Behavior**: KEDA waits for the `cooldown-period` (default: 30 seconds) before reducing replicas
+- **Scale to zero**: Occurs when lag is below `activation-lag-threshold` for the duration of the cooldown period
+
+### StatefulSet Pod Calculation
+Each resource type (KafkaSources, Triggers, etc.) has its own dispatcher StatefulSet. Within each StatefulSet, the number of dispatcher pods is calculated as:
+```
+dispatcher_pods = ceil(total_consumers_for_resource_type / POD_CAPACITY)
+```
+
+Where `POD_CAPACITY` defaults to 20, meaning each dispatcher pod can handle up to 20 Kafka consumers.
+
+**Important**: Each resource (KafkaSource, Trigger, Subscription) creates its own consumer group, and KEDA scales the number of consumers within that consumer group. All consumers for a given resource type are distributed across the dispatcher pods in that type's StatefulSet.
+
+To maintain a **fixed minimum number of StatefulSet pods** for a resource type, calculate the total consumers needed:
+```
+total_consumers = sum_of_all_min_scales_for_resource_type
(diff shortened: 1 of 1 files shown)


docs/versioned/eventing/configuration/keda-configuration.md
```diff
@@ -34,9 +34,77 @@ data:
...
+
+Note that the `min-scale` and `max-scale` parameters both refer to the number of Kafka consumers, not the number of dispatcher pods. The number of dispatcher pods in the StatefulSet is determined by `scale / POD_CAPACITY`, where `POD_CAPACITY` is an environment variable you can configure on the `kafka-controller` deployment, and defaults to `20`.
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-kafka-autoscaler
+  namespace: knative-eventing
+data:
+  # What autoscaling class should be used. Can either be keda.autoscaling.knative.dev or dispabled.
```
evankanderson, line 49:
```suggestion
  # What autoscaling class should be used. Can either be keda.autoscaling.knative.dev or deactivated.
```

evankanderson: (typo + inclusive language)

docs/versioned/eventing/configuration/keda-configuration.md
```diff
@@ -34,9 +34,77 @@ data:
...
+  class: keda.autoscaling.knative.dev
+  # The period in seconds the autoscaler waits until it scales down
+  cooldown-period: "30"
+  # The lag that is used for scaling (1<->N)
```
evankanderson, line 53: This should probably have a unit?

docs/versioned/eventing/configuration/keda-configuration.md
```diff
@@ -34,9 +34,77 @@ data:
...
+  lag-threshold: "100"
+  # The maximum number of replicas to scale up to
```
evankanderson, line 55: This says "replicas", but the comment above talks about "Kafka consumers" (which I assume means threads reading from a shared consumer group)?

docs/versioned/eventing/configuration/keda-configuration.md
```diff
@@ -34,9 +34,77 @@ data:
...
+  # The interval in seconds the autoscaler uses to poll metrics in order to inform its scaling decisions
+  polling-interval: "10"
+```
+
+## Understanding KEDA Scaling Patterns
+
+KEDA uses different lag thresholds to determine when to scale your Kafka components:
+
+### Scaling from 0 to 1 (Activation)
+- **Trigger**: When consumer lag exceeds the `activation-lag-threshold` (default: 0)
```
evankanderson, line 68: `activation-lag-threshold` doesn't seem to be in the above ConfigMap.  Is this another key?

docs/versioned/eventing/configuration/keda-configuration.md
```diff
@@ -34,9 +34,77 @@ data:
...
+- **Behavior**: KEDA waits for the `cooldown-period` (default: 30 seconds) before reducing replicas
+- **Scale to zero**: Occurs when lag is below `activation-lag-threshold` for the duration of the cooldown period
+
+### StatefulSet Pod Calculation
+Each resource type (KafkaSources, Triggers, etc.) has its own dispatcher StatefulSet. Within each StatefulSet, the number of dispatcher pods is calculated as:
+```
+dispatcher_pods = ceil(total_consumers_for_resource_type / POD_CAPACITY)
+```
+
+Where `POD_CAPACITY` defaults to 20, meaning each dispatcher pod can handle up to 20 Kafka consumers.
```
evankanderson, line 88: In the example above, `max-scale` was set to 50, which seems like it would mean 3 dispatcher pods (the same as if the value were 60).  Is there a practical difference between 41 and 60 here given the default value of `POD_CAPACITY`?

I ask because we had a few samples for Serving autoscaling where the end result was that a single request coming in for a scaled-down pod would sometimes cause 2 pods to start, because the pending request + the "spare capacity" factor ended up crossing the one-pod capacity threshold (i.e. the "spare capacity" number was exactly equal to "pod capacity", so `ceil(200 + 1 / 200) => 2`).

docs/versioned/eventing/configuration/keda-configuration.md
Cali0707: disabled is the key we are using in the code afaik - I just took it from the repo

matzew: +1 it is there since longer.

docs/versioned/eventing/configuration/keda-configuration.md
```diff
@@ -72,3 +140,32 @@ metadata:
...
+
+## Manual Scaling with KEDA Disabled
+
+When KEDA is disabled (either globally or for specific resources), you can manually control the number of StatefulSet pods by scaling the consumer group resources directly.
+
+Each Kafka resource creates a corresponding `ConsumerGroup` resource in the same namespace that controls the number of consumers. You can manually scale these to achieve your desired pod count:
+
+```bash
+# List all consumer groups in your workload namespace
+kubectl get consumergroups -n <your-namespace>
```
matzew, line 152: can we use the FQN of the resource, @Cali0707 ?

docs/versioned/eventing/configuration/keda-configuration.md
```diff
@@ -34,9 +34,77 @@ data:
...
+**Important**: Each resource (KafkaSource, Trigger, Subscription) creates its own consumer group, and KEDA scales the number of consumers within that consumer group. All consumers for a given resource type are distributed across the dispatcher pods in that type's StatefulSet.
+
+To maintain a **fixed minimum number of StatefulSet pods** for a resource type, calculate the total consumers needed:
+```
+total_consumers = sum_of_all_min_scales_for_resource_type
+min_dispatcher_pods = ceil(total_consumers / POD_CAPACITY)
+```
+
+For example:
+- 2 Triggers each with `min-scale: "40"` = 80 total consumers
```
twoGiants, line 99: Not rendered as a list. You probably need a blank line between the bulletpoint and the "For example:" header.

Screenshot From 2025-09-30 18-57-47

docs/versioned/eventing/configuration/keda-configuration.md
```diff
@@ -34,9 +34,77 @@ data:
...
+- With default `POD_CAPACITY: 20`, this creates `ceil(80/20) = 4` dispatcher pods for the Trigger StatefulSet
+- All 80 consumers are distributed across these 4 pods
+
 ## Configure Autoscaling for a Resource
 
-If you want to customize how KEDA scales a KafkaSource, trigger, or subscription you can set annotations on the resource:
+If you want to customize how KEDA scales a KafkaSource, trigger, or subscription you can set annotations on the resource.
```
twoGiants, line 105: nit: here `trigger, or subscription` have a small first letter, but [here](https://github.com/knative/docs/pull/6406/files#diff-95fa06ee23f165cbf1dbdc09e998c4cb8c8111be61c614b51ce28a45610f233fR90) a capital first letter :)

creydr:
@Cali0707 sorry, that this PR got stale and closed. I really like to improve the docs here. Could you address the comments and reopen the PR?

Issue #6380, opened 2025-09-17
Unable to navigate docs on mobile


Cali0707:
## Expected Behavior
When navigating to the docs section of the website on mobile, I should be able to select the components I want (e.g. functions, serving, etc.)

## Actual Behavior
I only see the options in the nav for the sub-categories on the home page of the docs:

<img width="337" height="502" alt="Image" src="https://github.com/user-attachments/assets/7f77c8a6-7eb6-43d5-bad8-4234a5910b7c" />

## Steps to Reproduce the Problem

1. Visit the docs page on mobile view

## Additional Info
**Additional context**

**Install information:**
 - Platform (GKE, IKS, AKS, etc.):
 - Knative Version:

Cali0707:
cc @dprotaso @evankanderson

Cali0707:
IMO if we can't find a way to solve this before the graduation announcement, we should rollback the docs navigation changes that led to this - otherwise users won't be able to find anything on mobile

evankanderson:
Thanks for flagging this!  I thought I had tried it earlier and the navigation worked. 

/accept

evankanderson:
For me, if you select the back arrow in the upper-left you can navigate to the other topics, but I agree that the result is non-optimal (though I'm trying to figure out what the optimal result would be for both showing where you are in the nav and showing the parent topics).

![Screenshot_20250917-075256.Arc Search.png](https://github.com/user-attachments/assets/2e3e0b4f-306f-43ab-9b2a-728f3b7eef37)

Cali0707:
> For me, if you select the back arrow in the upper-left you can navigate to the other topics, but I agree that the result is non-optimal 

Yeah, that worked for me as well - but the first time I tried the website on mobile that didn't occur to me. My fear is that users will look, and then leave the website when they can't find where they want to go quickly

Cali0707:
IMO, if you navigate to the _home_ for the docs on mobile, the side nav should give the "top-level" nav, not the nested nav for that page when you first visit it.

dprotaso:
@Cali0707 that might be a mkdocs specific thing - eg. see: https://gateway-api.sigs.k8s.io/

dprotaso:
and see - https://squidfunk.github.io/mkdocs-material/

Cali0707:
> [@Cali0707](https://github.com/Cali0707) that might be a mkdocs specific thing - eg. see: https://gateway-api.sigs.k8s.io/

[...]

PR #6413, opened 2025-09-30
Compress spacing on header at small sizes

evankanderson:
Part of #6380

(There will be more changes, but these were quick and improve the top bar a fair bit.)

## Proposed Changes 

- Reduce spacing around version drop-down at under-630px wide to about "6rem" (about 80px) from about "11rem" (about 150px).  This makes the first 4 topics (Home, Docs, About, Blog) visible at 400px wide.
- Reduce spacing around the search bar, this was consuming around 30px (2rem), which makes "Com" from "Community" visible at 400px wide.

### DIFF
--- a/docs/stylesheets/extra.css
+++ b/docs/stylesheets/extra.css
@@ -4,54 +4,63 @@
   font-weight: 400;
   src: local('Satoshi'), url('https://fonts.cdnfonts.com/s/85546/Satoshi-Regular.woff') format('woff');
 }
+
 @font-face {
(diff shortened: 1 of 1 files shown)


dprotaso:
/lgtm
/approve

Is there a way to make it so the search button doesn't cause the top nav to move around? eg. on the mkdocs material site clicking the search makes it a modal (https://squidfunk.github.io/mkdocs-material/)

dprotaso:
Also do we want to cherry pick this change?

PR #6415, opened 2025-09-30
Drop Kong in our documentation since they don't support Knative

dprotaso:
See: https://github.com/Kong/kubernetes-ingress-controller/issues/2813


evankanderson (commented):
/lgtm
/approve

PR #6428, opened 2025-10-07
Re-arrange navigation to fill out 'Administration' topic

evankanderson:
Related to #6305 and #6412

## Proposed Changes 

- The previous "install" pages (which included upgrade and uninstall) were moved to Administration in #6403, but there were still a lot of "how to install" documents scattered throughout the repo (particularly in eventing)
- Similarly, administration topics for configuring Knative were scattered between "Serving" and "Eventing", and pages which could be done by non-privileged users were mixed with documents that required administrator privileges to run.
  - I had to split the Kafka and RabbitMQ broker docs, but most docs clearly targeted one role or another.
- This PR **does not** generally move the actual pages (and URLs), but instead restructures the navigation.  This somewhat kicks the can of moving things down the road.  I chose to do this because:
  1. Given past history, this may need a little time to settle, and moving documents means we need to put in redirects (more toil)
  1. Moving document means updating links, and I'd like to fix a bunch of the internal link warnings in `--strict` mode, so we can detect when we break these links with higher reliability.


evankanderson:
New content starts here: https://deploy-preview-6428--knative.netlify.app/development/administer/

iRaindrop:
It's a good restructuring. Good to the Administrator node with content that applies to both Serving and Eventing. Navigation is more intuitive. Important concepts are more prominent.

PR #6446, opened 2025-10-14
Remove TriggerMesh to resolve invalid URL

johansja:
Seems like TriggerMesh has shut down according to this discussion.

https://github.com/triggermesh/triggermesh/issues/1544


dprotaso:
/lgtm
/approve

thanks!

Issue #6431, opened 2025-10-08
A lot of stutter in our left navigation


dprotaso:
I was wondering if we should aim to clean up our navigation a little bit. I notice there is quite a bit of 'stutter' among the sections and subsections

<img width="274" height="347" alt="Image" src="https://github.com/user-attachments/assets/ac1e1bea-d4fe-49f1-8c38-6d4225dc0777" />

Secondly, some navigation items appear to be 'nodes' that only group together a bunch of subtopics. This leads to excessive clicks to find content.

I don't know what the right solution is - but am looking for ideas.

dprotaso:
cc @iRaindrop @evankanderson for input

evankanderson:
Yes!  I think we may need to write and/or refactor content to improve the navigation to allow us to reduce this stutter.

dprotaso:
I also see that[ material mkdocs](https://squidfunk.github.io/mkdocs-material/getting-started/) supports having bolded headers - some of them can be pages like 'Tutorials' and 'Changelog'. 'Getting Started' doesn't have a link.

<img width="316" height="560" alt="Image" src="https://github.com/user-attachments/assets/a30465db-d91a-4adc-baef-620f122e8bbc" />

dprotaso:
Though - I guess you lose the indentation

lakxy:
Hi @evankanderson @dprotaso 

Is this issue still open for contribution ?
there is a linked PR but it is dated earlier than the issue.

evankanderson:
Hi,

This is still open for contribution.  The referenced PR was being asked to avoid introducing _more_ stutter, rather than cleaning up existing instances in the navigation.

evankanderson:
(you can use `/assign` to assign an issue to yourself)

lakxy:
Hey @evankanderson 
Are there any specific reference that I can follow in my implementation?
Any design pattern that should be followed...as other CNCF projects.
Once I have clarity over the design pattern I can move ahead to resolving this nav issue.

Thank you.

lakxy:
Hi @dprotaso  @evankanderson 
I need confirmation, Is the expected outcome supposed to align with patterns/practices in the `material/mkdocs` mentioned.

Please clarify and I will fix it!
Thank you.

evankanderson:
[...]

PR #6457, opened 2025-10-17
Fix rename pages 6303

### LINKED ISSUE #6303: Rename pages
## Overview

Well named pages and headings help users quickly find the right information.
Consistent naming aids in scanning the table of contents, and accurate names
make search more effective.

Audit page and heading titles for consistency and accuracy.

Audience: All

Type: All

## Context

This issue tracks recommended changes resulting from an analysis of the Knative
documentation commissioned by CNCF. The analysis and supporting documents are
here: https://github.com/cncf/techdocs/tree/main/analyses under `0015-knative`.

## Possible Implementation

Related material in the current doc:

- https://github.com/knative/docs/
- https://knative.dev/docs/client/
- https://knative.dev/docs/getting-started/tutorial/
- https://knative.dev/docs/serving/
- https://knative.dev/docs/eventing/

Suggested changes:

- Remove "About" from titles:
  - "About Installing Knative" > "Installing Knative"
  - "About Eventing Features" > "Eventing Features"
  - and so on.
- Rename "Tutorial" to "Tutorials" (there is more than one tutorial here).
- Make verb tense consistent in procedure titles. Use "-ing" verbs to title tasks
	and processes (thiis is already done in the majority of cases!):
  - [Configure Broker defaults] > "Configuring Broker defaults"
  - [Install Knative using quickstart] > "Installing Knative using quickstart"
  - and so on.
- In [Serving] and [Eventing], rename "Developer Topics" and
  "Admin Topics" to "Developer Tasks" and "Admin Tasks" to cue readers.
- Retitle this page:
  [Knative Eventing - The Event-driven application platform for Kubernetes]
  to "Eventing".
- Rename "Knative CLI" to "CLI tools".
- Make TOC titles agree with page titles. Some examples (not a complete list):
  - Change "Install the Knative CLI" in the TOC to match the page title:
    [Installing the Knative CLI].
  - Change the page title: [Metrics] to match "Configuring metrics" in the TOC.
  - Change "About Apache Kafka Broker" in the TOC to match the page title:
    [Knative Broker for Apache Kafka].
- Rename "About" in the landing page banner to something less misleading: "Why
	Knative?", "Testimonials and Case Studies", or "Endorsements".

skmahe1077:
Fixes #6303 

This PR updates documentation page titles and navigation labels for consistency across the Knative Docs site based on the recommendations in issue.

### DIFF
--- a/docs/versioned/.nav.yml
+++ b/docs/versioned/.nav.yml
@@ -12,7 +12,7 @@ nav:
           - Tutorial introduction: getting-started/README.md
           - Install Knative using quickstart: getting-started/quickstart-install.md
           - Knative Functions:
-            - Knative Functions: getting-started/about-knative-functions.md
+            - Getting Started: getting-started/about-knative-functions.md
             - Installing Knative Functions: getting-started/install-func.md
             - Creating a function: getting-started/create-a-function.md
             - Building, running, or deploying a function: getting-started/build-run-deploy-func.md
@@ -21,7 +21,7 @@ nav:
             - Autoscaling: getting-started/first-autoscale.md
             - Traffic splitting: getting-started/first-traffic-split.md
           - Knative Eventing:
-            - Knative Eventing: getting-started/getting-started-eventing.md
+            - Getting Started: getting-started/getting-started-eventing.md
             - Sources, Brokers, and Triggers: getting-started/first-broker.md
             - Using a Knative Service as a source: getting-started/first-source.md
             - Using Triggers and sinks: getting-started/first-trigger.md


dprotaso:
cc @dwelsch-esi @iRaindrop for review here

docs/versioned/.nav.yml
```diff
@@ -6,13 +6,13 @@ nav:
...
 ###############################################################################
-    - Tutorial:
+    - Tutorials:
         - Overview: getting-started/tutorial.md
         - Quickstart:
           - Tutorial introduction: getting-started/README.md
           - Install Knative using quickstart: getting-started/quickstart-install.md
           - Knative Functions:
-            - About Knative Functions: getting-started/about-knative-functions.md
+            - Knative Functions: getting-started/about-knative-functions.md
```
dprotaso: I don't want to introduce stutter  (eg. https://github.com/knative/docs/issues/6431)

Should we maybe title these sections `Getting Started`?

docs/versioned/.nav.yml
```diff
@@ -73,9 +73,9 @@ nav:
         - Additional autoscaling configuration for Knative Pod Autoscaler: serving/autoscaling/kpa-specific.md
         - Autoscale Sample App - Go: serving/autoscaling/autoscale-go/README.md
       # Serving - developer docs
-      - Developer Topics:
+      - Developer Tasks:
         - Services:
-          - About Knative Services: serving/services/README.md
+          - Knative Services: serving/services/README.md
```
dprotaso, line 78: Don't know the right title here

docs/versioned/.nav.yml
skmahe1077: I will change Knative Functions to "Getting Started"

docs/versioned/.nav.yml
skmahe1077: I’ve removed “About” from the title for consistency with other sections. Happy to update it if there’s a preferred alternative

docs/versioned/eventing/README.md
```diff
@@ -5,7 +5,7 @@ components:
 function: explanation
 ---
 
-# Knative Eventing - The Event-driven application platform for Kubernetes
+# Eventing
```
evankanderson, line 8: Note that because we have explicit `.nav.yaml`, the titles here aren't reflected in the side nav, so we can have longer titles here and shorter titles in the nav.

My $0.02 here would be to keep the longer name for SEO purposes.

docs/versioned/.nav.yml
```diff
@@ -21,7 +21,7 @@ nav:
             - Autoscaling: getting-started/first-autoscale.md
             - Traffic splitting: getting-started/first-traffic-split.md
           - Knative Eventing:
-            - About Knative Eventing: getting-started/getting-started-eventing.md
+            - Knative Eventing: getting-started/getting-started-eventing.md
```
evankanderson: This should use the same pattern as line 15.

docs/versioned/.nav.yml
evankanderson: It turns out that it doesn't matter in this case, because building the nav causes this README.md page to be organized under `Services` as a clickable link, and "About Knative Services" isn't shown at all.

Example: https://knative.dev/docs/serving/services/

docs/versioned/.nav.yml
```diff
@@ -130,7 +130,7 @@ nav:
             # add default IMC broker page, page explaining broker types
             - Channel based Broker: eventing/brokers/broker-types/channel-based-broker/README.md
             - Apache Kafka:
-              - About Apache Kafka Broker: eventing/brokers/broker-types/kafka-broker/README.md
+              - Knative Broker for Apache Kafka: eventing/brokers/broker-types/kafka-broker/README.md
```
evankanderson, line 133: This title change also doesn't matter, because the singleton with README.md gets parented to "Apache Kafka".  I do think the longer name in the title makes sense -- it just won't show up in the nav either way.

evankanderson (commented):
One real comment on the top-level title for Knative Eventing (though I could be convinced that we need metadata rather than a longer title), and I agree with Dave that "Getting Started" or moving the files to clickable parent navigation are better for the tutorial concept sections for Functions and Eventing (apparently we never wrote one for serving!)

/approve

(Will wait on the other changes for `/lgtm`)

docs/versioned/.nav.yml
dprotaso: oh - that's confusing - what's the best way to represent that in the nav yaml file so it's clear?

docs/versioned/.nav.yml
skmahe1077: Modified Getting Started from Knative Functions

docs/versioned/.nav.yml
skmahe1077: Modified Getting started from Knative Eventing

docs/versioned/.nav.yml
evankanderson: I think you could take out the title, see https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#section-index-pages

evankanderson (commented):
One answer for Dave, but I think this overall makes sense.

I slightly wonder about losing the SEO on "Event mesh", but not strongly enough to block the rest of the improvements.

/lgtm
/approve

dprotaso:
> I slightly wonder about losing the SEO on "Event mesh", but not strongly enough to block the rest of the improvements.

I don't see a diff removing "Event Mesh" - can you point that out?

PR #6458, opened 2025-10-18
Plugin: Contour

iRaindrop:
Need topic to install Contour plugin. 

Content derived from https://projectcontour.io/getting-started/

## Proposed Changes 

- Create Installing Contour to include in the Plugin section for Installing Knative.

### DIFF
--- a/docs/versioned/install/installing-contour.md
+++ b/docs/versioned/install/installing-contour.md
@@ -32,17 +32,19 @@ For information about Contour versions, see the Contour [Compatibility Matrix](h
     --patch '{"data":{"ingress-class":"contour.ingress.networking.knative.dev"}}'
     ```
 
-1. Get the External IP address or CNAME:
+1. Get the External IP address or full qualified domain name (CNAME):
 
     ```bash
     kubectl --namespace contour-external get service envoy
     ```
 
+    Use this value to configure your external DNS records.
+
 The adapter's configurations are performed natively through Contour. For more information and resources see [Contour](https://projectcontour.io/) home page.
 
 ## Visibility
 
-The following table shows the classes and services that expose Contour services.
+The following table shows the classes and services that expose Contour networking.
 
 | ExternalIP | ClusterLocal |
 | --- | --- |


docs/versioned/install/installing-contour.md
```diff
@@ -0,0 +1,209 @@
...
+function: how-to
+---
+
+# Installing Contour for Knative
+
+This page shows how to install Contour in three ways:
+
+- By using Contour’s example YAML.
+- By using the Helm chart for Contour.
+- By using the Contour gateway provisioner.
```
evankanderson: We should focus on installing the Knative adapter for Contour, rather than installing Contour itself.  I'd make "Contour [installed](https://projectcontour.io/docs/1.33/deploy-options/) on the cluster" a pre-requisite, and then just talk about installing the `net-contour` controller.

docs/versioned/install/installing-contour.md
```diff
@@ -0,0 +1,209 @@
...
+
+It then shows how to deploy a sample workload and route traffic to it through Contour.
+
+This guidance uses all default settings. No additional configuration is required.
```
evankanderson: I liked how the Kourier docs provided information about configuration options.  It may be that `net-contour` does not have any separate configuration options, but if so, we should say that the configuration is done natively through Contour configuration, and people should see the guides at https://projectcontour.io/

docs/versioned/install/installing-contour.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Before you begin
+
+This installation requires the following prerequisites:
+
+- A Kubernetes cluster with the Knative Serving component installed.
+- Knative [load balancing](../serving/load-balancing/README.md) is activated.
+- HELM installed locally, if selected as the installation method.
```
evankanderson: "Helm" as project is not an initialism, and shouldn't be in all-caps.  In any case, this seems like guidance for installing Contour, and we should simply make that a pre-requisite, rather than duplicating Contour's installation instructions.

docs/versioned/install/installing-contour.md
```diff
@@ -0,0 +1,209 @@
...
+
+    ```bash
+    kubectl -n projectcontour port-forward service/envoy-contour 8888:80
+    ```
+
+In a browser or via curl, make a request to `http://local.projectcontour.io:8888`. The `local.projectcontour.io` URL is a public DNS record resolving to `127.0.0.1` to make use of the forwarded port. You should see the httpbin home page.
+
+## See also
+
+Contour [Getting Started](https://projectcontour.io/getting-started/) documentation.
```
evankanderson: All of these instructions get Contour installed on the cluster, but they don't get `net-contour` (the Knative component that performs the adapter layer between Knative abstract Routes and Contour HTTPProxy objects) installed.  We have some instructions for that in `install-serving-with-yaml.md`:

```markdown
    1. Install the Knative Contour controller by running the command:
      ```bash
      kubectl apply -f {{ artifact(repo="net-contour",org="knative-extensions",file="net-contour.yaml")}}
      ```

    1. Configure Knative Serving to use Contour by default by running the command:
      ```bash
      kubectl patch configmap/config-network \
        --namespace knative-serving \
        --type merge \
        --patch '{"data":{"ingress-class":"contour.ingress.networking.knative.dev"}}'
      ```

    1. Fetch the External IP address or CNAME by running the command:

        ```bash
        kubectl --namespace contour-external get service envoy
        ```

        !!! tip
            Save this to use in the following [Configure DNS](#configure-dns) section.
```

(I removed the step 1 which used a `contour.yaml` which was published by Knative, because I think the necessary changes have already been contributed upstream.

You may want to document at least the `visibility` configuration from https://github.com/knative-extensions/net-contour/blob/main/config/config-contour.yaml#L50 as well, though I think most of the configuration for `net-contour` is inherited from Contour's defaults described in the Contour documentation.

evankanderson (commented):
I agree with Dave that we don't really want to duplicate the Contour installation documentation here.  I seem to recall that we have some sort of specialized Contour configuration in `net-contour` -- @dprotaso , is that `contour.yaml` still needed, or can users install the upstream Contour at this point?  (My recollection was that it installed two Contour installations, one for external addresses and one for internal services.)

dprotaso:
> is that contour.yaml still needed, or can users install the upstream Contour at this point? (My recollection was that it installed two Contour installations, one for external addresses and one for internal services.)

Yeah we still need both for external and internal routes.

iRaindrop:
This content will now be under Configuring Knative->Networking Options->Plugin: Contour

docs/versioned/install/installing-contour.md
evankanderson: See the discussion from Dave -- it sounds like the Contour installation either needs to the Knative-published `contour.yaml`, or two installations using different ingress class names.

docs/versioned/install/installing-contour.md
iRaindrop: Thanks. I left the "step 1" in regard to Dave's feedback below - and I'm not sure whether these instructions are complete in regard to that.

docs/versioned/.nav.yml
```diff
@@ -291,7 +291,7 @@ nav:
           - Configure Kafka Broker features: eventing/brokers/broker-types/kafka-broker/configuring-kafka-features.md
           - Configure event source defaults: eventing/configuration/sources-configuration.md
           - Configure Sugar Controller: eventing/configuration/sugar-configuration.md
-          - Configure KEDA Autoscaling of Knative Kafka Resources: eventing/configuration/keda-configuration.md
+          - Configure KEDA Autoscaling of Knative Kafka Resources: eventing/configuration/
```
dprotaso, line 294: this change seems unrelated

docs/versioned/.nav.yml
```diff
@@ -273,6 +272,7 @@ nav:
           - Configure domain names: serving/using-a-custom-domain.md
           - Istio Authorization: serving/istio-authorization.md
           - Extending Queue Proxy image with QPOptions: serving/queue-extensions.md
+          - Configure Contour adapter: ./install/installing-contour.md
```
dprotaso, line 275: I'd move this under `Installing plugins > Install Istio for Knative`

dprotaso: We also have existing content here: https://knative.dev/docs/install/yaml-install/serving/install-serving-with-yaml/#__tabbed_1_3

docs/versioned/install/installing-contour.md
```diff
@@ -0,0 +1,52 @@
...
+function: how-to
+---
+
+# Configure Contour adapter
+
+This page describes how to install and configure the Contour adapter for Knative: `net-contour`.
+
+## Before you begin
+
+This installation requires a Kubernetes cluster with [Contour](https://projectcontour.io/docs/) installed.
```
dprotaso, line 14: We prescribe how to install contour

contour.yaml is here: https://github.com/knative-extensions/net-contour/releases/tag/knative-v1.20.0

You can see what we do to the upstream yaml here: https://github.com/knative-extensions/net-contour/blob/6bbd6018a8ca5d1f774acae0af1deb6f734ba255/hack/update-deps.sh#L41

The main reason why we have our own 'installation' vs recommending upstream is in order to support private knative services

https://knative.dev/docs/serving/services/private-services/#making-individual-services-private

Without a separate deployment of contour we end up exposing internal routes to the public

docs/versioned/.nav.yml
iRaindrop: Seems like the change is gone now

PR #6459, opened 2025-10-19
Plugin: Gateway API

iRaindrop:
Need topic to install Gateway-API plugin.

Content derived from https://gateway-api.sigs.k8s.io/guides/

## Proposed Changes 

- Create Installing Gateway API to include in the Plugin section for Installing Knative.

### DIFF
--- a/docs/versioned/install/installing-gateway-api.md
+++ b/docs/versioned/install/installing-gateway-api.md
@@ -7,7 +7,7 @@ function: how-to
 
 # Installing Gateway API for Knative
 
-This guide shows how to install the Kubernetes Gateway API. You can install either the standard or experimental Gateway API channel.
+This page shows how to install the Kubernetes Gateway API. You can install either the standard or experimental Gateway API channel.
 
 ## Before you begin
 


docs/versioned/install/installing-gateway-api.md
```diff
@@ -0,0 +1,64 @@
+---
+audience: administrator
+components:
+  - serving
+function: how-to
+---
+
+# Installing Gateway API for Knative
+
+This page shows how to install the Kubernetes Gateway API. You can install either the standard or experimental Gateway API channel.
```
evankanderson, line 10: This should be documenting installing the Knative networking adapter for Gateway API.

I don't know if a picture is helpful, but here's some mermaid that might clarify:

```
flowchart TB
 subgraph s1["Gateway API"]
    HR[HTTPRoute]
    GW[Gateway listener]
  end
  KSvc["Knative Service"] -- owns --> Route
  Route -- creates --> HR
  Route -- creates --> GW
```

For this, we're talking about the component that implements the "creates"

docs/versioned/install/installing-gateway-api.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Before you begin
+
+This installation requires the following prerequisites:
+
+- A Kubernetes cluster with the Knative Serving component installed.
+- A Gateway Controller. You can install a Gateway Controller by using one of the following guides:
```
evankanderson, line 17: We should link to the Gateway API top-level guide, rather than duplicating all of the sub-guides.  I think https://gateway-api.sigs.k8s.io/implementations/ is probably the best page for linking to the options (yes, there are more than a dozen GA implementations).

docs/versioned/install/installing-gateway-api.md
```diff
@@ -0,0 +1,64 @@
...
+
+    - [Simple Gateway](https://gateway-api.sigs.k8s.io/guides/simple-gateway/) (recommended)
+    - [HTTP routing](https://gateway-api.sigs.k8s.io/guides/http-routing/)
+    - [HTTP redirects and rewrites](https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/)
+    - [HTTP traffic splitting](https://gateway-api.sigs.k8s.io/guides/traffic-splitting/)
+    - [Routing across Namespaces](https://gateway-api.sigs.k8s.io/guides/multiple-ns/)
+    - [Configuring TLS](https://gateway-api.sigs.k8s.io/guides/tls/)
+    - [TCP routing](https://gateway-api.sigs.k8s.io/guides/tcp/)
```
evankanderson, line 25: Some of these guides are for unrelated topics (TCP routing, for instance, is not relevant to Knative's scaling model).

docs/versioned/install/installing-gateway-api.md
```diff
@@ -0,0 +1,64 @@
...
+## Uninstall the Gateway API
+
+To uninstall the Gateway API, replace `kubectl apply --server-side -f` with `kubectl delete --server-side -f` in the commands above.
+
+Do not uninstall if the resources are in use or if they were installed by a Gateway controller. This command uninstalls the Gateway API resources for the entire cluster.
+
+## See also
+
+- [Kubernetes Gateway API](https://kubernetes.io/docs/concepts/services-networking/gateway/)
+- [Getting started with Gateway API](https://gateway-api.sigs.k8s.io/guides/)
```
evankanderson, line 64: These should cover installing https://github.com/knative-extensions/net-gateway-api

This should probably use a snippet like:

```markdown
    1. Install the Knative Gateway API controller by running the command:
      ```bash
      kubectl apply -f {{ artifact(repo="net-gateway-api",org="knative-extensions",file="net-gateawy-api.yaml")}}
      ```

    1. Configure Knative Serving to use Gateway API by default by running the command:
      ```bash
      kubectl patch configmap/config-network \
        --namespace knative-serving \
        --type merge \
        --patch '{"data":{"ingress-class":"gateway-api.ingress.networking.knative.dev"}}'
```

And then using `kubectl get gateway --all-namespaces` to find the IP address of the gateway.  This configuration will also need to be incorporated into the `config-gateway` ConfigMap to find the external and within-cluster (local) gateways.  I'm not sure whether that's well-documented anywhere, but it's the "istio-system" part of the "gateway" sub-key of the gateway class.

It's probably worth calling out that Istio and Contour are the currently-tested implementations, but that other Gateway API implementations should work.  https://github.com/knative-extensions/net-gateway-api/blob/main/docs/test-version.md has some additional information that it may be worth incorporating or linking.

Configuration for the installation is here:

https://github.com/knative-extensions/net-gateway-api/blob/main/config/config-gateway.yaml

PR #6461, opened 2025-10-21
add document for secure-pod-defaults feature and settings

nader-ziada:
## Proposed Changes 

- add a new page for secure-pod-defaults based on changes in PR: https://github.com/knative/serving/pull/16042 

/cc @dprotaso

### DIFF
--- a/docs/versioned/.nav.yml
+++ b/docs/versioned/.nav.yml
@@ -106,7 +106,6 @@ nav:
         - Installing Security-Guard: serving/app-security/security-guard-install.md
         - Security-guard quickstart: serving/app-security/security-guard-quickstart.md
         - Security-Guard example alerts: serving/app-security/security-guard-example-alerts.md
-        - SecurePodDefaults: serving/app-security/secure-pod-defaults.md
       # Serving - observability
       - Observability:
         - Collecting metrics: serving/observability/metrics/collecting-metrics.md
@@ -279,6 +278,7 @@ nav:
           - Configure gradual rollout of traffic to Revisions: serving/configuration/rolling-out-latest-revision-configmap.md
           - Config Revision Garbage Collection: serving/revisions/revision-admin-config-options.md
           - Configure the Defaults ConfigMap: serving/configuration/config-defaults.md
+          - Secure Pod Defaults: serving/app-security/secure-pod-defaults.md
           - Serving encryption configuration:
             - Overview: serving/encryption/encryption-overview.md
             - Configure cert-manager integration: serving/encryption/configure-certmanager-integration.md
--- a/docs/versioned/serving/app-security/secure-pod-defaults.md
+++ b/docs/versioned/serving/app-security/secure-pod-defaults.md
@@ -5,40 +5,37 @@ components:
 function: how-to
 ---
 
+!!! important
+    The default setting of `diabled` will be changed in the upcoming release of knative 1.21 to be more secure.
 
-## SecurePodDefaults
-
-Knative Serving provides a `secure-pod-defaults` configuration option that gives operators granular control over pod security defaults, enabling progressive adoption of more secure pod configurations while maintaining backward compatibility with existing workloads. This feature offers three security levels—`disabled`, `root-allowed`, and `enabled` allowing organizations to gradually adopt security best practices without breaking container images that require specific permissions. The default is `disabled` to ensure existing deployments continue to work without modification.
+Knative Serving provides a `secure-pod-defaults` configuration option that  allows the default Service configuration to run in the Kubernetes [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) Pod Security Standard profile without requiring application developers to explicitly set security properties. This feature offers three security levels: `disabled`, `root-allowed`, and `enabled` allowing organizations to gradually adopt security best practices without breaking container images that require specific permissions. The default is `disabled` to ensure existing deployments continue to work without modification.
 
 
-### Security Levels
+## Security Levels
 
 | Level | Description | Use Case |
 |-------|-------------|----------|
 | **disabled** | No security defaults applied | Legacy workloads, maximum compatibility |
-| **root-allowed** | Applies security defaults (blocks privilege escalation, runtime-default seccomp, drops capabilities) but allows root containers if not already set | Transition period, balanced security |
-| **enabled** | Applies all `root-allowed` defaults plus enforces non-root execution if not already set | Maximum security for production |
-
+| <span style="white-space:nowrap;">**root-allowed**</span> | Implements most of the Kubernetes [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) Pod Security Standard profile requirements, with the exception of `runAsNonRoot`, allowing containers to run as root when needed. | Transition period, balanced security |
+| **enabled** | Aligns with the Kubernetes [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted) Pod Security Standard in addition to enforcing non-root execution if not already set | Maximum security for production |
 
 ## Key Features
 
-### 1. **Backward Compatible Default**
-- Default remains `disabled` to prevent breaking existing deployments
-- Only affects newly created Revisions (admission webhook behavior)
-
-### 2. **Progressive Security Hardening**
+### **Progressive Security Hardening**
 When `root-allowed` is configured:
-- Sets `allowPrivilegeEscalation` to `false` if not specified
-- Sets `seccompProfile` to `RuntimeDefault` if not specified
-- Drops all capabilities if not specified
-- Conditionally adds `NET_BIND_SERVICE` capability if a container port below 1024 is detected and capabilities are not already configured
+security settings only apply if the field is not set -- if it is explicitly set to any value, it's assumed to be intentional, and not modified.
+- Sets `allowPrivilegeEscalation` to `false`
+- Sets `seccompProfile` to `RuntimeDefault`, see [Seccomp and Kubernetes](https://kubernetes.io/docs/reference/node/seccomp/) for more details
+- Drops all capabilities
+  - Conditionally adds `NET_BIND_SERVICE` capability if a container port below 1024 is detected and capabilities are not already configured
 - **Does NOT** enforce `runAsNonRoot` (allows root containers)
 
 When `enabled` is configured:
+
 - All of the above, PLUS
 - Sets `runAsNonRoot` to `true` if not already specified
 
-### 3. **Respects User Intent**
+### **Respects User Intent**
 - Only applies defaults when values are not explicitly set by users
 - Never overrides user-specified security contexts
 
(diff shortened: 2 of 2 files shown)


docs/versioned/.nav.yml
```diff
@@ -106,6 +106,7 @@ nav:
         - Installing Security-Guard: serving/app-security/security-guard-install.md
         - Security-guard quickstart: serving/app-security/security-guard-quickstart.md
         - Security-Guard example alerts: serving/app-security/security-guard-example-alerts.md
+        - SecurePodDefaults: serving/app-security/secure-pod-defaults.md
```
dprotaso:
```suggestion
        - Secure Pod Defaults: serving/app-security/secure-pod-defaults.md
```

evankanderson: I would move this to "Administration / Configuring Knative / Serving Configuration, around line 280.

Sorry for leaving this section here to confuse -- I think it may be removed altogether soon.

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+audience: administrator
+components:
+  - serving
+function: how-to
+---
+
+
+## SecurePodDefaults
+
+Knative Serving provides a `secure-pod-defaults` configuration option that gives operators granular control over pod security defaults, enabling progressive adoption of more secure pod configurations while maintaining backward compatibility with existing workloads. This feature offers three security levels—`disabled`, `root-allowed`, and `enabled` allowing organizations to gradually adopt security best practices without breaking container images that require specific permissions. The default is `disabled` to ensure existing deployments continue to work without modification.
```
evankanderson:
```suggestion
Knative Serving provides a `secure-pod-defaults` configuration option that gives operators granular control over pod security defaults, enabling progressive adoption of more secure pod configurations without needing to update all application definitions. This feature offers three security levels: `disabled`, `root-allowed`, and `enabled` allowing organizations to gradually adopt security best practices without breaking container images that require specific permissions. The default is `disabled` to ensure existing deployments continue to work without modification.
```

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+### Security Levels
+
+| Level | Description | Use Case |
+|-------|-------------|----------|
+| **disabled** | No security defaults applied | Legacy workloads, maximum compatibility |
+| **root-allowed** | Applies security defaults (blocks privilege escalation, runtime-default seccomp, drops capabilities) but allows root containers if not already set | Transition period, balanced security |
```
evankanderson: It seems like this should somewhere reference https://kubernetes.io/docs/concepts/security/pod-security-standards/.  In particular, **enabled** causes the default configuration of Knative Services to run in the "restricted" profile, and **root-allowed** implements _most_ of the "restricted" profile except for `runAsNonRoot`.

docs/versioned/serving/app-security/secure-pod-defaults.md
dprotaso: Can we mention it's part of the `config-defaults` config map and link out to this page. We should mention it's an administrator option.

https://deploy-preview-6461--knative.netlify.app/docs/serving/configuration/config-defaults/

I also think that page should have a 'short' entry that links out to this page.

evankanderson: I'd be careful about saying "granular control".  I'd rather link to Pod Security Standards, and say that it allows the default Service configuration to run in the "restricted" profile without requiring application developers to explicitly set security properties.  Another possible phrasing: **enabled** will default to the values permitted by the "restricted" profile, and **root-allowed** will default most values to the "restricted" profile while allowing several common application containers to run.

evankanderson (commented):
/assign

Thanks for writing this up!  A couple comments, but it's good to have this documented more clearly.

docs/versioned/serving/app-security/secure-pod-defaults.md
dprotaso: nit: is there a way to disable word wrapping? 

Screenshot 2025-10-21 at 4 28 00 PM

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+| **enabled** | Applies all `root-allowed` defaults plus enforces non-root execution if not already set | Maximum security for production |
+
+
+## Key Features
+
+### 1. **Backward Compatible Default**
```
dprotaso: Drop the number in the heading they don't render properly

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
+---
+audience: administrator
+components:
+  - serving
+function: how-to
+---
+
+
+## SecurePodDefaults
```
dprotaso: The title is already rendered at the top of the page so we don't need a duplicate

```suggestion

```

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+
+Knative Serving provides a `secure-pod-defaults` configuration option that gives operators granular control over pod security defaults, enabling progressive adoption of more secure pod configurations while maintaining backward compatibility with existing workloads. This feature offers three security levels—`disabled`, `root-allowed`, and `enabled` allowing organizations to gradually adopt security best practices without breaking container images that require specific permissions. The default is `disabled` to ensure existing deployments continue to work without modification.
+
+
+### Security Levels
```
dprotaso: Align headers format

```suggestion
## Security Levels
```

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Key Features
+
+### 1. **Backward Compatible Default**
+- Default remains `disabled` to prevent breaking existing deployments
+- Only affects newly created Revisions (admission webhook behavior)
+
+### 2. **Progressive Security Hardening**
+When `root-allowed` is configured:
+- Sets `allowPrivilegeEscalation` to `false` if not specified
```
dprotaso: The markdown I think needs an extra newline after `configured:`

It's not rendering the list properly

Screenshot 2025-10-21 at 4 31 03 PM

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+- Sets `seccompProfile` to `RuntimeDefault` if not specified
```
dprotaso: link to k8s docs for definition of `RuntimeDefault`

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+|-------|-------------|----------|
+| **disabled** | No security defaults applied | Legacy workloads, maximum compatibility |
+| **root-allowed** | Applies security defaults (blocks privilege escalation, runtime-default seccomp, drops capabilities) but allows root containers if not already set | Transition period, balanced security |
+| **enabled** | Applies all `root-allowed` defaults plus enforces non-root execution if not already set | Maximum security for production |
+
+
+## Key Features
+
+### 1. **Backward Compatible Default**
+- Default remains `disabled` to prevent breaking existing deployments
```
dprotaso: I think we should drop this key feature because we want to change the default in 1.21

docs/versioned/serving/app-security/secure-pod-defaults.md
dprotaso: Can we add a `!!! notice/important/warning` banner on this page that the current default will be changed in 1.21 and that users should explicitly disable this setting to retain the existing behaviour

Then we can link this section in the 1.20 release notes

dprotaso (commented):
Looks good - some minor stuff

docs/versioned/serving/app-security/secure-pod-defaults.md
dprotaso: I think in this paragraph we should mention and link out to the Pod Security Standards - https://kubernetes.io/docs/concepts/security/pod-security-standards/

docs/versioned/serving/app-security/secure-pod-defaults.md
evankanderson: Ooh, good point!

Yes, I think a warning banner would be a good call given that we know we'll be changing this.

dprotaso:
Also please see this issue comment: https://github.com/knative/serving/issues/16137#issuecomment-3429516437

I realize we might have an opportunity to simplify this a bit in 1.21

docs/versioned/.nav.yml
dprotaso: I suggest to Nader this section because I think it's relevant and we can have documentation in two places.

I'm going to make a separate PR that collapses security guard into a single page. Thus the section becomes about 'Application Security' instead of just exclusive to security guard.

(edit) done: https://github.com/knative/docs/pull/6464

dprotaso: As can we move `Secure Pod Defaults` to the top of the section

dprotaso: @evankanderson I'll defer to you what content you feel belongs in one spot vs the other

docs/versioned/serving/app-security/secure-pod-defaults.md
evankanderson: https://stackoverflow.com/a/30311040/1543502 suggests one solution with a non-breaking hyphen, but the `<span>` with a no-break class and supporting CSS seems more palatable.

docs/versioned/serving/app-security/secure-pod-defaults.md
evankanderson: Actually, I'm wondering if a little bit of philosophy would be useful here. (Defaults should be safe, but k8s defaults aren't always / values should be safe if but specified, but don't block explicit reduction of security stance.)

We may also want to point to k8s Pod Security Admission and/or policy frameworks like OPA gatekeeper and Kyverno for "enforce secure settings" (as opposed to "default to safe").  An additional benefit of these other frameworks is that they cover all types of pods, while secure-pod-defaults only covers Knative pods that haven't thought about security settings.

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+- Only affects newly created Revisions (admission webhook behavior)
+
+### 2. **Progressive Security Hardening**
+When `root-allowed` is configured:
```
evankanderson, line 25: Rather than saying "if not already specified" at the end of each item, how about saying once that security settings only apply if the field is not set -- if it is explicitly set to *any* value, it's assumed to be intentional, and not modified.

docs/versioned/serving/app-security/secure-pod-defaults.md
```diff
@@ -0,0 +1,89 @@
...
+- Sets `allowPrivilegeEscalation` to `false` if not specified
+- Sets `seccompProfile` to `RuntimeDefault` if not specified
+- Drops all capabilities if not specified
+- Conditionally adds `NET_BIND_SERVICE` capability if a container port below 1024 is detected and capabilities are not already configured
```
evankanderson: This feels like a sub-bullet of dropping all capabilities -- NET_BIND is "safe" and needed to listen on low-numbered ports in Linux.

nader-ziada:
@evankanderson @dprotaso I think I covered all of the comments in the discussion, please take another look 

note: I moved the page link in the nav.yaml, but didn't move the actual file location because I wasn't sure if github would be able to show the diff and comments, I can move it once you both think its ready.

evankanderson:
I think we'll need a big reorganization one day; in the meantime, the nav.yml magic picks up the first place the page appears in the nav bar

dprotaso:
thanks @nader-ziada

PR #6462, opened 2025-10-21
Remove instruction to remove the _example key from a ConfigMap

iRaindrop:
Although removing the _example key is not an issue per se, it shouldn't be a recommended practice.

## Proposed Changes 

- Removed clause to remove _example key in procedure

### DIFF
--- a/docs/versioned/serving/using-a-custom-domain.md
+++ b/docs/versioned/serving/using-a-custom-domain.md
@@ -31,6 +31,7 @@ You can change the default domain for all Knative Services on a cluster by modif
     apiVersion: v1
     data:
       knative.dev: ""
+      _example: | ...
     kind: ConfigMap
     [...]
     ```


iRaindrop:
Ready to merge

docs/versioned/serving/using-a-custom-domain.md
```diff
@@ -19,13 +19,13 @@ You can change the default domain for all Knative Services on a cluster by modif
...
 
-1. Open the `config-domain` ConfigMap in your default text editor:
+1. Open the `config-domain` ConfigMap in a text editor:
 
     ```bash
     kubectl edit configmap config-domain -n knative-serving
     ```
 
-1. Edit the file to replace `svc.cluster.local` with the domain you want to use, then remove the `_example` key and save your changes. In this example, `knative.dev` is configured as the domain for all routes:
+1. Edit the file to replace `svc.cluster.local` with the domain you want to use  and save your changes. In this example, `knative.dev` is configured as the domain for all routes:
```
evankanderson, line 28: You might want to call out that the `_example` key may be present, and not to edit the contents.  I would do this as:

```yaml
apiVersion: v1
    data:
      knative.dev: ""
      _example: | ...
    kind: ConfigMap
```

evankanderson (commented):
/approve

I think not mentioning the `_example` key may confuse people when they see it present.  Mentioning it but indicating that the domain name should be separate seems clearer to me.  (Though I could be convinced otherwise.)

docs/versioned/serving/using-a-custom-domain.md
iRaindrop: Done as suggested. I'll put a link in this topic to 'Editing ConfigMaps' when it gets checked in.

PR #6466, opened 2025-10-21
Create Editing ConfigMaps page

iRaindrop:
Need to alert users about not modifying the _example key, value propagation, versioning, and other tips about ConfigMaps.

## Proposed Changes 

- New topic in Administration: Editing ConfigMaps

### DIFF
--- a/docs/versioned/admin/admin-overview.md
+++ b/docs/versioned/admin/admin-overview.md
@@ -48,6 +48,8 @@ Knative configurations are performed by the following methods:
 
     Store and manage configuration data as key-value pairs. ConfigMaps are frequently used to tune platform-wide behavior. Most of the Knative ConfigMaps are in the `knative-serving` and `knative-eventing` namespaces. Their settings apply to all the relevant Knative components in all namespaces.
 
+    See [Working with ConfigMaps](editing-configmaps.md) for important usage information and best practices.
+
 - Using the Knative Operator
 
     Some platform-wide settings can be managed declaratively using the Knative Operator.


docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+  - eventing
+function: explanation
+---
+# Editing ConfigMaps
+
+This page provides information and best practices for editing ConfigMaps. While there is no technical difference in how Knative uses ConfigMaps in Kubernetes, you should be aware of the following points and tips while working with ConfigMaps:
+
+- The _example key. Don't modify.
+- Value propagation. Know when changes to default settings take affect.
+- Version control. Use ConfigMaps for version control.
```
evankanderson: I don't understand what this means / what is supposed to be done here.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+The most common way to use ConfigMaps is to configure settings for containers running in a Pod, or in multiple pods, that use a specified namespace. There are a number of other ways to use ConfigMaps as described in [Kubernetes ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/) documentation.
```
evankanderson: This seems misleading, as this is not how the Knative controllers use ConfigMaps.  Developers can still use ConfigMaps for these purposes, but that is a separate topic in a different area of the documentation (if needed).

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
+---
+audience: administrator
+components:
+  - serving
+  - eventing
+function: explanation
+---
+# Editing ConfigMaps
+
+This page provides information and best practices for editing ConfigMaps. While there is no technical difference in how Knative uses ConfigMaps in Kubernetes, you should be aware of the following points and tips while working with ConfigMaps:
```
evankanderson: "while there is no technical difference" is confusing, as there is no alternative that is being compared to.

I also feel like the opening sentence needs slightly more orientation.  Maybe something like:

> Knative usually uses ConfigMaps to manage most system-level configuration, including default values, minimum and maximum values, and names and addresses of connecting services.  Because Knative is implemented as a set of controllers, Knative watches the ConfigMaps and should update behavior live shortly after the ConfigMap is updated.  

In general, I'm not a big fan of the LLM "listicle" approach of providing a bunch of short bulleted lists; there are times when it's good to enumerate options, but this doesn't feel like a useful enumeration, it feels like a way to try to unify disparate topics that could be served just as well by actual sentences.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+- The _example key. Don't modify.
+- Value propagation. Know when changes to default settings take affect.
+- Version control. Use ConfigMaps for version control.
+
+The most common way to use ConfigMaps is to configure settings for containers running in a Pod, or in multiple pods, that use a specified namespace. There are a number of other ways to use ConfigMaps as described in [Kubernetes ConfigMaps](https://kubernetes.io/docs/concepts/configuration/configmap/) documentation.
+
+## The _example key
+
+ConfigMap files installed by Knative contain an `_example` key that shows the usage and purpose of a configuration key. The key itself is not a key, but a long comment.
```
evankanderson: "The key itself is not a key" doesn't make sense.  Maybe:

> This key does not affect Knative behavior, but contains a value which acts as a documentation comment.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+In case a user edits the `_example` key by mistakenly thinking their edits would have an affect, the administrator needs to be alerted. The Knative webhook server determines if the `_example key` has been altered. The edit is caught when the value of the checksum for the `_example` key is different when compared with the pod's template annotations. If the checksum is null however, it does not create the warning.
```
evankanderson: This phrasing is slightly confusing.  If you have a quickstart cluster, it might be informative to simply try editing one of the configmaps with something like `kubectl edit --namespace knative-serving config-default`, and see what message you get when you edit the `_example`.  That message has been fairly stable, and including it in the documentation would help for people searching for that error string.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+Accordingly, you cannot alter the contents of the `_example` key, but you can delete the `_example` key altogether or delete the annotation:
+
+```yml
+...
+annotations:
+    knative.dev/example-checksum: "47c2487f"
+```
+
+To accurately create a ConfigMap using an existing file, use the Kubernetes[Define the key to use when creating a configmap from a file](https://kubernetes.io/dos/tasks/configure-pod-container/configure-pod-configmap/#define-the-key-to-use-when-creating-a-configmap-from-a-file)
```
evankanderson: This sentence doesn't read correctly.  I'm also not sure what it's trying to say -- why does this matter?

evankanderson: Also, I'm not sure what work "accurately" is doing here -- what would it mean to "inaccurately" create a ConfigMap?

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Value propagation
+
+The ConfigMap change is immediate in the Kubernetes API, but its effect on Pods or applications depends on how the ConfigMap is consumed (volume, environment variables, or API) and whether the application or Pod is designed to pick up the change dynamically or requires a restart. For volumes, updates propagate within seconds, but for environment variables, a Pod restart is needed.
```
evankanderson: None of this applies to the Knative controllers, which use the Kubernetes Watch API to be informed of any changes to the ConfigMap roughly as soon as they happen.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+When a ConfigMap currently consumed in a volume is updated, projected keys are eventually updated as well. The `kubelet` checks whether the mounted ConfigMap is fresh on every periodic sync.
+
+Kubernetes does not automatically restart Pods when a ConfigMap changes, even if the ConfigMap is referenced. You may need to trigger a rolling update (e.g., via `kubectl rollout restart`) to ensure the changes are applied to running Pods.
```
evankanderson: We generally don't mount the ConfigMaps or feed them into environment variables, so these should not matter.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+A ConfigMap can be either propagated by watch (default), ttl-based, or by redirecting all requests directly to the API server. Because the `kubelet` uses its local cache for getting the current value of the ConfigMap, there is a delay from the moment when the ConfigMap is updated to the moment when new keys are projected to the Pod. The total delay can be as long as the `kubelet` sync period added to the cache propagation delay. The type of the cache is configurable using the `configMapAndSecretChangeDetectionStrategy` field in the [KubeletConfiguration struct](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/).
```
evankanderson: Again, we don't rely on the kubelet for detecting ConfigMap changes, so this doesn't apply.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+- Environment Variables - If the ConfigMap is used to set environment variables in a Pod, changes to the ConfigMap do not automatically propagate to the Pod. Pods must be restarted (e.g., by deleting and recreating them) for the updated ConfigMap values to take effect, as environment variables are set at container startup.
+- Direct API Access - If an application directly queries the ConfigMap via the Kubernetes API, it will see the updated values immediately after the change is applied.
+- Special Cases - Some applications or operators (e.g., those using tools like `reloader`) can watch for ConfigMap changes and automatically trigger Pod restarts or reloads.
+- If the ConfigMap is used by a controller (e.g., a Deployment), changes might not affect running Pods unless the controller reconciles the change, which depends on its implementation.
```
evankanderson: This seems like it is just rehashing Kubernetes documentation, without any specifics for Knative.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+        db.port=5432
+    ```
+
+- Commit these files to a Git repository to track changes over time.
+
+### Versioning ConfigMaps in Git
+
+- Use meaningful commit messages to describe changes to ConfigMap data (e.g., "Updated app.properties to version 1.2.4").
+- Tag commits in Git with version numbers (e.g., `git tag config-v1.2.3`) to mark specific ConfigMap versions.
+- Use branches for different environments (e.g., `dev`, `staging`, `prod`) or feature-specific ConfigMap changes.
```
evankanderson: These are general GitOps principles, and it's not clear that we need to repeat them here.  Either you've already figured this out, or Knative isn't going to suddenly change your mind.  (IMO)

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+    ```
+
+- Alternatively, use annotations in the ConfigMap’s metadata:
+
+    ```yaml
+    metadata:
+      name: my-app-config
+      annotations:
+        app-version: "1.2.3"
+    ```
```
evankanderson: I'm not sure what this is suggesting -- it records a version number, but it doesn't help me find previous versions, or compare changes between versions.  I'd stick with GitOps as the single recommended approach.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Use GitOps for Deployment
+
+- Implement a GitOps workflow with tools like ArgoCD or Flux to synchronize ConfigMaps from Git to your Kubernetes cluster.
+- These tools detect changes in the Git repository and automatically apply them to the cluster, ensuring the deployed ConfigMap matches the versioned configuration in Git.
```
evankanderson: Once the ConfigMaps are checked in to Git, there are three options for applying them:

1. Use a tool like ArgoCD or Flux to apply them.  This makes sense if you've got one of them set up already (and is generally considered a best practice).
2. Update the Git source of truth, then manually apply the ConfigMap with `kubectl apply --file $filename`.
3. Apply the new configuration with `kubectl apply --file $filename`, and then commit the change afterward if the results are as expected.

I'd probably use two headings for this -- "Automatic Application Via Continuous Delivery" and "Manual Application"

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+      template:
+        spec:
+          containers:
+          - name: my-app
+            envFrom:
+            - configMapRef:
+                name: my-app-config-v1.2.3
+    ```
+
+- This approach ensures ConfigMaps are not modified in-place, preserving historical versions.
```
evankanderson: This doesn't work for Knative, as the controllers are watching ConfigMaps with specific names.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Track Changes in Kubernetes
+
+- Use Kubernetes resource versioning (e.g., `resourceVersion` in metadata) to track changes to the ConfigMap object itself within the cluster. However, this is managed by Kubernetes and not suitable for application-level versioning.
+- Enable audit logging in Kubernetes to track who modified ConfigMaps and when.
```
evankanderson: I don't think this suggestion makes sense -- it's both advanced and lossy, so it's more likely to cause harm than prove useful.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Use Helm for Versioned Configurations
+
+- If using Helm, store ConfigMaps in Helm chart templates or values files and version the chart itself (e.g., `Chart.yaml` with `version: 1.2.3`).
+- Helm’s release history can track which ConfigMap versions were deployed.
```
evankanderson: Helm can be used this way, but neither Serving nor Eventing currently publishes a helm chart, so this is probably not helpful.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Automate Version Updates
+
+- Use CI/CD pipelines (e.g., GitHub Actions, Jenkins) to automate updates to the version field in ConfigMaps when the application version changes.
+- Example: A pipeline updates `app.properties` with `version=1.2.4` and commits the change to Git.
```
evankanderson: As mentioned above, the version string (or versioned resource names) doesn't make much sense for the pattern that Knative is using.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Validate and Test Changes
+
+- Before applying ConfigMaps, validate their syntax and content using tools like `kubeval` or `kubectl apply --dry-run=client`.
+- Test ConfigMap changes in a staging environment to ensure compatibility with the application version.
```
evankanderson: These are general best-practices, but I think the more useful thing to know is that Knative will do some validation when the ConfigMap is updated.  (Verify this in a `quickstart` cluster, and you might include the error messages provided.)

Note that ConfigMaps in general don't support this type of validation, but Knative has added validating admission webhooks to perform these checks before the configuration is submitted to the cluster.

docs/versioned/admin/editing-configmaps.md
evankanderson, line 1: This topic covers editing the ConfigMaps, which is great, but it doesn't explain how the operator interacts with ConfigMaps.  IIRC, the operator will write and update ConfigMaps, so if you're using the operator, you should update the KnativeServing and KnativeEventing custom resources instead of directly editing ConfigMap values.

docs/versioned/admin/editing-configmaps.md
iRaindrop: Removing an unneeded list of sections

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,138 @@
...
+
+Accordingly, you cannot alter the contents of the `_example` key, but you can delete the `_example` key altogether or delete the annotation:
+
+```yml
+...
+annotations:
+    knative.dev/example-checksum: "47c2487f"
+```
+
+To accurately create a ConfigMap using an existing file, use the Kubernetes[Define the key to use when creating a configmap from a file](https://kubernetes.io/dos/tasks/configure-pod-container/configure-pod-configmap/#define-the-key-to-use-when-creating-a-configmap-from-a-file)
```
evankanderson: This sentence doesn't scan.  I'm not sure what you're trying to link to or reference here.

docs/versioned/admin/editing-configmaps.md
evankanderson: I'm not sure if any of this "value propagation" information is necessary given how Knative watches ConfigMaps directly.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,138 @@
...
+- Environment Variables - If the ConfigMap is used to set environment variables in a Pod, changes to the ConfigMap do not automatically propagate to the Pod. Pods must be restarted (e.g., by deleting and recreating them) for the updated ConfigMap values to take effect, as environment variables are set at container startup.
+- Direct API Access - If an application directly queries the ConfigMap via the Kubernetes API, it will see the updated values immediately after the change is applied.
+- Special Cases - Some applications or operators (e.g., those using tools like `reloader`) can watch for ConfigMap changes and automatically trigger Pod restarts or reloads.
+- If the ConfigMap is used by a controller (e.g., a Deployment), changes might not affect running Pods unless the controller reconciles the change, which depends on its implementation.
```
evankanderson: This is worth calling out -- changes to default values (but not maximums or minimums) should only take effect at the next deployment of a Service.  (I'm not sure whether Brokers follow this pattern as well -- I suspect it only affects the version-controlled resources like Revisions.)

docs/versioned/admin/editing-configmaps.md
evankanderson: (this still seems to be an issue)

docs/versioned/admin/editing-configmaps.md
iRaindrop: Not applicable anymore - my thinking was wrong

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,176 @@
+---
+audience: administrator
+components:
+  - serving
+  - eventing
+function: explanation
+---
+# Editing ConfigMaps
+
+This page provides information and best practices for editing ConfigMaps. Knative uses ConfigMaps to manage most system-level configuration, including default values, minimum and maximum values, and names and addresses of connecting services. Because Knative is implemented as a set of controllers, Knative watches the ConfigMaps and updates behavior live shortly after the ConfigMap is updated.
```
evankanderson: IMO, we should add a line here explaining that if you use the Knative Operator, the fields in the Operator resources (KnativeServing and KnativeEventing) are used to fill out the ConfigMap, and you shouldn't edit the ConfigMaps directly -- the operator will overwrite your changes.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Knative Operator and YAML installations
+
+To be written.
+
+### Brokers
+
+When creating a Broker class, you can define default values and settings on different levels of scope. You can specify a ConfigMap that defines the implementation of the Broker, and apply it in different ways.
+
+Knative uses `MTChannelBasedBroker` as the default class for creating Brokers. For scopes at the namespace level, you can specify which Broker class to use for a particular namespace.
```
evankanderson: I'm not sure I understand the second sentence here -- where are you picking that content up from?

docs/versioned/admin/editing-configmaps.md
```diff
@@ -25,33 +26,31 @@ annotations:
...
 - If the ConfigMap is used by a controller (e.g., a Deployment), changes might not affect running Pods unless the controller reconciles the change, which depends on its implementation.
 
-## Version control
+## Best practices
+
+This section provides recommended procedures for storage, monitoring, and version control of ConfigMaps.
+
+### Validate and Test Changes
+
+- Before applying ConfigMaps, validate their syntax and content using tools like `kubeval` or `kubectl apply --dry-run=client`.
```
evankanderson: The `--dry-run=server` option should run the webhook validation, and is probably better than `--dry-run=client`.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -67,40 +66,18 @@ Define ConfigMaps in YAML or JSON files and store them in a version control syst
...
-
-### Use GitOps for Deployment
-
 - Implement a GitOps workflow with tools like ArgoCD or Flux to synchronize ConfigMaps from Git to your Kubernetes cluster.
 - These tools detect changes in the Git repository and automatically apply them to the cluster, ensuring the deployed ConfigMap matches the versioned configuration in Git.
 
-### Immutable ConfigMaps with Versioned Names
+### Immutable ConfigMaps
+
+If you have services processing information that must not change, such as for finances and payment processing, consider using immutable ConfigMaps. This approach ensures ConfigMaps are not modified in-place, preserving historical versions.
```
evankanderson: Immutable ConfigMaps don't work for the Knative configuration, because the Knative controllers look up specific ConfigMap names, and immutable ConfigMaps need a new name each time.

docs/versioned/admin/editing-configmaps.md
iRaindrop: It was from aggregated from eventing/brokers/broker-developer-config-options.md and eventing/brokers/broker-types/channel-based-broker/README.md - but won't be using

docs/versioned/admin/editing-configmaps.md
iRaindrop: Got the Quickstart going and made a change to `config-defaults` - here's the error text:

```
```bash
error: configmaps "config-defaults" could not be patched: admission webhook "config.webhook.serving.knative.dev" denied the request: validation failed: the update modifies a key in "_example" which is probably not what you want. Instead, copy the respective setting to the top-level of the ConfigMap, directly below "data"
You can run `kubectl replace -f /var/folders/9t/yzwp6zrx765clbvl1c_dqc2r0000gn/T/kubectl-edit-2068368769.yaml` to try this update again.
```

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,83 @@
...
+audience: administrator
+components:
+  - serving
+  - eventing
+function: explanation
+---
+
+# Using ConfigMaps
+
+This page provides important information and best practices for working with Kubernetes ConfigMaps. ConfigMaps and YAML resource files are the primary means for storing configuration values in Knative.
```
evankanderson:
```suggestion
This page provides important information and best practices for working with Kubernetes ConfigMaps. ConfigMaps and YAML resource files are the primary means for managing configuration values for Knative controllers.
```

I'm trying here to distinguish here between "driving the car" (developers) and "tuning the engine" (administrators).  In particular, I worry about pages getting pulled together in AI slop without sufficient context, and then mixing developer and operator concerns.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,83 @@
...
+
+## The _example key
+
+ConfigMap files installed by Knative contain an `_example` key that shows the usage and purpose of a configuration key. This key does not affect Knative behavior, but contains a value which acts as a documentation comment.
+
+If a user edits the `_example` key by mistakenly thinking their edits will have an affect, the Knative webhook server catches the error with the following alert text (using the `config-defaults` ConfigMap as a example):
```
evankanderson:
```suggestion
If a user edits the `_example` key by mistakenly thinking their edits will have an affect, the Knative webhook catches the error with the following alert text (using the `config-defaults` ConfigMap as a example):
```

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,83 @@
...
+
+```bash
+error: configmaps "config-defaults" could not be patched: admission webhook "config.webhook.serving.knative.dev" denied the request: validation failed: the update modifies a key in "_example" which is probably not what you want. Instead, copy the respective setting to the top-level of the ConfigMap, directly below "data"
+You can run `kubectl replace -f /var/folders/9t/yzwp6zrx765clbvl1c_dqc2r0000gn/T/kubectl-edit-2068368769.yaml` to try this update again.
+```
```
evankanderson: This output is valid, but given the code is in a `pre` or `code` block, we end up with a _very_ long horizontal line.

I suggest using a `<code>` tag explicitly here with a `style=` attribute:


```suggestion
<!-- use code here to wrap long command-line output -->
<code style="white-space: pre-wrap">
error: configmaps "config-defaults" could not be patched: admission webhook "config.webhook.serving.knative.dev" denied the request: validation failed: the update modifies a key in "_example" which is probably not what you want. Instead, copy the respective setting to the top-level of the ConfigMap, directly below "data"
You can run `kubectl replace -f /var/folders/9t/yzwp6zrx765clbvl1c_dqc2r0000gn/T/kubectl-edit-2068368769.yaml` to try this update again.
</code>
```

Though I'd check that the back-tick in the second line doesn't get escaped if you do that...

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,83 @@
...
+
+More specifically, the edit is caught when the value of the checksum for the `_example` key is different when compared with the pod's template annotations. If the checksum is null or missing, the webhook server does not create the warning.
+
+Accordingly, you cannot alter the contents of the `_example` key, but you can delete the `_example` key altogether or delete the annotation. 
+
+### Example
+
+The following example shows the abbreviated content of the `config-defaults` ConfigMap with most of the file removed except for the last four lines. The checksum is in the annotations as `Knative.dev/example-checksum: "5b64ff5c"`
```
evankanderson: I'm not sure what this example is supposed to illustrate.

Note that you might be able to use annotations to make this clearer: https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-annotations

You can also [highlight specific lines](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#highlighting-specific-lines)

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,83 @@
...
+    # See https://github.com/knative/serving/issues/8498.
+    enable-service-links: "false"
+```
+
+## Best practices
+
+### Validate and Test Changes
+
+- Before applying ConfigMaps, validate their syntax and content using tools like `kubeval` or `kubectl apply --dry-run=server`.
+- Test ConfigMap changes in a staging environment to ensure compatibility with the application version.
```
evankanderson, line 59: It's probably worth highlighting here that Knative controllers should pick up the new value (or a roll-back to the old value) within a few seconds of application.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Storage and versioning
+
+- Periodically export ConfigMaps from the cluster (`kubectl get configmap -o yaml`) and commit them to Git for recovery purposes. Include applicable version numbers in `app.properties` as needed.
```
evankanderson: I'd suggest that the first thing to decide is whether the Git or cluster is considered the _canonical_ location.  If the cluster is canonical, then you're exporting or backing up the configuration to Git.  If Git is canonical, then you're practicing GitOps, and you should make changes in Git and then apply the files from Git to your cluster.  For GitOps, you can either apply the changes manually, or use automation (for example Flux or ArgoCD) to apply the changes once they are committed.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,86 @@
...
+audience: administrator
+components:
+  - serving
+  - eventing
+function: explanation
+---
+
+# Working with ConfigMaps
+
+This page provides important information and best practices for working with Kubernetes ConfigMaps. ConfigMaps and YAML resource files are the primary means for managing configuration values for Knative controllers.
```
evankanderson: Add a sentence or two here about "if you are using the operator to install and manage your Knative cluster, see [operator documentation] for information on how to manage ConfigMaps using the operator".

I'd also remove the reference to "YAML resource files" in the second sentence -- YAML is a format for structured information which is used to manage all types of Kubernetes resources, and we specifically want to focus on the ConfigMap resources (a subset of all resources) on this page.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,86 @@
...
+
+## The _example key
+
+ConfigMap files installed by Knative contain an `_example` key that shows the usage and purpose of a configuration key. This key does not affect Knative behavior, but contains a value which acts as a documentation comment.
```
evankanderson:
```suggestion
ConfigMap files installed by Knative contain an `_example` key that shows the usage and purpose of all the known configuration keys in that ConfigMap. This key does not affect Knative behavior, but contains a value which acts as a documentation comment.
```

evankanderson: Typically, each ConfigMap has _many_ recognized configuration keys, but `_example` is a single string which holds documentation for all of them.

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,86 @@
...
+
+If a user edits the `_example` key by mistakenly thinking their edits will have an affect, the Knative webhook catches the error with the following alert text (using the `config-defaults` ConfigMap as a example):
```
evankanderson:
```suggestion
If a user edits the `_example` key by mistakenly thinking their edits will have no effect on the cluster operation. The Knative webhook catches the error with the following alert text (using the `config-defaults` ConfigMap as a example):
```

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,86 @@
...
+
+<!-- use code here to wrap long command-line output -->
+<code style="white-space: pre-wrap">
+error: configmaps "config-defaults" could not be patched: admission webhook "config.webhook.serving.knative.dev" denied the request: validation failed: the update modifies a key in "_example" which is probably not what you want. Instead, copy the respective setting to the top-level of the ConfigMap, directly below "data"
+You can run `kubectl replace -f /var/folders/9t/yzwp6zrx765clbvl1c_dqc2r0000gn/T/kubectl-edit-2068368769.yaml` to try this update again.
+</code>
+
+More specifically, the edit is caught when the value of the checksum for the `_example` key is different when compared with the pod's template annotations. If the checksum is null or missing, the webhook server does not create the warning.
```
evankanderson: There isn't a pod template here -- the annotation is part of the metadata of the ConfigMap.


```suggestion
More specifically, the edit is caught when the value of the checksum for the `_example` key is different when compared with the `knative.dev/example-checksum` annotation on the ConfigMap. If the checksum is null or missing, the webhook server does not create the warning.
```

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,86 @@
...
+    # to users that `kubectl edit` this config map.
+    #
+  . . . 
+    # In environments with large number of services it is suggested
+    # to set this value to `false`.
+    # See https://github.com/knative/serving/issues/8498.
+    enable-service-links: "false"
+```
+
+(1) Testing
```
evankanderson: Is this a stub of a section  or a leftover?

docs/versioned/admin/editing-configmaps.md
evankanderson: More concretely, I'd suggest prefixing this sentence with something like "If you manage the ConfigMap by using `kubectl edit`, periodically ..."

And then I'd add a second bullet point saying "If you manage the ConfigMap by keeping the definition in Git and automatically applying it to the cluster (GitOps), ..."

docs/versioned/admin/editing-configmaps.md
iRaindrop: A footnote annotation test. Removed now.

docs/versioned/admin/editing-configmaps.md
iRaindrop: Thx - temp fix but redo coming

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,75 @@
...
+audience: administrator
+components:
+  - serving
+  - eventing
+function: explanation
+---
+
+# Working with ConfigMaps
+
+This page provides important information and best practices for working with Kubernetes ConfigMaps. ConfigMaps are the primary means for managing configuration values for Knative controllers.
```
evankanderson:
```suggestion
This page provides important Knative-specific information and best practices for managing cluster configuration using ConfigMaps. ConfigMaps are the primary means for managing configuration values for Knative controllers.
```

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,75 @@
...
+
+If you are using the Knative Operator to install and manage your Knative cluster, you should not edit the ConfigMaps directly as the Operator will overwrite the changes. See [Configuring Knative by using the Operator](../install/operator/configuring-with-operator.md) for information on how to manage ConfigMaps using the operator.
+
+## The _example key
+
+ConfigMap files installed by Knative contain an `_example` key that shows the usage and purpose of all the known configuration keys in that ConfigMap. This key does not affect Knative behavior, but contains a value which acts as a documentation comment.
+
+If a user edits the `_example` key by mistakenly thinking their edits don't effect on the cluster operation, the Knative webhook catches the error and alerts the user that their update could not be patched. More specifically, the edit is caught when the value of the checksum for the `_example` key is different when compared with the `knative.dev/example-checksum` annotation on the ConfigMap. If the checksum is null or missing, the webhook server does not create the warning.
```
evankanderson:
```suggestion
If a user edits the `_example` key rather than creating a new key in the ConfigMap, the changes won't affect the cluster cluster configuration, which can be confusing. The Knative webhook flags this error and alerts the user that their update could not be patched. More specifically, the edit is caught when the checksum for the `_example` key differs from the `knative.dev/example-checksum` annotation on the ConfigMap. If the checksum is null or missing, the webhook does not create the warning.
```

The sentence here had some awkward wording "by mistakenly thinking ... don't effect on ...".  I _think_ this reads more smoothly?

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,75 @@
...
+
+Accordingly, you cannot alter the contents of the `_example` key, but you can delete the `_example` key altogether or delete the annotation.
+
+For example, the following YAML code shows the first 24 lines of the `config-defaults` ConfigMap with the checkum highlighted.
```
evankanderson:
```suggestion
For example, the following YAML code shows the first 24 lines of the `config-defaults` ConfigMap with the checksum highlighted.
```

docs/versioned/admin/editing-configmaps.md
```diff
@@ -0,0 +1,75 @@
...
+- Before applying ConfigMaps, validate their syntax and content using tools like `kubeval` or `kubectl apply --dry-run=server`.
+- Test ConfigMap changes in a staging environment to ensure compatibility with the application version.
+
+### Storage and versioning
+
+How you manage storage relates the canonical locations of both the cluster and Git. If your cluster is canonical, then you're exporting or backing up the configuration to Git. If Git is canonical, then you're practicing GitOps, and you should make changes in Git and then apply the files from Git to your cluster.
+
+If you manage the ConfigMap by using `kubectl edit`, periodically export ConfigMaps from the cluster (`kubectl get configmap -o yaml`) and commit them to Git for recovery purposes. Include applicable version numbers in `app.properties` as needed.
+
+If you manage the ConfigMap by keeping the definition in Git and automatically applying it to the cluster (GitOps), you can apply the changes manually  sor use automation (for example Flux or ArgoCD) to apply the changes once they are committed.
```
evankanderson:
```suggestion
If you manage the ConfigMap by keeping the definition in Git and automatically applying it to the cluster (GitOps), you can apply the changes manually or use automation (for example Flux or ArgoCD) to apply the changes after they are committed.
```

iRaindrop:
Ready for PR approval. Thanks for the great edits and fixes. I also added a link to this page from the Admin Overview.

PR #6473, opened 2025-10-29
Fix typos/casing in a few places

dprotaso:


dprotaso:
/hold cancel 

all good

PR #6477, opened 2025-10-29
Clean up absolute links in docs

evankanderson:
Absolute links like `/docs/...` will always link to the current release, even when in older or pre-release version documentation.  This switches all `(/docs` links (except one snippet, which has inclusions at different depths).

At some point, we should probably look at combining the serving and eventing observability documentation (which are now both under "administration"), but I didn't want to blow up the size of this PR.

## Proposed Changes 

- Replace `(/docs` links with relative links within the versioned directory.


evankanderson:
/hold

I need to check a couple pages before this is ready to merge.

evankanderson:
It turns out that `security-prereqs-images.md` was used by previous versions, which was flagged by `mkdocs strict`.  😁

PR #6478, opened 2025-10-29
allow build on mac

dprotaso:
- make sure users set GITHUB_TOKEN
- require gsed on mac
- if an `upstream` remote exist use that over `origin` for branch pulling. This allows `gh repo fork` flows which sets up `origin` and `upstream` remotes


evankanderson (commented):
/lgtm
/approve

PR #6501, opened 2025-11-14
Fixed port mapping.

munjal-gc:
Default project deployment is having wrong port mapping of 8080. Node application runs on port 3000.

This PR will fix - front-end-deployment when initial setup is done...

### DIFF
--- a/code-samples/eventing/bookstore-sample-app/start/frontend/config/100-front-end-deployment.yaml
+++ b/code-samples/eventing/bookstore-sample-app/start/frontend/config/100-front-end-deployment.yaml
@@ -18,7 +18,7 @@ spec:
         - name: bookstore-frontend
           image: ghcr.io/knative/bookstore-frontend:latest
           ports:
-            - containerPort: 8080
+            - containerPort: 3000
           env:
           - name: POD_NAMESPACE
             valueFrom:
@@ -33,7 +33,7 @@ metadata:
 spec:
   ports:
     - port: 80
-      targetPort: 8080
+      targetPort: 3000
   selector:
     app: bookstore-frontend
   type: LoadBalancer


munjal-gc:
Yes, this first PR as we are trying out the quickstart and bumped in to
this issue.

dprotaso:
The Dockerfile has `PORT` enviroment variable set to `8080` which I believe next.js will use

https://github.com/knative/docs/blob/eee8ba83522bcf8a11ced3ff0c64fcf4c29367e6/code-samples/eventing/bookstore-sample-app/start/frontend/Dockerfile#L38

Thus I think the configuration is correct - though if you're running the server locally I think it will use the local port `3000` because we haven't updated the package.json

munjal-gc:
make sense so issue may be in latest image uploaded at ghcr.io/knative/bookstore-frontend:latest
which may not have been built / published post above chances...

dprotaso:
Looks like our CI has an issue and the latest changes weren't pushed maybe

eg. https://github.com/knative/docs/actions/runs/19432715858/job/55595454488

let me fix this

dprotaso:
I have a PR out here: https://github.com/knative/docs/pull/6508

cdsre:
Nice catch with this one. I have just started the E2E tutorial and faced this issue too. for now I have changed the port mapping locally.

dprotaso:
Ok new images are out - can folks try it out?

dprotaso:
I see 8080 as a port here now

https://explore.ggcr.dev/?blob=ghcr.io/knative/bookstore-frontend@sha256:8b89a6f735cb5b4b1e991141fe0edba499afa0c884e4a02324f6c4b39d98a4fa&mt=application%2Fvnd.oci.image.config.v1%2Bjson&size=8127&manifest=ghcr.io/knative/bookstore-frontend@sha256:f62aff150b43562b47eba6840e11c601d7bf63987d239422f6b3a36f6670bb52

dprotaso:
I think what's left is if you want to update the package.json to change the local port to 8080 I'd be ok with that.

Otherwise I think we can close out the PR

munjal-gc:
I think we are good to close the PR as root cause of the issue is addressed... 

Thanks for propmtly addressing same.

PR #6518, opened 2025-11-19
Create Configure Knative Networking page

iRaindrop:
The content for the Kourier, Contour, Gateway API, and Istio is could be helpful consolidated as opposed to separate install and config topics. 

## Proposed Changes 

- Create "Configure network adapters" under "Administration->Configuring Knative->Networking options

### DIFF
--- a/docs/versioned/serving/config-network-adapters.md
+++ b/docs/versioned/serving/config-network-adapters.md
@@ -40,7 +40,7 @@ Although Knative uses the same interface for different ingress providers, they b
 
 ## Ingress providers
 
-Review the tabbed content in this section to determine the optimal networking layer for your cluster. If you already have one of the ingress controllers installed in your cluster, we recommend using your existing installation. For most users without a supported ingress, the Kourier ingress controller is sufficient. You can expand your capabilities with the Contour ingress, a full-feature service mesh with Istio, and the Kubernetes Gateway API.
+Review the tabbed content in this section to determine the optimal provider for your cluster. If you already have one of the ingress controllers installed in your cluster, we recommend using your existing installation. For most users without a supported ingress, the Kourier ingress controller is sufficient. You can expand your capabilities with the Contour ingress, a full-feature service mesh with Istio, and the Kubernetes Gateway API.
 
 === "Kourier"
 


docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,183 @@
...
+      look: neo
+    ---
+    flowchart LR
+    K1["Knative<br>net-kourier"] -- creates --> K2["Ingress&nbsp;objects"]
+    K2 --> K3["Class: kourier.ingress.networking.knative.dev"]
+    ```
+
+    The Knative `net-kourier` ingress is installed with Knative Serving. Kourier is a lightweight alternative for the Istio ingress as its deployment consists only of an envoy proxy and a control plane. If Kourier is satisfactory, no further configurations are required.
+
+    Kourier is the only supported ingress controller for the IBM Z and IBM Power platforms, and requires additional steps as described in [Install Serving with YAML on IBM-Z and IBM-P](/versioned/install/yaml-install/serving/install-serving-with-yaml-on-IBM-Z-and-IBM-P.md).
```
iRaindrop: The link to install-serving-with-yaml-on-IBM-Z-and-IBM-P.md returns a 404 because it is not yet in the Nav.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,183 @@
...
+      look: neo
+    ---
+    flowchart LR
+        Client["External&nbsp;Client"] --> CGW["Custom&nbsp;Ingress&nbsp;Gateway"]
+        CGW --> KIGW["Knative&nbsp;Ingress&nbsp;Gateway"] & Client
+        KIGW --> Revision["Knative&nbsp;Revision"] & CGW
+        Revision --> KIGW
+    ```
+
+    Knative has a default Istio integration without the full-feature service mesh. The `knative-ingress-gateway` in the `knative-serving` namespace is a shared Istio gateway resource that handles all incoming (north-south) traffic to Knative services. This gateway points to the underlying `istio-ingressgateway` service in the `istio-system` namespace. You can replace this gateway with one of your own, see [Configuring the Ingress gateway](setting-up-custom-ingress-gateway.md).
```
iRaindrop: Decision to make - bring the topic into this tab or not?

docs/snippets/netadapter-istio.md
```diff
@@ -0,0 +1,35 @@
+Use the following steps to install Istio and set it as the ingress conroller.
+
+1. Install a properly configured Istio:
```
iRaindrop, line 6: The artifact "macros" are not rendering to provide the URL. I wasn't able to find this one.

docs/snippets/netadapter-istio.md
```diff
@@ -0,0 +1,35 @@
...
+    kubectl apply -f {{ artifact(repo="net-istio",org="knative-extensions",file="istio.yaml")}}
+    ```
+
+1. Install the Knative Istio controller:
+
+    ```bash
+    kubectl apply -f {{ artifact(repo="net-istio",file="net-istio.yaml")}}
+    ```
+
+<!-- 1. Set the `config-network` ConfigMap to use Istio:
```
iRaindrop: I added this ConfigMap patch step that might not be correct. Or does Istio make the ingress controller connection without the ConfigMap?

docs/snippets/netadapter-gatewayapi.md
```diff
@@ -0,0 +1,22 @@
+Use the following steps to install and configure the Knative Gateway API.
+
+1. Install the Knative Gateway API channel:
+
+    ```bash
+      kubectl apply -f {{ artifact(repo="net-gateway-api",org="knative-extensions",file="net-gateawy-api.yaml")}}
+    ```
+
+1. Configure Knative Serving to use Knative Gateway API channel:
```
iRaindrop: Is channel the right word? Doesn't the ConfigMap need to configure the underlying ingress controller?

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,187 @@
...
+- Contour: `contour-*`
+- Istio: `istio-webhook-*`. The main Istio control plane pods such as `istiod-*` are in the `istio-system` namespace. Knative adds the `istio-webhook-*` pod in the `knative-serving` namespace when Istio is the chosen networking layer.
+
+The `network-config` ConfigMap sets which controller to use with the ingress controller key. This key is patched with the name of any new controller. See [Changing the ingress controller](#change-the-controller) for important information about using this key.
+
+## Network layer options
+
+Review the following tabs to determine the optimal networking layer for your cluster. For most users, the Kourier ingress controller is sufficient in conjunction the default Istio gateway, which is also included in the Knative Serving installation. You can expand your capabilities with the Contour ingress, a full-feature service mesh with Istio, and the Kubernetes Gateway API.
+
+=== "Kourier"
```
iRaindrop, line 45: A little context on these mermaid charts. Evan created the one for Gateway API and I did the others with AI help and I also added the Controller to the Gateway API.

Hence, they need tech reviewed as they could be misleading!

docs/versioned/serving/config-network-adapters.md
evankanderson: This should be a relative link, e.g. `../install/yaml-install/serving/install-serving-with-yaml-on-IBM-Z-and-IBM-P.yaml`.

Absolute links like this won't work when we build the different documentation versions, _and_ they wouldn't work right between newer and older documentation versions even if you got the URL to line up at one point in time.

docs/snippets/dns.md
```diff
@@ -1,7 +1,6 @@
-You can configure DNS to prevent the need to run curl commands with a host header.
+You can configure DNS to avoid running curl commands with a host header.
```
evankanderson:
```suggestion
You can configure DNS to avoid specifying the host header in curl commands, or to access the content with a web browser.
```

docs/snippets/netadapter-contour.md
```diff
@@ -0,0 +1,28 @@
+Use the following steps to install and enable Contour and set it as the ingress conroller.
+
+1. Install the Knative Contour controller:
+
+    ```bash
+    kubectl apply -f https://github.com/knative/net-kourier/releases/latest/download/kourier.yaml
```
evankanderson: We have a macro which will use the same version as the documentation:

```suggestion
    kubectl apply -f {{ artifact(org="knative-extensions", repo="net-kourier", file="kourier.yaml" }}
```

evankanderson: Although, I'm realizing this should probably be `net-contour` and `contour.yaml`...

docs/snippets/netadapter-gatewayapi.md
evankanderson: Knative Serving reads the configmap, so updating the ConfigMap configures Knative Serving.  (This usage is correct.)

docs/snippets/netadapter-istio.md
evankanderson: You may need a space between `) }}` for the macro to work.  I can take a look this weekend.

docs/snippets/netadapter-istio.md
evankanderson: I believe that Istio may be the default value if one is not set, but doing this explicitly shouldn't hurt.

docs/snippets/netadapter-istio.md
```diff
@@ -0,0 +1,35 @@
...
+    kubectl patch configmap/config-network \
+    --namespace knative-serving \
+    --type merge \
+    --patch '{"data":{"ingress-class":"istio.ingress.networking.knative.dev"}}'
+  ``` -->
+
+1. Verify the installation by having pods with the base name of `istio` and `istio-webhook` in the results.
+
+    ```bash
+    kubectl get pods -n knative-serving
```
evankanderson: Is `knative-serving` the right namespace for these pods?  (I'm not sure, I haven't set this up recently)

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Determine current state
+
+Use the following command to determine which ingress controllers are installed and their status.
+
+```bash
+kubectl get pods -n knative-serving
+```
+
+The ingress controllers, that have been tested for Knative, have the following base names:
```
evankanderson: The commas here don't seem quite right.

Maybe phrase this as the "The Knative team tests the following ingress controllers:"?

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,187 @@
...
+
+- Kourier: `kourier-control-*`, and `kourier-gateway-*`. Kourier is included in the Knative Serving installation should appear in the results when your cluster is first created.
+- Contour: `contour-*`
+- Istio: `istio-webhook-*`. The main Istio control plane pods such as `istiod-*` are in the `istio-system` namespace. Knative adds the `istio-webhook-*` pod in the `knative-serving` namespace when Istio is the chosen networking layer.
+
+The `network-config` ConfigMap sets which controller to use with the ingress controller key. This key is patched with the name of any new controller. See [Changing the ingress controller](#change-the-controller) for important information about using this key.
```
evankanderson: This is actually slightly different -- each ingress controller only manages ingress objects that are annotated with its key.  Knative Serving uses a default value of the key based on the `network-config` ConfigMap.  So this is somewhat of a cooperative process between the ingress controller and the core Knative controller.

evankanderson: (The end result is effectively the same, but the mechanism is different, which seemed important to correct in the documentation.)

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Network layer options
+
+Review the following tabs to determine the optimal networking layer for your cluster. For most users, the Kourier ingress controller is sufficient in conjunction the default Istio gateway, which is also included in the Knative Serving installation. You can expand your capabilities with the Contour ingress, a full-feature service mesh with Istio, and the Kubernetes Gateway API.
```
evankanderson: Folks shouldn't need _both_ Kourier and Istio -- one or the other should be sufficient, depending on need.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,187 @@
...
+
+    ```mermaid
+    ---
+    config:
+      theme: default
+      layout: elk
+      look: neo
+    ---
+    flowchart LR
+    K1["Knative<br>net-kourier"] -- creates --> K2["Ingress&nbsp;objects"]
```
evankanderson: "Ingress" is confusing here, as Kubernetes has an "Ingress" object type which is _different_ from the Knative `networking.internal.knative.dev` Ingress type.  We called it "Ingress" because it was what we _wanted_ Ingress to be, and every actual Ingress implementation had a different name for the "better Ingress" implementation.

We generally refer to the Knative version of Ingress from `networking.internal` as "KIngress".

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,187 @@
...
+    K2 --> K3["Class: kourier.ingress.networking.knative.dev"]
+    ```
+
+    The Kourier ingress controller, `net-kourier`, is installed with Knative Serving. Kourier is a lightweight alternative for the Istio ingress as its deployment consists only of an envoy proxy and a control plane. If Kourier is satisfactory, no further configurations are required.
+
+    Kourier is the only supported ingress controller for the IBM Z and IBM Power platforms, and requires additional steps as documented in [Install Serving with YAML on IBM-Z and IBM-P](/versioned/install/yaml-install/serving/install-serving-with-yaml-on-IBM-Z-and-IBM-P.md).
```
evankanderson: We should make clear that Kourier is a perfectly fine choice for all platforms, but for these IBM platforms, it's the only option.

docs/versioned/serving/config-network-adapters.md
iRaindrop: Great edit Evan - now is:

Each ingress controller manages only those ingress objects that are annotated with its key. Knative Serving uses a default value of the key based on the `network-config` ConfigMap.

iRaindrop:
Moved the page up a level next to the Networking options node.
Update: Moved the page back.

docs/snippets/dns.md
```diff
@@ -11,10 +10,4 @@ Follow the procedure for the DNS of your choice:
...
     ```
 
-    !!! warning
-        This will only work if the cluster `LoadBalancer` Service exposes an
-        IPv4 address or hostname, so it will not work with IPv6 clusters or local setups
-        like minikube unless [`minikube tunnel`](https://minikube.sigs.k8s.io/docs/commands/tunnel/)
-        is running.
-
-        In these cases, see the "Real DNS" or "No DNS" tabs.
+    This configuration works only if the cluster `LoadBalancer` Service exposes an IPv4 address or hostname. It does not work with IPv6 clusters or local setups such as minikube, unless [`minikube tunnel`](https://minikube.sigs.k8s.io/docs/commands/tunnel/) is running, and should consider using the "Real DNS" or "No DNS" tabs.
```
evankanderson: "It does not work with ...., and should consider" seems to have changed the subject of the sentence.

docs/versioned/.nav.yml
```diff
@@ -266,6 +266,7 @@ nav:
         - Working with ConfigMaps: admin/editing-configmaps.md
         - Configure high-availability components: serving/config-ha.md
         - Exclude namespaces from the Knative webhook: serving/webhook-customizations.md
+        - Configure Knative networking: serving/config-network-adapters.md
         - Networking Options:
           - Configure the ingress gateway: serving/setting-up-custom-ingress-gateway.md
```
evankanderson: It's odd to have this up one level from "Networking options" and "Configure the ingress gateway", which feel like related tasks.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,189 @@
...
+
+    ```mermaid
+    ---
+    config:
+      theme: default
+      layout: elk
+      look: neo
+    ---
+    flowchart LR
+    K1["Knative<br>net-kourier"] -- creates --> K2["KIngress&nbsp;objects"]
```
evankanderson: `net-kourier` doesn't create KIngress objects; it reads objects created by serving.  I _think_ a more accurate diagram would have two parts.  One is common for all network layers:

```mermaid
flowchart LR

route["Route object"] -- "read by" --> serving-core("Serving<br>controller") -- creates --> KIngress["Ingress object<br>networking.internal.knative.dev<br>(KIngress)"]
```

And then for Kourier, the chart would look like

```mermaid
flowchart LR

KIngress["KIngress<br>Class:kourier.ingress.networking.knative.dev"] -- "read by" --> controller("net-kourier<br>controller") -- programs --> envoy("Envoy deployment<br>kourier-system namespace")
```

I'll update the other diagrams with just the second-part diagrams.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,189 @@
...
+    K2 --> K3["Class: kourier.ingress.networking.knative.dev"]
+    ```
+
+    The Kourier ingress controller, `net-kourier`, is installed with Knative Serving. Kourier is a lightweight alternative for the Istio ingress as its deployment consists only of an envoy proxy and a control plane. If Kourier is satisfactory, no further configurations are required.
```
evankanderson: I believe that Kourier is installed by the quickstart, but other installations need to explicitly install it.

I wouldn't say it's an alternative for the Istio ingress so much as it's a lightweight implementation of the KIngress resource for clusters which don't need other ingress features.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,189 @@
...
+
+    Kourier is a fine choice for all platforms, but for these IBM platforms it's the only option and requires additional steps as documented in [Install Serving with YAML on IBM-Z and IBM-P](../install/yaml-install/serving/install-serving-with-yaml-on-IBM-Z-and-IBM-P.md).
```
evankanderson:
```suggestion
    Kourier is a fine choice for all platforms, but for IBM-Z and IBM-P platforms it's the only supported option and requires additional steps as documented in [Install Serving with YAML on IBM-Z and IBM-P](../install/yaml-install/serving/install-serving-with-yaml-on-IBM-Z-and-IBM-P.md).
```

docs/snippets/dns.md
```diff
@@ -11,10 +10,4 @@ Follow the procedure for the DNS of your choice:
     kubectl apply -f {{artifact(repo="serving",file="serving-default-domain.yaml")}}
```
evankanderson:
```suggestion
    kubectl apply -f {{ artifact(repo="serving",file="serving-default-domain.yaml") }}
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,189 @@
...
+    ```mermaid
+    ---
+    config:
+      theme: default
+      layout: elk
+      look: neo
+    ---
+    flowchart LR
+    C1["Knative<br>net-contour"] -- creates --> C2["KIngress&nbsp;objects"]
+    C2 --> C3["Class: contour.ingress.networking.knative.dev"]
```
evankanderson: This should be:  (May need "TD" formatting)

```mermaid
    flowchart LR
    
    C1["KIngress objects"] -- "read by" --> C2("Knative<br>net-contour") -- creates --> C3["HTTPProxy<br>projectcontour.io"] -- "read by" --> C4("Contour")
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,189 @@
...
+    ```
+
+    The Contour ingress controller, `net-contour`, bridges Knative's KIngress resources to Contour's HTTPProxy resources. A good choice for clusters that already run non-Knative apps, teams who want to use a single Ingress controller, and are already using Contour envoy but don't need a full-feature service mesh.
```
evankanderson:
```suggestion
    The Contour ingress controller, `net-contour`, bridges Knative's KIngress resources to Contour's HTTPProxy resources. A good choice for clusters that already run non-Knative apps, teams who want to use a single Ingress controller, and are already using Contour or don't need a full-feature service mesh.
```

docs/snippets/netadapter-contour.md
```diff
@@ -0,0 +1,28 @@
+Use the following steps to install and enable Contour and set it as the ingress conroller.
+
+1. Install the Knative Contour controller:
+
+    ```bash
+    kubectl apply -f {{ artifact(org="knative-extensions", repo="net-contour", file="contour.yaml" }}
+    ```
```
evankanderson: This installs Contour in an internal/external configuration.  You need to install both the `contour.yaml` and the `net-contour.yaml` files from the release.  They are separate because the `contour.yaml` is a re-packaging of the base Contour project to set up both internal (cluster-local) and external (public-facing) HTTP routers.

(Specifically, we need an extra step which installs `net-contour.yaml`; without this, nothing will convert the KIngress to HTTPRoutes, and deployments will stall on the "checking that the Ingress is working" step.)

docs/snippets/netadapter-gatewayapi.md
```diff
@@ -0,0 +1,22 @@
+Use the following steps to install and configure the Knative Gateway API.
```
evankanderson:
```suggestion
Use the following steps to install and configure the Knative Gateway API adapter.  Note that you already need to have a [Gateway API implementation](https://gateway-api.sigs.k8s.io/implementations/) installed in your cluster.  The Knative team currently tests the Istio, Contour, and Envoy-Gateway implementations of Gateway API.
```

docs/snippets/netadapter-gatewayapi.md
```diff
@@ -0,0 +1,22 @@
...
+
+1. Install the Knative Gateway API channel:
```
evankanderson: This isn't a "channel".  If anything, it's an "Ingress implementation" or "controller".

docs/snippets/netadapter-gatewayapi.md
evankanderson: Oh, I misread the comment.  You're right that channel (here and above) isn't right, and should be controller.

docs/snippets/netadapter-gatewayapi.md
```diff
@@ -0,0 +1,22 @@
...
+
+    ```bash
+      kubectl apply -f {{ artifact(repo="net-gateway-api",org="knative-extensions",file="net-gateawy-api.yaml") }}
+    ```
```
evankanderson, line 12: We need an additional step here (not currently well-documented) to align the `config-gateway` ConfigMap with the installed Gateways.  Knative assumes two Gateways in the cluster: one which is externally exposed, and one which is only exposed within the cluster ("internal").

Creation of these gateways is alluded to in this section of the `net-gateway-api` documentation: https://github.com/knative-extensions/net-gateway-api?tab=readme-ov-file#install-a-supported-implementation (In particular, the `kubectl apply -f third_party/$IMPLEMENTATION` directions in each section.)

More concretely, the `config-gateway` ConfigMap needs two keys:

* **external-gateways**
* **local-gateways**

Each value is a string which contains additional YAML content (a list of objects, typically a single item) in the following format:

```yaml
- class: $GATEWAY_CLASS_NAME
  gateway: $NAMESPACE/$GATEWAY_NAME
  service: $NAMESPACE/$SERVICE_NAME
```

Where `$SERVICE_NAME` is the Kubernetes Service name which points to the pods in the Gateway implementation.  (Note: _Kubernetes_ Service name, not _Knative_ Service name.)

docs/snippets/netadapter-istio.md
evankanderson: Yep, I see this not working, but I haven't yet figured out why.  I'm guessing it's the combination of macros and snippets that is doing it.

docs/snippets/netadapter-istio.md
```diff
@@ -0,0 +1,35 @@
+Use the following steps to install Istio and set it as the ingress conroller.
+
+1. Install a properly configured Istio:
+
+    ```bash
+    kubectl apply -l knative.dev/crd-install=true -f {{ artifact(repo="net-istio",org="knative-extensions",file="istio.yaml") }}
```
evankanderson, line 9: @dprotaso -- do we still need our own `istio.yaml`?

docs/snippets/netadapter-istio.md
```diff
@@ -0,0 +1,35 @@
...
+    kubectl apply -f {{ artifact(repo="net-istio",org="knative-extensions",file="istio.yaml") }}
+    ```
+
+1. Install the Knative Istio controller:
+
+    ```bash
+    kubectl apply -f {{ artifact(repo="net-istio",file="net-istio.yaml") }}
+    ```
+
+1. Set the `config-network` ConfigMap to use Istio:
```
evankanderson: Elsewhere, we call this "Configure...".  Can we use the same language in each tab for this step?

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,189 @@
...
+        CGW --> KIGW["Knative&nbsp;Ingress&nbsp;Gateway"] & Client
+        KIGW --> Revision["Knative&nbsp;Revision"] & CGW
+        Revision --> KIGW
+    ```
+
+    Knative has a default Istio integration without the full-feature service mesh. The `knative-ingress-gateway` in the `knative-serving` namespace is a shared Istio gateway resource that handles all incoming (north-south) traffic to Knative services. This gateway points to the underlying `istio-ingressgateway` service in the `istio-system` namespace. You can replace this gateway with one of your own.
+
+    **Install and configure Ingress Gateway**
+
+    See [Configuring the Ingress gateway](setting-up-custom-ingress-gateway.md).
```
evankanderson: The referenced document contains configuration details for customizing the Istio configuration.  It's not separate from the "Istio" configuration, but instead provides additional configuration detail.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,189 @@
...
+        HR --> Controller
+        Controller -- routes&nbsp;traffic&nbsp;to --> Pods["Your&nbsp;Pods"]
+
+    style net-gateway-api fill:#e3f2fd,stroke:#1976d2
+    style underlying fill:#fff3e0,stroke:#ef6c00
+    ```
+
+    The Knative `net-gateway-api` is a KIngress implementation and testing for Knative integration with the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/). Good for teams adopting the Gateway API to unify ingress across Kubernetes.
+
+    The Kubernetes Gateway API requires a controller or service mesh. Istio and Contour implementations are tested though other Gateway API implementations should work. Currently, there is no native Gateway API support for Kourier. For more information see [Tested Gateway API version and Ingress](https://github.com/knative-extensions/net-gateway-api/blob/main/docs/test-version.md).
```
evankanderson: We test Istio, Contour, and Envoy Gateway implementation at this point.  I don't think we ever intend to implement Gateway API support for Kourier.  (Also, Envoy Gateway and other Gateway API implementations will probably never get a separate support path like Istio and Contour did -- that was basically a stepping stone before Gateway API existed.)

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,189 @@
...
+
+    The controller that Knative uses is determined by which Gateway API-compatible controller you install and configure in your cluster.
```
evankanderson: In particular, it's determined by the configuration in `config-gateway`, that I described in an earlier comment.

docs/versioned/.nav.yml
iRaindrop: I see your point. I'll put it back.

docs/snippets/netadapter-kourier.md
```diff
@@ -0,0 +1,28 @@
+Use the following steps to install Kourier and set it as the ingress controller.
+
+1. Install the Knative Kourier controller:
+
+    ```bash
+    kubectl apply -f https://github.com/knative/net-kourier/releases/latest/download/kourier.yaml
```
dsimansk: Same placeholder pattern should be usable here:
```suggestion
     kubectl apply -f {{ artifact(repo="net-kourier",file="kourier.yaml") }}
```

docs/snippets/dns.md
iRaindrop: Removed that "and should consider..." clause as the tabs of DNS choice are inherently being compared.

dprotaso:
FYI - I just merged https://github.com/knative/docs/pull/6540 and didn't realize this was still outstanding.

Unsure if this changes anything in this PR

iRaindrop:
> FYI - I just merged #6540 and didn't realize this was still outstanding.
> 
> Unsure if this changes anything in this PR

It doesn't. No worries, as I can just incorporate it into the content. 

I do suggest that that note, because it's large and at the bottom of the list, be changed to the following to avoid confusion with the other bullets in the list:

"Third party integrations are not managed by the Knative community."

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+
+# Configure Knative networking
+
+This page provides configuration guidance for Knative networking. You can configure Ingress controls, service-meshes, and gateways.
+
+For installation instructions, see [Install serving with YAML](../install/yaml-install/serving/install-serving-with-yaml.md).
+
+## Network layer options
+
+Review the tabbed content in this section to determine the optimal networking layer for your cluster. For most users, the Kourier ingress controller is sufficient. You can expand your capabilities with the Contour ingress, a full-feature service mesh with Istio, and the Kubernetes Gateway API.
```
evankanderson: I'd recommend people use what they already have if they have one of the solutions.

```suggestion
Review the tabbed content in this section to determine the optimal networking layer for your cluster. If you already have one of the ingress controllers installed in your cluster, we recommend using your existing installation. For most users without a supported ingress, the Kourier ingress controller is sufficient. You can expand your capabilities with the Contour ingress, a full-feature service mesh with Istio, and the Kubernetes Gateway API.
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
+---
+audience: administrator
+components:
+  - serving
+function: how-to
+---
+
+# Configure Knative networking
+
+This page provides configuration guidance for Knative networking. You can configure Ingress controls, service-meshes, and gateways.
```
evankanderson: There are two parts of Kubernetes networking -- in cluster networking (which Knative doesn't need to do anything special with), and into-cluster networking (aka ingress), which Knative _does_ need to integrate with.  Clarify that this is for ingress networking:

```suggestion
This page provides configuration guidance for Knative Serving's integration with Kubernetes ingress controllers. Knative leverages existing ingress controls in your Kubernetes cluster, allowing you to use the same monitoring, .
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+
+## Network layer options
+
+Review the tabbed content in this section to determine the optimal networking layer for your cluster. For most users, the Kourier ingress controller is sufficient. You can expand your capabilities with the Contour ingress, a full-feature service mesh with Istio, and the Kubernetes Gateway API.
+
+The Knative tested ingress controllers (Contour, Istio, and Kourier) have the following common configurations:
+
+- Certificate management: Configurable secrets for TLS encrypted traffic.
+- Timeout policies: Controls for idle, and response stream timeouts.
+- Traffic visibility: Mechanisms to expose services externally or cluster-locally.
```
evankanderson: Link to these topics?

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+        direction LR
+            kingress1["Ingress object (KIngress)<br>networking.internal.knative.dev"]
+            serving["Serving<br>controller"]
+            route["Route object"]
+      end
+        route -- read by --> serving
+        serving -- creates --> kingress1
+        style kingress1 fill:#BBDEFB,stroke-width:1px,stroke-dasharray: 0
+        style serving fill:#FFE0B2
+        style top fill:transparent
```
evankanderson: The mapping of Route objects to `networking.internal.knative.dev` objects is common across all the controllers -- you might want to pull it out of the tabs, and put it in a common section above the "Network layer options".

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+        kourier -- programs --> envoy
+    
+        style envoy fill:#BBDEFB
+        style kourier fill:#FFE0B2
+        style bottom fill:transparent
+    ```
+
+    Kourier is a lightweight implementation of the KIngress resource and is suitable for clusters that don't need other ingress features. Kourier is optimal for learning and prototyping, and is installed by the Knative [Quickstart](../getting-started/README.md).
+
+    Kourier is is the only supported option for IBM-Z and IBM-P platforms. These IBM platforms require additional steps as documented in [Install Serving th YAML on IBM-Z and IBM-P](../install/yaml-install/serving/install-serving-with-yaml-on-IBM-Z-and-IBM-P.md).
```
evankanderson:
```suggestion
    Kourier is is the only supported option for IBM-Z and IBM-P platforms. These IBM platforms require additional steps as documented in [Install Serving with YAML on IBM-Z and IBM-P](../install/yaml-install/serving/install-serving-with-yaml-on-IBM-Z-and-IBM-P.md).
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+
+    Kourier provides the following additional configuration options:
```
evankanderson, line 74: Do you want to link to documentation on any of these?

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+            C2 -- "creates" --> C3
+            C3 -- "read by" --> C4
+        
+            style C2 fill:#FFE0B2
+            style C3 fill:#BBDEFB
+            style C4 fill:#BBDEFB
+            style bottom fill:transparent
+    ```
+
+    The Contour ingress controller, `net-contour`, bridges Knative's KIngress resources to Contour's HTTPProxy resources. A good choice for clusters that already run non-Knative apps, are already using a Contour envoy, or don't need a full-feature service mesh.
```
evankanderson:
```suggestion
    The Contour ingress controller, `net-contour`, bridges Knative's KIngress resources to Contour's HTTPProxy resources. Contour is good choice for clusters that already run non-Knative apps, are already using a Contour envoy, or don't need a full-feature service mesh.
```

docs/snippets/netadapter-contour.md
```diff
@@ -0,0 +1,32 @@
...
+
+1. Install the Knative Contour configuration:
+
+    ```bash
+    kubectl apply -f {{ artifact(repo="net-contour",org="knative-extensions",file="contour.yaml")}}
+    ```
+
+    <!-- ```bash
+    kubectl apply -f https://github.com/knative-extensions/net-contour/releases/download/knative-v1.20.0/contour.yaml
+    ```   -->
```
evankanderson: I think you want to remove this commented line.

docs/snippets/netadapter-gatewayapi.md
evankanderson: I would say "class" or "ingress class" instead of "channel":


```suggestion
1. Configure Knative Serving to use Knative Gateway API class:
```

docs/snippets/netadapter-gatewayapi.md
evankanderson: We still need some documentation about filling in the `config-gateway` Configmap with `external-gateways` and `local-gateways` objects.

docs/snippets/netadapter-istio.md
evankanderson: Okay, I figured it out.  We have _two_ ways of including snippets.  `--8<--` comes from [`pymdownx.snippets`](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/) and seems to run _after_ the `macros` extension.  `{% include ... %}` comes from [`mkdocs-macros`](https://mkdocs-macros-plugin.readthedocs.io/en/latest/advanced/#including-external-files-in-pages), and _will_ render the macros like `artifact(...)` in the markdown.

You can see the use of `{% include "dns.md" %} in `install-serving-with-yaml.md` as an example.

Note that you may need to indent the snippet content by 4 spaces when using `{% include ... %}`, based on experimentation.  You'll note that `dns.md` and `no-dns.md` both do this (and also include the tab header in the document).

docs/snippets/no-dns.md
```diff
@@ -1,7 +1,3 @@
-<!-- Referenced by:
-- install/operator/knative-with-operators.md
-- install/yaml-install/serving/install-serving-with-yaml.md
--->
```
evankanderson, line 4: Why remove this information?

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+
+    Contour provides the following additional configuration options:
+
+    - CORS policy configuration.
+    - Direct visibility classes for external and internal traffic.
```
evankanderson: Contour also supports a number of cluster-wide configuration options on its own: https://projectcontour.io/docs/1.33/configuration/

It's probably worth linking that here, so that it's clear that you can configure many options using the native Contour configuration.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+            envoy["Envoy deployment<br>istio-system namespace"]
+            kourier["net-istio<br>controller"]
+            kingress2["KIngress class:<br>istio.ingress.networking.knative.dev"]
+      end
+        kingress2 -- read by --> kourier
+        kourier -- programs --> envoy
+    
+        style envoy fill:#BBDEFB
+        style kourier fill:#FFE0B2
+        style bottom fill:transparent
```
evankanderson: The pattern here is equivalent to Contour, except using the VirtualService resource type.  The diagram should be:

```suggestion
     subgraph bottom[" "]
        direction LR
            VirtualService["Istio<br>VirtualService"]
            istiod("Istio")
            net-istio("net-istio<br>controller")
            kingress["KIngress class:<br>istio.ingress.networking.knative.dev"]
      end
        kingress -- read by --> net-istio
        net-istio -- creates --> VirtualService
        VirtualService -- read by --> istiod
    
        style VirtualService fill:#BBDEFB
        style istiod fill:#BBDEFB
        style net-istio fill:#FFE0B2
        style bottom fill:transparent
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+    ```
+
+    The Knative `net-istio` is a KIngress controller for Istio. It's a full-feature service mesh that also functions as a Knative ingress. Well suited for enterprises already running Istio or who need advanced service mesh features.
```
evankanderson:
```suggestion
    The Knative `net-istio` is a KIngress controller for Istio. Istio is a full-feature service mesh that also functions as a Knative ingress. `net-istio` is well suited for enterprises already running Istio or who need advanced service mesh features.
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+
+    Knative has a default Istio integration without the full-feature service mesh. The `knative-ingress-gateway` in the `knative-serving` namespace is a shared Istio gateway resource that handles all incoming (north-south) traffic to Knative services. This gateway points to the underlying `istio-ingressgateway` service in the `istio-system` namespace. You can replace this gateway with one of your own. See [Configuring the Ingress gateway](setting-up-custom-ingress-gateway.md).
```
evankanderson:
```suggestion
    Knative has a default Istio integration without the full-feature service mesh. The `knative-ingress-gateway` in the `knative-serving` namespace is a shared Istio gateway resource that handles all incoming (north-south) traffic to Knative services. This gateway points to the underlying `istio-ingressgateway` service in the `istio-system` namespace. You can replace this gateway with an existing or customized gateway, see [Configuring the Ingress gateway](setting-up-custom-ingress-gateway.md).
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+
+    Istio provides the following additional configuration options:
+
+    - Advanced gateway selection with label selectors for fine-grained routing.
+    - Support for mesh-aware and cluster-local access.
```
evankanderson, line 148:
```suggestion
    In addition to [Istio's configuration options](https://istio.io/latest/docs/reference/config/), `net-istio` provides the following additional configuration options for mapping Knative Routes to Istio VirtualServices:

    - Advanced gateway selection with label selectors for fine-grained routing.
    - Support for mesh-aware and cluster-local access.
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+        Controller["GatewayClass&nbsp;Controller"]
+      end
+        KSvc["Knative&nbsp;Service"] --> Route
+        Route -- translates&nbsp;to --> GW & HR
+        GW --> Controller
+        HR --> Controller
+        Controller -- routes&nbsp;traffic&nbsp;to --> Pods["Your&nbsp;Pods"]
+
+    style net-gateway-api fill:#e3f2fd,stroke:#1976d2
+    style underlying fill:#fff3e0,stroke:#ef6c00
```
evankanderson, line 182: We should re-use the `Route --> KIngress` diagram, as it's still the mechanism for `net-gateway-api` as well:

```suggestion
    flowchart LR
     net-gateway["net-gateway-api controller"]
     GW["Gateway"]
     KIngress["KIngress objects"]
     HR["HTTPRoute"]
     subgraph underlying["Gateway-API Implementation<br>(Contour, Istio, Envoy Gateway, …)"]
        Controller("GatewayClass&nbsp;Controller")
      end
        KIngress -- read by --> net-gateway
        net-gateway -- updates --> GW
        net-gateway -- creates --> HR
        GW -- read by --> Controller
        HR -- read by --> Controller

    style net-gateway fill:#FFE0B2
    style GW fill:#BBDEFB
    style HR fill:#BBDEFB
    style Controller fill:#BBDEFB
    style underlying fill:#fff3e0,stroke:#ef6c00
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+    ```
+
+    The Knative `net-gateway-api` is a KIngress implementation for Knative integration with the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/). A good choice for teams adopting the Gateway API to unify ingress across Kubernetes.
```
evankanderson: I would say that Gateway API is recommended for many teams, but may experience limits with large numbers of Knative Services.  @dprotaso can correct me here.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+
+``` bash
+kubectl get pods -n knative-serving
+```
+
+The ingress controllers have the following names:
+
+- Kourier: `kourier-control-*`, and `kourier-gateway-*`. Kourier is included in the Knative Serving installation should appear in the results when your cluster is first created.
+- Contour: `contour-*`
+- Istio: `istio-webhook-*`. The main Istio control plane pods such as `istiod-*` are in the `istio-system` namespace. Knative adds the `istio-webhook-*` pod in the `knative-serving` namespace when Istio is the chosen networking layer.
```
evankanderson, line 212: For Gateway API, running `kubectl get gatewayclass` should list which gateway implementations are available in the cluster.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+
+Each ingress controller manages only those ingress objects that are annotated with its key. Knative Serving uses the controller based on the `network-config` ConfigMap.
```
evankanderson: I think "selects" here is more precise, because Serving encompasses both the core controller and the various networking plugins.

```suggestion
Each ingress controller manages only those ingress objects that are annotated with its key. Knative Serving selects the controller based on the rules in the `network-config` ConfigMap.
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+If you want to install a new controller, see [Install serving with YAML](../install/yaml-install/serving/install-serving-with-yaml.md). Be aware that changing the Ingress class of an existing Route may result in undefined behavior.
+
+## Gateway configurations
+
+Knative assumes two gateways in the cluster:
+
+- Externally exposed - defines the gateway for external traffic.
+- Internally-only exposed - defines the gateway for local traffic.
```
evankanderson: Can you add a TODO here to add a diagram?  I don't think we need to add it now, but it would be a handy future addition.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,315 @@
...
+
+When gateways are installed, the `config-gateway` ConfigMap is updated to track these two gateways set to the following environment variables:
+
+- class: `$GATEWAY_CLASS_NAME`
+- gateway: `$NAMESPACE/$GATEWAY_NAME`
+- service: `$NAMESPACE/$SERVICE_NAME`
```
evankanderson: I think this is trying to describe the configuration for Gateway API, but it somehow got here.

Also, these are fields in the configuration object, not environment variables.  We happen to use a similar familiar syntax to environment variables, but the user needs to **fill in** these values, not consume them.

docs/versioned/.nav.yml
```diff
@@ -267,6 +267,7 @@ nav:
         - Configure high-availability components: serving/config-ha.md
         - Exclude namespaces from the Knative webhook: serving/webhook-customizations.md
         - Networking Options:
+          - Configure Knative networking: serving/config-network-adapters.md
           - Configure the ingress gateway: serving/setting-up-custom-ingress-gateway.md
```
evankanderson: We should name this "Configure Istio's ingress gateway", to avoid confusion with the other KIngress implementations.  (Yes, this is existing, but it gets more confusing with a "Configure Knative Networking" item.)

evankanderson (commented):
Sorry for the long delay!  I didn't have a chance to dig into the bad macro rendering until today, and it took much longer than it should have to realize that we had _two_ snippet injection systems.  Sorry!

A bunch of comments, but I think the remaining work is actually fairly short/straightforward.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,348 @@
+---
+audience: administrator
+components:
+  - serving
+function: how-to
+---
+
+# Configure Knative networking
+
+This page provides configuration guidance for Knative Serving's integration with Kubernetes ingress controllers. Knative leverages existing ingress controls in your Kubernetes cluster, allowing you to use the same monitoring features and capabilities.
```
iRaindrop: added "features and capabilities" at end of sentence.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,348 @@
...
+The Knative tested ingress controllers (Contour, Istio, and Kourier) have the following common configurations:
+
+- Certificate management: Configurable secrets for TLS encrypted traffic. See [Using a custom TLS certificate for DomainMapping](./services/custom-tls-certificate-domain-mapping.md).
+- Timeout policies: Controls for idle, and response stream timeouts. See [Configuring the Defaults ConfigMap](./configuration/config-defaults.md) to review timeout settings.
+- Traffic visibility: Mechanisms to expose services externally or cluster-locally. See [Traffic management](./traffic-management.md).
+- The mapping of Route objects to `networking.internal.knative.dev` objects.
+
+The Knative `networking.internal.knative.dev` Ingress type is generally referred to as KIngress objects.
+
+In scenarios where there are multiple networking implementations, you can create a custom ingress class to specify different ingress class annotations for each service. For more information, see [Configuring Services custom ingress class](./services/ingress-class.md)
```
iRaindrop: FYI - I added this link to the custom ingress class (new).

docs/versioned/serving/config-network-adapters.md
iRaindrop: Changed to: 

"A recommended choice for many teams adopting the Gateway API to unify ingress across Kubernetes. Clusters with large numbers of Knative Services however, may experience limitations."

docs/snippets/no-dns.md
iRaindrop: STET -  I must have cut instead of copied for use as a template

docs/snippets/netadapter-gatewayapi.md
```diff
@@ -0,0 +1,74 @@
...
+    ```bash
+    export KO_DOCKER_REPO=kind.local
+    export KIND_CLUSTER_NAME=knative  # (keep this if your cluster is named "knative")
+    ```
+
+1. Deploy Gateway API resources:
+
+    ```bash
+    ko apply -f config/
+    ```
```
evankanderson: We don't need these if people are installing the release version (which they should).

docs/snippets/netadapter-gatewayapi.md
evankanderson: Actually, looking at the release artifacts at https://github.com/knative-extensions/net-gateway-api/releases/tag/knative-v1.20.1, it looks like configuration for Contour / Istio are already shipped as `contour-gateway.yaml` and `istio-gateway.yaml`.  We'll still need directions for using a different Gateway API implementation (which involves creating the Gateways and pointing to them in the `config-gateway` ConfigMap).

evankanderson: (It looks like this is mostly resolved starting on line 38)

docs/snippets/netadapter-gatewayapi.md
```diff
@@ -0,0 +1,74 @@
...
+1. Configure Knative Serving to use the Knative Gateway API ingress class:
+
+    ```bash
+    kubectl patch configmap/config-network \
+      --namespace knative-serving \
+      --type merge \
+      --patch '{"data":{"ingress-class":"gateway-api.ingress.networking.knative.dev"}}'
+    ```
+
+1. Edit the `config-gateway` ConfigMap in the knative-serving namespace to specify gateway resources for external and local traffic. For the `external-gateways` key, specify the value for `name` and for `service` as needed. Do the same for the `local gateways`. The `namespace` should be kept at `knative-serving`.
```
evankanderson:
```suggestion
1. Create Gateway resources for use by external ("north-south") Knative traffic, and local ("east-west") traffic.  If you do not need separate routing for local traffic (or [private Knative services](../../../serving/services/private-services.md)), you can use the external Gateway for both.  Note both the Kubernetes namespace and name of the Gateways, and an underlying DNS name (such as a Kubernetes service DNS name) corresponding to the Gateway, which will be used by Knative to verify traffic routing.

1. Edit the `config-gateway` ConfigMap in the `knative-serving` namespace to specify the Gateway resources for external and local traffic. For the `external-gateways` key, specify the values for `name`, `namespace` and for `service` corresponding to the external Gateway resource. Do the same for the `local-gateways` key:
```

docs/snippets/netadapter-gatewayapi.md
```diff
@@ -0,0 +1,74 @@
...
+    ```bash
+    cat <<EOF | kubectl apply -f -
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: config-gateway
+      namespace: knative-serving
+      labels:
+        app.kubernetes.io/component: net-gateway-api
+        app.kubernetes.io/name: knative-serving
```
evankanderson: You don't need these labels on the ConfigMap.


```suggestion
```

docs/snippets/netadapter-istio.md
```diff
@@ -0,0 +1,33 @@
...
+1. Install the Knative Istio controller:
+
+    ```bash
+    kubectl apply -f {{ artifact(repo="net-istio",org="knative-extensions", file="net-istio.yaml") }}
+    ```
+
+    <!-- ```bash
+    kubectl apply -f https://github.com/knative/net-istio/releases/download/knative-v1.20.1/net-istio.yaml
+    ``` -->
+
```
evankanderson:
```suggestion
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+
+The Knative tested ingress controllers (Contour, Istio, and Kourier) have the following common configurations:
+
+- Certificate management: Configurable secrets for TLS encrypted traffic. See [Using a custom TLS certificate for DomainMapping](./services/custom-tls-certificate-domain-mapping.md).
+- Timeout policies: Controls for idle, and response stream timeouts. To review timeout settings, see [Configuring the Defaults ConfigMap](./configuration/config-defaults.md).
+- Traffic visibility: Mechanisms to expose services externally or cluster-locally. See [Traffic management](./traffic-management.md).
+- The mapping of Route objects to `networking.internal.knative.dev` objects is common to these controllers.
+
+The Knative `networking.internal.knative.dev` Ingress type is generally referred to as KIngress objects.
```
evankanderson: Given that the mapping of Knative Route objects to KIngress is uniform, why not move the diagram up here?

```suggestion
The Knative `networking.internal.knative.dev` Ingress type is generally referred to as KIngress objects.

    ```mermaid
    ---
    config:
      layout: elk
      theme: default
    ---
    flowchart LR
      subgraph top[" "]
        direction LR
            kingress1["Ingress object (KIngress)<br>networking.internal.knative.dev"]
            serving("Serving<br>controller")
            route["Route object<br>serving.knative.dev"]
      end
        route -- read by --> serving
        serving -- creates --> kingress1
        style kingress1 fill:#BBDEFB,stroke-width:1px,stroke-dasharray: 0
        style serving fill:#FFE0B2
        style top fill:transparent
    ```
```

docs/versioned/serving/config-network-adapters.md
evankanderson: (Suggested where to add it above, and then remove from here, Contour, and Istio diagrams.  It's already missing from Gateway API.)

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+
+    Istio provides the following additional configuration options:
+
+    - Advanced gateway selection with label selectors for fine-grained routing.
+    - Support for mesh-aware and cluster-local access.
+    
+    In addition to [Istio's configuration options](https://istio.io/latest/docs/reference/config/), `net-istio` provides the following additional configuration options for mapping Knative Routes to Istio VirtualServices:
+
+    - Advanced gateway selection with label selectors for fine-grained routing.
+    - Support for mesh-aware and cluster-local access.
```
evankanderson, line 153: This list is a repeat of the previous.  You may need to adjust the previous paragraph given the removal of the list.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+=== "Gateway API"
+
+    ```mermaid
+    ---
+    config:
+      layout: elk
+      theme: default
+    ---
+    flowchart LR
+     net-gateway["net-gateway-api controller"]
```
evankanderson: The diagram gets kind of small, this might give it a little room to breathe.
```suggestion
     net-gateway["net-gateway-api<br>controller"]
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+     GW["Gateway"]
+     KIngress["KIngress objects"]
+     HR["HTTPRoute"]
+     subgraph underlying["Gateway-API Implementation<br>(Contour, Istio, Envoy Gateway, …)"]
+        Controller("GatewayClass&nbsp;Controller")
```
evankanderson, line 171: Same thing here:
```suggestion
        Controller("GatewayClass<br>Controller")
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+    style net-gateway fill:#FFE0B2
+    style GW fill:#BBDEFB
+    style HR fill:#BBDEFB
+    style Controller fill:#BBDEFB
+    style underlying fill:#fff3e0,stroke:#ef6c00
+    ```
+
+    The Knative `net-gateway-api` is a KIngress implementation for Knative integration with the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/). A recommended choice for many teams adopting the Gateway API to unify ingress across Kubernetes. Clusters with large numbers of Knative Services however, may experience limitations.
+
+    The Kubernetes Gateway API requires a controller or service mesh. Istio and Contour implementations are tested. For more information see [Tested Gateway API version and Ingress](https://github.com/knative-extensions/net-gateway-api/blob/main/docs/test-version.md).
```
evankanderson:
```suggestion
    The Kubernetes Gateway API requires a controller or service mesh. Istio, Contour, and Envoy Gateway implementations are tested. For more information see [Tested Gateway API version and Ingress](https://github.com/knative-extensions/net-gateway-api/blob/main/docs/test-version.md).
    
    Each Gateway API implementation supports its own configuration options and policy configuration; Knative explicitly supports [attaching Gateway policies](https://gateway-api.sigs.k8s.io/reference/policy-attachment/) to the HTTPRoutes created by Knative, as well as the Gateways objects referenced by Knative.
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+``` bash
+kubectl get pods -n knative-serving
+```
+
+The ingress controllers have the following names:
+
+- Kourier: `kourier-control-*`, and `kourier-gateway-*`. Kourier is included in the Knative Serving installation should appear in the results when your cluster is first created.
+- Contour: `contour-*`
+- Istio: `istio-webhook-*`. The main Istio control plane pods such as `istiod-*` are in the `istio-system` namespace. Knative adds the `istio-webhook-*` pod in the `knative-serving` namespace when Istio is the chosen networking layer.
+
```
evankanderson, line 214:
```suggestion
- Gateway API: `net-gateway-api-controller` is the deployment used to interface with the Gateway API. It uses the configuration from `config-gateway` to select existing Gateway resources on the cluster.

```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+You can remove an unused key with a dot with the following command:
+
+```bash
+kubectl patch configmap config-network -n knative-serving \                                                    
+--type=json -p='[{"op": "remove", "path": "/data/ingress.class"}]'
+```
+
+If you want to install a new controller, see [Install serving with YAML](../install/yaml-install/serving/install-serving-with-yaml.md). Be aware that changing the Ingress class of an existing Route may result in undefined behavior.
+
+## Gateway configurations
```
evankanderson, line 252: We need something here to call out this this only applies for `net-gateway-api` / using the `gateway-api.ingress.networking.knative.dev` ingress type.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+ subgraph Cluster["Knative Cluster"]
+    direction TB
+        ExtGateway["External Gateway<br>Exposed to external traffic"]
+        IntGateway["Internal Gateway<br>Only for local internal traffic"]
+        KServe["Knative Services<br>Serving and Eventing"]
+  end
+    ExtGateway -- Routes external HTTP/HTTPS traffic --> KServe
+    IntGateway -- Routes internal traffic --> KServe
+    ExternalClient["External Client<br>Internet / Outside cluster"] -- Ingress --> ExtGateway
+    InternalClient["Internal Client<br>Pod / Service inside cluster"] -- "Cluster-internal" --> IntGateway
```
evankanderson: "InternalClient" needs to be in the "Knative Cluster" subgraph.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+
+```mermaid
+---
+config:
+  look: neo
+  theme: redux
+  layout: elk
+---
+flowchart LR
+ subgraph Cluster["Knative Cluster"]
```
evankanderson, line 273:
```suggestion
 subgraph Cluster["Kubernetes Cluster"]
```

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+    direction TB
+        ExtGateway["External Gateway<br>Exposed to external traffic"]
+        IntGateway["Internal Gateway<br>Only for local internal traffic"]
+        KServe["Knative Services<br>Serving and Eventing"]
```
evankanderson:
```suggestion
        KServe["Knative Services"]
```

This is all Serving.  Unfortunately, Eventing has a somewhat different model and doesn't have clear abstractions for exposing the event ingress from outside the cluster.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,358 @@
...
+
+Example results:
+
+```bash
+NAME           CONTROLLER                    ACCEPTED   AGE
+istio          istio.io/gateway-controller   True       11d
+istio-remote   istio.io/unmanaged-gateway    True       11d
+```
+
+When gateways are installed, the `config-gateway` ConfigMap is updated to track the `class`, `gateway`, and `service`. The service is the Kubernetes Service name that points to the pods in the Gateway implementation.
```
evankanderson: This doesn't happen automatically -- this needs to be done by a cluster administrator when they want to change the Gateways used by a certain type of traffic.  There's one extra layer here which I didn't describe previously (because it's Gateway-API general, rather than Knative-specific):

```$ kubectl get gatewayclass --context prod
NAME      CONTROLLER                                      ACCEPTED   AGE
cilium    io.cilium/gateway-controller                    True       134d
default   gateway.envoyproxy.io/gatewayclass-controller   True       127d

$ kubectl get gateway --context prod --all-namespaces
NAMESPACE              NAME     CLASS     ADDRESS          PROGRAMMED   AGE
envoy-gateway-system   shared   default   100.100.100.100   True         127d
```

So specific Gateways (which go in `config-network`) will reference a GatewayClass, which says which implementation that Gateway should use.  But it may make more sense to simply have people list the Gateway objects in their cluster, and ignore the GatewayClass in this document.

docs/versioned/install/yaml-install/serving/install-serving-with-yaml.md
```diff
@@ -41,104 +41,23 @@ To install the Knative Serving component:
...
-
-        ```bash
-        kubectl --namespace istio-system get service istio-ingressgateway
-        ```
-
-        !!! tip
-            Save this to use in the following [Configure DNS](#configure-dns) section.
+=== "Kourier"
 
+    {% include "netadapter-kourier.md" %}
```
evankanderson: I did a little more digging, and it looks like we can use the `indent` jinja2 filter here to avoid needing to indent the contents of each `netadapter-*` file:


```suggestion
{% filter indent(width=4) %}
{% include "netadapter-kourier.md" %}
{% endfilter %}
```

docs/versioned/serving/config-network-adapters.md
iRaindrop: Removed that paragraph. The "Edit the `config-gateway` ConfigMap" step in netadapter-gateway.api (in the installation instructions) show editing the YAML for configuring the external and internal gateways.

docs/versioned/install/yaml-install/serving/install-serving-with-yaml.md
```diff
@@ -41,104 +41,14 @@ To install the Knative Serving component:
...
-            Save this to use in the following [Configure DNS](#configure-dns) section.
-
+Expand the following tabs for instructions on installing network layers. For an overview of network layer options, architecture, and configurations see [Configure Knative networking](../../../serving/config-network-adapters.md).
+
+{% filter indent(width=4) %}
+{% include "netadapter-kourier.md" %}
+{% include "netadapter-contour.md" %}
+{% include "netadapter-istio.md" %}
+{% include "netadapter-gatewayapi.md" %}
+{% endfilter %}
```
evankanderson, line 61: You don't need (or want) `{% filter indent(width=4) %}` if you're going to put the tab headers in the included documents.  A 4-space indent (absent other context) in Markdown is going to mean "preformatted text", which is what we see in the render.

docs/versioned/install/yaml-install/serving/install-serving-with-yaml.md
```diff
@@ -41,104 +41,14 @@ To install the Knative Serving component:
...
-            Save this to use in the following [Configure DNS](#configure-dns) section.
-
+Expand the following tabs for instructions on installing network layers. For an overview of network layer options, architecture, and configurations see [Configure Knative networking](../../../serving/config-network-adapters.md).
+
+{% include "netadapter-kourier.md" %}
+{% filter indent(width=4) %}
+{% include "netadapter-contour.md" %}
+{% include "netadapter-istio.md" %}
+{% include "netadapter-gatewayapi.md" %}
+{% endfilter %}
```
evankanderson, line 61: You don't want to include the `{% filter indent(width=4) %}` directive if you're putting the tab headers into the included documents.  Absent other context, a 4-space indent means "preformatted text" in markdown, which is what you get in the render.  So, *either*:

```
=== Kourier
{% filter indent(width=4) %}
{% include "netadapter-kourier.md" %}
{% endfilter %}
```

with _no_ indent in the `netadapter-kourier.md` file, **OR**:

```
{% include "netadapter-kourier.md" %}
```

Where the included file looks like:

```
=== Kourier

    Some words about kourier...
```

iRaindrop: Yeah, I've been trying the right combo. But not this yet:
=== Kourier
{% filter indent(width=4) %}
{% include "netadapter-kourier.md" %}
{% endfilter %}

Fingers crossed!

docs/versioned/serving/config-network-adapters.md
iRaindrop: A great developmental edit observation.

docs/snippets/netadapter-gatewayapi.md
```diff
@@ -0,0 +1,58 @@
...
+-->
+Use the following steps to install and configure the Knative Gateway API adapter.  Note that you need to have a [Gateway API implementation](https://gateway-api.sigs.k8s.io/implementations/) installed in your cluster.  The Knative team currently tests the Istio, Contour, and Envoy-Gateway implementations of Gateway API.
+
+1. Install the Knative Gateway API:
+
+    ```bash
+      kubectl apply -f {{ artifact(repo="net-gateway-api",org="knative-extensions",file="net-gateway-api.yaml") }}
+    ```
+
+    Alternatively, you can also install a preconfigured Gateways for either Contour or Istio. Replace the filename in this step with either `contour-gateway.yaml` or `istio-gateway.yaml`. If you use either file  you can skip the next step to configure gateway resources.
```
evankanderson: The `contour-gateway.yaml` and `istio-gateway.yaml` don't replace the `net-gateway-api.yaml` file, but they replace steps 2 and 3.  Not sure how to make that clearer here, but right now, it reads as if you can just install those files _instead_ of the `net-gateway-api.yaml` (which won't produce a working system, but the commands _will_ succeed).

docs/snippets/netadapter-istio.md
evankanderson: We got this working, resolving the comment.

docs/versioned/serving/config-network-adapters.md
evankanderson: Moving this content out of each tab still needs to happen.  I'd suggest moving the diagram up to the first section (maybe around line 11), and also moving line 222 explaining what a KIngress is up to that area.  I'd probably also include an explanation of why the type exists:

> The Knative `networking.internal.knative.dev` Ingress type is generally referred to as KIngress objects. **Knative uses the KIngress type to provide a common interface across multiple ingress implementations.**

(bold added)

I'd also use this diagram, which includes the Knative Route object to make it clear what user-visible resource maps to the "hidden" KIngress resource:

```
    ```mermaid
    ---
    config:
      layout: elk
      theme: default
    ---
    flowchart LR
      subgraph top[" "]
        direction LR
            kingress1["Ingress object (KIngress)<br>networking.internal.knative.dev"]
            serving("Serving<br>controller")
            route["Route object<br>serving.knative.dev"]
      end
        route -- read by --> serving
        serving -- creates --> kingress1
        style kingress1 fill:#BBDEFB,stroke-width:1px,stroke-dasharray: 0
        style serving fill:#FFE0B2
        style top fill:transparent
    ```
```

docs/versioned/serving/config-network-adapters.md
evankanderson: https://github.com/knative-extensions/net-kourier?tab=readme-ov-file#deployment would be a reasonable link, or https://github.com/knative-extensions/net-kourier/blob/main/README.md#deployment

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,341 @@
...
+    style net-gateway fill:#FFE0B2
+    style GW fill:#BBDEFB
+    style HR fill:#BBDEFB
+    style Controller fill:#BBDEFB
+    style underlying fill:#fff3e0,stroke:#ef6c00
+    ```
+
+    The Knative `net-gateway-api` is a KIngress implementation for Knative integration with the [Kubernetes Gateway API](https://gateway-api.sigs.k8s.io/). A recommended choice for many teams adopting the Gateway API to unify ingress across Kubernetes. Clusters with large numbers of Knative Services however, may experience limitations.
+
+    The Kubernetes Gateway API requires a controller or service mesh. Istio and Contour implementations are tested. For more information see [Tested Gateway API version and Ingress](https://github.com/knative-extensions/net-gateway-api/blob/main/docs/test-version.md).
```
evankanderson:
```suggestion
    The Kubernetes Gateway API supports a number of implementations. Istio, Contour, and Envoy-Gateway implementations are tested. For more information see [Tested Gateway API version and Ingress](https://github.com/knative-extensions/net-gateway-api/blob/main/docs/test-version.md).
```

evankanderson: Service meshes and controllers are _different_ types of things.  It would be like saying "gardens need fruit or paths".  Let's call them implementations, and include envoy-gateway.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,341 @@
...
+=== "Gateway API"
+
+    ```mermaid
+    ---
+    config:
+      layout: elk
+      theme: default
+    ---
+    flowchart LR
+     net-gateway["net-gateway-api controller"]
```
evankanderson:
```suggestion
     net-gateway("net-gateway-api<br>controller")
```

Let's narrow this text so we can get slightly bigger font here.  Also, let's use rounded-boxes for software components like controllers, and square boxes for Kubernetes resources (objects).

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,341 @@
...
+    ---
+    config:
+      layout: elk
+      theme: default
+    ---
+    flowchart LR
+     subgraph bottom[" "]
+        direction LR
+            envoy["Envoy deployment<br>kourier-system namespace"]
+            kourier["net-kourier<br>controller"]
```
evankanderson:
```suggestion
            envoy("Envoy deployment<br>kourier-system namespace")
            kourier("net-kourier<br>controller")
```

Let's use rounded-boxes for software components.

docs/versioned/serving/config-network-adapters.md
evankanderson: I'd also suggested another sentence here covering how options were configured, to match the other tabs.  In this case, I'd probably say:

> Knative is designed to support [attaching Gateway API policies](https://gateway-api.sigs.k8s.io/reference/policy-attachment/) to both Knative-created HTTPRoutes and the existing Gateway resources which manage with those routes.

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,341 @@
...
+
+``` bash
+kubectl get pods -n knative-serving
+```
+
+The ingress controllers have the following names:
+
+- Kourier: `kourier-control-*`, and `kourier-gateway-*`. Kourier is included in the Knative Serving installation should appear in the results when your cluster is first created.
+- Contour: `contour-*`
+- Istio: `istio-webhook-*`. The main Istio control plane pods such as `istiod-*` are in the `istio-system` namespace. Knative adds the `istio-webhook-*` pod in the `knative-serving` namespace when Istio is the chosen networking layer.
```
evankanderson, line 212: Again, I think I had a comment here, but Gateway API will create `net-gateway-api-*` pods in the `knative-serving` namespace, which can be used to recognize that Gateway API KIngress class is available.  (The same as the other directions.)

docs/versioned/serving/config-network-adapters.md
```diff
@@ -0,0 +1,341 @@
...
+--type=json -p='[{"op": "remove", "path": "/data/ingress.class"}]'
+```
+
+If you want to install a new controller, see [Install serving with YAML](../install/yaml-install/serving/install-serving-with-yaml.md). Be aware that changing the Ingress class of an existing Route may result in undefined behavior.
+
+In scenarios where there are multiple networking implementations, you can create a custom ingress class to specify different ingress class annotations for each service. For more information, see [Configuring Services custom ingress class](./services/ingress-class.md).
+
+There are also third-party Knative networking options and Knative products available but are not tested or managed by the Knative community. For more information, see [Knative offerings](../install/knative-offerings.md).
+
+## Gateway configurations
```
evankanderson, line 252: This is all Gateway-API specific.  I don't know if you want to create a parallel page with "Configure Istio's ingress gateway" (e.g. "Configure Gateway API's ingress gateway"), or if you just want to add a clarification banner here that the following section only applies to Gateway API.

iRaindrop: Your call, I can create a new page - would it need additional content? Added the clarification for now.

docs/versioned/serving/config-network-adapters.md
iRaindrop: you did - thanks for catching it

docs/snippets/netadapter-gatewayapi.md
iRaindrop: I removed that "Alternative ...." paragraph (I tested it on my quickstart and it failed) added a the following at the top of the file:
This component in beta. For the latest information including supported implementations, see the [Knative net-gateway-api](https://github.com/knative-extensions/net-gateway-api).

docs/snippets/dns.md
```diff
@@ -1,20 +1,11 @@
-You can configure DNS to prevent the need to run curl commands with a host header.
+<!-- Referenced by:
+- install/operator/knative-with-operators.md
```
evankanderson, line 2: Sorry, I just noticed this, but you removed the tab header and prelude from the "Install by using the Knative Operator" page.  This also affects the "No DNS" tab on that page:

Image

evankanderson: It previously was wrong like this, so removing the `!!! warning` is probably correct:

Image

docs/snippets/real-dns-yaml.md
```diff
@@ -1,30 +1,28 @@
-=== "Real DNS"
+<!-- Referenced by:
+- install/yaml-install/serving/install-serving-with-yaml.md
```
evankanderson, line 2: This is also referenced by `install/operator/knative-with-operators.md`, and needs the update there like the other DNS entries.

docs/snippets/dns.md
iRaindrop: I'm sorry I overlooked that. Thanks for catching it!

dprotaso:
Do we want to cherry pick these changes to 1.20?

evankanderson:
Sure, why not.

/cherrypick 1.20

PR #6533, opened 2025-12-03
Fix Serving Metric Queue Names + add HTTP Metrics

dprotaso:
Part of https://github.com/knative/serving/issues/16216


dprotaso:
cc @Cali0707

dprotaso:
/assign @Cali0707 
cc @linkvt

PR #6556, opened 2026-01-14
docs: prevent header tabs from overlapping on search expand

### LINKED ISSUE #6332: When search bar is expanded it causes top section titles to be scrunched
You can see in the below image the top section titles and version picker and skewed

Image

jijo-OO7:
## Header tabs overlap when search expands [Fixes #6332]

problem

## Proposed Changes

- Fixed navigation layout shift when search is activated
- Restructured header to use separate sticky navigation container
- Maintained responsive design across all breakpoints

**Reference:** https://squidfunk.github.io/mkdocs-material/tutorials/

### DIFF
--- a/docs/stylesheets/extra.css
+++ b/docs/stylesheets/extra.css
@@ -673,6 +673,49 @@ a.md-header__button.md-logo img {
   }
 }
 
+@media (min-width: 1025px) {
+  .md-header__source {
+    color: var(--md-primary-fg-color--dark);
+    flex: 0 0 auto !important;   
+    margin-left: 1rem;          
+  }
+  
+  [data-md-toggle="search"]:checked ~ .md-header .md-header__source {
+    transform: none !important;
+    margin-left: 1rem !important;
+  }
+  
+  .md-search__form {
+    border-radius: 10px;
+  }
+  
+  .md-search__inner {
+    max-width: 31rem;
+  }
+  
+  .md-search__input {
+    background-color: white;
+    border: black 1px solid;
+    border-radius: 10px;
+    width: 100%;
+  }
+  
+  .md-header__title {
+    flex: 0 0 auto !important;
+    min-width: 11rem;            
+    max-width: 16rem;             
+    margin-right: 0.75rem;
+  }
+  
+  .md-header__list {
+    gap: 0.75rem;
+  }
+}
+
+.md-search-result__article {
+  max-height: none;
+}
+
 .md-search-result__article {
   max-height: none;
 }


lakxy:
Hi @jijo-OO7 I don't think you have actually followed the issue discussion, read it then open the PR.

lakxy:
This is totally mis-aligned from what @dprotaso described there.

dprotaso:
Yeah I don't see any difference on Chrome Desktop.

dprotaso:
Note the idea is for the home, docs etc in the top row to _not_ move when the search bar has focus

jijo-OO7:
Thanks for the pointer. I’ve re-read the original issue discussion and I see where the intent differs from the current implementation.
Based on @dprotaso’s clarification, I’ll update the PR so the top navigation remains static when the search bar gains focus

jijo-OO7:
Hi @dprotaso Updated the implementation based on your feedback. The top navigation now remains static when search gains focus by moving to a separate sticky container, following the Material for MkDocs pattern.

Would appreciate your review to ensure this matches the intended behavior. Let me know if any further adjustments are needed.

lakxy:
@jijo-OO7 I am not sure if this is what was described? the before version is not complete, initially there were elements in nav beside the search bar, which have been removed in after version,
You explicitly hid that part in PR description? why?

lakxy:
Final review to @dprotaso . Does not seem optimal to me.

lakxy:
> Thanks for the pointer. I’ve re-read the original issue discussion and I see where the intent differs from the current implementation. Based on @dprotaso’s clarification, I’ll update the PR so the top navigation remains static when the search bar gains focus

@jijo-OO7 Again, you did not follow the issue discussion, as Far as I know the nav elements were described to be overlayed by search bar expansion and not removed.

dprotaso:
Yeah this doesn't look correct


Screenshot 2026-01-15 at 12 33 10 PM

jijo-OO7:
@dprotaso Could you try reloading the preview? The navigation elements remain visible on my local build and the search expands as expected per the Material for MkDocs pattern. A cache refresh might resolve any discrepancies.
Screenshot (19)

dprotaso:
I would expect `Home` etc. to be at the same horizontal line as the search bar

jijo-OO7:
Got it, thank you for clarifying. I understand now, the navigation items should remain aligned on the same horizontal line as the search bar. I'll revisit the implementation to ensure proper alignment and will push an updated commit shortly with Home, Docs, About, Blog, and Community properly aligned alongside the search bar.

jijo-OO7:
Do we expect the top nav to be static as well upon search expansion ?

dprotaso:
> Do we expect the top nav to be static as well upon search expansion ?

Yeah

jijo-OO7:
Hi @dprotaso, I’ve updated the PR with the latest changes. Could you please take a look when you have a moment and let me know if this aligns with the intended vision, or if any further tweaks are needed?

jijo-OO7:
Fixed a broken redirect for the contributing docs that was causing the strict docs build to fail after rebasing onto the latest upstream/main

jijo-OO7:
@dprotaso I ran .``/hack/build.sh --strict`` locally and the build fails with:
``WARNING - Redirect target 'community/contributing/README.md'`` does not exist!

I verified the current docs structure and we have docs/community/contributing.md (file), not docs/community/contributing/README.md (directory + file). I tried updating the redirect target to community/contributing.md, but the strict check still fails.

This makes me think the CI check might be validating redirects against a different path layout (or across release branches where the structure differs). Could you confirm the expected redirect target here, or whether the strict check needs an update to reflect the current docs structure?

dprotaso:
> WARNING - Redirect target 'community/contributing/README.md' does not exist!

You can ignore that check - it's not blocking the PR

dprotaso:
I fixed it in another PR (https://github.com/knative/docs/pull/6570) - but it didn't help the warning and I don't have time to dig into it

dprotaso:
FYI - I'm still seeing the same behaviour as before

jijo-OO7:
Thanks for flagging this I’ll take a closer look and dig into what might still be causing it.

dprotaso:
Note - I'm looking at the preview mentioned in this comment and still see the issue

https://github.com/knative/docs/pull/6556#issuecomment-3750972475

I'm going to mark this PR as a draft - until you see things working in the preview as expected then please switch it be ready

jijo-OO7:
I’ve pushed a follow-up that fixes the version selector overlap when the search bar is expanded.

I’ve verified the latest preview on my end and it now behaves as expected. Could you please take another look and let me know if this works for us? I’m happy to make further tweaks if needed.
Thanks!

dprotaso:
I still see home, docs etc shifting around in the preview

jijo-OO7:
In that case do we expect the  knative logo and  the version selector to be static as well cause right now they just hide on search bar expansion or we are okay with the behavior the way it is at present?

dprotaso:
Clicking the search bar shouldn't move/hide anything. there should just be an gray overlay ontop of everything

jijo-OO7:
@dprotaso i have updated the pr with expected changes .
can you check on this once on your end .
Thank you for your time and patience.

dprotaso:
I see the menu doesn't move but everything else is broken.

@jijo-OO7 At this point I don't want to keep reviewing this PR while things remain in a broken state. This has happened on about 10 times now.

I'm moving this back to draft and I want _you_ to verify things are working as expected prior to marking this as done. To make open source managable I can't be 'testing' your changes when clearly you have not.

dprotaso:
Everytime you push a change the preview is updated here: https://github.com/knative/docs/pull/6556#issuecomment-3750972475

You should be using that to verify your work.

jijo-OO7:
> I see the menu doesn't move but everything else is broken.
> 
> @jijo-OO7 At this point I don't want to keep reviewing this PR while things remain in a broken state. This has happened on about 10 times now.
> 
> I'm moving this back to draft and I want _you_ to verify things are working as expected prior to marking this as done. To make open source managable I can't be 'testing' your changes when clearly you have not.

I understand the concern, and apologies for the repeated broken states I’ll make sure this is properly validated against the preview going forward before marking it ready again.

jijo-OO7:
I’ve made some progress on reducing the overlap behavior, but resolving this cleanly across viewports is turning into a larger layout refactor than initially anticipated.

To avoid blocking other work, I’m going to pause on this for now. I’m happy to revisit and iterate further if this is still considered a priority.

Thanks!

jijo-OO7:
### UPDATE 
I revisited this and scoped the header/search layout fixes to desktop breakpoints only, so mobile behavior remains unchanged.
I’ve verified the overlap is resolved on desktop and that mobile viewports behave as before (including on deploy preview).

dprotaso:
The search functionality is still broken - please test your changes before marking the PR as ready

jijo-OO7:
Hi @dprotaso — thanks for the continued feedback and for your patience here.

I’ve been rechecking the latest Netlify preview and confirmed the search interaction is still not behaving as expected with my recent changes. I’m keeping this in draft until that’s resolved.

While debugging, I also noticed that even after reverting my CSS updates, the search remains in an “Initializing search” state in:

>Docker preview (./hack/docker/run.sh)

>Netlify preview

Since this differs from production, I might be missing something in the preview/dev setup. If there’s a known setup or reference branch you’d recommend for validating search behavior locally, I’d appreciate the pointer.

Thanks again for the guidance.

dprotaso:
unsure @evankanderson do you know why the search might be broken in the preview?

jijo-OO7:
@dprotaso I looked into the search issue to make sure I hadn't introduced a regression.

I pulled the latest`` upstream: main`` and ran the local Docker preview it turns out the ``"Initializing search" ``hang is present on the base branch as well. My Network tab shows that ``search_index.json`` isn't being requested even on a clean build of main, so it looks like an existing environment or worker issue.

Since my changes are limited to CSS locking header widths with ``flex: 0 0 auto ``and ``min-width ``they shouldn't be able to impact the JavaScript search logic or index fetching.

I’ve verified the layout fix by manually toggling the search state in the inspector; the nav tabs now stay static and aligned as intended. If the visual behavior looks good to you on the preview, let me know and I'll rebase this if needed  so we can move it out of draft.

dprotaso:
ok - thanks for checking.

When the search box expands can we make it wider similar to how it is here: 
https://squidfunk.github.io/mkdocs-material/tutorials/

jijo-OO7:
> ok - thanks for checking.
> 
> When the search box expands can we make it wider similar to how it is here: https://squidfunk.github.io/mkdocs-material/tutorials/

@dprotaso Certainly ,  allowing the expanded search to overlay the navigation tabs instead of pushing the flex layout. This keeps the header elements stable while still allowing the search field to expand wider, similar to the Material example you shared.

I’ve updated the PR with the new changes for preview. Please let me know if this approach works for us.

jijo-OO7:
@dprotaso friendly ping I’ve addressed the previous feedback and updated the search expansion behavior to overlay the navigation while keeping the header items static.

Could you take another look when you have a moment? Happy to make further adjustments if needed.

dprotaso:
Looks like the search bar spacing cuts off some of the navigation see the screenshot

Screenshot 2026-03-11 at 2 10 29 PM

jijo-OO7:
@dprotaso just a gentle  ping
 whenever you have a moment, I'd really appreciate your  feedback on the current behavior .

jijo-OO7:
Hi @dprotaso, I have pushed an update addressing the search bar's behavior across various viewports. I'd appreciate your review and feedback on these changes when you have a moment PTAL .
Thanks !

dprotaso:
/lgtm
/approve

search is broken @ main now but the change here seems good now and we can make adjustments in future PRs

thanks!

PR #6689, opened 2026-08-27
Mark multi-cluster deployment as alpha

kahirokunn:
Update the multi-cluster Operator documentation to mark the feature as alpha while the Cluster Inventory API is still `v1alpha1`.

## Proposed Changes

- Add the alpha stage to the multi-cluster guide and update the related blog post to match.


kahirokunn:
Originally, this functionality was marked as beta, but the dependent API was marked as alpha, so it should have been marked accordingly.
Only after upstream upgraded to beta could we finally mark this feature as beta—until then, it should never have been marked as beta. I only recently realized this rule.

kahirokunn:
@dsimansk Hi ✋ PTAL. Thank you 🙏

